PR #48, opened 2015-02-13
Use bash interpreter instead of sh

vadmeste:
/bin/sh points to light shell interpreter like dash in Debian, use env bash instead, check_version won't work otherwise


harshavardhana:
LGTM

PR #116, opened 2015-03-13
Add recursive directory upload and bucket download   support for './mc cp' with -r, deprecate './mc sync'

harshavardhana:


harshavardhana:
LGTM

PR #173, opened 2015-03-25
Implement donut client to be used with mc now supports creating bucket, cp of objects in and out

harshavardhana:


harshavardhana:
LGTM

PR #369, opened 2015-04-14
cmd-ls now treats urlType=file as list objects instead of list buckets always. Closes #365.

### LINKED ISSUE #365: mc ls . does not traverse into directories
Expected:
```sh
$ ms ls .
foo/hello
foo/world
```

Actual:
```sh
$ mc ls .
foo
```

fkautz:


fkautz:
LGTM

PR #386, opened 2015-04-17
Implement GetPartial() for fsClient, and renamed files `_unix.go` as we bring in windows specific support

harshavardhana:


harshavardhana:
LGTM

PR #423, opened 2015-04-18
Make all lexical paths file:///<path> ../<path> <path>/../../<path - behave similar - fixes #419

### LINKED ISSUE #419: Add all path styles relative, absolute - through ls, cp, mb

harshavardhana:


harshavardhana:
LGTM

PR #457, opened 2015-04-20
For ./mc cp if target doesn't have a bucket try to create one, just like its recursive version

harshavardhana:


harshavardhana:
LGTM

PR #458, opened 2015-04-20
Fix a bug introduced in qdb migration commit 27d87fdd8ec268446c238009cb86af1b255582dc when config file doesn't exist

harshavardhana:


harshavardhana:
LGTM

PR #461, opened 2015-04-20
Make sure to properly exit if requested theme not available, also re-structure console package a bit

harshavardhana:


harshavardhana:
LGTM

PR #537, opened 2015-04-25
Simplify cat by using Put(), since now Put() supports calculating md5sum internally and report errors on Close()

harshavardhana:


harshavardhana:
LGTM

PR #581, opened 2015-04-30
Access validation only for ListBuckets(), all others do not explicitly require Auth verification

harshavardhana:


harshavardhana:
LGTM

PR #599, opened 2015-05-09
Add new option for bucket ACL named ``authenticated-read`` to provide read access to all users who are part of AuthenticatedUsers group

harshavardhana:


harshavardhana:
LGTM

PR #610, opened 2015-05-15
Ignore multipart object Etag string, amazon doesn't guarantee them to be md5sum of the object

harshavardhana:


harshavardhana:
LGTM

PR #616, opened 2015-05-16
Right fix for progressbar segfault is not not initiating Finish() for an unstarted progress bar

harshavardhana:


harshavardhana:
LGTM

PR #623, opened 2015-05-16
Returning right after doCopy error would hang progressbar, we should just print the error and move forward

harshavardhana:


harshavardhana:
LGTM

PR #624, opened 2015-05-16
Fix copy cmd when the host has only one CPU core

vadmeste:
Some VPS providers offer a single core VM, that's how I detected the error :)


harshavardhana:
Thanks for the fix, inline comments added..

harshavardhana:
LGTM

PR #632, opened 2015-05-17
Add function isatty() to detect if Stdout for console is a terminal, if not disable coloring.

harshavardhana:


harshavardhana:
LGTM

PR #638, opened 2015-05-17
Bring in new changes from objectstorage-go which allow for recursive infinite putObject() operation

harshavardhana:


harshavardhana:
LGTM

PR #669, opened 2015-05-21
Fix output coloring in Windows Command Prompt

vadmeste:
fatih/color library doesn't help to distinguish between stdout/stderr, needed to import ansicolor directly to quickly fix the issue


harshavardhana:
LGTM

PR #677, opened 2015-05-22
Fix the coloring of File, Dir, Size, Time and Retry strings in the Windows console

vadmeste:


harshavardhana:
New changes on top of console package make this changeset deferred, closing this. Thanks for your help BTW.

PR #703, opened 2015-05-27
Fix a nil reference remove erroneous usage of url.Parse() use client.Parse(), unexport progress() method

harshavardhana:


harshavardhana:
LGTM

PR #848, opened 2015-06-30
Add fakeReader to sessions to populate progress bar as it builds and skips previously uploaded files

harshavardhana:


harshavardhana:
LGTM

PR #849, opened 2015-07-01
Protect progress bar printing by the mutex

vadmeste:
Use print() instead of color.Print function to protect progress bar by a mutex


harshavardhana:
This is nice, thanks.

PR #851, opened 2015-07-03
Asterisk in session clear cmd is interpreted by bash if not escaped. Use 'all' keyword instead.

vadmeste:


harshavardhana:
LGTM

PR #852, opened 2015-07-04
Prettify the progress bar output

vadmeste:
*) "➩" character  is deleted because progress bar dependency doesn't know how to count it (len("➩") = 3 though it takes only one column in the terminal)
*) One more change in bar.Callback to avoid overlapping between progress bar output and error messages

### DIFF
--- a/pb.go
+++ b/pb.go
@@ -152,7 +152,7 @@ func newCpBar() barSend {
 		bar.NotPrint = true
 		bar.ShowSpeed = true
 		bar.Callback = func(s string) {
-			console.Bar(s + "\r")
+			console.Bar("\r" + s)
 		}
 		cursorCh := cursorAnimate()
 		// Feels like wget


harshavardhana:
> Sorry, it was my bad description, I meant to say avoid the use of this set of characters "➩➪➫➬➭➮➯➱" in the code,

We do wish to keep these characters, will fix pb package instead.  The best fix is to get actual Rune length from utf8.RuneCountInString or len([]rune(string))

harshavardhana:
> We do wish to keep these characters, will fix pb package instead. The best fix is to get actual Rune length from utf8.RuneCountInString or len([]rune(string))

https://github.com/cheggaaa/pb/pull/46 an upstream pull request is submitted, can you separately provide a patch for item 2), i will merge it in.  Thanks again.

vadmeste:
well, actually this patch depends on cheggaaa/pb#46, you merged it too fast before I can comment this pull request :)

harshavardhana:
Don't worry, i migrated to minio/pb which has the necessary changes in place.

PR #853, opened 2015-07-05
Going to use minio/pb forked version as primary, since we might make more changes in future.

harshavardhana:


harshavardhana:
LGTM

PR #856, opened 2015-07-07
Avoid possible corrupted data when saving with quick

vadmeste:
The user could kill mc when saving data to a quick config data (such as large session file) which could lead to a possible corrupted data. os.Rename is atomic, so I save data in a tmp file before renaming the latter to the original file name

### DIFF
--- a/pkg/quick/quick.go
+++ b/pkg/quick/quick.go
@@ -115,7 +115,9 @@ func (d config) Save(filename string) (err error) {
 		return iodine.New(err, nil)
 	}
 
-	file, err := os.OpenFile(filename, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
+	tmpfile := filename + ".tmp"
+
+	file, err := os.OpenFile(tmpfile, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
 	if err != nil {
 		return iodine.New(err, nil)
 	}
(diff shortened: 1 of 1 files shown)


harshavardhana:
Thanks was about to do it! :-) .. YOU beat me to it.

harshavardhana:
Let's do it this way.. 

~~~
// File container provided by disk for atomic file writes
type File struct {
	*os.File
	path string
}

// Close the file replacing the configured file.
func (f *File) Close() error {
	if err := f.File.Close(); err != nil {
		return err
	}
	if err := os.Rename(f.Name(), f.path); err != nil {
		return err
	}
	return nil
}

// Abort closes the file and removes it instead of replacing the configured
// file. This is useful if after starting to write to the file you decide you
// don't want it anymore.
func (f *File) Abort() error {
	if err := f.File.Close(); err != nil {
		return err
	}
	if err := os.Remove(f.Name()); err != nil {
		return err
	}
	return nil
}
// CreateFile - create a file inside disk root path, replies with custome disk.File which provides atomic writes
func (disk Disk) CreateFile(filename string) (*File, error) {
	disk.lock.Lock()
	defer disk.lock.Unlock()

	if filename == "" {
		return nil, iodine.New(InvalidArgument{}, nil)
	}

	filePath := filepath.Join(disk.path, filename)
	// Create directories if they don't exist
	if err := os.MkdirAll(filepath.Dir(filePath), 0700); err != nil {
		return nil, iodine.New(err, nil)
	}
	f, err := ioutil.TempFile(filepath.Dir(filePath), filepath.Base(filePath))
	if err != nil {
		return nil, iodine.New(err, nil)
	}
	if err := os.Chmod(f.Name(), 0600); err != nil {
		os.Remove(f.Name())
		return nil, iodine.New(err, nil)
	}

	return &File{File: f, path: filePath}, nil
}
~~~

harshavardhana:
LGTM, merging my changes on top of this.. taking this in.

PR #858, opened 2015-07-08
Update to latest minio-go with Listobjects() handling for listing of more than 1000 entries when there are delimiters and prefixes

harshavardhana:


harshavardhana:
LGTM

PR #861, opened 2015-07-09
Correct definition of diff

bosky101:
hat-tip everyone.

Just checked it out today, made a small change.

Make it more clear that size difference is done

Looking to contribute more in the future.

### DIFF
--- a/cmd-diff.go
+++ b/cmd-diff.go
@@ -27,7 +27,7 @@ import (
 // Help message.
 var diffCmd = cli.Command{
 	Name:        "diff",
-	Usage:       "Compute differences between two files or folders",
+	Usage:       "Compute size differences between two files or folders",
 	Description: "NOTE: This command *DOES NOT* check for content similarity, which means objects with same size, but different content will not be spotted",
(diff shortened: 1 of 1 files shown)


vadmeste:
Thanks for you contribution,

Actually, `mc diff` can compute between two folders and doesn't only check for size differences.
Try this `mc diff /usr/bin /usr/sbin` to have more ideas about diff command

harshavardhana:
mc diff for the time being does simple fast verification with following scenarios

```
- names
- file type
- file size 
```

We would love to do checksum verification but restrained from it, perhaps if we find fast checksums with better scalablity would be nice.

harshavardhana:
Thanks merged!

abperiasamy:
Both Bosky and Anis are right. I struggled with that line when I wrote it.  I picked a simple definition over an accurate one. Users may not read the note in "Description" section and expect diff to compare content. If we mislead the user, it may result in a data-loss. 

We need to find a better definition for diff which satisfies both Anis and Bosky's condition.

PR #877, opened 2015-07-13
Add Close() methods for proxy readers

harshavardhana:


abperiasamy:
Don't merge it yet. I specifically kept it as reader..  Creater closes it. Because caller is only a a reader, not sender. Only sender should close.

harshavardhana:
> Don't merge it yet. I specifically kept it as reader.. Creater closes it. Because caller is only a a reader, not sender. Only sender should close

This is just a cosmetic change, which i was thinking to be the issue at hand. Looks like its not. Need to keep digging.

abperiasamy:
Actually this patch is correct. I like this change.. closing the tail end.

PR #913, opened 2015-07-23
If hostconfig not found for 127.0.0.1 or localhost, auth/keys not needed

krishnasrinivas:


harshavardhana:
@krishnasrinivas - try the latest master and see if it helps for you. You will have to regenerate the configs.

krishnasrinivas:
@harshavardhana that would work too and better solution

harshavardhana:
> @harshavardhana that would work too and better solution

Earlier there was no real support for access key generation, with new JSONRPC changes on server there is an easier way to generate and test signature v4 locally too. 

So it was silently ignored, now i think is the right time ot add them to regular configs and have them take access keys. 

Thanks for contributing @krishnasrinivas :-)

PR #948, opened 2015-07-30
Fix a config corruption bug when config.json has wrong json format, would lead to overwriting the previous config.json

harshavardhana:


harshavardhana:
LGTM

PR #1036, opened 2015-08-23
Show correct needed dependencies versions in checkdeps

vadmeste:


harshavardhana:
Thanks! this was missed.

PR #1083, opened 2015-09-08
Print folder names when the user asks to list several folders

vadmeste:
Without this PR, it is not possible to know which listed entry belongs to which source folder

### DIFF
--- a/ls-main.go
+++ b/ls-main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"github.com/minio/cli"
+	"github.com/minio/mc/pkg/console"
 	"github.com/minio/minio/pkg/probe"
 )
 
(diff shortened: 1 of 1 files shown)


vadmeste:
Sorry for not testing this patch enough before sending it.
Well, not sure but maybe we could add a flag which enforces to print the full path of listed files and directories each time we print it, somewhat like what mc does when it recursively lists a directory

harshavardhana:
> Well, not sure but maybe we could add a flag which enforces to print the full path of listed files and directories each time we print it, somewhat like what mc does when it recursively lists a directory

Adding a new flag is a bad idea we should incorporate your change, do not worry about being 'unix ls' output compatible.

This patch needs following 

- differentiate args with file v/s directories and behave appropriately 
- strip recursive suffix for all directory arguments. 

Do not worry about json it requires some refactoring on how arguments are passed down, for now covering above two would suffice.

harshavardhana:
https://github.com/harshavardhana/mc/commit/b6bcba4828d7ccc5d6ec73f6b4ab85d89418cba4 - take a look at the following change, clean it and incorporate it in your patch and resubmit.

vadmeste:
I looked at your change but it still seems to have something missing. Try with 'mc ls pkg c*.go vendor/ m*.go' and you can notice that the output is confusing. The unix ls sorts entries and starts to list files before directories. @harshavardhana, you can close this PR and go ahead with your implementation if you have a clear idea :)

harshavardhana:
> . The unix ls sorts entries and starts to list files before directories. @harshavardhana, you can close this PR and go ahead with your implementation if you have a clear idea :)

Yes sorting is something we shouldn't do and its useless in case of large directories - .. it is better for us to not do it or do it cleanly. 

Let me think about it.

harshavardhana:
> you can close this PR and go ahead with your implementation if you have a clear idea :)

I have made improvements over your patch and merged it. Going to think about how to get this working for --json output. 

Thanks for the initial work.

PR #1112, opened 2015-09-19
Fix printing fatal message prefix when json flag is enabled

vadmeste:
This patch fixes primarily the bug of printing fatal and error prefix when json flag is enabled.. (``` mc --json ls /tmp/non_existant_directory /usr/```)
but also it could help later to :
*) control mc exit codes since we can control fatal/error post hook callbacks
*) move out the logic of IsTesting/IsError/IsExit outside of the console package

### DIFF
--- a/pkg/console/console.go
+++ b/pkg/console/console.go
@@ -83,72 +83,66 @@ var (
 		return
 	}
 
-	// Fatal print a error message and exit
-	Fatal = func(data ...interface{}) {
-		consolePrint("Fatal", Theme["Fatal"], data...)
+	// fatalQuitRoutine called after fatal message printing
+	// by the user specific version
+	fatalQuitRoutine = func() {
 		if !IsTesting {
 			os.Exit(1)
 		}
 		defer func() {
 			IsExited = true
 		}()
+	}
+
+	// Fatal print a error message and exit
+	Fatal = func(data ...interface{}) {
+		consolePrint("Fatal", Theme["Fatal"], data...)
+		fatalQuitRoutine()
 		return
 	}
 
 	// Fatalf print a error message with a format specified and exit
 	Fatalf = func(format string, data ...interface{}) {
 		consolePrintf("Fatal", Theme["Fatal"], format, data...)
-		if !IsTesting {
-			os.Exit(1)
-		}
-		defer func() {
-			IsExited = true
-		}()
+		fatalQuitRoutine()
 		return
 	}
 
 	// Fatalln print a error message with a new line and exit
 	Fatalln = func(data ...interface{}) {
 		consolePrintln("Fatal", Theme["Fatal"], data...)
-		if !IsTesting {
-			os.Exit(1)
-		}
-		defer func() {
-			IsExited = true
-		}()
+		fatalQuitRoutine()
 		return
 	}
 
-	// Error prints a error message
-	Error = func(data ...interface{}) {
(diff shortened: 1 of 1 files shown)


harshavardhana:
Interesting fix, but there are some changes..

vadmeste:
I called that function callback because the developer who could use the console package later will be able to override the behaviour of fatal and error messages. So, this change is based on the will to separate mc and its testing logic from the console package itself

harshavardhana:
> I called that function callback because the developer who could use the console package later will be able to override the behaviour of fatal and error messages. So, this change is based on the will to separate mc and its testing logic from the console package itself

Yes that part is correct but it isn't rightfully called as callback function,  should only be called as such if it is called in a callback style. The testing part of console is provided solely from the point that it can be tested even outside of `mc`, when one wishes to use console as part of their package. 

The only way to override the function is by writing a new one, since this function doesn't take any arguments. So it generally defeats the purpose. 

It would be better in following manner where a user is supplying a callback function as an argument, which would make it generic. 

```
FatalWithCb(cb func(), data ...interface{}) {
         consolePrint(data)
         cb()
}

ErrorWithCb(cb func(), data ...interface{}) {
}
```

vadmeste:
Sure, I agree with you about the callback arguments, however, I thought about writing the following code in main() for example
``` 
func main() {
    console.FatalPostCallback = func(data ...interface{}) {
        os.Exit(1);
    }
    ...
    console.Fatal("message..")
}
```

harshavardhana:
> console.FatalPostCallback = func(data ...interface{}) {
>         os.Exit(1);
>     }

The code you see in console.Fatal() is only to test Fatal() function itself, moving it out makes it dangerous and some one can override the behavior of testing console.Fatal() unknowingly which would not provide the context of what it was meant to be used. 

Moving code in a function like this is equivalent of doing 

```
func A() {
   B()
}

func B() {
   os.Exit(1)
}
```

Now here the ``func B()`` is pointless, just to do something for ``A()`` without having its context in ``B()`` is wrong style. Unless ``B()`` provides more purpose to ``A()``

Extrapolating that to console.Fatal() the expectancy is that it should os.Exit(1) by default and one shouldn't be able to override this behavior only if its for testing purposes only and that is what the extra code does. That it simulates the console.Fatal() behavior.  So now you see console.Fatal() is not meant to have a generic purpose outside of printing and exiting itself. 

You could write a new generic Printer function which takes a callback function which can do something of custom printing and users can override that behavior.  But i do not see the real purpose of that, it is usually simple to not provide many ways of doing things. 

A console package purpose is printing and printing alone and simple stub code to test itself.

harshavardhana:
> *) move out the logic of IsTesting/IsError/IsExit outside of the console package

Okay after thinking over it - here is what i think you should implement for console package - there are definitely some improvements. 

But we don't need a callback function or a function which should override this behavior. 

```
 // Fatal print a error message and exit
        Fatal = func(data ...interface{}) {
                consolePrint("Fatal", Theme["Fatal"], data...)
                if !IsTesting {
                        os.Exit(1)
                }
                defer func() {
                        IsExited = true
                }()
                return
        }
```

The part of the code is repetitive, but it shouldn't be exposed outside in a way that its modifiable. 

```
              if !IsTesting {
                        os.Exit(1)
              }
              defer func() {
                        IsExited = true
              }()
              return
```

Can you squash this where you implement this as internal function?

PR #1168, opened 2015-10-08
mc share upload

krishnasrinivas:
1)
func doShareList() *probe.Error {
        sURLs, err := loadSharedURLsV2()
        if err != nil {
                return err.Trace()
        }
        for i, data := range sURLs.URLs {
                if time.Since(data.Date) > data.Message.Expiry {
                        sURLs.URLs = append(sURLs.URLs[:i], sURLs.URLs[i+1:]...)
                        continue
                }

doing append inside range was causing problem, fixed it

2)
Lot of types like ShareMessage should be unexported shareMessage we should fix it

3)
"share list" does not show full presigned download URL or full curl upload command, i think it's better to show that too?

4)
s3v2.go does not implement presigned post yet

5)
presigned put not implemented yet for s3v4 and s3v2

6)
content-range argument not implemented for mc

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -36,7 +36,8 @@ type Client interface {
 	SetBucketACL(acl string) *probe.Error
 
 	// Object operations
-	Share(expires time.Duration) (string, *probe.Error)
+	ShareDownload(expires time.Duration) (string, *probe.Error)
+	ShareUpload(bool, time.Duration, string) (map[string]string, *probe.Error)
 	GetObject(offset, length int64) (body io.ReadCloser, size int64, err *probe.Error)
 	PutObject(size int64, data io.Reader) *probe.Error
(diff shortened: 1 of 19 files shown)


harshavardhana:
> Lot of types like ShareMessage should be unexported shareMessage we should fix it

There is a reason to keep it that way,  golint and others do not complain about lack of documentation for such structs.  With "Message" it was kind of prudent to have a proper documentation at the top on what it is meant to be. 


> "share list" does not show full presigned download URL or full curl upload command, i think it's better to show that too?

That can be shown too shouldn't be a problem.

harshavardhana:
Oddly "Remove github.com/tchap/go-patricia/patricia from vendor.json" is part of the commit message, can you --amend "minio-go" vendor update commit message?

PR #1189, opened 2015-10-17
support for "rm" of objects and buckets

krishnasrinivas:


harshavardhana:
You should also write a test case for this as rm_test.go - so that we can get it tested with our windows build.

Since we are manipulating quite a lot of paths and trimming off separators we should do it in a platform neutral way.

harshavardhana:
Also there are some frivolous commits remove them..

PR #1193, opened 2015-10-21
Support for "mc rm s3/bucket/dir/... incomplete force"

krishnasrinivas:


harshavardhana:
Actually thinking about this further.. actually leave this for now. I will fix it properly.

Issue #1196, opened 2015-10-21
mc ls output error


krishnasrinivas:
krishna@escape:~/gocode/src/github.com/minio/mc$ ./mc ls awss3/krisupload/emacs/... awss3/krisupload/ema...
[2015-10-20 23:18:40 PDT]     0B emacs/fff
[2015-10-20 23:18:41 PDT] 1.8KiB emacs/site-start.d/00debian-vars.el
[2015-10-20 23:18:41 PDT]   729B emacs/site-start.d/50autoconf.el
[2015-10-20 23:18:41 PDT] 1.5KiB emacs/site-start.d/50dictionaries-common.el
[2015-10-20 23:18:41 PDT]   412B emacs/site-start.el
[2015-10-20 23:18:40 PDT]     0B ema/cs/fff
[2015-10-20 23:18:41 PDT] 1.8KiB ema/cs/site-start.d/00debian-vars.el
[2015-10-20 23:18:41 PDT]   729B ema/cs/site-start.d/50autoconf.el
[2015-10-20 23:18:41 PDT] 1.5KiB ema/cs/site-start.d/50dictionaries-common.el
[2015-10-20 23:18:41 PDT]   412B ema/cs/site-start.el

harshavardhana:
I am not able to reproduce the issue 

```
$ mc ls s3/miniocloud/ema... | wc -l
     712
```
```
$ mc ls s3/miniocloud/ema... s3/miniocloud/emacs... | wc -l
    1190
```

krishnasrinivas:
if you do "wc" you cant see the output, here it is:
here is the outout:
```
krishna@escape:~/gocode/src/github.com/minio/mc$ ./mc ls awss3/krisupload/ema... awss3/krisupload/ema...
[2015-10-20 23:18:40 PDT]     0B ema/cs/fff
[2015-10-20 23:18:41 PDT] 1.8KiB ema/cs/site-start.d/00debian-vars.el
[2015-10-20 23:18:41 PDT]   729B ema/cs/site-start.d/50autoconf.el
[2015-10-20 23:18:41 PDT] 1.5KiB ema/cs/site-start.d/50dictionaries-common.el
[2015-10-20 23:18:41 PDT]   412B ema/cs/site-start.el
[2015-10-20 23:18:40 PDT]     0B ema/cs/fff
[2015-10-20 23:18:41 PDT] 1.8KiB ema/cs/site-start.d/00debian-vars.el
[...]

PR #1197, opened 2015-10-21
recursive ls on object storage was giving incorrect output fix for #1196

krishnasrinivas:


harshavardhana:
Can you rebase this again?

harshavardhana:
There is a simpler fix which avoids doing url2Dir() and avoids the problems of "bucket" being appended to the output. 

```
if !strings.HasPrefix(contentCh.Content.Name, parentContent.Name) {
    contentCh.Content.Name = filepath.Join(parentContent.Name, strings.TrimPrefix(contentCh.Content.Name, parentContent.Name))
}
```

harshavardhana:
> if !strings.HasPrefix(contentCh.Content.Name, parentContent.Name) {
>     contentCh.Content.Name = filepath.Join(parentContent.Name, strings.TrimPrefix(contentCh.Content.Name, parentContent.Name))
> }

This is still incomplete, re-looked at multiple cases - this fix is valid. Just that we have to modify the rm- url construction to do filepath.Join() properly.

PR #1225, opened 2015-10-29
"rm" progress output

krishnasrinivas:
Note that we don't have complete list of files to be removed hence can't show progress bar i.e 0 to 100%
But we can show the file that is being removed using the "scanBar" it also shows the current file count.
I will rename scanBar to a more generic name.

### DIFF
--- a/rm-main.go
+++ b/rm-main.go
@@ -17,11 +17,15 @@
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


harshavardhana:
Rebase this patch.. LGTM

PR #1226, opened 2015-10-29
rename scanBar to a more generic updateBar as it is being used by "rm…

krishnasrinivas:
…" for updating current file being removed

### DIFF
--- a/cp-main.go
+++ b/cp-main.go
@@ -160,9 +160,9 @@ func doPrepareCopyURLs(session *sessionV2, trapCh <-chan bool) {
 	// Create a session data file to store the processed URLs.
 	dataFP := session.NewDataWriter()
 
-	var scanBar scanBarFunc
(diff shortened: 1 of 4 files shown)


krishnasrinivas:
i have put a comment in source "[X] ◓ randomstring" containing unicode character not sure how fine that is. go source code is unicode stream and all modern editors support unicodes. so thought it'll be cool to put a comment that way.

abperiasamy:
UpdateBar is not appropriate. ScanBar is similar to ProgressBar. It is used to display scanning (in case of storage, it is ftw or list) progress. UpdateBar means you are updating a progress bar. It is will confuse the reader thinking that it is an accessory function to progress bar.

krishnasrinivas:
@abperiasamy i am using scanBar to show the current deleted file, not current scanned file. Essentially I am updating it to the user not scanning anything.

PR #1227, opened 2015-10-30
mirror: refactor session creation

winchram:
Previously session creation was written in a way to work unsorted
source and targets content, but reality is source and targets give
contents in sorted order always.

This patch fixes session creation to use pre-sorted source and targets
content.


harshavardhana:
LGTM

PR #1337, opened 2015-11-16
progress summary should not be shown for cp/mirror when --json is set

krishnasrinivas:


harshavardhana:
Closing this - it should be fixed in a different way.

PR #1409, opened 2015-11-18
diff - continue on non-fatal error. refactor to remove unneeded goroutine

krishnasrinivas:


harshavardhana:
> urls for trace is not needed as the user already knows it. also we are not returning it to another function so storing this context in trace is not useful, what do you think?

The reasoning is that we add whenever we have a string which can provide more info, we can say that because in the call trace one can observe if URL was ever mangled internally.  The reasoning to provide .Trace() was to provide those arguments where it actually failed. 

It is not consistent everywhere but it is slowly being fixed, but where-ever there is a string URL we should add it.

PR #1444, opened 2015-11-19
new recursive cp and session

abperiasamy:
* New session v5 with support for globals and some fixes.
* Cp url processing is rewritten. 
* Cp now supports -r instead of ...

I am also thinking about type copyURLsTypeE (cp -r [](f|d) d). That will be next.


harshavardhana:
Fixes #1377

PR #1447, opened 2015-11-19
check bucketname validity

krishnasrinivas:


harshavardhana:
Let's rework this patch what is required is not just throwing a 'InvalidBucketName' error. Here is what is required for s3.go to do 

Different types of errors.

- BucketNames cannot be less than '3' characters. Please read <link_to_s3>
- BucketNames cannot be more than '63' characters. Please read <link_to_s3>
- BucketNames cannot have special characters. Please read <link_to_s3>

Even after this 's3' sends and error we show that as well by looking at 'minio.ErrorResponse' - 'InvalidBucketName' and reply properly with providing the 'BucketName'

PR #1497, opened 2015-11-30
check for pipe args

abperiasamy:


abperiasamy:
Fixes crash in mc pipe when no args are passed.

PR #1533, opened 2015-12-03
Supporting Alpine Linux

arschles:
I tried to build `mc` for Alpine Linux (inside an `alpine:3.1` docker container). As [these instructions](http://dominik.honnef.co/posts/2015/06/statically_compiled_go_programs__always__even_with_cgo__using_musl/) say, musl doesn't support the [`os/user` package](http://godoc.org/os/user) yet. As this program uses `os/user` to get the home directory (to get the config file), Alpine builds fail at runtime, even when `-C` is passed via the CLI.

Right now, this PR removes the entire dependency on `os/user`, but the final version could use [build constraints](https://golang.org/pkg/go/build/#hdr-Build_Constraints) to toggle usage of `os/user`. Or perhaps there's an easier way to get this working that I haven't thought of. Thoughts?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,10 @@ pkg-remove:
 
 install: gomake-all
 
+all-tests: test
+	# TODO disable them for now.
+	#@./tests/test-minio.sh
+
 release:
 	@./release.sh
 
--- a/tests/.gitignore
+++ b/tests/.gitignore
@@ -1,0 +1,24 @@
+# Compiled Object files, Static and Dynamic libs (Shared Objects)
+*.o
+*.a
+*.so
+
+# Folders
+_obj
+_test
+
+# Architecture specific extensions/prefixes
+*.[568vq]
+[568vq].out
+
+*.cgo1.go
+*.cgo2.c
+_cgo_defun.c
+_cgo_gotypes.go
+_cgo_export.*
+
+_testmain.go
+
+*.exe
+*.test
+*.prof
--- a/tests/LICENSE
+++ b/tests/LICENSE
@@ -1,0 +1,202 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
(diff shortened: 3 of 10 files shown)


harshavardhana:
The right fix is to use - https://github.com/minio/minio/blob/8046b24b975a26d659bda12797a30631bb427e8f/utils.go#L63 - which we did with our minio container.

Can you bring this change in this patch?

harshavardhana:
@arschles please incorporate this diff into 'mc'

```
-// user.Current is not implemented on 32bit, falling back and using a workaround instead.
+// user.Current is not implemented on 32bit and statically compiled binaries.
+// Following code provides valid workaround for each of these situations.
 func userCurrent() (*user.User, *probe.Error) {
+       // Set env variable to indicate 'mc' is being run inside DOCKERIMAGE.
+       if os.Getenv("DOCKERIMAGE") == "1" {
+               wd, err := os.Getwd()
+               if err != nil {
+                       return nil, probe.NewError(err)
+               }
+               return &user.User{Uid: "0", Gid: "0", Username: "root", Name: "root", HomeDir: wd}, nil
+       }
        // Remove this check if golang fixes their code to support 32bit properly for user.Current.
        if runtime.GOARCH == "386" && runtime.GOOS == "linux" {
                return &user.User{
```

arschles:
@harshavardhana thanks for the direction.

I ported the [code that you suggested](https://github.com/minio/minio/blob/8046b24b975a26d659bda12797a30631bb427e8f/utils.go#L63) and it looks like the code already has the pertinent changes in the diff you pasted.

I've also merged from `master`. Can you please let me know if you prefer I rebase?

harshavardhana:
@arschles  - can you squash these commits into single change?

harshavardhana:
Thanks for your work.

arschles:
@harshavardhana I wasn't paying attention to what I was doing and messed up the squash. I'll open a new PR for this change

PR #1539, opened 2015-12-09
porting userCurrent from the minio project

arschles:
I accidentally closed https://github.com/minio/mc/pull/1533, so this replaces it with a single commit.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -47,9 +47,19 @@ func isBucketVirtualStyle(host string) bool {
 	return s3Virtual || googleVirtual
 }
 
-// user.Current is not implemented on 32bit, falling back and using a workaround instead.
+// Workaround for docker images with fully static binary and 32bit linux operating systems.
+// For static binaries NSS library will not be a part of the static binary hence user.Current() fails.
+// For 32bit linux CGO is not enabled so it will not provide linux specific codebase.
+// This code is taken from the minio project:
+// (https://github.com/minio/minio/blob/8046b24b975a26d659bda12797a30631bb427e8f/utils.go#L60-L85)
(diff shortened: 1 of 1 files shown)


arschles:
cc/ @harshavardhana

harshavardhana:
This looks good to me. :+1:

arschles:
@harshavardhana this doesn't look to be in `master` at https://github.com/minio/mc/blob/master/utils.go. did it get moved in a more recent commit?

harshavardhana:
> @harshavardhana this doesn't look to be in master at https://github.com/minio/mc/blob/master/utils.go. did it get moved in a more recent commit?

Yes i created a comprehensive package pkg/user which 'mc' inherits now. and uses 'user.Current()' - https://github.com/minio/minio/tree/master/pkg/user

arschles:
got it, thanks!

harshavardhana:
> got it, thanks!

Your welcome!

arschles:
@harshavardhana just an FYI, [three](https://github.com/deis/builder) [deis v2](https://github.com/deis/slugbuilder) [components](https://github.com/deis/minio) will be using `mc`, all in docker containers with `DOCKERIMAGE=1` set. In my testing it's been working very well.

I will be investigating whether it would be useful to run the minio server with that env variable set as well.

harshavardhana:
> @harshavardhana just an FYI, three deis v2 components will be using mc, all in docker containers with DOCKERIMAGE=1 set. In my testing it's been working very well.
> 

That is great news, happy to help anytime if you face any issues - you can find us on https://gitter.im/minio/minio

> I will be investigating whether it would be useful to run the minio server with that env variable set as well.

You could use our official docker container.

PR #1540, opened 2015-12-09
pkg/user: Bring user package from server and use it here in place of …

harshavardhana:
…userCurrent.


harshavardhana:
ping @krishnasrinivas - review and merge.

PR #1555, opened 2015-12-28
new alias style and numerous cleanups.

abperiasamy:
This patch implements Jitesh's idea (support for multiple accounts aws accounts). User can now add multiple accounts belonging to same domain.
* New config format v7 (all hosts have alias by default)
* New session format v6 (data files also stores alias)
* Rewrite config migration code.


harshavardhana:
LGTM.

PR #1567, opened 2016-01-08
session: Add new flags to session and properly restore old flags if any.

harshavardhana:


abperiasamy:
Session resume should retain the original state. We should not let the user introduce new flags in the middle of execution (session resume). Session resume is like checkpoint restart.

PR #1568, opened 2016-01-08
mirror/cp: Remove aliases from being printed in json output. - fixes #1563

### LINKED ISSUE #1563: Improve json output remove unnecessary fields

harshavardhana:


abperiasamy:
Length should be added to JSON since it is not part of the console message. Alternatively, cp and mirror should not report length as well. They should also output original aliased URL instead of parsed alias and  HTTP URL.

harshavardhana:
> Length should be added to JSON since it is not part of the console message. Alternatively, cp and mirror should not report length as well. They should also output original aliased URL instead of parsed alias and HTTP URL.

Done

PR #1571, opened 2016-01-09
Fix ls error msg

vadmeste:
Not documented, but according to this https://github.com/golang/go/blob/master/src/path/filepath/symlink.go#L100, EvalSymlinks always returns an empty string if there is an error.

This patch will fix this minor ls error msg
$ mc ls /path_not_found
mc: <ERROR> Unable to stat ‘/path_not_found’. Requested file `` not found

So it will be shown correctly
$ mc ls /path_not_found
mc: <ERROR> Unable to stat ‘/path_not_found’. Requested file `/path_not_found` not found

Not sure about how this patch affects this line f.handleWindowsSymlinks(f.PathURL.Path), but I am pretty sure that it wasn't working at all before now


harshavardhana:
LGTM

PR #1595, opened 2016-01-28
url: Error out if the aliased URL has host entry.

harshavardhana:
```
mc ls http://localhost:9000/
```

ends up listing `mc ls /`, This is confusing and should be
treated as an error for filesystem.

### DIFF
--- a/common-methods.go
+++ b/common-methods.go
@@ -19,6 +19,7 @@ package main
 import (
 	"io"
 	"os"
+	"regexp"
 	"runtime"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


abperiasamy:
We dont support URL at all. It should always be alias.

harshavardhana:
> We dont support URL at all. It should always be alias. when a URL is passed, we should display Invalid argument or appropriate useful error message.

You are saying we should fail before even going down till FS?

krishnasrinivas:
Probably it is a better idea to display the message:
```
Use `mc config host add` to add an alias to the host and use the alias for S3 operations
```

harshavardhana:
> Use `mc config host add` to add an alias to the host and use the alias for S3 operations

Detecting early is better.. let me do a simple parsing and then fail.

krishnasrinivas:
looks good.

pkg/client/fs/fs.go
```diff
@@ -45,8 +47,12 @@ func New(path string) (client.Client, *probe.Error) {
 	if strings.TrimSpace(path) == "" {
 		return nil, probe.NewError(client.EmptyPath{})
 	}
+	newPathURL := client.NewURL(normalizePath(path))
+	if newPathURL.Host != "" {
+		return nil, probe.NewError(client.InvalidFSPath{path})
```
abperiasamy: It is better for client.NewURL to return newURL and error.

url.go
```diff
@@ -51,9 +51,25 @@ func url2Stat(urlStr string) (client client.Client, content *client.Content, err
...
+		return true
+	}
+	return false
+}
+
 // url2Alias separates alias and path from the URL. Aliased URL is of
 // the form [/]alias/path/to/blah.
 func url2Alias(aliasedURL string) (alias, path string) {
+	// If incoming aliased url is of type object, error out quickly.
+	if isAliasedURLObject(aliasedURL) {
```
abperiasamy: It is not the right fix.

common-methods.go
```diff
@@ -131,11 +132,19 @@ func newClientFromAlias(alias string, urlStr string) (client.Client, *probe.Erro
 	return s3Client, nil
 }
 
+// isURL - verify if aliased url is real URL.
+var isURL = regexp.MustCompile("^https?://")
+
```
abperiasamy: isURL -> urlRgx or urlRegex

PR #1622, opened 2016-02-24
Retrieve secretKey from the environment first

### LINKED ISSUE #1607: Allow reading AWS credentials from ENV
> As a system administrator managing multiple AWS accounts with certain security requirements I need to have the tools I'm using to be able to access those accounts without pasting the credentials into unencrypted plaintext files on my harddisk.

I'm using a tool to set my environment variables to access those different AWS accounts from an encrypted storage after having to enter a master password. Sure I could add several `s3+accountname` entries to the `~/.mc/config.json` file but that would lead to a plain text file having all the credentials in them. Though they are of limited access I would need to suspect every program / script on my system is able to retrieve them.

If `mc` would retrieve the credentials from ENV I only need to trust the tool storing/setting the ENV variables and `mc`. No other program has access to them as they are stored in the current shells environment.

## Expected

If left out from the configuration file `mc` tries to read the credentials from default AWS environment variables and the actions are executed on their behalf.

## Observed

```bash
# awsenv run private mc ls s3
mc: <ERROR> Unable to stat ‘s3’. XML syntax error on line 164: invalid character entity &ntilde;
INFO[0001] Unclean exit with exit-code != 0

# awsenv run demo env | grep AWS
AWS_ACCESS_KEY_ID=demo-key
AWS_SECRET_ACCESS_KEY=demo-secret
AWS_ACCESS_KEY=demo-key
AWS_SECRET_KEY=demo-secret
AWS_REGION=us-east-1
AWS_DEFAULT_REGION=us-east-1
```

vadmeste:
For users who are not comfortable with storing secret keys in the disk, this is another way to pass credentials:

```
$ export MC_SECRET_${ACCESS_KEY}=${SECRET_KEY}
$ mc ls ..
```

Fixes #1607

### DIFF
--- a/common-methods.go
+++ b/common-methods.go
@@ -23,6 +23,7 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/minio/mc/pkg/console"
 	"github.com/minio/minio/pkg/probe"
 )
 
@@ -113,8 +114,32 @@ func newClientFromAlias(alias string, urlStr string) (Client, *probe.Error) {
 	// We have a valid alias and hostConfig. We populate the
 	// credentials from the match found in the config file.
 	s3Config := new(Config)
-	s3Config.AccessKey = hostCfg.AccessKey
-	s3Config.SecretKey = hostCfg.SecretKey
+
+	// secretKey retrieved from the environement overrides the one
+	// present in the config file
+	keysPairEnv := os.Getenv("MC_SECRET_" + alias)
+	keysPairArray := strings.Split(keysPairEnv, ":")
+	var accessKeyEnv, secretKeyEnv string
+	if len(keysPairArray) >= 1 {
+		accessKeyEnv = keysPairArray[0]
+	}
+	if len(keysPairArray) >= 2 {
+		secretKeyEnv = keysPairArray[1]
+	}
+	if len(keysPairEnv) > 0 &&
+		isValidAccessKey(accessKeyEnv) && isValidSecretKey(secretKeyEnv) {
+		s3Config.AccessKey = accessKeyEnv
+		s3Config.SecretKey = secretKeyEnv
+	} else {
+		if len(keysPairEnv) > 0 {
+			console.Errorln("Access/Secret keys associated to `" + alias + "' " +
+				"are found in your environment but not suitable for use. " +
+				"Falling back to the standard config.")
+		}
+		s3Config.AccessKey = hostCfg.AccessKey
+		s3Config.SecretKey = hostCfg.SecretKey
+	}
+
 	s3Config.Signature = hostCfg.API
 	s3Config.AppName = "mc"
 	s3Config.AppVersion = mcVersion


harshavardhana:
> $ export MC_SECRET_${ACCESS_KEY}=${SECRET_KEY}
> $ mc ls ..

Sorry for being late here, but the problem here is alias names. We have to cater for that since an access key and secret key are not global. 

Any 'mc' operation like cp, mirror can happen across different s3 accounts with different access and secret key. 

I don't see an easier way to incorporate them easily, do you have any other way in mind?

vadmeste:
> Sorry for being late here,

Really no problem,

> Any 'mc' operation like cp, mirror can happen across different s3 accounts with different access and secret key.

Not sure if I did get your point, so I will show you a better example of the target scenario:
```
$ export MC_SECRET_1AZ98AACAERZ="A/33AB"
$ export MC_SECRET_9BF00DC0CGE="XXXXX"
$ mc cp alias1/dir1/file1.txt alias2/dir2/file2.txt
```

won't that simply work as expected?

harshavardhana:
> Not sure if I did get your point, so I will show you a better example of the target scenario:
> 
> $ export MC_SECRET_1AZ98AACAERZ="A/33AB"
> $ export MC_SECRET_9BF00DC0CGE="XXXXX"
> $ mc cp alias1/dir1/file1.txt alias2/dir2/file2.txt
> won't that simply work as expected?

No it won't alias is associated in the file whereas this secret/access is outside of it. So the code change is necessary to cleanly handle this together with aliases.

It is little harder to achieve.

vadmeste:
> No it won't

I've just tested with two instances of minio and it worked, so I guess that there is another use case that doesn't work or did you just mean that this pull request is bad designed? 

Here, we are just overriding the secret (which could be specified or not in ~/.mc/config.json), it could be not perfect but I can't see any other way to do that

common-methods.go
```diff
@@ -114,7 +114,15 @@ func newClientFromAlias(alias string, urlStr string) (Client, *probe.Error) {
 	// credentials from the match found in the config file.
 	s3Config := new(Config)
 	s3Config.AccessKey = hostCfg.AccessKey
-	s3Config.SecretKey = hostCfg.SecretKey
+
+	// secretKey retrieved from the environement overrides the one
+	// present in the config file
+	if secretKeyEnv := os.Getenv("MC_SECRET_" + s3Config.AccessKey); isValidSecretKey(secretKeyEnv) {
```
harshavardhana: We can perhaps throw a warning message, if 'isValidSecretKey' fails, we are falling back to the config file.

vadmeste: yes, you are right, I'll do it..

vadmeste:
There was a bug because I thought that isValidSecretKey doesn't accept empty strings, now it is fixed. However, I don't think it is a good idea to consider an empty string as a secret key, it would be better to indicate the non presence of a secret key with another way more suitable to the mind.

common-methods.go
```diff
@@ -114,7 +115,20 @@ func newClientFromAlias(alias string, urlStr string) (Client, *probe.Error) {
 	// credentials from the match found in the config file.
 	s3Config := new(Config)
 	s3Config.AccessKey = hostCfg.AccessKey
-	s3Config.SecretKey = hostCfg.SecretKey
+
+	// secretKey retrieved from the environement overrides the one
+	// present in the config file
+	secretKeyEnv := os.Getenv("MC_SECRET_" + s3Config.AccessKey)
```
abperiasamy: We can get both secret key and access key from the env. Can you please name them as 
MC_SECRETKEY & MC_ACCESSKEY

harshavardhana: > We can get both secret key and access key from the env. Can you please name them as 
> MC_SECRETKEY & MC_ACCESSKEY

The reason to use MC_SECRET_<ACESSKEY_ID> = SECRET
so that being able to support multiple access keys using just MC_SECRETKEY and MC_ACCESSKEY will not be able to provide such differentiation.

abperiasamy:
If we use host alias as key, we can support fetching both access and secret keys from the env. 
> MC_SECRET_$ALIAS = $ACCESSKEY:$SECRETKEY

common-methods.go
```diff
@@ -113,8 +114,32 @@ func newClientFromAlias(alias string, urlStr string) (Client, *probe.Error) {
 	// We have a valid alias and hostConfig. We populate the
 	// credentials from the match found in the config file.
 	s3Config := new(Config)
-	s3Config.AccessKey = hostCfg.AccessKey
-	s3Config.SecretKey = hostCfg.SecretKey
+
+	// secretKey retrieved from the environement overrides the one
+	// present in the config file
+	KeysPairEnv := os.Getenv("MC_SECRET_" + alias)
```
harshavardhana: These variable names should be lower case.

vadmeste: Yes, I didn't pay attention, you mean only the first character, right? 
By the way, gofmt didn't complain at all, do you you have any tool that detect that ?

harshavardhana: > Yes, I didn't pay attention, you mean only the first character, right? 
> By the way, gofmt didn't complain at all, do you you have any tool that detect that ?

It doesn't detect everything, yes only the first character.

vadmeste: Should be fine

harshavardhana:
LGTM.

Issue #1530, opened 2015-12-03
Add tests for 'mc cat/pipe' command.


harshavardhana:
- mc cat should read from stdin. 
- mc pipe should behave like 'mc cat' with no argument.
- mc cat | mc pipe should work properly. 
- mc cat fails for non existant file. 
- mc pipe doesn't have any failure condition other than network related errors.

PR #1650, opened 2016-03-27
tests: add tests for 'mc cat' and 'mc config host'

sreeram-boyapati:
@harshavardhana I have written tests for 'mc cat' and 'mc pipe'. In test-minio.sh, it validates using test_script's stdout.
How do I integrate these to test-minio.sh?

I have not written tests for the 5th condition in #1530. Needs more clarity on what cases it should succeed except network failure.

### DIFF
--- a/tests/tests/cat-test.sh
+++ b/tests/tests/cat-test.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+
+# Testing mc cat
+# 1. Pass a variable as stdin
+# 2. Assert result returned by mc cat is the same
+# 3. Repeat the same for mc pipe with no arguments
+# 4. Take a random string to test for non existant files
+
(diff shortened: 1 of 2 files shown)


harshavardhana:
> I have not written tests for the 5th condition in #1530. Needs more clarity on what cases it should succeed except network failure.

mc cat and pipe do network calls,  config related do not have network related operation.  For network calls we need to have a sample data set which can be used on a downloaded minio server.

harshavardhana:
Thanks a lot on working on this and spending time @sreeram-boyapati

harshavardhana:
Will merge this @sreeram-boyapati - there is no reason to wire this up with test-minio.sh, we need to do lot more work to get this integrated well. 

Thanks for your contribution.

PR #1653, opened 2016-03-30
ListObjects on a prefix with 10 million entries takes around 8 minutes.

krishnasrinivas:


harshavardhana:
Can you squash the commits?

PR #1657, opened 2016-04-01
realpath() alternative implementation doesn't really work under linux and freebsd, use a simpler and more portable method

vadmeste:
To check under Darwin/OSX before merging


harshavardhana:
LGTM

PR #1662, opened 2016-04-02
Make --json and --quiet work as expected with mc mirror cmd

vadmeste:


mirror-main.go
```diff
@@ -174,8 +180,8 @@ func doMirror(sURLs mirrorURLs, progressReader *progressBar, accountingReader *a
 	return sURLs
 }
 
-// doMirrorFake - Perform a fake mirror to update the progress bar appropriately.
-func doMirrorFake(sURLs mirrorURLs, progressReader *progressBar) mirrorURLs {
+// doMirrorFakeProgressUpdate - Update the progress bar to represent a fake mirror
+func doMirrorFakeProgressUpdate(sURLs mirrorURLs, progressReader *progressBar) mirrorURLs {
```
harshavardhana: During 'fake' it means that we are not updating the progress bar, we should update rather than returning dummy.

vadmeste: umm, is this function name is confusing? I preferred to change it because doMirror() has isFake parameter now so it supports fake mirroring. doMirrorFake would imply the same functionality that's why I decided to change its name

harshavardhana: > umm, is this function name confusing? I preferred to change it because doMirror() has isFake parameter now so it supports fake mirroring. doMirrorFake would imply the same functionality that's why I decided to change its name

No if sURLs returns for 'isFake' it wouldn't update the progress bar.

vadmeste:
I am going to merge doMirror and doMirrorFake. I thought about separating them with a little code duplication but I think we don't need separate code paths anytime soon, any comments?

harshavardhana:
> I am going to merge doMirror and doMirrorFake. I thought about separating them with a little code duplication but I think we don't need separate code paths anytime soon, any comments?

You can merge them together.

mirror-main.go
```diff
@@ -330,15 +331,9 @@ func doMirrorSession(session *sessionV6) {
 		json.Unmarshal([]byte(urlScanner.Text()), &sURLs)
 		// Verify if previously copied, notify progress bar.
 		if isCopied(sURLs.SourceContent.URL.String()) {
-			statusCh <- doMirrorFake(sURLs, progressReader)
```
harshavardhana: Lets simplify this further like this 

```
if isCopied(sURLs.SourceContent.URL.String()) {
            isFake = true
}
doMirror(sURLs, progressReader, accntReader, isFake)
```

vadmeste: done,

by the way,  progressReader.ProgressBar.Add64(sURLs.SourceContent.Size) is fast, it takes 10 seconds for 100 millions entries in my little old system (i5, from 2011)

harshavardhana:
> by the way, progressReader.ProgressBar.Add64(sURLs.SourceContent.Size) is fast, it takes 10 seconds for 100 millions entries in my little old system (i5, from 2011)

Yeah its a dummy arithmetic operation, it will be fast.

Issue #1621, opened 2016-02-24
minio client config host add <alias> not working properly


kmala:
minio client config host add <alias> is not working properly.It is properly updating the config.json but other client commands aren't working.
Temporary fix: Delete the ~/.mc folder and try adding again

harshavardhana:
> minio client config host add is not working properly.It is properly updating the config.json but other client commands aren't working.
> Temporary fix: Delete the ~/.mc folder and try adding again

Easy to reproduce as mentioned related to 'loadMcConfig()' and caching.

sreeram-boyapati:
I would like to take this up, looks like an easy one to fix. When I wrote tests for mc config host. I was able to list, add and remove hosts. Is the issue still open?

harshavardhana:
> I would like to take this up, looks like an easy one to fix. When I wrote tests for mc config host. I was able to list, add and remove hosts. Is the issue still open?

It might not be fixed just yet, it might be still reproducible if you have a custom folder. Basically if you have a "~/.mc" directory then "--config-dir" option adds the host but fails to list or complains about alias being missing. 

To reproduce you should already have a ".mc" directory.

sreeram-boyapati:
Will try to fix it. The Issue is still reproducible.

harshavardhana:
> Will try to fix it. The Issue is still reproducible.

It is a little tricky! , ping us if you have more questions @sreeram-boyapati :-)

harshavardhana:
> Will try to fix it. The Issue is still reproducible.

Thanks a lot! :cake:

sreeram-boyapati:
@harshavardhana 
```go
func getMcConfigPath() (string, *probe.Error) {
	if mcCustomConfigDir == "" {
		fmt.Println("Custom Config is NUll")
	}
	if mcCustomConfigDir != "" {
		return filepath.Join(mcCustomConfigDir, globalMCConfigFile), nil
	}
	dir, err := getMcConfigDir()
	if err != nil {
		return "", err.Trace()
	}
	return filepath.Join(dir, globalMCConfigFile), nil
}
```
mcCustomConfigDir is always empty

harshavardhana:
> func getMcConfigPath() (string, *probe.Error) {
>     if mcCustomConfigDir == "" {
>         fmt.Println("Custom Config is NUll")
>     }
>     if mcCustomConfigDir != "" {
>         return filepath.Join(mcCustomConfigDir, globalMCConfigFile), nil
>     }
[...]

PR #1666, opened 2016-04-05
config: (tmp fix) move loadConfigV8() inside closure to solve issue #1621

sreeram-boyapati:
-Move loadConfigV8() inside closure. This should solve the issue for now.
I was not able to detect the origin of the bug.

This solved the issue.
I guess quick.Load saves new config if earlier config does not exist use config path as key.
So repeated calls to loadConfigV8 will fetch from the cache. I am not completely sure, though.

### DIFF
--- a/config.go
+++ b/config.go
@@ -111,7 +111,8 @@ func loadMcConfigFactory() func() (*configV8, *probe.Error) {
 	}
 }
 
-var loadMcConfig = loadMcConfigFactory()
+// loadMcConfig - returns configuration, initialized later.
+var loadMcConfig func() (*configV8, *probe.Error)
 
 // saveMcConfig - saves configuration file and returns error if any.
 func saveMcConfig(config *configV8) *probe.Error {


sreeram-boyapati:
Yes, There is no reason to have the closure. But still caches in cacheCfgV8. you need no fetch from key value store again and again.
closure is caching wrong values. 
I called loadMcConfig after app.RunAndExitonError() . It still returned the incorrect config. 

But may I ask why are you caching it? "minio/pkg/quick" Key-Value store is fast enough right?

If this is completely wrong, I will close it. :)

harshavardhana:
> Yes, There is no reason to have the closure. But still caches in cacheCfgV8. you need no fetch from key value store again and again.
> closure is caching wrong values. 
> I called loadMcConfig after app.RunAndExitonError() . It still returned the incorrect config.
> 
> But may I ask why are you caching it? "minio/pkg/quick" Key-Value store is fast enough right?
> 
> If this is completely wrong, I will close it. :)

Caching is perhaps optimization in terms of loading config, since in 'mc' each operation loads config.  Your patch is alright it fixes an important bug, it is just that i want to see if there is a better way and keep the cache.

vadmeste:
@sreeram-boyapati, actually, if you disable setting loadMcConfig variable in https://github.com/minio/mc/blob/master/config.go#L114 it will work fine.. The problem is that go is setting this variable before parsing mc command arguments, so when you run mc, the default mc config path will be configured and config will be cached.

harshavardhana:
> @sreeram-boyapati, actually, if you disable setting loadMcConfig variable in https://github.com/minio/mc/blob/master/config.go#L114 it will work fine.. The problem is that go is setting this variable before parsing mc command arguments, so when you run mc, the default mc config path will be configured and config will be cached.

This makes more sense @sreeram-boyapati can you do that?

sreeram-boyapati:
@harshavardhana  Yes, I can do that.

PR #1667, opened 2016-04-05
Rename --quiet flag to --no-progressbar

vadmeste:
*) '--quiet' claims that it is generating less chatting messages but it doesn't, 'mc ls' can fill many pages of my screen
*) '--quiet is little vague, the user cannot speculate its effect on the command
*) --no-progressbar is much more specific and the user could easily know its effect

### DIFF
--- a/session-v7.go
+++ b/session-v7.go
@@ -32,14 +32,14 @@ import (
 	"github.com/minio/minio/pkg/quick"
 )
 
-func migrateSessionV5ToV6() {
+func migrateSessionV6ToV7() {
 	for _, sid := range getSessionIDs() {
-		sessionV6, err := loadSessionV6(sid)
-		fatalIf(err.Trace(sid), "Unable to load version ‘6’. Migration failed please report this issue at https://github.com/minio/mc/issues.")
-		if sessionV6.Header.Version == "6" { // It is new format.
+		sessionV7, err := loadSessionV7(sid)
+		fatalIf(err.Trace(sid), "Unable to load version ‘7’. Migration failed please report this issue at https://github.com/minio/mc/issues.")
+		if sessionV7.Header.Version == "7" { // It is new format.
 			return
 		}
-		/*** Remove all session files older than v6 ***/
+		/*** Remove all session files older than v7 ***/
 
 		sessionFile, err := getSessionFile(sid)
 		fatalIf(err.Trace(sid), "Unable to get session file.")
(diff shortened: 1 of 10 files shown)


harshavardhana:
> vadmeste commented 2 hours ago
> *) '--quiet' claims that it is generating less chatting messages but it doesn't, 'mc ls' can fill many pages of my screen
> *) '--quiet is little vague, the user cannot speculate its effect on the command
> *) --no-progressbar is much more specific and the user could easily know its effect

The problem here is 

```
mc ls --no-progressbar
``` 

Has no real meaning, in that sense '--quiet' is a better choice since it's more generic.

vadmeste:
yes, I understand that but I think that the user won't invoke this flag since he already knows that it does nothing in the case of ls, at the same time --quiet is not quiet at all with ls..

harshavardhana:
> yes, I understand that but I think that the user won't invoke this flag since he already knows that it does nothing in the case of ls, at the same time --quiet is not quiet at all with ls..

Then it shouldn't be a global option, '--quiet' should be moved into only '--no-progressbar'  for cp and mirror.

session-v7.go
```diff
@@ -32,14 +32,14 @@ import (
 	"github.com/minio/minio/pkg/quick"
 )
 
-func migrateSessionV5ToV6() {
+func migrateSessionV6ToV7() {
```
harshavardhana: Another thing is we cannot just do migrate and forget older v6 versions, we need to add migration for v5-->v6 and then v6-->v7 - just like how we do for config files.

abperiasamy:
Reason to name it as --quiet is keep it generic. It would converge multiple related options into one, such as --no-progress-bar.

session-v7.go
vadmeste: yes, I saw you did that with config migration, however in session, migration does nothing more than deleting old session files, this code  will clean all versions inferior to 7

harshavardhana: > yes, I saw you did that with config migration, however in session, migration does nothing more than deleting old session files, this code will clean all versions inferior to 7

It shouldn't be it was only for V5 --> V6, but V6 onwards it needs to migrate.

vadmeste: but I know that in general, sequential migration is a good idea to avoid bugs and hidden problems

vadmeste: oh okay, are you saying that i should replace quiet with no progress bar ?

vadmeste: I mean in session migration

vadmeste:
yes but still, mc --quiet mirror is really chatty for the current mc.

What about doing the thing in reverse ? Making mc quiet by default but it uses the progress bar, and if the user wants to see more details, he can invoke  --chatty for example, then we will not need --quiet or --no-progressbar

Issue #1543, opened 2015-12-11
mc config syntax verification should be added. 


harshavardhana:
Verify config syntax validating, aliasnames, host config, endpoint.

sreeram-boyapati:
Hey I would like to work on this. Can you add more details with an example?

harshavardhana:
> Hey I would like to work on this. Can you add more details with an example?

What this needs is validating config file, if its intact. Currently it is done via 'checkConfig()'  - i.e it just validates if the config file can be read or not, what we need to validate is if the config entries are intact and have valid format for example 

To start by validating if version is correct.

```
	"version": "8",
	"hosts": {
```

Then next is to validate hosts for example 

Validating if 'aliases' were hand edited and contains unsupported characters,
```
    "church": {
			"url": "https://churchofminio.com",
			"accessKey": "W891COK7VCTZA55PY0PN",
			"secretKey": "ieStcv8GvbeqGdQrIdDutMicbymOuMYvj3jdlIyA",
			"api": "S3v4"
		},
``` 

Next is validating individual fields for example 

- 'api' can only have 'S3v2' or 'S3v4' and should always be non-empty.  
- 'url' should be of a proper URL format and should always be non-empty. 

All this is to validate existing config values and reporting to user if there are problems, exit 'mc' and giving valid suggestions to fix them accordingly.

harshavardhana:
Thanks again @sreeram-boyapati for taking a look.

sreeram-boyapati:
Ah! I get to write more LoC. cool on it.

PR #1676, opened 2016-04-10
config: validate config.json file

sreeram-boyapati:
Minio Client validates only config host add parameters
    Not the config file. Use the same utils to validate config file|
#1543

### DIFF
--- a/config-validate.go
+++ b/config-validate.go
@@ -1,0 +1,78 @@
+/*
+ * Minio Client (C) 2015 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


harshavardhana:
Can you also add how the errors would look like here in the commit logs?

config-validate.go
```diff
@@ -0,0 +1,78 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import (
+	"bytes"
+	"container/list"
```
harshavardhana: It is better to use a simple slice of error messages here, container/list is an overkill, even the order and popping can be done easily. 

https://github.com/golang/go/wiki/SliceTricks

sreeram-boyapati:
I will add how the error messages will look like and slices instead of container/list. But they do not stop the execution if validation fails. Is it ok?

harshavardhana:
> I will add how the error messages will look like and slices instead of container/list. But they do not stop the execution if validation fails. Is it ok?

They should stop, we can error out properly.

PR #1679, opened 2016-04-10
url: There is no reason to split '?' delimiter.

### LINKED ISSUE #1677: mc cannot treat file system files whose names contain odd but legitimate characters like '?'
you can find below how to reproduce the problem:

```
$ touch /tmp/foo\?.txt
$ mc cp /tmp/foo\?.txt /tmp/foo.backup
mc: <ERROR> Unable to guess the type of copy operation. Invalid arguments provided, cannot proceed.
```

harshavardhana:
The reasoning is that we are not providing query
params through 'mc' and we don't need to handle it
specially. Since all paths are absolute with only
delimiter of '/' all other characters should be allowed
including '?'.

Fixes #1677

### DIFF
--- a/client-fs.go
+++ b/client-fs.go
@@ -55,7 +55,7 @@ func fsNew(path string) (Client, *probe.Error) {
 		return nil, probe.NewError(EmptyPath{})
 	}
 	return &fsClient{
-		PathURL: newURL(normalizePath(path)),
+		PathURL: newClientURL(normalizePath(path)),
 	}, nil
 }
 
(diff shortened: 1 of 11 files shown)


vadmeste:
I suggest to replace newURL() with newRawURL()

harshavardhana:
> I suggest to replace newURL() with newRawURL()

Unless we have newURL() also, newRawURL() doesn't make much sense.

harshavardhana:
> I suggest to replace newURL() with newRawURL()

Also since internally it does split based on `/` , rawURL brings a new meaning altogether - we are not trying replace this URL with `*url.URL`

vadmeste:
yes.. maybe newClientURL() ? the reason is that newURL() is too vague

harshavardhana:
> yes.. maybe newClientURL() ? the reason is that newURL() is too vague

This makes sense.

PR #1681, opened 2016-04-12
Rewrite diff code

vadmeste:
This is the first version of the diff rewrite. It only works with diff cmd, mirror cmd is still using the old way (so no function names are not properly chosen etc..)

The code is compilable but not deeply tested, I need to reread it again especially when errors occur,

Reviewing and testing are appreciated :D

### DIFF
--- a/difference.go
+++ b/difference.go
@@ -17,7 +17,9 @@
 package main
 
 import (
+	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/minio/minio/pkg/probe"
@@ -32,11 +34,11 @@ type objectDifference func(string, string, os.FileMode, int64, time.Time) (diffe
 type differType string
 
 const (
-	differInSize  differType = "size"          // differs in size
-	differInTime             = "time"          // differs in time
-	differInFirst            = "only-in-first" // only on first source
-	differInType             = "type"          // differs in type, exfile/directory
-	differInNone             = ""              // does not differ
+	differInSize   differType = "size"           // differs in size
+	differInFirst             = "only-in-first"  // only in source
+	differInSecond            = "only-in-second" // only in target
+	differInType              = "type"           // differs in type, exfile/directory
+	differInNone              = ""               // does not differ
 )
 
 // objectDifferenceFactory returns objectDifference function to check for difference
@@ -63,7 +65,6 @@ func objectDifferenceFactory(targetClnt Client) objectDifference {
 			if expected == current {
 				tgtType := content.Type
 				tgtSize := content.Size
-				tgtTime := content.Time
 				if srcType.IsRegular() && !tgtType.IsRegular() {
 					// Type differes. Source is never a directory.
 					return differInType, nil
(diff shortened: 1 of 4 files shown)


difference.go
```diff
@@ -90,3 +87,129 @@ func objectDifferenceFactory(targetClnt Client) objectDifference {
...
+				}
+				srcCtnt, srcOk = <-srcCh
+				continue
+			}
+			if expected == current {
+				srcType, tgtType := srcCtnt.Type, tgtCtnt.Type
+				srcSize, tgtSize := srcCtnt.Size, tgtCtnt.Size
+				if srcType.IsRegular() && !tgtType.IsRegular() ||
+					!srcType.IsRegular() && tgtType.IsRegular() {
+					// Type differes. Source is never a directory.
```
harshavardhana: "Source and target differs in type." is a better comment.

difference.go
```diff
@@ -90,3 +87,129 @@ func objectDifferenceFactory(targetClnt Client) objectDifference {
...
+					d <- diffMessage{
+						FirstURL:  srcCtnt.URL.String(),
+						SecondURL: tgtCtnt.URL.String(),
+						Diff:      differInType,
+					}
+				} else if (srcType.IsRegular() && tgtType.IsRegular()) && srcSize != tgtSize {
+					// Regular files differing in size.
```
harshavardhana: "// Source and target are of same type, but differ in content size. "

difference.go
```diff
@@ -90,3 +87,129 @@ func objectDifferenceFactory(targetClnt Client) objectDifference {
...
+					d <- diffMessage{
+						FirstURL:  srcCtnt.URL.String(),
+						SecondURL: tgtCtnt.URL.String(),
+						Diff:      differInSize,
+					}
+				}
+				// No differ
```
harshavardhana: "// Source and target are same, move to the next one."

difference.go
```diff
@@ -90,3 +87,129 @@ func objectDifferenceFactory(targetClnt Client) objectDifference {
...
+			if srcEOF {
+				d <- diffMessage{
+					SecondURL: tgtCtnt.URL.String(),
+					Diff:      differInSecond,
+				}
+				tgtCtnt, tgtOk = <-tgtCh
+				continue
+			}
+
+			// The same for target
```
harshavardhana: "// If target doesn't have object anymore, comparison becomes obvious. "

difference.go
```diff
@@ -90,3 +87,129 @@ func objectDifferenceFactory(targetClnt Client) objectDifference {
...
+			// Differ in second
+			d <- diffMessage{
+				SecondURL: tgtCtnt.URL.String(),
+				Diff:      differInSecond,
+			}
+			tgtCtnt, tgtOk = <-tgtCh
+			continue
+		}
+	}()
+
```
harshavardhana: Overall this code is complete in approach and very clean.

difference.go
```diff
@@ -32,11 +34,11 @@ type objectDifference func(string, string, os.FileMode, int64, time.Time) (diffe
 type differType string
 
 const (
-	differInSize  differType = "size"          // differs in size
-	differInTime             = "time"          // differs in time
-	differInFirst            = "only-in-first" // only on first source
-	differInType             = "type"          // differs in type, exfile/directory
-	differInNone             = ""              // does not differ
+	differInSize   differType = "size"           // differs in size
```
harshavardhana: Let's change this from string to 'iota' style 

```
type differType int 

const (
     differInNone differType = iota
     differInSize
     differInType
     differInFirst
     differInSecond
)
```

This helps in not passing around strings and also simpler matching techniques.

difference.go
```diff
@@ -90,3 +87,129 @@ func objectDifferenceFactory(targetClnt Client) objectDifference {
 		}
 	}
 }
+
+func objectDifferenceNewImpl(sourceClnt, targetClnt Client, sourceURL, targetURL string) (d chan diffMessage) {
```
harshavardhana: Use "diffCh chan" instead of "d chan" - "d" is too generic.

PR #1687, opened 2016-04-17
mirror: Implement '--remove' feature.

### LINKED ISSUE #1628: mirror: Implement 'mirror --remove'

harshavardhana:
Fixes #1628

### DIFF
--- a/mirror-url.go
+++ b/mirror-url.go
@@ -36,8 +36,11 @@ func (m mirrorURLs) isEmpty() bool {
 	if m.SourceContent == nil && m.TargetContent == nil && m.Error == nil {
 		return true
 	}
-	if m.SourceContent.Size == 0 && m.TargetContent == nil && m.Error == nil {
-		return true
+	// If remove flag is set then sourceContent is usually nil.
+	if m.SourceContent != nil {
+		if m.SourceContent.Size == 0 && m.TargetContent == nil && m.Error == nil {
+			return true
+		}
 	}
 	return false
 }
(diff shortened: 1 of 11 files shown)


mirror-url.go
```diff
@@ -157,13 +160,25 @@ func deltaSourceTarget(sourceURL string, targetURL string, isForce bool, isFake
...
+				mirrorURLsCh <- mirrorURLs{
+					Error: errDeleteNotAllowed(diffMsg.SecondURL),
+				}
+				continue
+			}
+			mirrorURLsCh <- mirrorURLs{
+				TargetAlias:   targetAlias,
+				TargetContent: diffMsg.secondContent,
+			}
+		}
```
abperiasamy: If none of the conditions match, we should still handle it and throw an error or panic. If not, such errors can be hard to catch. Also switch-case may be better here.

session-migrate.go
```diff
@@ -0,0 +1,97 @@
...
+		fatalIf(err.Trace(sid), "Unable to initialize quick config for session '7' header.")
+
+		err = qs.Save(sessionFile)
+		fatalIf(err.Trace(sid, sessionFile), "Unable to migrate session from '6' to '7'.")
+
+		console.Println("Successfully migrated ‘" + sessionFile + "’ from version ‘" + sV6Header.Version + "’ to " + "‘" + sV7Header.Version + "’.")
+	}
+}
+
+func migrateSessionV5ToV6() {
```
abperiasamy: Requires comment, even if we are dropping the session files.

PR #1688, opened 2016-04-17
Move ioutils pkg from minio to mc and fix a sorting list bug

vadmeste:
1. ioutils was removed from the upstream minio but actually it is still used in mc (https://github.com/minio/mc/blob/master/client-fs.go#L783)
2. It seems that there is bug in sorting FS entries needed in recursive listing, diff, mirror etc.. so strange how we didn't detect it before, I am running freebsd maybe that helped to detect the bug, please check twice as I did before validating this commit

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ deadcode:
 build: getdeps verifiers
 	@echo "Installing mc:"
 
-test: verifiers
+test: getdeps verifiers
 	@echo "Running all testing:"
 	@GO15VENDOREXPERIMENT=1 go test $(GOFLAGS) ./
 	@GO15VENDOREXPERIMENT=1 go test $(GOFLAGS) github.com/minio/mc/pkg...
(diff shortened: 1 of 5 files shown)


harshavardhana:
> It seems that there is bug in sorting FS entries needed in recursive listing, diff, mirror etc.. so strange how we didn't detect it before, I am running freebsd maybe that helped to detect the bug, please check twice as I did before validating this commit

The reason is that it's not used internally for any purposes, since 'diff and mirror' are recursive. Directories are ignored for 'FS'

vadmeste:
You can test this under FreeBSD:
$ mkdir ~/testdir
$ mc mirror /usr/local/include ~/testdir/
$ mc diff /usr/local/include ~/testdir/

You'll have some diffs already though that shouldn't happen.

harshavardhana:
> You can test this under FreeBSD:
> $ mkdir ~/testdir
> $ mc mirror /usr/local/include ~/testdir/
> $ mc diff /usr/local/include ~/testdir/
> 
> You'll have some diffs already though that shouldn't happen.

Can you fix the appveyor changes in client-fs_test.go ?

PR #1727, opened 2016-05-23
Ignore BucketEmpty error for top-level listing

### LINKED ISSUE #1725: `mc ls <alias>` fails
`mc ls <alias>` fails on current master. It fails with the following error 

`$mc ls s3/`
`mc: <ERROR> Unable to stat target ‘s3/’. Bucket name cannot be empty.`

krisis:
This is relevant in the case of non-fs backend like s3, play etc.

Closes #1725

### DIFF
--- a/ls-main.go
+++ b/ls-main.go
@@ -135,10 +135,13 @@ func mainList(ctx *cli.Context) {
 		clnt, err := newClient(targetURL)
 		fatalIf(err.Trace(targetURL), "Unable to initialize target ‘"+targetURL+"’.")
 
-		st, err := clnt.Stat()
-		fatalIf(err.Trace(targetURL), "Unable to stat target ‘"+targetURL+"’.")
-
-		if st.Type.IsDir() {
+		var st *clientContent
+		if st, err = clnt.Stat(); err != nil {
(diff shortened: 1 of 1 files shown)


ls-main.go
```diff
@@ -135,10 +135,13 @@ func mainList(ctx *cli.Context) {
...
 		fatalIf(err.Trace(targetURL), "Unable to initialize target ‘"+targetURL+"’.")
 
-		st, err := clnt.Stat()
-		fatalIf(err.Trace(targetURL), "Unable to stat target ‘"+targetURL+"’.")
-
-		if st.Type.IsDir() {
+		var st *clientContent
+		if st, err = clnt.Stat(); err != nil {
+			// For aliases like ``mc ls s3`` it's acceptable to receive BucketNameEmpty error.
+			if _, bucketEmpty := err.Cause.(BucketNameEmpty); !bucketEmpty {
```
harshavardhana: You shouldn't access the probe errors internal semantics like this. It should be done as below to be consistent with other parts of the code so that the underlying API change or names doesn't affect the future functionalities. 

```
switch e := err.ToGoError().(type) {
case BucketNameEmpty:
```

PR #1728, opened 2016-05-23
Made session list ignore broken sessions

### LINKED ISSUE #1723: `mc session list` errors out when one or more session data files are missing - May 24
```
mc: <ERROR> Unable to open session data file "/home/minio/.mc/session/JwaZuaWY.data". open /home/minio/.mc/session/JwaZuaWY.data: no such file or directory
```

krisis:
- `mc session list` - skips partial sessions[1]

- `mc session clear all` - doesn't clear partial sessions.
The idea is to prevent the user from clearing all sessions without
realising that there were partial sessions.

When there is one or more partial sessions, clear each one of them individually
as follows,
 `mc session clear <broken-session-id>` - clears the partial sessions.

[1] - partial session - a session with missing `.data` file in the sessions directory.

N B The presence of partial sessions indicate some underlying bug in session mangement
code. This PR provides relief for users in the event of such a bug.

Closes #1723

### DIFF
--- a/session-main.go
+++ b/session-main.go
@@ -90,9 +90,8 @@ func listSessions() *probe.Error {
 		session, err := loadSessionV7(sid)
 		if err != nil {
 			continue // Skip 'broken' session during listing
-		} else {
-			session.Close() // Session close right here.
 		}
+		session.Close() // Session close right here.
 		bySessions = append(bySessions, session)
 	}
 	// sort sessions based on time.
@@ -150,6 +149,8 @@ func clearSession(sid string) {
 
 	session, err := loadSessionV7(sid)
 	if err != nil {
+		// `mc session clear <broken-session-id>` assumes that user is aware that the session is unuseful
+		// and wants the associated session files to be removed
 		removeSessionFile(sid)
 		removeSessionDataFile(sid)
 		printMsg(clearSessionMessage{Status: "forced", SessionID: sid})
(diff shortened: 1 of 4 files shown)


session.go
```diff
@@ -115,3 +115,19 @@ func getSessionIDs() (sids []string) {
 	}
 	return sids
 }
+
+func removeSessionFile(sid string) {
```
harshavardhana: Needs a comment

session.go
```diff
@@ -115,3 +115,19 @@ func getSessionIDs() (sids []string) {
...
+	sessionFile, err := getSessionFile(sid)
+	if err != nil {
+		return
+	}
+	os.Remove(sessionFile)
+}
+
+func removeSessionDataFile(sid string) {
```
harshavardhana: Same here.

session-main.go
```diff
@@ -140,7 +149,12 @@ func clearSession(sid string) {
 	}
 
 	session, err := loadSessionV7(sid)
-	fatalIf(err.Trace(sid), "Unable to load session ‘"+sid+"’.")
+	if err != nil {
+		removeSessionFile(sid)
```
harshavardhana: Needs a comment for the conditional.

session-main.go
```diff
@@ -89,9 +89,10 @@ func listSessions() *probe.Error {
 	for _, sid := range getSessionIDs() {
 		session, err := loadSessionV7(sid)
 		if err != nil {
-			return err.Trace(sid)
+			continue // Skip 'broken' session during listing
+		} else {
```
harshavardhana: If there is a continue no need to add 'else' below code should suffice. 

```
if err != nil {
   continue
}
session.Close()
```

PR #1742, opened 2016-06-22
ls/cp/mirror: Ignore objects with StorageClass GLACIER.

### LINKED ISSUE #1741: mirror try to sync glacier files
I think it's the origin of the problem, we should be able to ignore those files that can't be mirrored anyway.

`mc: <ERROR> Failed to copy ‘http://s3.amazonaws.com/xxx/00002c2100782297d38eb8e77f7dfce9/1280x720.jpg’. The operation is not valid for the object's storage class`

harshavardhana:
Fixes #1741

### DIFF
--- a/client-errors.go
+++ b/client-errors.go
@@ -74,6 +74,15 @@ func (e ObjectAlreadyExists) Error() string {
 	return "Object ‘" + e.Object + "’ already exists."
 }
 
+// ObjectOnGlacier - object is of storage class glacier.
+type ObjectOnGlacier struct {
+	Object string
+}
+
+func (e ObjectOnGlacier) Error() string {
+	return "Object ‘" + e.Object + "’ is on Glacier storage."
+}
+
 // BucketNameTopLevel - generic error
 type BucketNameTopLevel struct{}
 
--- a/client-s3.go
+++ b/client-s3.go
@@ -678,6 +678,19 @@ func (c *s3Client) listInRoutine(contentCh chan *clientContent) {
 	}
 }
 
+// S3 offers a range of storage classes designed for
+// different use cases, following list captures these.
+const (
+	// General purpose.
+	// s3StorageClassStandard = "STANDARD"
+	// Infrequent access.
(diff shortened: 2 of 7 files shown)


client-s3.go
```diff
@@ -678,6 +678,19 @@ func (c *s3Client) listInRoutine(contentCh chan *clientContent) {
...
+// different use cases, following list captures these.
+const (
+	// General purpose.
+	// s3StandardStorageClass = "STANDARD"
+	// Infrequent access.
+	// s3InfrequentStorageClass = "STANDARD_IA"
+	// Reduced redundancy access.
+	// s3RedundancyStorageClass = "REDUCED_REDUNDANCY"
+	// Archive access.
+	s3GlacierStorageClass = "GLACIER"
```
abperiasamy: Rename it to s3StorageClassGlacier. Variable component comes at the end. This will also allow IDEs to autocomplete.

client-errors.go
```diff
@@ -74,6 +74,15 @@ func (e ObjectAlreadyExists) Error() string {
 	return "Object ‘" + e.Object + "’ already exists."
 }
 
+// ObjectOnGlacier - object is of storage class glacier.
+type ObjectOnGlacier struct {
+	Object string
+}
+
+func (e ObjectOnGlacier) Error() string {
+	return "Object ‘" + e.Object + "’ is of storage class GLACIER."
```
abperiasamy: return "Object ‘" + e.Object + "’ is on Glacier storage."

ls.go
```diff
@@ -126,6 +126,9 @@ func doList(clnt Client, isRecursive, isIncomplete bool) *probe.Error {
 			case PathInsufficientPermission:
 				errorIf(content.Err.Trace(clnt.GetURL().String()), "Unable to list folder.")
 				continue
+			case ObjectOnGlacier:
+				errorIf(content.Err.Trace(clnt.GetURL().String()), "Unable to list GLACIER object, you must restore the object to be able to list.")
+				continue
```
abperiasamy: ObjectOnGlacier error message will do.

PR #1754, opened 2016-07-08
Fix broken documentation links in README.md

Flimm:
This is only broken in GitHub, not on the documentation site, but still.


harshavardhana:
Thank you!

PR #1756, opened 2016-07-09
Add support of golang tip

### LINKED ISSUE #1709: Binary built with tip Go version is completely unusable because of some runtime checks
```
go (master) ¶ mc
mc: <ERROR> Unable to determine current go version. Malformed version: vel +61602b0 Fri May 6 17:04:11 2016 +0000
1 go (master) ¶ go version
go version devel +61602b0 Fri May 6 17:04:11 2016 +0000 darwin/amd64
```

This is the first time I've ever see such runtime checks against Go version. Can't such checks be implemented as build-time ones (version-specific build tags)?

vadmeste:
* Fixes #1709 
* Support compiling with Makefile too, though it will fail for the moment due to a go bug (https://github.com/golang/go/issues/16310), you have to manually disable lint to make it work

### DIFF
--- a/buildscripts/checkdeps.sh
+++ b/buildscripts/checkdeps.sh
@@ -53,8 +53,8 @@ readlink() {
 ###
 #
 # Takes two arguments
-# arg1: version number in `x.x.x` format
-# arg2: version number in `x.x.x` format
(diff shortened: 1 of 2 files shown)


harshavardhana:
LGTM, Thanks :+1:

PR #1762, opened 2016-07-11
client-s3 Stat() returns proper error in case of an object or dir are…

vadmeste:
… not found


harshavardhana:
LGTM, Thanks :+1:

PR #1763, opened 2016-07-11
Fix S3 Stat() dir mode marking

vadmeste:
A Stat() bug happens when ListObjects returns more than one entry (in case there are many entries with common prefix). In that case, the object is marked as directory which is wrong. Also we should trim trailing slashes to make ListObjects works as we need.


krishnasrinivas:
Looks good, but I see that this is a fix on top of harsha's previous fix, so harsha can review.

PR #1769, opened 2016-07-18
mc removes incomplete downloads .part.minio when rm is invoked with --incomplete flag

vadmeste:


harshavardhana:
Actually this could be implemented because 'Listing' already lists incomplete upload properly even for FS.

vadmeste:
> Actually this could be implemented because 'Listing' already lists incomplete upload properly even for FS.

This is local FS and not Minio FS (if I understood you well)

harshavardhana:
> This is local FS and not Minio FS (if I understood you well)

yes 'mc ls -I'  lists incompletely download files.

harshavardhana:
what else is missing in this patch @vadmeste ?

vadmeste:
This patch is updated so it will remove *.part.minio for incomplete downloaded files

harshavardhana:
> This patch is updated so it will remove *.part.minio for incomplete downloaded files

Update the patch description, it doesn't reflect what the patch is doing.

vadmeste:
:+1:

Issue #1767, opened 2016-07-16
mc: Implement asynchronous replication support 


harshavardhana:

GarimaKapoor:
Please update the timeline for the fix.

nl5887:
I'll add an extra option `-monitor` to cp (create, update) and mirror (create, update, delete) which will monitor the source folder for any changes. If there are any changes those will be reflected on destination as well.

nl5887:
Wednesday 27th I'll have this all implemented.

nl5887:
Almost there.... Working on the latest bugs and optimizations.

nl5887:
Ok, fixed last deadlock. Everything working great now. Optimizing, cleaning up code and documenting.

nl5887:
- the cp command will stay (at least for now) the same, the command is quite complex, and don't know what side-effects changes will have
- the mirror command will have a -watch option, which will watch for changes
- the s3 client will support watch using bucket notifications

PR #1784, opened 2016-07-30
async replication support for mirror

nl5887:
resolves minio/mc#1767

* disk will be monitored for changes, monitoring is client pluggable, so we can use notification for this as well
* refactored code
* needs to be tested thoroughly


harshavardhana:
Will review and test

cp-main.go
```diff
@@ -45,6 +52,14 @@ var (
 			Name:  "recursive, r",
 			Usage: "Copy recursively.",
 		},
+		cli.BoolFlag{
+			Name:  "skip-copy",
```
abperiasamy: --fake is consistently used (see mirror --fake)

nl5887: I think fake should be added as well. Maybe `skip-copy` is being named wrong, but what I want with this option is to skip the initial copy and scan, and only start monitoring.

nl5887:
fsnotify doesn't support multi-platform close notifications (yet), fsnotify/fsnotify#22. There is a linux platform approach fsnotify/fsnotify#91.

harshavardhana:
@nl5887 can you fix the build issues here? related to 'golint and govet' complaints?

status.go
```diff
@@ -0,0 +1,162 @@
+package main
```
harshavardhana: Add missing license header.

status.go
```diff
@@ -0,0 +1,162 @@
...
+		console.Println(msg.String())
+	} else {
+		console.Println(msg.JSON())
+	}
+}
+
+func (s *QuitStatus) Start() {
+}
+
+func (s *QuitStatus) Finish() {
```
harshavardhana: Should this be ``QuiteStatus`` ?

session-v7.go
```diff
@@ -194,6 +194,7 @@ func (s *sessionV7) NewDataReader() io.Reader {
 func (s *sessionV7) NewDataWriter() io.Writer {
 	// DataFP is always intitialized, either via new or load functions.
 	s.DataFP.Seek(0, os.SEEK_SET)
+	s.DataFP.Truncate(0)
```
harshavardhana: Needs a comment on why are we truncating.

scan-bar.go
```diff
@@ -43,6 +43,9 @@ func fixateScanBar(text string, width int) string {
 // Progress bar function report objects being scaned.
 type scanBarFunc func(string)
 
+func discardScanBarFactory(s string) {
```
harshavardhana: is this intentional?

mirror-url.go
```diff
@@ -32,6 +32,39 @@ type mirrorURLs struct {
...
 }
 
+func (m mirrorURLs) WithError(err *probe.Error) mirrorURLs {
+	m.Error = err
+	return m
+}
+
+func (m mirrorURLs) Equal(o interface{}) bool {
+	n, ok := o.(mirrorURLs)
+	if !ok {
```
harshavardhana: wouldn't this be solved by not taking an input as 'interface ? - why would you type assert which a compiler can do this for you?

cp-url.go
```diff
@@ -31,6 +31,15 @@ type copyURLs struct {
 	Error         *probe.Error `json:"-"`
 }
 
+func (m *copyURLs) Equal(o interface{}) bool {
```
harshavardhana: Same here why would you take an interface? when one can avoid compile time problems with proper types?

cp-url-syntax.go
```diff
@@ -40,10 +40,15 @@ func checkCopySyntax(ctx *cli.Context) {
...
 	// Verify if source(s) exists.
 	for _, srcURL := range srcURLs {
-		_, _, err := url2Stat(srcURL)
+		_, sourceContent, err := url2Stat(srcURL)
 		if err != nil {
 			fatalIf(err.Trace(srcURL), fmt.Sprintf("Unable to stat '%s'.", srcURL))
 		}
+
+		if !ctx.Bool("monitor") {
+		} else if sourceContent.URL.Type != fileSystem {
```
harshavardhana: May be this is not needed anymore?

client-s3.go
```diff
@@ -142,6 +144,93 @@ func (c *s3Client) GetURL() clientURL {
...
+
+func (c *s3Client) Watch(recursive bool) (*watchObject, *probe.Error) {
+	eventChan := make(chan Event)
+	errorChan := make(chan *probe.Error)
+	doneChan := make(chan bool)
+
+	bucket, _ := c.url2BucketAndObject()
+
+	// todo(nl5887): correct arn creation
+	accountARN := minio.NewArn("minio", "lambda", "us-east-1", "1", "minio")
```
harshavardhana: The Arn should be changed to 

```
minio.NewArn("minio", "lambda", "us-east-1", "1", "lambda")
```

client-s3.go
```diff
@@ -142,6 +144,93 @@ func (c *s3Client) GetURL() clientURL {
...
+
+		close(doneCh)
+		close(eventChan)
+		close(errorChan)
+	}()
+
+	eventsCh := c.api.ListenBucketNotification(bucket, accountARN, doneCh)
+	go func() {
+		for {
+			if notificationInfo, ok := <-eventsCh; !ok {
```
harshavardhana: You could simply do 

```
for notificationInfo := range eventsCh {
}
```

This would simplify the large 'if else' statements.

client-fs.go
```diff
@@ -85,6 +86,102 @@ func (f *fsClient) GetURL() clientURL {
...
+			select {
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return
+				}
+
+				switch {
+				case event.Op&fsnotify.Create == fsnotify.Create:
+					if i, err := os.Stat(event.Name); err != nil {
+						errorChan <- probe.NewError(err)
```
harshavardhana: We can ignore the file which doesn't exist. for example `os.IsNotExist(err)` and not send errors. Since it is a good condition.

client-fs.go
```diff
@@ -85,6 +86,102 @@ func (f *fsClient) GetURL() clientURL {
...
+					} else if !i.IsDir() {
+						// we want files
+					} else if !recursive {
```
harshavardhana: Wouldn't it make sense to not watch at all? - do not understand the logic here btw. Can you explain?

client-s3.go
```diff
@@ -33,7 +34,8 @@ import (
 
 	"github.com/minio/mc/pkg/httptracer"
 	"github.com/minio/minio/pkg/probe"
-	"gopkg.in/minio/minio-go.v2"
+	// todo(nl5887): http://gopkg.in/minio/minio-go.v2
```
harshavardhana: You can re-vendorize 'minio-go' btw we will make a v3 release soon.

mirror-url.go
nl5887: This was written to be used with copyURLs as well, using the same interface.

cp-url.go
nl5887: The interface is being implemented is Comparer (see queue.go)

cp-url-syntax.go
nl5887: true, resetted.

client-s3.go
nl5887: fixed

client-fs.go
nl5887: this part of the code is to enable watching recursively on new folders. If we don't do this, then newly created folders in target, won't be watched.

client-fs.go
nl5887: Don't completely agree here, it would be strange when getting a create event and the file or folder won't exist seconds later anymore. But we can ignore them.

client-s3.go
nl5887: good one, changed.

cp-url-syntax.go
nl5887: removed

scan-bar.go
nl5887: Yes, the discardScanBarFactory is a function that won't show the output, so we don't have to write all ifs and elses.

status.go
nl5887: fixed.

client-fs.go
```diff
@@ -85,6 +86,109 @@ func (f *fsClient) GetURL() clientURL {
 	return *f.PathURL
 }
```
harshavardhana: All these functions need comments as well.

client-fs.go
```diff
@@ -85,6 +86,109 @@ func (f *fsClient) GetURL() clientURL {
...
+	}
+
+	go func() {
+		<-doneChan
+
+		watcher.Close()
+		close(eventChan)
+		close(errorChan)
+	}()
+
```
harshavardhana: We need a comment as soon as we enter a go-routine.

client-s3.go
```diff
@@ -142,6 +143,92 @@ func (c *s3Client) GetURL() clientURL {
 	return *c.targetURL
 }
 
+type watchObject struct {
+	events chan Event
+	errors chan *probe.Error
+	done   chan bool
+}
+
```
harshavardhana: Each of these functions would need comments, including the struct datatypes.

cp-url.go
harshavardhana: > The interface is being implemented is Comparer (see queue.go)

In such a case i think copyURLs and mirrorURLs can be of same type, i would generally avoid taking interfaces like this.

```
type mirrorURLs struct {
        SourceAlias   string
        SourceContent *clientContent
        TargetAlias   string
        TargetContent *clientContent
        Error         *probe.Error `json:"-"`
}

type copyURLs struct {
        SourceAlias   string
        SourceContent *clientContent
        TargetAlias   string
        TargetContent *clientContent
        Error         *probe.Error `json:"-"`
}

```
Both structs are same here.

mirror-main.go
```diff
@@ -92,6 +97,31 @@ EXAMPLES:
 `,
 }
```
harshavardhana: We need to document each fields here especially the channels.

mirror-main.go
```diff
@@ -245,221 +266,422 @@ func doMirror(sURLs mirrorURLs, progressReader *progressBar, accountingReader *a
...
 			}
-			totalObjects++
-		case <-trapCh:
-			// Print in new line and adjust to top so that we don't print over the ongoing scan bar
-			if !globalQuiet && !globalJSON {
-				console.Eraseline()
+
+			ms.Save()
+
+			if sURLs.SourceContent != nil {
```
harshavardhana: Is this `sURLs.SourceContent` for future? why not do 

```
if sURLs.SourceContent == nil && sURLs.TargetContent != nil {
```

mirror-main.go
```diff
@@ -245,221 +266,422 @@ func doMirror(sURLs mirrorURLs, progressReader *progressBar, accountingReader *a
...
-	var progressReader *progressBar
-	if !globalQuiet && !globalJSON {
-		progressReader = newProgressBar(session.Header.TotalBytes)
-	}
+			v := ms.queue.Pop()
+			if v == nil {
+				break
+			}
+
+			sURLs := v.(mirrorURLs)
```
harshavardhana: Needs proper type assertion here, with 'ok' better we avoid non-typed returns in the first place.

mirror-main.go
```diff
@@ -245,221 +266,422 @@ func doMirror(sURLs mirrorURLs, progressReader *progressBar, accountingReader *a
...
 			}
+
+		case err := <-ms.watcher.Errors():
+			errorIf(err,
+				fmt.Sprintf("Got error during monitoring."))
 		}
-	}()
+	}
+}
+
```
harshavardhana: Comment here..

mirror-main.go
```diff
@@ -245,221 +266,422 @@ func doMirror(sURLs mirrorURLs, progressReader *progressBar, accountingReader *a
...
+	var totalBytes int64
+	var totalObjects int
+
+loop:
+	for {
+		select {
+		case sURLs, ok := <-ms.harvestCh:
+			if !ok {
+				// finished harvesting urls
+				break loop
```
harshavardhana: Is this 'break loop' necessary?

scan-bar.go
harshavardhana: Yes it would be misleading to call it factory.  you can use a local 'func()' closure rather than declaring it here.

mirror-main.go
nl5887: this breaks the for loop, otherwise it will break the case.

cp-url.go
nl5887: Because this was already in the code, I thought it was done for a reason. I've done quite some refactoring already, but wanted to keep as close as the original code as possible.

mirror-main.go
nl5887: the sURLs.SourceContent is being used to determine if the file is being deleted or created. Depending on that it will set LastCopied or LastRemoved.

harshavardhana: This logic is not apparent in this case.

harshavardhana:
This needs full rebase.. now with refactor merged.

PR #1791, opened 2016-08-10
Add notify cmd to listen on bucket notifications

vadmeste:


harshavardhana:
You might have to rebase this with latest master and also re-vendorize the 'minio-go'

notify-main.go
```diff
@@ -0,0 +1,108 @@
...
+	fatalIf(probe.NewError(err), "Cannot watch on the specified bucket.")
+
+	// Print all notifications as we receive them
+	for notification := range notificationCh {
+		if notification.Err != nil {
+			// Ignore errors
+			continue
+		}
+		for _, info := range notification.Records {
+			fmt.Printf("%s\t%s\t%s\n", info.EventName, info.S3.Bucket.Name, info.S3.Object.Key)
```
harshavardhana: We should pick a color here for proper printing.  Also the info.EventName can be converted to a meaningful message for 'mc'. 

EventName with 's3:ObjectCreatedPut' is too much detail on a CLI. 

Additionally we can have a '--extended' option as well which prints 'sourceIPAddress'

notify-main.go
```diff
@@ -0,0 +1,108 @@
...
+	region := ctx.String("account-region")
+	accountId := ctx.String("account-id")
+
+	client, pErr := newClient(path)
+	if pErr != nil {
+		fatalIf(pErr.Trace(), "Cannot parse the provided url.")
+	}
+
+	// For the moment, we only support s3
+	s3Client, ok := client.(*s3Client)
```
harshavardhana: We don't need this in current master, Watch() api is combined.

vadmeste: yes, but currently, Watch of S3 and FS don't have the same input and output parameters, we need to discuss this a bit

harshavardhana: Make them same. They have to comply to a common interface.

vadmeste:
This is still not finished

vadmeste:
Should be ready for testing and merge

mc/client-s3.go
```diff
@@ -158,27 +157,40 @@ func (w *watchObject) Close() {
...
-	lc.AddEvents(minio.ObjectCreatedAll, minio.ObjectRemovedAll)
+	accountARN := minio.NewArn("minio", "lambda", region, accountID, "lambda")
 	mb := minio.BucketNotification{}
-	mb.AddLambda(lc)
+	if enable {
+		// enable bucket notifications
+		lc := minio.NewNotificationConfig(accountARN)
+		lc.AddEvents(minio.ObjectCreatedAll, minio.ObjectRemovedAll)
+		mb.AddLambda(lc)
+	}
```
harshavardhana: the toggle should remove only the aforementioned lambda config not the entire bucket notification we shouldn't edit other configs.

PR #1793, opened 2016-08-11
Refacture structure proposal.

nl5887:
in discussion with @krishnasrinivas, we're experimenting with a new structure. This will have less files in root.


command/api_handlers_test.go.bak
```diff
@@ -0,0 +1,167 @@
+/*
+ * Minio Client (C) 2015 Minio, Inc.
```
harshavardhana: looks like '.bak' file?

command/main.go
```diff
@@ -0,0 +1,300 @@
...
+	app.CommandNotFound = commandNotFound // handler function declared above.
+	return app
+}
+
+// mustGetProfilePath must get location that the profile will be written to.
+func mustGetProfileDir() string {
+	return filepath.Join(mustGetMcConfigDir(), globalProfileDir)
+}
+
+func main() {
```
harshavardhana: Looks like this main is not needed anymore.

PR #1799, opened 2016-08-16
mirror: Watch() for events based on session time.

harshavardhana:
Enable notification if not already set, avoid overriding
or append to existing notification set - all of this
needs to be done to make this transparent.


harshavardhana:
// cc @nl5887 for your review.

mc/client-s3.go
```diff
@@ -190,16 +191,92 @@ func (c *s3Client) ToogleLambdaNotificationStatus(enable bool, region, accountID
...
+	shouldSetNotification := len(nConfig.TopicConfigs) == 0
+	if !shouldSetNotification {
+		// We found previously configure SNS topics, validate if current account-id is the same.
+		for _, topicConfig := range nConfig.TopicConfigs {
+			if topicConfig.Topic == accountARN.String() {
+				shouldSetNotification = false
+				break
+			}
+		}
+		shouldSetNotification = true
```
nl5887: This will set shouldSetNotification always to true, ignoring the evaluation.

mc/client-s3.go
```diff
@@ -190,16 +191,92 @@ func (c *s3Client) ToogleLambdaNotificationStatus(enable bool, region, accountID
...
+	// Fetch any existing bucket notification on the bucket.
+	nConfig, e := c.api.GetBucketNotification(bucket)
+	if e != nil {
+		return probe.NewError(e)
+	}
+
+	// Construct account ARN.
+	accountARN := minio.NewArn("minio", "sns", location, params.accountID, "listen")
+
+	// Remote account ARN if any.
```
nl5887: Typo? Remove instead of Remote

mc/mirror-main.go
```diff
@@ -462,23 +460,37 @@ func (ms *mirrorSession) watch() {
...
+		return nil
+	} // Failed to initialize client.
+	return err
+}
 
-	return nil
+func (ms *mirrorSession) watchSourceURL(recursive bool) *probe.Error {
+	sourceClient, err := newClient(ms.sourceURL)
+	if err == nil {
+		err = ms.watcher.Join(sourceClient, recursive)
```
nl5887: return ms.watcher.Join(sourceClient, recursive)

mc/watch.go
```diff
@@ -119,14 +115,24 @@ func (w *Watcher) Wait() {
 	w.wg.Wait()
 }
 
-// Join the watcher with client
-func (w *Watcher) Join(client Client, recursive bool) *probe.Error {
```
nl5887: You've removed the check for ClientWatcher, I added this because in theory it could be possible to have a Client not supporting a watcher.

harshavardhana: We will not have such clients. We shouldnt bring in new interfaces in such a manner.

mc/client-s3.go
harshavardhana: Yes by default. Only if the account was found previously.

harshavardhana: It would be false.

nl5887: maybe I'm looking not good, but it sets shouldSetNotification to false, breaks the for and immediately sets it to true again.

harshavardhana: Oops yes you are right let me fix it

PR #1807, opened 2016-08-22
Move the notify events to new package - rjeczalik/notify.

harshavardhana:


harshavardhana:
Anything else that you see here missing @nl5887 ?

nl5887:
Nope, looks good to me.

PR #1811, opened 2016-08-28
Add policy list command

### LINKED ISSUE #1736: Feature: mc policy list
I was thinking it would be nice to have a command `mc access list <alias>/<bucket-name>` which would list the policies in a certain bucket. This command would provide an easy way for an admin to know the policy status of a particular bucket. It would come in handy when trying to modify/remove policies.
example:
```
$ mc access readonly myminio/mybucket/abc/
Access permission for ‘myminio/mybucket/abc/' is set to ‘readonly’
$ mc access writeonly myminio/mybucket/def
Access permission for ‘myminio/mybucket/def’ is set to ‘writeonly’
$ mc access list myminio/mybucket
mybucket/abc/*   : 'readonly'
mybucket/def*    : 'writeonly'
```

vadmeste:
Fixes #1736

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -979,6 +979,14 @@ func (f *fsClient) MakeBucket(region string) *probe.Error {
 	return nil
 }
 
+// GetAccessRules - unsupported API
+func (f *fsClient) GetAccessRules() (map[string]string, *probe.Error) {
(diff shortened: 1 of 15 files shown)


vadmeste:
Ready for review and merge

PR #1814, opened 2016-09-01
[wip] Add --insecure flag to skip certificate validation

vadmeste:
Remarks:
*) I upgraded to session v8 because we have the new insecure flag
*) I print certificate information when --debug is enabled.. otherwise I guess it will be just annoying for the user to always see it

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -47,6 +47,10 @@ var globalFlags = []cli.Flag{
 		Name:  "debug",
 		Usage: "Enable debugging output.",
 	},
+	cli.BoolFlag{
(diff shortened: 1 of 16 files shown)


cmd/flags.go
```diff
@@ -47,6 +47,10 @@ var globalFlags = []cli.Flag{
 		Name:  "debug",
 		Usage: "Enable debugging output.",
 	},
+	cli.BoolFlag{
+		Name:  "insecure",
+		Usage: "Skip SSl certificate verification.",
```
harshavardhana: 's/SSl/SSL/g'

Issue #1828, opened 2016-09-19
Add support to `rm` to remove objects older than given time.


balamurugana:

harshavardhana:
Fixed by #1837

PR #1837, opened 2016-09-27
rm: add timebased delete and object list from stdin

### LINKED ISSUE #1827: Add multi-delete support by taking object list from stdin

balamurugana:
* Add multi-delete support by taking object list from stdin
* Add support to `rm` to remove objects older than given time

Fixes #1827 #1828

### DIFF
--- a/cmd/rm-main.go
+++ b/cmd/rm-main.go
@@ -17,10 +17,15 @@
 package cmd
 
 import (
+	"bufio"
 	"encoding/json"
 	"fmt"
+	"os"
+	"path"
 	"path/filepath"
+	"runtime"
 	"strings"
+	"time"
 
 	"github.com/fatih/color"
 	"github.com/minio/cli"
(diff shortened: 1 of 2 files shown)


cmd/rm-main.go
```diff
@@ -169,14 +230,27 @@ func rmAll(targetAlias, targetURL string, isRecursive, isIncomplete, isFake bool
 			url.Path = strings.TrimSuffix(entry.URL.Path, string(entry.URL.Separator)) + string(entry.URL.Separator)
 
 			// Recursively remove contents of this directory.
-			rmAll(targetAlias, url.String(), isRecursive, isIncomplete, isFake)
+			rmAll(targetAlias, url.String(), isRecursive, isIncomplete, isFake, older)
+		}
+
+		// Check whether object is created older than given time only if older is >= one hour.
+		if older >= defaultOlderTime {
+			now := time.Now()
```
harshavardhana: We should use 'time.Now().UTC()'

harshavardhana:
> We should use 'time.Now().UTC()'

ping @balamurugana can you address the last comment?

cmd/rm-main.go
balamurugana: Done

PR #1870, opened 2016-11-05
In json or quiet we print the full path of the target. Instead with a…

harshavardhana:
…lias.


harshavardhana:
This is deferred to https://github.com/minio/mc/issues/1873

PR #1877, opened 2016-11-08
Rename 'both' policy to 'public'

vadmeste:


harshavardhana:
You might have to update - https://github.com/minio/mc/blob/master/docs/minio-client-complete-guide.md as well.

vadmeste:
> You might have to update - https://github.com/minio/mc/blob/master/docs/minio-client-complete-guide.md as well.

Already done.

PR #1886, opened 2016-11-19
we removed GOINSTALL.md some time ago, point at canonical golang.org …

rremer:
…page with the same instructions


harshavardhana:
Thanks for the patch @rremer - we actually prefer our link instead - https://docs.minio.io/docs/how-to-install-golang

rremer:
As you wish, opened #1888 with the same commit but the url you provided @harshavardhana 

See also #1887 which is why I was in that script to begin with. :)

PR #1888, opened 2016-11-19
we removed GOINSTALL.md some time ago, point at new minio docs withe …

rremer:
…same instructions


harshavardhana:
Thanks @rremer

PR #1892, opened 2016-11-21
Fix downloading files with exact ranges requested in client FS.

harshavardhana:
This is due to the fact that ReadAt() will not send io.EOF
when reading past the offset, server instead returns
Range error. Handle the error by only reading until the
requested size.

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -246,12 +246,13 @@ func (f *fsClient) Put(reader io.Reader, size int64, contentType string, progres
 				return 0, probe.NewError(e)
 			}
 		}
+
 		// Allocate buffer of 10MiB once.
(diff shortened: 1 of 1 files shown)


cmd/client-fs.go
```diff
@@ -246,12 +342,20 @@ func (f *fsClient) Put(reader io.Reader, size int64, contentType string, progres
 				return 0, probe.NewError(e)
 			}
 		}
+
 		// Allocate buffer of 10MiB once.
 		readAtBuffer := make([]byte, 10*1024*1024)
 
+		// Fetch the size of the reader.
+		var rsize int64
```
vadmeste: Actually, size is already passed as a parameter in Put(), do we need to calculate it again?

harshavardhana: Yep, did that now.

PR #1894, opened 2016-11-24
Relax access/secret key requirements as minio server.

### LINKED ISSUE #1868: Review which characters should be accepted in access and secret keys
minio & mc don't have the same restriction rule in terms of accepted characters in access and secret keys, we need to review this.

harshavardhana:
Also add tests in the process. Fixes #1868

### DIFF
--- a/cmd/config-utils.go
+++ b/cmd/config-utils.go
@@ -16,52 +16,45 @@
 
 package cmd
 
-import (
-	"regexp"
-	"strings"
-)
+import "strings"
 
 var validAPIs = []string{"S3v4", "S3v2"}
 
-// isValidSecretKey - validate secret key.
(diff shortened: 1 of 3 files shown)


cmd/config-utils.go
```diff
@@ -16,52 +16,45 @@
...
-	if strings.TrimSpace(hostURL) == "" {
-		return false
-	}
-	url := newClientURL(hostURL)
-	if url.Scheme != "https" && url.Scheme != "http" {
-		return false
-	}
-	if url.Path != "" && url.Path != "/" {
-		return false
+func isValidHostURL(hostURL string) (ok bool) {
```
krishnasrinivas: @harshavardhana previous function code flow was more readable. i.e single line indentation and eliminate false conditions one by one.

PR #1895, opened 2016-11-25
mc: Consistent exit codes across all commands.

### LINKED ISSUE #1880: `mc` returns inconsistent exit status on same kind of error
## Expected behaviour

mc should exit with non-zero exit code for "connection refused" errors (i.e. when the server is down).

## Actual behaviour

In some cases mc exits with 0 exit status when server is down. This makes it difficult to use in scripts.

```
aditya@donatello-laptop:~/Code/minio-code/data$ echo "tiny object" | mc pipe dminio1/testbucket/one
mc: <ERROR> Unable to write to one or more targets. Get http://127.0.0.1:9001/testbucket/?location=: dial tcp 127.0.0.1:9001: getsockopt: connection refused
aditya@donatello-laptop:~/Code/minio-code/data$ echo $?
1
aditya@donatello-laptop:~/Code/minio-code/data$ mc ls dminio1
mc: <ERROR> Unable to list folder. Get http://127.0.0.1:9001/: dial tcp 127.0.0.1:9001: getsockopt: connection refused
aditya@donatello-laptop:~/Code/minio-code/data$ echo $?
0

```

## mc version
```
$ mc version
Version: 2016-11-11T19:01:44Z
Release-tag: DEVELOPMENT.2016-11-11T19-01-44Z
Commit-id: 2895b0accaad4465128e4b950bf048494de36a9f
```
## System information
Linux.

harshavardhana:
Also brings in much needed CLI changes from
upstream and backporting our custom changes on
top of those changes.

Fixes #1880


cmd/ls-main.go
```diff
@@ -151,11 +148,7 @@ func mainList(ctx *cli.Context) {
 			clnt, err = newClient(targetURL)
 			fatalIf(err.Trace(targetURL), "Unable to initialize target ‘"+targetURL+"’.")
 		}
-
-		err = doList(clnt, isRecursive, isIncomplete)
-		if err != nil {
-			errorIf(err.Trace(clnt.GetURL().String()), "Unable to list target ‘"+clnt.GetURL().String()+"’.")
-			continue
-		}
+		cErr = doList(clnt, isRecursive, isIncomplete)
```
vadmeste: We are in a loop here, this means cErr can be set with 'nil' in the next loop and the error will be hidden.

cmd/ls.go
```diff
@@ -131,6 +132,7 @@ func doList(clnt Client, isRecursive, isIncomplete bool) *probe.Error {
 				continue
 			}
 			errorIf(content.Err.Trace(clnt.GetURL().String()), "Unable to list folder.")
+			cErr = exitCoder(1) // Set the exit status.
```
vadmeste: We are in a loop, cErr could be set to nil in the next loop

cmd/mb-main.go
```diff
@@ -113,23 +109,27 @@ func mainMakeBucket(ctx *cli.Context) {
...
 	region := ctx.String("region")
 
+	var cErr error
 	for i := range ctx.Args() {
 		targetURL := ctx.Args().Get(i)
 		// Instantiate client for URL.
 		clnt, err := newClient(targetURL)
 		if err != nil {
 			errorIf(err.Trace(targetURL), "Invalid target ‘"+targetURL+"’.")
+			cErr = exitCoder(1)
```
vadmeste: Same as above

cmd/mb-main.go
```diff
@@ -113,23 +109,27 @@ func mainMakeBucket(ctx *cli.Context) {
...
 			continue
 		}
 
 		// Make bucket.
 		err = clnt.MakeBucket(region)
 		if err != nil {
 			errorIf(err.Trace(targetURL), "Unable to make bucket ‘"+targetURL+"’.")
+			cErr = exitCoder(1)
```
vadmeste: Same as above

cmd/events-main.go
```diff
@@ -53,7 +48,7 @@ COMMANDS:
 }
 
 // mainEvents is the handle for "mc events" command.
-func mainEvents(ctx *cli.Context) {
+func mainEvents(ctx *cli.Context) error {
```
krishnasrinivas: this function returns nil always, what is the thought behind this way of doing things? several other functions also are behaving the same way, so wondering.

cmd/ls-main.go
abperiasamy: yeah set it only if cErr is nill.

cmd/error.go
```diff
@@ -76,6 +77,15 @@ func fatalIf(err *probe.Error, msg string) {
...
 
+// Exit coder wraps cli new exit error with a
+// custom exitStatus number. cli package requires
+// an error with `cli.ExitCoder` compatibility
+// after an action. Which woud allow cli package to
+// exit with the specified `exitStatus`.
+func exitCoder(exitStatus int) error {
+	return cli.NewExitError("", exitStatus)
+}
+
```
abperiasamy: rename it to exitStatus(status int) error.

cmd/events-main.go
harshavardhana: It is to generate consistent exit codes . It's a new CLI feature.

harshavardhana: Basically errorIf never exited the program so we didn't have a consistent exit code.

cmd/ls.go
harshavardhana: how will this happen?  content.Err is non nil.

cmd/mb-main.go
harshavardhana: how will this happen?

cmd/mb-main.go
harshavardhana: how will this happen?

PR #1897, opened 2016-11-26
cp/mirror: Use proper paths for windows during copyObject.

### LINKED ISSUE #1857: Recursive copy between buckets using mc fails
## Expected behaviour

Recursive copy between buckets must succeed. 
## Actual behaviour

```
$ mc --quiet cp  --recursive localxl/test1/ localxl/test2/
‘localxl/test1/hello.txt’ -> ‘localxl/test2/hello.txt’
mc.exe: <ERROR> Failed to copy ‘http://localhost:9000/test1/hello.txt’. Object key is missing, object key cannot be empty
Total: 6 B, Transferred: 0 B, Speed: 0 B/s
```

While the same works with aws,

```
$ aws --endpoint http://localhost:9000 s3 cp --recursive s3://test1 s3://test2
copy: s3://test1/hello.txt to s3://test2/hello.txt
```
- Make two buckets `test1` and `test2`
- Copy a small object to `test1`
- `mc --quiet cp  --recursive localxl/test1/ localxl/test2/` , where `localxl` is alias for local Minio server
## mc version

`17eeec6`
## System information

MS Windows

harshavardhana:
Fixes #1857

### DIFF
--- a/cmd/cp-main.go
+++ b/cmd/cp-main.go
@@ -183,7 +183,8 @@ func doCopy(cpURLs URLs, progressReader *progressBar, accountingReader *accounte
 			// If source/target are object storage their aliases must be the same.
 			if sourceAlias == targetAlias {
 				// Do not include alias inside path for ObjStore -> ObjStore.
-				err := copySourceStreamFromAlias(targetAlias, targetURL.String(), sourceURL.Path, length, progress)
+				sourcePath := filepath.ToSlash(sourceURL.Path)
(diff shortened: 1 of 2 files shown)


harshavardhana:
This fix is complete for the present form of path handling, what we need to do here is fix the newURL() to always provide paths with "/" not with "\\" in them. Currently i see that `mc` requires an overhaul in terms of path handling across the board which will do subsequently after this merge.  Such a change needs to proceed before we bring in `mc admin` capabilities.

PR #1901, opened 2016-11-27
tls: load custom public certificates from CAs dir

vadmeste:
When a user needs to setup his own TLS configuration, he can put his
custom authority certificates under ~/.mc/CAs/. These certificates will
be automatically loaded at startup and will be set to all HTTPS connections

### DIFF
--- a/cmd/certs.go
+++ b/cmd/certs.go
@@ -1,0 +1,144 @@
+/*
+ * Minio Client (C) 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


harshavardhana:
The dir structure for this should be `~/.mc/certs/CAs/` makes it easier to understand the purpose.

vadmeste:
> The dir structure for this should be ~/.mc/certs/CAs/ makes it easier to understand the purpose.

yes but certs will be always empty as I know. Probably it should name ~/.mc/rootCAs/ ?

harshavardhana:
> yes but certs will be always empty as I know. Probably it should be named ~/.mc/rootCAs/ ?

No certs can be provided too, it is a placeholder to acquire the namespace so that we don't have to migrate things in future if we bring in custom certs to be used from `mc`

PR #1927, opened 2016-12-14
mirror/cp/pipe: Preserve metadata while copying s3 to s3.

### LINKED ISSUE #1923: mirror:  Include user metadata (headers) from objects when mirroring
## Expected behaviour
If custom headers are present in the source object, these should be mirrored over to the new object. 

## Actual behaviour
Custom headers are not present on destination object therefore this data is missing. 

Create an object on source bucket with custom headers. 
Mirror to another bucket.
Confirm headers are not present. 

## mc version
# ./mc --version
mc version 0.0.0

That must be a joke.  :) I used code from @harshavardhana which was designed to improve throughput performance for mc mirror. 

## System information

CentOS 7.2.  
Scality RS2 connector to Scality S3 connector.

harshavardhana:
Fixes #1923

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -192,8 +192,7 @@ func (f *fsClient) Watch(params watchParams) (*watchObject, *probe.Error) {
 
 /// Object operations.
 
-// Put - create a new file.
-func (f *fsClient) Put(reader io.Reader, size int64, contentType string, progress io.Reader) (int64, *probe.Error) {
+func (f *fsClient) put(reader io.Reader, size int64, metadata map[string][]string, progress io.Reader) (int64, *probe.Error) {
 	// ContentType is not handled on purpose.
 	// For filesystem this is a redundant information.
 
@@ -352,6 +351,11 @@ func (f *fsClient) Put(reader io.Reader, size int64, contentType string, progres
 	return totalWritten, nil
 }
 
+// Put - create a new file with metadata.
+func (f *fsClient) Put(reader io.Reader, size int64, metadata map[string][]string, progress io.Reader) (int64, *probe.Error) {
+	return f.put(reader, size, nil, progress)
+}
+
 // ShareDownload - share download not implemented for filesystem.
 func (f *fsClient) ShareDownload(expires time.Duration) (string, *probe.Error) {
 	return "", probe.NewError(APINotImplemented{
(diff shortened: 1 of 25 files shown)


cmd/client-fs.go
```diff
@@ -352,6 +351,19 @@ func (f *fsClient) Put(reader io.Reader, size int64, contentType string, progres
...
+func (f *fsClient) PutWithMetadata(reader io.Reader, size int64, metadata map[string][]string, progress io.Reader) (int64, *probe.Error) {
+	return f.put(reader, size, metadata, progress)
+}
+
+// Put - create a new file.
+func (f *fsClient) Put(reader io.Reader, size int64, contentType string, progress io.Reader) (int64, *probe.Error) {
+	metadata := map[string][]string{
+		"Content-Type": {contentType},
+	}
+	return f.put(reader, size, metadata, progress)
```
vadmeste: We are not doing anything with metadata in FS, I think it is better to avoid passing metadata here, just for better code clarity.

cmd/client-s3.go
```diff
@@ -467,6 +467,39 @@ func (c *s3Client) Get() (io.Reader, *probe.Error) {
...
+			return nil, nil, probe.NewError(ObjectMissing{})
+		}
+		return nil, nil, probe.NewError(e)
+	}
+	objInfo, e := reader.Stat()
+	if e != nil {
+		return nil, nil, probe.NewError(e)
+	}
+	metadata := objInfo.Metadata
+	metadata.Set("Content-Type", objInfo.ContentType)
```
vadmeste: Not able to guess why we are setting content-type in metadata, could you add a comment ?

cmd/client-s3.go
```diff
@@ -507,6 +540,58 @@ func (c *s3Client) Copy(source string, size int64, progress io.Reader) *probe.Er
 	return nil
 }
 
+// PutObjectMetadata - upload an object with custom metadata.
+func (c *s3Client) PutWithMetadata(reader io.Reader, size int64, metadata map[string][]string, progress io.Reader) (int64, *probe.Error) {
+	bucket, object := c.url2BucketAndObject()
+	_, ok := metadata["Content-Type"]
```
harshavardhana: That is okay but client should set one of its own.  S3 doesn't say that if client should set or not set.   So a server implementation doesn't govern the client behavior.

cmd/client-s3.go
harshavardhana: We have to since Put needs it, we are kind of extracting the ContentType field back into metadata for consumption by caller.

cmd/client-fs.go
harshavardhana: We decided to avoid it. So can set this to nil.

PR #1943, opened 2017-01-02
Add policy links command

### LINKED ISSUE #1873: [policy] mc policy links s3/mybucket to list fully qualified URLs
This was requested by @atomi to list public URLs for external processing.

vadmeste:
Fixes #1873

### DIFF
--- a/cmd/policy-main.go
+++ b/cmd/policy-main.go
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"encoding/json"
+	"strings"
 
 	"github.com/fatih/color"
 	"github.com/minio/cli"
@@ -26,7 +27,12 @@ import (
 )
 
 var (
-	policyFlags = []cli.Flag{}
+	policyFlags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "recursive, r",
+			Usage: "List recursively.",
+		},
+	}
 )
 
 // Set public policy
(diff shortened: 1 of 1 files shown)


vadmeste:
This PR is ready for review, the only existing problem is that it accepts --recursive and --json flag when they are before 'links' command and not after, let's discuss other issues if found first.

cmd/policy-main.go
```diff
@@ -206,72 +255,151 @@ func doGetAccessRules(targetURL string) (r map[string]string, err *probe.Error)
...
+		for content := range clnt.List(isRecursive, isIncomplete, DirFirst) {
+			if content.Err != nil {
+				errorIf(content.Err.Trace(clnt.GetURL().String()), "Unable to list folder.")
+				continue
+			}
+			// Construct the message to be displayed to the user
+			msg := policyLinksMessage{}
+			msg.Status = "success"
+			// Convert any os specific delimiters to "/".
+			msg.FullURL = filepath.ToSlash(content.URL.String())
```
harshavardhana: Don't do 'ToSlash' for URLs on windows the output will have '\' which is unexpected.

cmd/policy-main.go
```diff
@@ -206,72 +255,151 @@ func doGetAccessRules(targetURL string) (r map[string]string, err *probe.Error)
...
+			if content.Type.IsDir() {
+				msg.Type = "folder"
```
harshavardhana: We can skip the 'CommonPrefix' since that is not useful in downloading the object itself.

harshavardhana: We can remove the Type itself it just prints fully qualified object URLs.

vadmeste:
All comments are addressed

PR #1950, opened 2017-01-06
service/status: Improve json output depending on the backend

### LINKED ISSUE #1948: admin: status shouldn't print XL related struct in json for FS
## Expected behaviour
```
$ mc admin service status home --json
{"status":"success","storageInfo":{"Total":177038229504,"Free":115952594944,"Backend":{"Type":FS"}}}
```

## Actual behaviour
```
$ mc admin service status home --json
{"status":"success","storageInfo":{"Total":177038229504,"Free":115952594944,"Backend":{"Type":1,"OnlineDisks":0,"OfflineDisks":0,"ReadQuorum":0,"WriteQuorum":0}}}
```

Should just have type and should point to something meaningful not "1" to "FS or "XL" or something and Disks related info shouldn't be printed if absent.

Start master minio server and use the above command

## mc version
master

vadmeste:
Type field in the json output contains 'FS' or 'XL', also ead and write quorums information are hidden and online and offline disks are shown only when the backend storage is XL

Fixes #1948

### DIFF
--- a/cmd/admin-service-status.go
+++ b/cmd/admin-service-status.go
@@ -21,7 +21,9 @@ import (
 	"fmt"
 
 	humanize "github.com/dustin/go-humanize"
+	"github.com/fatih/color"
 	"github.com/minio/cli"
+	"github.com/minio/mc/pkg/console"
 	"github.com/minio/minio/pkg/madmin"
 	"github.com/minio/minio/pkg/probe"
 )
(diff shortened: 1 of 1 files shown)


cmd/admin-service-status.go
```diff
@@ -122,10 +139,27 @@ func mainAdminServiceStatus(ctx *cli.Context) error {
...
 
-	printMsg(serviceStatusMessage{StorageInfo: status})
+	storageInfo := backendStatus{
+		Total: st.Total,
+		Free:  st.Free,
+	}
+
+	if st.Backend.Type == madmin.XL {
+		storageInfo.Backend = xlBackend{
+			Type:         "XL",
```
harshavardhana: Keep "XL" and "FS" as types instead of simple strings to avoid typos.

cmd/admin-service-status.go
```diff
@@ -51,10 +51,29 @@ EXAMPLES:
 `,
 }
```
harshavardhana: Can you have a commit message concise i.e it doesn't not wrap around.  and also have a 50-70 column width so that the commit messages are warping. 

Best style is to observe linux kernel commit logs.

cmd/admin-service-status.go
```diff
@@ -51,10 +51,37 @@ EXAMPLES:
...
+
+// fsBackend contains specific FS storage information
+type fsBackend struct {
+	Type backendType
+}
+
+// xlBackend contains specific XL storage information
+type xlBackend struct {
+	Type         backendType
+	OnlineDisks  int
```
harshavardhana: For json we can use the lower case like how we use with rest of the cli operations.

```
Type backendType `json:backendType`
```

cmd/admin-service-status.go
```diff
@@ -63,10 +90,8 @@ func (u serviceStatusMessage) String() string {
 		humanize.IBytes(uint64(u.StorageInfo.Total)),
 		humanize.IBytes(uint64(u.StorageInfo.Free)),
 	)
-	if u.StorageInfo.Backend.Type == madmin.XL {
-		msg += fmt.Sprintf(" Online Disks: %d, Offline Disks: %d\n",
-			u.StorageInfo.Backend.OnlineDisks,
-			u.StorageInfo.Backend.OfflineDisks)
+	if v, ok := u.StorageInfo.Backend.(xlBackend); ok {
+		msg += fmt.Sprintf(" Online Disks: %d, Offline Disks: %d\n", v.OnlineDisks, v.OfflineDisks)
```
harshavardhana: Can we also choose a color combination for this? You can choose something which is closer to `mc ls`

vadmeste:
@harshavardhana, all comments are addressed.

PR #1955, opened 2017-01-09
fixed formatting for  config host list command

poornas:


harshavardhana:
Looks like i didn't see the alignment @poornas - you might have to send another PR to fix the alignment to be left justified not right justified. 

```
fmt.Sprintf("%-*.*s", maxAlias, maxAlias, k)
```

harshavardhana:
also we need to sort this based on the Alias names perhaps something like this 

```
+// byAlias is a collection satisfying sort.Interface.
+type byAlias []hostMessage
+
+func (d byAlias) Len() int           { return len(d) }
+func (d byAlias) Swap(i, j int)      { d[i], d[j] = d[j], d[i] }
+func (d byAlias) Less(i, j int) bool { return d[i].Alias < d[j].Alias }
+
 // listHosts - list all host URLs.
 func listHosts() {
        conf, err := loadMcConfig()
@@ -275,14 +283,20 @@ func listHosts() {
                        maxAlias = len(k)
                }
        }
+       var hosts []hostMessage
        for k, v := range conf.Hosts {
-               printMsg(hostMessage{
+               hosts = append(hosts, hostMessage{
                        op:        "list",
-                       Alias:     fmt.Sprintf("%*.*s", maxAlias, maxAlias, k),
+                       Alias:     k,
                        URL:       v.URL,
                        AccessKey: v.AccessKey,
                        SecretKey: v.SecretKey,
                        API:       v.API,
                })
        }
+       sort.Sort(byAlias(hosts))
+       for _, host := range hosts {
+               host.Alias = fmt.Sprintf("%-*.*s", maxAlias, maxAlias, host.Alias)
+               printMsg(host)
+       }
 }
```

harshavardhana:
without sorting is also fine.. just that sorting will make it a consistent output.

PR #1956, opened 2017-01-10
sort  and format list host output

poornas:


harshavardhana:
You might have submit only new changes since your previous PR is already merged. You can rebase locally.

cmd/config-host-main.go
```diff
@@ -275,14 +283,22 @@ func listHosts() {
...
 	}
+	var hosts []hostMessage
+	sort.Sort(byAlias(hosts))
 	for k, v := range conf.Hosts {
-		printMsg(hostMessage{
+
+		hosts = append(hosts, hostMessage{
 			op:        "list",
-			Alias:     fmt.Sprintf("%*.*s", maxAlias, maxAlias, k),
+			Alias:     fmt.Sprintf("%-*.*s", maxAlias, maxAlias, k),
```
harshavardhana: You don't need to do it twice.. Do it once below and here keep it as 'Alias' itself. So that we can sort it properly.

cmd/config-host-main.go
```diff
@@ -275,14 +283,22 @@ func listHosts() {
 			maxAlias = len(k)
 		}
 	}
+	var hosts []hostMessage
+	sort.Sort(byAlias(hosts))
```
harshavardhana: This has to come later, its not useful to add here..

PR #1958, opened 2017-01-10
Add new heal command

### LINKED ISSUE #1921: admin: Implement heal command
## Expected behaviour
Having a heal command to list buckets/objects that need to be healed and effectively heal them.

## Actual behaviour
Not having heal command

## mc version
All
## System information
All

vadmeste:
Fixes #1921

### DIFF
--- a/cmd/admin-heal-list.go
+++ b/cmd/admin-heal-list.go
@@ -1,0 +1,213 @@
+/*
+ * Minio Client (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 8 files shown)


harshavardhana:
What we also need is a fake option

```mc admin heal --fake --recursive alias``` 

heals '`format.json``, ``buckets`` and then subsequently ``objects`` 

Otherwise 

```mc admin heal --fake alias``` only attempts to heal `format.json` you will have to manually provide buckets that needs to be healed.  Closer analogy to borrow from is to think like `fsck` 

```
$ fsck
Phase 1 (metadata)
Phase 2 (directories)
Phase 3 (inodes)
``` 

```mc admin heal --fake --recursive alias/bucket/prefix``` would heal everything inside `prefix`.

harshavardhana:
We do not need a sub command `repair`

harshavardhana:
Can this is rebased @vadmeste - since healing code on server side is merged..

vadmeste:
> Can this is rebased @vadmeste - since healing code on server side is merged..

I need to add --fake flag and other stuffs that you described earlier in this issue.

I just wanted to confirm, we'll remove repair command and keep list command ?

harshavardhana:
> I just wanted to confirm, we'll remove repair command and keep list command ?

Yes 'repair' is not needed..

vadmeste:
This is ready for review:

The only missing thing is healing format which is not ready yet in madmin package.

Also, when a bucket is specified in argument, I just heal it directly. ListBucketsHeal() is only called in heal list command when bucket is not specified, like `mc admin heal list myminio/`

PR #1977, opened 2017-01-22
client: S3 and Admin clients should have TLSConfig.

harshavardhana:
TLSConfig should be allocated regardless of the
URL scheme so that the re-direction works properly.

Without this change if a server is configured with
self signed certs redirection would fail with

```
x509: certificate signed by unknown authority
```

Allocate TLSConfig always with default values.

### DIFF
--- a/cmd/client-admin.go
+++ b/cmd/client-admin.go
@@ -19,9 +19,11 @@ package cmd
 import (
 	"crypto/tls"
 	"hash/fnv"
+	"net"
 	"net/http"
 	"net/url"
 	"sync"
+	"time"
 
 	"github.com/minio/mc/pkg/httptracer"
 	"github.com/minio/minio/pkg/madmin"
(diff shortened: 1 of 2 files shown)


cmd/client-admin.go
```diff
@@ -67,16 +67,14 @@ func newAdminFactory() func(config *Config) (*madmin.AdminClient, *probe.Error)
...
-				transport = &http.Transport{
-					TLSClientConfig: tlsConfig,
-				}
+			// Keep TLS config.
+			tlsConfig := &tls.Config{RootCAs: globalRootCAs}
+			if config.Insecure {
+				tlsConfig.InsecureSkipVerify = true
+			}
+
+			var transport http.RoundTripper = &http.Transport{
```
vadmeste: This way, only TLSClientConfig is set but all other http.Transport fields are nil, which is not so good for us, can you use default values from http.DefaultTransport (https://golang.org/pkg/net/http/#RoundTripper)

harshavardhana: Oops yes

Issue #1818, opened 2016-09-03
mc: Missing normalization for UTF-8 in keys causes mismatch on mirroring


felixbuenemann:
If files containing UTF-8 characters that have multiple possible representations are uploaded to an s3 bucket and then mirrored from a client with a different representation, the filename on disk will use the different representation, which causes the files to be re-transfered every time a mirror operation started, because the binary representation of the local filename and the s3 key differs.

Event worse, if those files are transferred back to s3 it will create duplicate keys in the bucket with different representations which is very confusing for both clients and humans, because it looks like the same key exists multiple times.

A common example for this is umlauts. Umlauts can be stored precomposed in utf-8 (eg. ü) or decomposed (eg. u+¨). Because Windows or Linux uses the precomposed UTF-8 presentation and Mac OS X (at least on HFS/HFS+) uses the decomposed representation, the bytes for both presentations will differ causing the above mentioned issue.

One solution to this problem would be to convert the source key to decomposed utf-8 before comparing on download and check for both a precomposed and decomposed form of the key on upload.
## mc version

Version: 2016-08-21T03:02:49Z
Release-tag: RELEASE.2016-08-21T03-02-49Z
Commit-id: 768be74f74578137951f65874cfc2e454b64aca0
## System information

Mac OS X 10.11.6

felixbuenemann:
Btw. this issue also exists in other tools, eg. s3cmd (see s3tools/s3cmd#639).

According to some info I google Mac OS X filesystems use the UTF-8 NFD normalization and Linux uses NFC normalization and Mac OS X will convert all filenames to NFD when saved to a HFS/HFS+ or UFS filesystem.

Windows NTFS apparently allows storing both NFC and NFD normalized versions of the same filenames along each other.

harshavardhana:
We don't do any special UTF-8 conversions, natively Go supports UTF-8 do not yet know if Go indeed supports UTF-8 OS X as well.

harshavardhana:
We have to use something like - https://godoc.org/golang.org/x/text/unicode/norm

harshavardhana:
This is fixed and  new mc has been released.

PR #1978, opened 2017-01-23
Mirror: normalize keys before comparing

awwalker:
Fix for #1818


harshavardhana:
You need to vendorize the dependencies use `make pkg-add PKG=golang.org/x/text/unicode/norm`

harshavardhana:
Can you add an example for a filename where normalization helps?

PR #1979, opened 2017-01-23
	admin: service status uses new response format

vadmeste:


vadmeste:
Ready for review and merge

PR #1980, opened 2017-01-24
admin: Heal format in heal command

vadmeste:


cmd/admin-heal.go
```diff
@@ -128,17 +128,18 @@ func mainAdminHeal(ctx *cli.Context) error {
...
 	bucket, object := splits[1], splits[2]
 
-	// Heal format if bucket is not specified
+	var e error
+
+	// Heal format if bucket is not specified and quit immediately
 	if bucket == "" {
-		// TBD, Heal format - not ready yet
-		console.Println("Healing format is not implemented yet.")
+		e = client.HealFormat()
```
harshavardhana: HealFormat should have a fake behavior as well.. i think this is not fixed yet.

harshavardhana:
Can you re-vendorize @vadmeste

vadmeste:
> Can you re-vendorize @vadmeste

Done

PR #1984, opened 2017-01-25
[Need review] mirror: Delete session data before watching events

vadmeste:
When the regular mirroring completes successfully, remove immediately
session file before starting to watch the server.

What's done here, after a successful phase one mirror, indicate that we don't need anymore to save session file when the user presses control-c

### DIFF
--- a/cmd/mirror-main.go
+++ b/cmd/mirror-main.go
@@ -102,6 +102,9 @@ type mirrorSession struct {
 	// the channel to trap SIGKILL signals
 	trapCh <-chan bool
 
+	// Indicate if the current session file is removed from disk
+	saveOnDisk bool
+
 	// channel for status messages
 	statusCh chan URLs
 	// channel for urls to harvest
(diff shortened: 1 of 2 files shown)


harshavardhana:
Can you explore if the queue.go is needed anymore.. Since we build the session harvestingURLs in the current style would not be needed..

vadmeste:
> Can you explore if the queue.go is needed anymore.. Since we build the session harvestingURLs in the current style would not be needed..

I made some investigations: Queue is similar to Channel here, however the good thing about it is that when  closed, it doesn't crash in contrary of channel. However, one missing thing is that Queue has unlimited size which is not good.

I tried to replace queue with channel as a first try. It looks like we need little work to do. However I have many questions, like, why are we using session in the first place since mirror command is also incremental.

Let's discuss internally.

harshavardhana:
This PR is not useful anymore. Can you close this @vadmeste ?

PR #1985, opened 2017-01-25
heal: Show help when alias is not specified

vadmeste:


cmd/admin-heal.go
```diff
@@ -59,6 +60,7 @@ USAGE:
 FLAGS:
   {{range .Flags}}{{.}}
   {{end}}
+
```
harshavardhana: why this space? - this would add double space.

cmd/admin-heal.go
```diff
@@ -99,10 +101,12 @@ func (u healObjectMessage) JSON() string {
 // checkAdminHealSyntax - validate all the passed arguments
 func checkAdminHealSyntax(ctx *cli.Context) {
 	if len(ctx.Args()) != 1 {
-		cli.ShowCommandHelpAndExit(ctx, "heal", 1) // last argument is exit code
+		cli.ShowSubcommandHelp(ctx)
```
harshavardhana: We can enhance cli to have `cli.ShowSubcommandHelpAndExit()`

cmd/admin-heal.go
vadmeste: > why this space? - this would add double space.

:+1:

cmd/admin-heal.go
vadmeste: I found a hacky change but it seems that my approach doesn't work well with cli package.


```
  1 package main
  2 
  3 import (
  4         "fmt"
  5         "os"
  6 
  7         "github.com/urfave/cli"
  8 )
  9 
 10 func main() {
 11 
 12         adminServiceCmd := cli.Command{
 13                 Name: "service",
 14                 Action: func(c *cli.Context) {
 15                         fmt.Printf("Run '%s'\n", c.Command.Name)
 16                 },
 17         }
 18 
 19         adminCmd := cli.Command{
 20                 Name:        "admin",
 21                 Subcommands: []cli.Command{adminServiceCmd},
 22                 Action: func(c *cli.Context) {
 23                         fmt.Printf("Run '%s'\n", c.Command.Name)
 24                 },
 25         }
 26 
 27         app := cli.NewApp()
 28         app.Name = "lookup"
 29         app.Commands = []cli.Command{adminCmd}
 30 
 31         app.Run(os.Args)
 32 }
```
```
$ ./a.out admin
Run ''
$ ./a.out admin service
Run 'service'
```

If you type `$ ./a.out admin`, it doesn't show the command name as I expect.

vadmeste: That's why `cli.ShowCommandHelpAndExit(ctx, "heal", 1)` wasn't working in the first place. I'll open a new bug in the cli upstream. Meanwhile, probably I'll reform the structure of admin heal command and make it similar to what we do with policy command.

vadmeste: https://github.com/urfave/cli/issues/590

PR #1992, opened 2017-01-30
mirror: Avoid session and enable continuous mirror

vadmeste:


vadmeste:
Ready for review.

felixbuenemann:
Enabling watch mode by default is imho a bad idea, it breaks existing scripts which depend on mirror exiting when it's done. It also looks like there's no way to disable watch mode.

harshavardhana:
The breaking change was introduced on purpose and intentional is there a reason that you need this @felixbuenemann . For S3  mc would exit normally but for Minio it would start watching and wait.

felixbuenemann:
> For S3 mc would exit normally but for Minio it would start watching and wait.

I tried mc mirror local_folder s3/my-bucket and it just sat there waiting for changes instead of exiting, which means mc mirror can no longer be used for scripting.

harshavardhana:
> I tried mc mirror local_folder s3/my-bucket and it just sat there waiting for changes instead of exiting, which means mc mirror can no longer be used for scripting.

It is going to watch for FS on new events..

felixbuenemann:
I know, and it should not, because it makes scripting impossible.

felixbuenemann:
To be clear, it's fine for me to enable watch behavior with a switch, but it shouldn't be the default.

PR #1995, opened 2017-02-03
mirror: For non-s3 make sure we copy and exit properly.

harshavardhana:


harshavardhana:
Fixes #1996

vadmeste:
:+1: for removing graceful shutdown.

However, doing mirroring in two phases can miss some objects uploaded during mirroring and before starting to watch because prepareMirrorURLs() can miss them.

harshavardhana:
That was intentional @vadmeste we don't need to do that.

harshavardhana:
> However, doing mirroring in two phases can miss some objects uploaded during mirroring and before starting to watch because prepareMirrorURLs() can miss them.

Done added @vadmeste

PR #2006, opened 2017-02-08
Mirror entire instance

vadmeste:
Ready for first review

### DIFF
--- a/cmd/mirror-main.go
+++ b/cmd/mirror-main.go
@@ -52,6 +52,15 @@ var (
 			Name:  "remove",
 			Usage: "Remove extraneous file(s) on target.",
 		},
+		cli.StringFlag{
+			Name:  "region",
+			Usage: "Specify which region to select when creating new buckets.",
+			Value: "us-east-1",
+		},
+		cli.BoolFlag{
+			Name:  "a",
+			Usage: "Preserve bucket policies rules.",
+		},
 	}
 )
 
@@ -99,9 +108,6 @@ type mirrorJob struct {
 	// the channel to trap SIGKILL signals
 	trapCh <-chan bool
 
-	// Hold context information
-	context *cli.Context
-
 	// mutex for shutdown, this prevents the shutdown
 	// to be initiated multiple times
 	m *sync.Mutex
@@ -131,6 +137,8 @@ type mirrorJob struct {
 
 	sourceURL string
 	targetURL string
+
+	isFake, isForce, isRemove, isWatch bool
 }
 
 // mirrorMessage container for file mirror messages
@@ -179,8 +187,7 @@ func (c mirrorStatMessage) String() string {
 
 // doRemove - removes files on target.
 func (mj *mirrorJob) doRemove(sURLs URLs) URLs {
-	isFake := mj.context.Bool("fake")
-	if isFake {
+	if mj.isFake {
 		return sURLs.WithError(nil)
 	}
 
@@ -191,13 +198,12 @@ func (mj *mirrorJob) doRemove(sURLs URLs) URLs {
 	targetWithAlias := filepath.Join(sURLs.TargetAlias, sURLs.TargetContent.URL.Path)
 
 	// Remove extraneous file on target.
-	err := probe.NewError(removeSingle(targetWithAlias, isIncomplete, isFake, 0))
+	err := probe.NewError(removeSingle(targetWithAlias, isIncomplete, mj.isFake, 0))
 	return sURLs.WithError(err)
 }
 
 // doMirror - Mirror an object to multiple destination. URLs status contains a copy of sURLs and error if any.
 func (mj *mirrorJob) doMirror(sURLs URLs) URLs {
-	isFake := mj.context.Bool("fake")
 
 	if sURLs.Error != nil { // Erroneous sURLs passed.
 		return sURLs.WithError(sURLs.Error.Trace())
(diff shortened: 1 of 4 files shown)


vadmeste:
Ready for review

cmd/common-methods.go
```diff
@@ -180,3 +181,16 @@ func newClient(aliasedURL string) (Client, *probe.Error) {
 	}
 	return newClientFromAlias(alias, urlStrFull)
 }
+
+// newS3Client gives S3 client if the url represents an S3 server
+func newS3Client(url string) (*s3Client, *probe.Error) {
```
harshavardhana: why a function like this?

cmd/mirror-main.go
```diff
@@ -52,6 +52,15 @@ var (
 			Name:  "remove",
 			Usage: "Remove extraneous file(s) on target.",
 		},
+		cli.StringFlag{
+			Name:  "region",
+			Usage: "Specify which region to select when creating new buckets.",
+			Value: "us-east-1",
+		},
+		cli.BoolFlag{
+			Name:  "preserve-policies, a",
```
harshavardhana: we don't need the long option just keep it "a"

cmd/mirror-main.go
```diff
@@ -522,16 +516,148 @@ func newMirrorJob(ctx *cli.Context) *mirrorJob {
...
+		if d.Diff == differInFirst {
+			// Bucket only exists in the source, create the same bucket in the destination
+			err := newDstClt.MakeBucket(ctx.String("region"))
+			fatalIf(err, "Cannot created bucket in `"+newTgtURL+"`")
+		}
+
+		// Copy policy rules from source to dest if flag is activated
+		if ctx.Bool("preserve-policies") {
+			rules, err := newSrcClt.GetAccessRules()
+			fatalIf(err, "Cannot get policy rules")
```
harshavardhana: We don't need to fail we can proceed forward to the next one.

cmd/mirror-main.go
```diff
@@ -522,16 +516,148 @@ func newMirrorJob(ctx *cli.Context) *mirrorJob {
...
+		newSrcURL := srcURL + d.firstContent.URL.Path
+		newTgtURL := dstURL + d.firstContent.URL.Path
+
+		newSrcClt, _ := newS3Client(newSrcURL)
+		newDstClt, _ := newS3Client(newTgtURL)
+
+		if d.Diff == differInFirst {
+			// Bucket only exists in the source, create the same bucket in the destination
+			err := newDstClt.MakeBucket(ctx.String("region"))
+			fatalIf(err, "Cannot created bucket in `"+newTgtURL+"`")
```
harshavardhana: Don't fail, proceed to next one.

cmd/mirror-main.go
```diff
@@ -522,16 +516,148 @@ func newMirrorJob(ctx *cli.Context) *mirrorJob {
...
+		}
+
+		// Copy policy rules from source to dest if flag is activated
+		if ctx.Bool("preserve-policies") {
+			rules, err := newSrcClt.GetAccessRules()
+			fatalIf(err, "Cannot get policy rules")
+			// Set found rules to target bucket if permitted
+			for _, r := range rules {
+				originalRule, err := newDstClt.GetAccess()
+				fatalIf(err, "Cannot get bucket policy rules from `"+newTgtURL+"`")
```
harshavardhana: Don't fail proceed to next one.

cmd/mirror-main.go
```diff
@@ -522,16 +516,148 @@ func newMirrorJob(ctx *cli.Context) *mirrorJob {
...
+
+	wg.Wait()
+	close(work)
+
+	return nil
+}
+
+// mirrorSingleBucket
+func mirrorSingleBucket(srcURL, tgtURL string, ctx *cli.Context) error {
+	mj := newMirrorJob(srcURL, tgtURL, 1,
```
harshavardhana: Even in single bucket if "-a" is specified should be honored.

cmd/mirror-main.go
```diff
@@ -522,16 +516,148 @@ func newMirrorJob(ctx *cli.Context) *mirrorJob {
...
+
+	dstClt, err := newClient(tgtURL)
+	fatalIf(err, "Unable to initialize `"+tgtURL+"`")
+
+	mirrorAll := false
+
+	// Check if passed arguments could trigger a complete mirror instance
+	if srcClt.GetURL().Type == objectStorage && dstClt.GetURL().Type == objectStorage {
+		// Check if buckets are not passed
+		srcBucket, _ := srcClt.(*s3Client).url2BucketAndObject()
```
harshavardhana: We should never have code like this.. unless really really needed. what problem are you trying to solve here?

vadmeste: > We should never have code like this.. unless really really needed. what problem are you trying to solve here?

Here, I needed to check if the user passed any buckets in arguments to know if we are going to mirror all buckets or it is just a simple mirror

cmd/common-methods.go
vadmeste: > why a function like this?

Simplification purpose, this returns an error when this is not an s3 url.

vadmeste:
@harshavardhana, I rewrite some codes after you gave some comments, but I didn't address all of them until we discuss them

cmd/common-methods.go
harshavardhana: Actually this is spilling the abstraction ..

vadmeste: > Actually this is spilling the abstraction ..

You mean mirroring instance can be somehow abstracted too ?

PR #2009, opened 2017-02-09
admin: Rename creds to password subcommand

### LINKED ISSUE #2002: Move `mc admin service creds` to `mc admin password`
## Expected behaviour
```
mc admin password ALIAS ACCESS-KEY SECRET-KEY with description 
```

and Description should be 
```
Change server access and secret keys.
```

## Actual behaviour
Currently the command is at the wrong place `mc admin service creds` 

## mc version
Current master

vadmeste:
Use new command "mc admin password" to change creds

Fixes #2002


vadmeste:
@harshavardhana, Done

PR #2018, opened 2017-02-17
Rename mc to minioc

slaxor:
mc is already occupied by the still popular "midnight commander"
therefore he who comes second must suffer ;)

I have changed all /mc/i refs to minioc (obeyeing the given upper and
lower case)

Since the refactor works under the assumption that this repo will be renamed to minioc as well testing was a little bit flaky. The dockerfile for example didn't build because the repo doesn't exist (yet)

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -4,8 +4,8 @@
 
 ## Steps to reproduce the behaviour
 
-## mc version
-- (paste output of `mc version`)
+## minioc version
+- (paste output of `minioc version`)
 
 ## System information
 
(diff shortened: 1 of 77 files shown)


harshavardhana:
Thanks for the PR @slaxor historically we debated this and have closed bugs in this [regard](https://github.com/minio/mc/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20midnight%20commander) , We won't be doing this or changing the name of the project.

slaxor:
ah ok thats a shame. i like your tool but renaming it on my systems is a bit of a nuisance.

PR #2024, opened 2017-02-19
cli: Make sure to add "--help" flag for subcommands.

harshavardhana:


vadmeste:
This change works but I didnt' understand what's going on, HelpFlag is already defined in cli package.

harshavardhana:
What we need is --help to be listed for commands and subcommands this is what the patch does.

PR #2026, opened 2017-02-19
mc: Rewrite update and version command.

harshavardhana:
- Supports docker image
- Supports showing how old your current
  `mc` version is.

### DIFF
--- a/cmd/update-main.go
+++ b/cmd/update-main.go
@@ -1,5 +1,5 @@
 /*
- * Minio Client (C) 2015 Minio, Inc.
+ * Minio Client (C) 2015, 2016, 2017 Minio, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,10 +17,13 @@
 package cmd
 
 import (
+	"bytes"
 	"encoding/json"
-	"errors"
+	"fmt"
 	"io/ioutil"
 	"net/http"
+	"os"
+	"os/exec"
 	"runtime"
 	"strings"
 	"time"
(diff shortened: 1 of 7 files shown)


cmd/update-main.go
```diff
@@ -31,204 +33,254 @@ import (
...
-	if releaseDateSplits[0] != "mc" {
-		return time.Time{}, probe.NewError(errors.New("Update data malformed, missing mc tag"))
+
+	return releaseTime, probe.NewError(e)
+}
+
+// GetCurrentReleaseTime - returns this process's release time.  If it is official minio version,
+// parsed version is returned else minio binary's mod time is returned.
+func GetCurrentReleaseTime() (releaseTime time.Time, err *probe.Error) {
+	return getCurrentReleaseTime(Version, os.Args[0])
```
vadmeste: os.Args[0] is sometimes 'mc' if the binary is in one of PATH directories, os.Stat() would fail because we are not providing it an absolute path. With go 1.8, there is os.Executable() which helps.

cmd/update-main.go
```diff
@@ -31,204 +33,254 @@ import (
...
 
-	var updateMsg updateMessage
-	var errMsg string
-	var err *probe.Error
-	// Check for update.
-	if ctx.Bool("experimental") {
-		updateMsg, errMsg, err = getReleaseUpdate(mcUpdateExperimentalURL)
-	} else {
-		updateMsg, errMsg, err = getReleaseUpdate(mcUpdateStableURL)
+	setGlobalsFromContext(ctx)
```
vadmeste: We changed recently to setting setGlobalsFromContext in Command.Before, that way we can support global flags being inserted in any place.

cmd/update-main.go
harshavardhana: This would be a bug on server too then can you fix it there @vadmeste ?

cmd/update-main.go
harshavardhana: Can you make that change for Minio server too @vadmeste ?

PR #2036, opened 2017-02-25
admin: Add info cmd and simplify service status response

vadmeste:
info command will show:
uptime, storage info (total, freel), server's region etc..

status command will only show if server online/offline status and uptime

### DIFF
--- a/cmd/admin-info.go
+++ b/cmd/admin-info.go
@@ -1,0 +1,207 @@
+/*
+ * Minio Client (C) 2016 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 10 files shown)


cmd/admin-info.go
```diff
@@ -0,0 +1,207 @@
...
+	Name:   "info",
+	Usage:  "Get information of a Minio server",
+	Action: mainAdminInfo,
+	Before: setGlobalsFromContext,
+	Flags:  append(adminInfoFlags, globalFlags...),
+	CustomHelpTemplate: `NAME:
+   {{.HelpName}} - {{.Usage}}
+
+USAGE:
+   {{.HelpName}} ALIAS
```
harshavardhana: One space character can you keep this consistent with flags as well?

vadmeste: > One space character can you keep this consistent with flags as well?

Done

PR #2037, opened 2017-02-26
admin: Add config get command

### LINKED ISSUE #2034: Implement get config command
## Expected behaviour
Having admin get config command

## Actual behaviour
Not having get config command

Obvious

## mc version
All

## System information
All

vadmeste:
The new command is `mc admin config get alias/` and returns a plain config file 

Fixes #2034

### DIFF
--- a/cmd/admin-config-get.go
+++ b/cmd/admin-config-get.go
@@ -1,0 +1,95 @@
+/*
+ * Minio Client (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"encoding/json"
+
+	"github.com/minio/cli"
+	"github.com/minio/minio/pkg/probe"
+)
+
+var adminConfigGetCmd = cli.Command{
+	Name:   "get",
+	Usage:  "Get config of a Minio server/cluster.",
+	Before: setGlobalsFromContext,
+	Action: mainAdminConfigGet,
+	Flags:  globalFlags,
+	CustomHelpTemplate: `NAME:
+  {{.HelpName}} - {{.Usage}}
+
+USAGE:
+  {{.HelpName}} ALIAS/
+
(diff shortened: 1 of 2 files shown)


vadmeste:
Ready for review.

cmd/admin-config-get.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"encoding/json"
+
+	"github.com/minio/cli"
+	"github.com/minio/minio/pkg/probe"
+)
+
+var (
+	adminConfigGetFlags = []cli.Flag{}
```
harshavardhana: We don't need this as placeholder.

cmd/admin-config-get.go
```diff
@@ -0,0 +1,99 @@
...
+	Name:   "get",
+	Usage:  "Get config of a Minio server/cluster.",
+	Before: setGlobalsFromContext,
+	Action: mainAdminConfigGet,
+	Flags:  append(adminConfigGetFlags, globalFlags...),
+	CustomHelpTemplate: `NAME:
+   {{.HelpName}} - {{.Usage}}
+
+USAGE:
+   {{.HelpName}} ALIAS/
```
harshavardhana: Space character here..

cmd/admin-config.go
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import "github.com/minio/cli"
+
+var (
+	adminConfigFlags = []cli.Flag{}
```
harshavardhana: Same we don't need..

cmd/admin-config-get.go
```diff
@@ -0,0 +1,99 @@
...
+
+// String colorized service status message.
+func (u configGetMessage) String() string {
+	return string(u.Config)
+}
+
+// JSON jsonified service status Message message.
+func (u configGetMessage) JSON() string {
+	u.Status = "success"
+	statusJSONBytes, e := json.Marshal(u)
```
harshavardhana: what happens with `mc --json` for this? because the `[]byte` is already JSON.

vadmeste: > what happens with mc --json for this? because the []byte is already JSON.

json.Marshal already transforms []byte to base64

vadmeste:
All done.

cmd/admin-config-get.go
harshavardhana: Oh we shouldn't do that @vadmeste

vadmeste: > Oh we shouldn't do that @vadmeste

In all cases, we need to apply a transformation of that []byte because it is already inside a json response. For example, quoting ".

harshavardhana: I think base64 is fine because from an API point of view we can't do anything else. They can easily script this out. So i guess --json printing base64 is fine.

vadmeste:
Can we merge this @harshavardhana ?

Issue #2029, opened 2017-02-20
status error seems to be false when connection cannot create


limbo127:
[wso]mc  mirror  s3/${_awsBucket}/upgrade/ /data/upload/upgrade/                               
mc: <ERROR> Unable to stat source ‘s3/ceb/upgrade/’. Get https://s3.amazonaws.com/ceb/?location=: read tcp 127.0.0.1:49546->127.0.0.1:14380: read: connection reset by peer
 0 B / ? ┃░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░▓┃mc: <ERROR> Failed on 'https://s3.amazonaws.com/ceb/upgrade/' Get https://s3.amazonaws.com/ceb/?location=: read tcp 127.0.0.1:49550->127.0.0.1:14380: read: connection reset by peer
[wso]echo $?
0

In case of connection reset by peer du to a proxy error ( not present for example), status is 0 , must be 1.

If I stop proxy during process, mc status error is > 0

Regards,
Nicolas

limbo127:
sorry, 
when i'm stopping proxy during process, mc status error is also 0 . 
But when I stop mc , with ctrl+c , status is > 0 .

PR #2043, opened 2017-02-27
mirror: Return exit code after mirror/watch error

vadmeste:
Fixes https://github.com/minio/mc/issues/2029

### DIFF
--- a/cmd/mirror-main.go
+++ b/cmd/mirror-main.go
@@ -133,6 +133,11 @@ type mirrorJob struct {
 	// channel for status messages
 	statusCh chan URLs
 
+	// Store last watch error
+	watchErr *probe.Error
+	// Store last mirror error
+	mirrorErr *probe.Error
+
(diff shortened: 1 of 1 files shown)


cmd/mirror-main.go
```diff
@@ -637,7 +658,10 @@ func mainMirror(ctx *cli.Context) error {
 	srcURL := args[0]
 	tgtURL := args[1]
 
-	runMirror(srcURL, tgtURL, ctx)
+	if err := runMirror(srcURL, tgtURL, ctx); err != nil {
+		exitStatus(globalErrorExitStatus)
```
harshavardhana: How will the subsequent return work here?

vadmeste: yeah you are right, but even like that it works (it seems the next line is enough), I'll fix that to be more generic.

vadmeste: Done

Issue #2048, opened 2017-02-28
mc mirror --force --delete when source bucket doesn't exist will remove target objects


vadmeste:
## Expected behaviour
If I pass a non existent bucket, mc mirror just quits and avoid doing anything.

## Actual behaviour
Passing a non existent bucket will just remove all objects in target

## Steps to reproduce the behaviour
`mc mirror --force --delete --watch myminio/wrongbucket/ ./\ will remove files in the current dir

## mc version
master (0934a39d731e8f80adcc66e67007eb223662abc3)
## System information
All

vadmeste:
Fixed by #2049

PR #2049, opened 2017-02-28
	difference() sends errors on channel

vadmeste:
It doesn't make sense to continue to calculate to the difference
between the source and target when source or target returns a listing error

Fixes 2^11 (#2048)

### DIFF
--- a/cmd/diff-main.go
+++ b/cmd/diff-main.go
@@ -173,6 +173,10 @@ func doDiffMain(firstURL, secondURL string) error {
 
 	// Diff first and second urls.
 	for diffMsg := range objectDifference(firstClient, secondClient, firstURL, secondURL) {
+		if diffMsg.Error != nil {
+			errorIf(diffMsg.Error, "Unable to calculate objects difference.")
+			break
+		}
 		printMsg(diffMsg)
 	}
 
--- a/cmd/difference.go
+++ b/cmd/difference.go
@@ -17,7 +17,6 @@
 package cmd
 
 import (
-	"fmt"
 	"strings"
 	"unicode/utf8"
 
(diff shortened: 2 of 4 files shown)


vadmeste:
@harshavardhana, when an error occurs in difference():
*) If --remove flag is passed, objects in target will be removed which is dangerous
*) If --remove is not passed, mirror can try to upload 1000 objects though they could be already existing in target.

harshavardhana:
> *) If --remove flag is passed, objects in target will be removed which is dangerous

But the error in difference is only per object? - why would a target be deleted if difference() function stops returning the diff. ?

vadmeste:
> *) If --remove flag is passed, objects in target will be removed which is dangerous
> But the error in difference is only per object? - why would a target be deleted if difference() function stops returning the diff. ?

yeah my point is what happens if we continue to apply difference().

harshavardhana:
> yeah my point is what happens if we continue to apply difference().

We can continue the error only happened for one object. We should continue if we can - why would it be dangerous?

vadmeste:
> We can continue the error only happened for one object. 

This is a listing error, this is not only about one object, it could be about one thousand objects or more. In case of --remove, mirror will try to delete objects that it didn't receive when an error occur.

harshavardhana:
> This is a listing error, this is not only about one object, it could be about one thousand objects or more. In case of --remove, mirror will try to delete objects that it didn't receive when an error occur.

I see what you mean bi-directional diff causes a problem if source actually kept returning error all the files in the 'target' would be wrongly considered as extraneous entries and deleted.   In a situation like this i think the ideal way is to fail and not proceed further as you suggested.  Since we can't tell for sure if the target does have extra files not in source, since we have partial data from the source anyways.

vadmeste:
> I see what you mean bi-directional diff causes a problem if source actually kept returning error all the files in the 'target' would be wrongly considered as extraneous entries and deleted. In a situation like this i think the ideal way is to fail and not proceed further as you suggested. Since we can't tell for sure if the target does have extra files not in source, since we have partial data from the source anyways.

This is ready for review.

PR #2051, opened 2017-02-28
admin: Add set config command

vadmeste:
Currently, it only work this way `cat config.json | mc admin config set myminio/`, 

Can we discuss if we need to add other way to pass the new config file ?

### DIFF
--- a/cmd/admin-config-set.go
+++ b/cmd/admin-config-set.go
@@ -1,0 +1,134 @@
+/*
+ * Minio Client (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/fatih/color"
+	"github.com/minio/cli"
+	"github.com/minio/mc/pkg/console"
+	"github.com/minio/minio/pkg/madmin"
+	"github.com/minio/minio/pkg/probe"
+)
+
+var adminConfigSetCmd = cli.Command{
+	Name:   "set",
+	Usage:  "Set new config file to a Minio server/cluster.",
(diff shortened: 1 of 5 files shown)


harshavardhana:
> Currently, it only work this way cat config.json | mc admin config set myminio/,
> 
> Can we discuss if we need to add other way to pass the new config file ?

It behaves like `mc pipe`  without any piped input it reads from stdin so that things like this work 

```
mc admin config set myminio/ < config.json-mod
```

```
mc cat config.json-mod | mc config set myminio/
```

```
mc admin config set myminio/ 
{
	"version": "14",
	"credential": {
		"accessKey": "USWUXHGYZQYFYFFIT3RE",
		"secretKey": "MOJRH0mkL1IPauahWITSVvyDrQbEEIwljvmxdq03"
	},
	"region": "us-east-1",
....
```

cmd/admin-config-set.go
```diff
@@ -0,0 +1,128 @@
...
+  1. Set server configuration of a Minio server/cluster.
+     $ cat myconfig | {{.HelpName}} myminio/
+
+`,
+}
+
+// configSetMessage container to hold locks information.
+type configSetMessage struct {
+	Status               string               `json:"status"`
+	SetConfigStatus      bool                 `json:"-"`
```
harshavardhana: If there is no plan to jsonify this you can keep this lower case.

cmd/admin-config-set.go
```diff
@@ -0,0 +1,128 @@
...
+	defer func() {
+		msg = console.Colorize("SetConfig", msg)
+	}()
+	if u.SetConfigStatus {
+		msg += "Setting new configuration file has been successful.\n"
+	} else {
+		msg += "Setting new configuration file has failed.\n"
+	}
+	for _, s := range u.SetConfigNodeSummary {
+		msg += fmt.Sprintf("  %s.. ", s.Name)
```
harshavardhana: Can you paste a sample output how this looks ?

vadmeste:
```
$ mc admin config set myminiold1/ < config.json
Setting new configuration file has been successful.
  172.17.0.2:9000.. OK.
  172.17.0.3:9000.. OK.
  172.17.0.4:9000.. OK.
  172.17.0.5:9000.. OK.
```

harshavardhana: Actually here the output would be nicer to be like - success is printed at the end based on the errors..

```
$ mc admin config set myminiold1/ < config.json
New configuration applied to `172.17.0.2:9000`  successfully.
New configuration applied to `172.17.0.3:9000`  successfully.
New configuration applied to `172.17.0.4:9000`  successfully.
New configuration applied to `172.17.0.5:9000`  successfully.
Setting new Minio configuration has been successful.
```

vadmeste: Yeah it is nicer like that.

PR #2056, opened 2017-03-02
[PROPOSAL] Add size output for `diff` command.

harshavardhana:
It is useful to know before hand what will be total
size to be copied when a user initiates `mc mirror`

### DIFF
--- a/cmd/diff-main.go
+++ b/cmd/diff-main.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"strings"
 
+	humanize "github.com/dustin/go-humanize"
 	"github.com/fatih/color"
 	"github.com/minio/cli"
 	"github.com/minio/mc/pkg/console"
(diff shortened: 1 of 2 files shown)


cmd/diff-main.go
```diff
@@ -67,11 +73,33 @@ type diffMessage struct {
...
 }
 
+type diffTotalSize struct {
+	Status     string `json:"status"`
+	SourceSize int64  `json:"sourceSize"`
+	TargetSize int64  `json:"targetSize"`
+}
+
+func (t diffTotalSize) String() string {
+	msg := console.Colorize("DiffSize", fmt.Sprintf("SourceSize: %s TargetSize: %s",
```
vadmeste: Probably we should only print the difference between sizes, easier for the user to read.

LGTM otherwise.

deekoder:
Is this PR relevant. Can we either merge it or should we close it?

harshavardhana:
> Is this PR relevant. Can we either merge it or should we close it?

This is relevant waiting on @abperiasamy input.

harshavardhana:
This is not going to be implemented instead we will do a UI change in #2090

PR #2069, opened 2017-03-11
Split config-host-main into remove, list, add commands

harshavardhana:
This PR also adds convenient aliases for

 - remove as `rm`
 - list as `ls`
 - add as `a`

Also brings much needed cleanup.


cmd/config-host-add.go
```diff
@@ -0,0 +1,140 @@
...
+USAGE:
+  {{.HelpName}} ALIAS URL ACCESS-KEY SECRET-KEY [API]
+
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}
+EXAMPLES:
+  1. Add Amazon S3 storage service under "mys3" alias. For security reasons turn off bash history momentarily.
+     $ set +o history
+     $ {{.HelpName}} add mys3 https://s3.amazonaws.com \
```
vadmeste: 'add' is not needed here (it is already shown with {{.HelpName}})

cmd/config-host-add.go
```diff
@@ -0,0 +1,140 @@
...
+                 BKIKJAA5BMMU2RHO6IBB V8f1CwQqAcwo80UEIJEjc5gVQUSSx5ohQ9GSrr12
+     $ set -o history
+
+  2. Add Amazon S3 accelerated storage service under "mys3-accel" alias. For security reasons turn off bash history momentarily.
+     $ set +o history
+     $ {{.HelpName}} add mys3-accel https://s3-accelerate.amazonaws.com \
```
vadmeste: Same as above.

cmd/config-host-list.go
```diff
@@ -0,0 +1,143 @@
...
+	var hosts []hostMessage
+	for k := range conf.Hosts {
+		if len(k) > maxAlias {
+			maxAlias = len(k)
+		}
+	}
+
+	for k, v := range conf.Hosts {
+		// If specific alias is requested, only print the requested one.
+		if alias != "" && alias == k {
```
vadmeste: I think it is more readable if we move out this block outside of for, besides it is better to print an error message when alias is specified but not found

```
if alias != "" {
  if v, ok := conf.Hosts[alias] && ok {
    hosts = []hostsMessage{{  .... }}
  } else {
     error("no alias is found")
  }
}
```

cmd/config-host-remove.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"github.com/fatih/color"
+	"github.com/minio/cli"
+	"github.com/minio/mc/pkg/console"
+)
+
+var configHostRemoveCmd = cli.Command{
+	Name:            "remove",
+	ShortName:       "rm",
+	Usage:           "Remove hosts from configuration file.",
```
vadmeste: You mean "remove a host" ?

cmd/config-host-add.go
harshavardhana: ah yes..

cmd/config-host-list.go
harshavardhana: We already skip that part in checkSyntaxt() where we throw an error when alias is wrong.

cmd/config-host-remove.go
harshavardhana: Yes 'a host'

PR #2070, opened 2017-03-14
policy/links: Do not list prefixes in recursive mode.

harshavardhana:


cmd/policy-main.go
```diff
@@ -320,6 +320,10 @@ func runPolicyLinksCmd(ctx *cli.Context) {
 				continue
 			}
 
+			if content.Type.IsDir() && isRecursive {
```
vadmeste: Why not hiding directories in all cases ?

harshavardhana: In-case of non-recursive we would want to show.. if there are other directories available for navigation. It is expected..

vadmeste: > In-case of non-recursive we would want to show.. if there are other directories available for navigation. It is expected..

Yes, it makes sense.

PR #2095, opened 2017-03-28
cli: Add bash and zsh autocomplete

### LINKED ISSUE #2094: mc should have auto-completion facility
## Expected behaviour

Like many common CLI utilities that are structured as a single program with numerous subcommands (e.g. git/docker), mc will benefit from auto-completion support, especially for the bash shell.

## mc version

harshavardhana:
Fixes #2094

### DIFF
--- a/autocomplete/bash_autocomplete
+++ b/autocomplete/bash_autocomplete
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+: ${PROG:=$(basename ${BASH_SOURCE})}
+
+_cli_bash_autocomplete() {
+     local cur opts base
+     COMPREPLY=()
+     cur="${COMP_WORDS[COMP_CWORD]}"
+     opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} --compgen )
+     COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
+     return 0
+ }
+
+ complete -F _cli_bash_autocomplete $PROG -q
--- a/README.md
+++ b/README.md
@@ -147,6 +147,8 @@ mc ls play
 ```
 <a name="everyday-use"></a>
 ## Everyday Use
+
+### Shell aliases
 You may add shell aliases to override your common Unix tools.
 
 ```sh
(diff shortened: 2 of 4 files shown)


autocomplete/bash_autocomplete
```diff
@@ -0,0 +1,14 @@
+#! /bin/bash
```
vadmeste: Unneeded space?

vadmeste: between `"!` and `/bin/bash`

autocomplete/bash_autocomplete
```diff
@@ -0,0 +1,14 @@
...
+_cli_bash_autocomplete() {
+     local cur opts base
+     COMPREPLY=()
+     cur="${COMP_WORDS[COMP_CWORD]}"
+     opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} --compgen )
+     COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
+     return 0
+ }
+  
+ complete -F _cli_bash_autocomplete $PROG
```
vadmeste: We probably need to add `-q` flag here to avoid printing update message when compgen is called.

harshavardhana: where does it show that?

harshavardhana:
```sh
$ PROG=mc source autocomplete/bash_autocomplete 
$ 
```

Don't see any message.

harshavardhana: oh you mean the `mc update` message?

vadmeste: > oh you mean the mc update message?

Besides, we used to have that banner which is shown for each command when mc binary becames outdated, we don't have that thing anymore?

harshavardhana: Fixed that too..

autocomplete/bash_autocomplete
harshavardhana: Fixed

PR #2097, opened 2017-03-29
diff: Change output UI

### LINKED ISSUE #2090: diff: ui change move from `only in first` to new format
## Expected behaviour

```
> /home/harsha/Photos/alice (11).jpg
> /home/harsha/Photos/alice (12).jpg
< https://play.minio.io:9000/testbucket/bin/7z
< https://play.minio.io:9000/testbucket/bin/7za
v https://play.minio.io:9000/testbucket/new-image.jpg
```

Here 

```
> - indicates file should be copied
< - indicates file should be deleted
v - indicates file differs in size. 
```

## Actual behaviour

Current output needs to be improved.

Just build from master

## mc version
master

## System information
All operating systems.

vadmeste:
Use >, < and ! marks to represent differences between
source and destination.

Fixes #2090

### DIFF
--- a/cmd/diff-main.go
+++ b/cmd/diff-main.go
@@ -52,12 +52,19 @@ FLAGS:
 DESCRIPTION:
   {{.Description}}
 
-EXAMPLES:
-   1. Compare a local folder with a folder on Amazon S3 cloud storage.
-      $ {{.HelpName}} ~/Photos s3/MyBucket/Photos
+OUTPUT:
+  Differences between source and destination are represented by mark notations with following meaning:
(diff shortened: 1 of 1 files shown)


cmd/diff-main.go
```diff
@@ -52,12 +52,19 @@ FLAGS:
 DESCRIPTION:
   {{.Description}}
 
+OUTPUT:
+  Differences between source and destination are represented by some mark notations which have the following meaning:
```
harshavardhana:
```
Differences between source and destination are represented by mark notations with following meaning:
```

PR #2098, opened 2017-03-30
Improve admin subcommand structure

krisis:
The proposal for changed `mc-admin` command structure is captured here: https://gist.github.com/krisis/98f4171e12bed9e9c022cd79b74db754. Removal of `heal-list` in favour of `--fake` flag is missing.


harshavardhana:
can you fix the build issue @krisis ?

Issue #2093, opened 2017-03-28
`mc share list` help text is missleading


badscooter23:
## Expected behaviour

`mc share list` should print out the `help` text but should not be an error
the `help` text for the command should inidcate that the user needs to specify a command (either
`upload` or `download`) after `list`

Usage
   mc share list COMMAND

## Actual behaviour

Today the help text is incomplete / incorrect... it does not indicate that the user needs to 
specify `upload` or `download` after `list`. This is confusing.... it looks like `mc share list` is 
valid and (sort of) implies that `mc share list` would do something useful (like list all non-expired shared urls). 

## Steps to reproduce the behaviour

`mc share list` 

look at output

## mc version
- (paste output of `mc version`)

Version: 2017-03-26T00:24:12Z
Release-tag: DEVELOPMENT.2017-03-26T00-24-12Z
Commit-id: a15d54c4a812ecc7183565720a6ae26db07fec2e

## System information

Mac OS, but this would happen on any system

PR #2100, opened 2017-03-31
Fix for issue #2093 NAME:

badscooter23:
The `mc share list` help text misleading...


badscooter23:
Simple change to the help text for the `mc share list` command ... 

The old help text did not explicitly indicate that the COMMAND (either upload or download) was required when the user did an `mc share list`

krishnasrinivas:
@badscooter23 thanks for the PR.

PR #2101, opened 2017-03-31
Replace admin-heal-list with admin-heal --fake

krisis:
- Replace admin-heal-list with admin-heal --fake
 - Improve admin-heal message for partial healing
  
    * Vendorize pkg/madmin changes to include improved messaging for
      partial healing of objects/uploads.
    
    * Modify message strings to include additional heal result information
      like number of disks where heal was successful.


cmd/admin-heal.go
```diff
@@ -225,6 +304,12 @@ func mainAdminHeal(ctx *cli.Context) error {
 
 	// Heal format if bucket is not specified and quit immediately
 	if bucket == "" {
+		// If --fake was given, print all objects/uploads that
+		// need to be healed.
+		if isFake {
+			listAllToBeHealed(client, aliasedURL, isIncomplete)
```
vadmeste: But if this function returns an error instead.

krisis: >Better if this function returns an error instead.

@vadmeste while it would make `if isFake {` block shorter, this function doesn't fail (excl. the fatalIf), so it doesn't make sense to return `error`.

vadmeste: > @vadmeste while it would make if isFake { block shorter, this function doesn't fail (excl. the fatalIf), so it doesn't make sense to return error.

yeah, it is just little more readable to avoid calling fatalIf() inside listAllToBeHealed()

krisis:
Sample output:
```
$ mc admin heal --fake s1/testbucket
Object: testbucket/go1.6.2.linux-amd64.tar.gz, can be partially healed.
Object: testbucket/obj1, can be partially healed.
Object: testbucket/obj2, can be partially healed.

$ mc admin heal --fake -I s1/testbucket
Upload: testbucket/obj2, can be partially healed.

```

cmd/admin-heal.go
```diff
@@ -96,20 +96,36 @@ var adminHealCmd = cli.Command{
 
 // healMessage container to hold repair information.
 type healMessage struct {
-	Status string             `json:"status"`
-	Bucket string             `json:"bucket"`
-	Object *madmin.ObjectInfo `json:"object"`
-	Upload *madmin.UploadInfo `json:"upload"`
+	Status string                  `json:"status"`
+	Result madmin.HealObjectResult `json:"result"`
```
vadmeste: Do you mean madmin.HealResult here? (without object because it could be upload sometimes?)

vadmeste:
LGTM, except one comment.

cmd/admin-heal.go
krisis: @vadmeste I meant `madmin.HealObjectResult `. I agree that the name suggests that it returns result of the healing of objects and it can be confusing when used with result of healing uploads. I have named the member variable `Result` to make it applicable for both objects and uploads and left the type as-is. Do you have a different approach in mind?

vadmeste: @krisis, I guess we should rename HealObjectResut to HealResult in madmin package

krisis: @vadmeste I have a [PR](https://github.com/minio/minio/pull/4035) that renames HealObjectResult to HealResult in pkg/madmin.

vadmeste: > @vadmeste I have a PR that renames HealObjectResult to HealResult in pkg/madmin.

Thanks

harshavardhana:
Can you vendorize the pkg/madmin @krisis ?

PR #2102, opened 2017-03-31
trace: Avoid httputil for less memory consumption

### LINKED ISSUE #2086: Multi-part upload fails back to single PUT on some (non-minio) servers
## Expected behaviour
When uploading a large file, `mc` should use multi-part uploads
## Actual behaviour
`mc` attempts to use multi-part upload, gets a `HTTP 403` response, and fails back to a single `PUT`
When `--debug` is specified, this results in a large spike in memory usage, which seems to hit `[filesize]*2`
Tested combinations:
* `mc cp` -> riak-cs: single `PUT`, RSS ~8.1MB
* `mc cp --debug` -> riak-cs: single `PUT`, RSS ~1700MB
* `mc cp` -> minio: multi-part, RSS ~8.5MB
* `mc cp --debug` -> minio: multi-part, RSS ~160MB
* `aws s3 --debug cp` -> riak-cs: multi-part, RSS ~60MB

1. Start a riak-cs server (docker steps here) `docker run -d --env 'RIAK_CS_BUCKETS=test' -p 80:8080 --name riakcs ianbytchek/riak-cs `
1. After a few seconds for startup, get credentials: `docker logs riakcs` 
1. Add configuration to `mc`: `mc config host add riak http://[docker_host_ip]:80 [AccessKey] [SecretKey] S3v2`
1. Copy a large file into the bucket: `mc cp ubuntu-16.04.2-server-amd64.iso riak/test/`
    * If you add `--debug` there is a memory usage spike toward the end of the transfer

## mc version
Version: 2017-02-06T20:16:19Z
Release-tag: RELEASE.2017-02-06T20-16-19Z
Commit-id: 2c8115de4edc5612525488b1b3b804689d336d01

## System information
x86_64 VM running RHEL 6

vadmeste:
Rewrite tracing without httputil since it consumes memory with
requests that have huge data.

Copying 829 MiB file when --debug is enabled reduces RSS from ~400 Mb to 14 Mb

Fixes #2086

### DIFF
--- a/cmd/client-s3-trace_v2.go
+++ b/cmd/client-s3-trace_v2.go
@@ -18,8 +18,8 @@ package cmd
 
 import (
 	"bytes"
+	"io/ioutil"
 	"net/http"
-	"net/http/httputil"
 	"strings"
 
 	"github.com/minio/mc/pkg/console"
@@ -45,11 +45,17 @@ func (t traceV2) Request(req *http.Request) (err error) {
 		// Set a temporary redacted auth
 		req.Header.Set("Authorization", "AWS **REDACTED**:**REDACTED**")
 
-		var reqTrace []byte
-		reqTrace, err = httputil.DumpRequestOut(req, false) // Only display header
-		if err == nil {
-			console.Debug(string(reqTrace))
+		// Store headers in buf
+		var buf bytes.Buffer
+		if err = req.Header.Write(&buf); err != nil {
+			return err
 		}
+
+		// Add headers to the debug text
+		reqTrace := append(buf.Bytes(), []byte("\r\n")...)
+
+		// Print debug information
+		console.Debug(string(reqTrace))
 
 		// Undo
 		req.Header.Set("Authorization", origAuth)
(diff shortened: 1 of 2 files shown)


cmd/client-s3-trace_v2.go
```diff
@@ -45,12 +45,18 @@ func (t traceV2) Request(req *http.Request) (err error) {
 		// Set a temporary redacted auth
 		req.Header.Set("Authorization", "AWS **REDACTED**:**REDACTED**")
 
-		var reqTrace []byte
-		reqTrace, err = httputil.DumpRequestOut(req, false) // Only display header
```
harshavardhana: With `false` as last argument. it shouldn't consume so much memory at all.

vadmeste: > With false as last argument. it shouldn't consume so much memory at all.

https://github.com/golang/go/blob/master/src/net/http/httputil/dump.go#L135

For some reason, when false is specified, that code replaces the actual req.Body with a dummy one which has the same content length.

harshavardhana: That is a bug. @vadmeste

Issue #2099, opened 2017-03-31
share: `mc share download` clean up; `--recursive option` and error messages


badscooter23:
## Expected behaviour

Assuming: 
bucket `foo` has objects `bar1` and `bar2`

(NOTE: I updated assumption, the original text had a mistake.)

Then: 

`mc share download play/foo/bar`
should give an error (and it does), but the current error message is not good

`mc share download play/foo/bar --recursive`
should succeed and return two download URLs, one each for `bar1` and `bar2`; currently it returns no URLs and and the same error message

`mc share download play/foo/invalid`
should fail (which it does) but the current error message is not good

`mc share download play/foo/invalid --recursive`
should fail, currently it does fail but the error message is not good

## Actual behaviour

`mc share download play/foo/bar`
`mc share download play/foo/bar --recursive`
`mc share download play/foo/invalid`
`mc share download play/foo/invalid --recursive`

all generate the same error:
mc: <ERROR> Unable to stat `play/foo/bar`. Object key is missing, object key cannot be empty
or
mc: <ERROR> Unable to stat `play/foo/invalid`. Object key is missing, object key cannot be empty

The message should be more descriptive: 

mc <ERROR> Object `play/foo/invalid` does not exist.

Also the second case should NOT be an error, it should succeed.

## Steps to reproduce the behaviour

see above

## mc version
- (paste output of `mc version`)

Scotts-MacBook-Pro-2:mc scott$ mc version
Version: 2017-03-31T00:31:47Z
Release-tag: DEVELOPMENT.2017-03-31T00-31-47Z
Commit-id: 3c0094a5fe84d632b87030a46f3f2d696947cc13

## System information

can be reproduced on any system
done using mc version (above) against `play.minio.io:9000`

vadmeste:
Hey @badscooter23,



> Assuming:
> bucket foo has objects bar1 and bar
> 
> Then:
> 
> mc share download play/foo/bar
> should give an error (and it does), but the current error message is not good

If we are assuming that bucket `foo` bucket has `bar` object, why `mc share download play/foo/bar` should return an error ?

badscooter23:
Hey @vadmeste 

You are correct. The original text had a mistake  so it did not make sense. 
It should have said:

Assuming:
bucket foo has objects `bar1` and `bar2`

With this correction it should make sense do you. `mc share download play/foo/bar` should return an error because `play/foo/bar` does not exist and the user did not specify `--recursive`. 

[...]

PR #2106, opened 2017-04-03
minor fixes to `mc share download` command; improved error messages and issue with `--recursive` option. (#2099)

badscooter23:
Verified that all the test cases below are working as expected... 

`mc ls play/foo` should show two objects `bar1`, `bar2`

```
Scotts-MacBook-Pro-2:mc scott$ mc ls play/foo
[2017-04-03 13:10:00 PDT] 2.0MiB bar1
[2017-04-03 13:10:19 PDT] 1.1MiB bar2
```

All of these `mc share download` commands should generate an error ... 

```
Scotts-MacBook-Pro-2:mc scott$ mc share download play/foo/invalid
mc: <ERROR> Object `play/foo/invalid` does not exist. 
Scotts-MacBook-Pro-2:mc scott$ mc share download play/foo/invalid --recursive
mc: <ERROR> Object `play/foo/invalid` does not exist. 
Scotts-MacBook-Pro-2:mc scott$ mc share download play/foo/bar
mc: <ERROR> Object `play/foo/bar` does not exist. 
```
Prior to this fix, the error message above was confusing, and the middle case failed without an error message. 

`mc share download play/foo/bar --recursive` should generate two download share URLs... 

```
Scotts-MacBook-Pro-2:mc scott$ mc share download play/foo/bar --recursive
URL: https://play.minio.io:9000/foo/bar1
Expire: 7 days 0 hours 0 minutes 0 seconds
Share: <.... URL deleted...>

URL: https://play.minio.io:9000/foo/bar2
Expire: 7 days 0 hours 0 minutes 0 seconds
Share: <.... URL deleted...>
```

Prior to this fix, this case would give an error instead of returning the two download share URLs.

### DIFF
--- a/cmd/share-download-main.go
+++ b/cmd/share-download-main.go
@@ -89,9 +89,15 @@ func checkShareDownloadSyntax(ctx *cli.Context) {
 		fatalIf(errDummy().Trace(expiry.String()), "Expiry cannot be larger than 7 days.")
 	}
 
-	for _, url := range ctx.Args() {
-		_, _, err := url2Stat(url)
-		fatalIf(err.Trace(url), "Unable to stat `"+url+"`.")
+	// Validate if object exists only if the `--recursive` option was NOT specified
+	isRecursive := ctx.Bool("recursive")
+	if !isRecursive {
+		for _, url := range ctx.Args() {
+			_, _, err := url2Stat(url)
+			if err != nil {
+				fatalIf(err.Trace(url), "Unable to stat `"+url+"`.")
+			}
+		}
 	}
 }
 
--- a/cmd/client-errors.go
+++ b/cmd/client-errors.go
@@ -150,7 +150,7 @@ func (e EmptyPath) Error() string {
 type ObjectMissing struct{}
 
 func (e ObjectMissing) Error() string {
-	return "Object key is missing, object key cannot be empty"
+	return "Object does not exist"
 }
 
 // UnexpectedShortWrite - write wrote less bytes than expected.


harshavardhana:
@badscooter23 Thanks for the patch, - can you have a more meaningful title "changes for pull request #2099 #2106" really doesn't say much.

vadmeste:
Can you also modify the commit title ?

cmd/share-download-main.go
```diff
@@ -89,9 +89,15 @@ func checkShareDownloadSyntax(ctx *cli.Context) {
...
-	for _, url := range ctx.Args() {
-		_, _, err := url2Stat(url)
-		fatalIf(err.Trace(url), "Unable to stat `"+url+"`.")
+	// Validate if object exists only if the `--recursive` option was NOT specified
+	isRecursive := ctx.Bool("recursive")
+	if !isRecursive {
+		for _, url := range ctx.Args() {
+			_, _, err := url2Stat(url)
+			if err != nil {
+				fatalIf(errDummy(), "Object `"+url+"` does not exist.")
```
harshavardhana: url2Stat can send different errors other than just object may not exist. Also doesn't use the traced err. So the origination of the error is lost

To fix the message we need to fix the the odd message is originated inside client-s3.go

cmd/share-download-main.go
```diff
@@ -134,8 +140,15 @@ func doShareDownloadURL(targetURL string, isRecursive bool, expiry time.Duration
 		// Recursive mode: Share list of objects
 		go func() {
 			defer close(objectsCh)
+			objectCount := 0
 			for content := range clnt.List(isRecursive, isIncomplete, DirNone) {
 				objectsCh <- content
+				objectCount ++
```
harshavardhana: The ++ here should be `objectCount++`

deekoder: @badscooter23 - If you can amend your commit with this fix, we can take this forward.

PR #2109, opened 2017-04-04
fs watch: Use a dynamically sized channel

vadmeste:
Write and use a growing/shrinking channel to be able to absorb file system
notify events even when they come in large numbers.

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -101,7 +101,7 @@ func (f *fsClient) Watch(params watchParams) (*watchObject, *probe.Error) {
 	doneChan := make(chan bool)
 	// Make the channel buffered to ensure no event is dropped. Notify will drop
 	// an event if the receiver is not able to keep up the sending pace.
-	neventChan := make(chan notify.EventInfo, 1)
+	in, out := PipeChan(1000)
 
 	var fsEvents []notify.Event
 	for _, event := range params.events {
@@ -121,7 +121,7 @@ func (f *fsClient) Watch(params watchParams) (*watchObject, *probe.Error) {
 	if params.recursive {
 		recursivePath = f.PathURL.Path + "..."
 	}
-	if e := notify.Watch(recursivePath, neventChan, fsEvents...); e != nil {
+	if e := notify.Watch(recursivePath, in, fsEvents...); e != nil {
 		return nil, probe.NewError(e)
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/dynchan/pipechan.go
```diff
@@ -0,0 +1,74 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dynchan
+
+// Dynamically sized logical channel: a pipe which never blocks even when
+// it receives too many elements. Memory consumption is increased and decresed
+// on the fly following the number of elements.
+func PipeChan(capacity int) (inputCh, outputCh chan interface{}) {
```
harshavardhana: why take capacity? if its going to increase in size anyways. Perhaps i am seeing that you want it something like `bytes.NewBuffer([capacity]byte)` .

pkg/dynchan/README.md
```diff
@@ -0,0 +1,47 @@
+
+# Dynchan
+
+## Go code
+
+```go
+
+inputCh, outputCh := dynchan.PipeChan(1000)
```
harshavardhana: We need to document more perhaps like take the `1000` as capacity.

pkg/dynchan/pipechan.go
```diff
@@ -0,0 +1,74 @@
...
+			currCh, ok := <-channels
+			if !ok {
+				break
+			}
+			for v := range currCh {
+				outputCh <- v
+			}
+		}
+		close(outputCh)
+	}()
```
harshavardhana: This function helps in not loosing any events? does it minimize?

pkg/dynchan/README.md
```diff
@@ -0,0 +1,47 @@
...
+}
+
+
+// Receive and process objects
+for m := range outputCh {
+    // ...
+}
+
+
+```
```
harshavardhana: Can you show an example of integrated work like, for example use this in `client-fs.go` for events..

pkg/dynchan/pipechan.go
vadmeste: > This function helps in not loosing any events? does it minimize?

It does minimize. There is no generic way to totally avoid losing events if sender doesn't wait at all for blocked channels.

If the developer chooses 1000 as a starting channel size and the sender was able to send 1500 events before getting scheduled  by the scheduler, 500 events will be lost because we won't get a chance to grow our channel.

pkg/dynchan/README.md
vadmeste: I did.

Also it looks like I need to avoid using this feature as a library :( I failed to make it work in a generic way with `chan interface{}` with client-fs.

vadmeste:
@harshavardhana, can you review this again ?

harshavardhana:
Will do @vadmeste

PR #2120, opened 2017-04-10
watch/events: Add eventInfo struct for brevity.

harshavardhana:


harshavardhana:
> why not merge Event and Source? It will be simpler.

Done!

cmd/client-s3.go
```diff
@@ -448,76 +445,48 @@ func (c *s3Client) Watch(params watchParams) (*watchObject, *probe.Error) {
...
-							Port:      record.Source.Port,
-							UserAgent: record.Source.UserAgent,
-						},
+					eventChan <- EventInfo{
+						Time:      record.EventTime,
+						Size:      record.S3.Object.Size,
+						Path:      u.String(),
+						Client:    c,
+						Type:      EventCreate,
+						IP:        record.Source.Host,
```
krishnasrinivas: `IP` can be renamed to `Host` to be consistent

cmd/watch-main.go
```diff
@@ -97,8 +97,17 @@ func checkWatchSyntax(ctx *cli.Context) {
...
+		Time string    `json:"time"`
+		Size int64     `json:"size"`
+		Path string    `json:"path"`
+		Type EventType `json:"type"`
+	} `json:"events"`
+	Source struct {
+		IP        string `json:"ip,omitempty"`
+		Port      string `json:"port,omitempty"`
+		UserAgent string `json:"userAgent,omitempty"`
+	} `json:"source,omitempty"`
```
krishnasrinivas: Status is not useful.

EvenInfo can be used for printing and `watchMessage` can be eliminated completely.

harshavardhana: No we need the structure of json to be different. It is not a one to one mapping.

harshavardhana: Status is consistent with all mc output it is part of json API.

cmd/client-s3.go
harshavardhana: Will do.

PR #2126, opened 2017-04-14
session: Close data session file when migrating

### LINKED ISSUE #1904: session: [windows] mc session resume/clear fails due to file locking
When using `mc session resume <id>` or `mc session clear all` on Windows, the command fails with the error "The file is in use by another process".

## mc version
mc 2016-10-07

## System information
Windows 10 Anniversary x64

vadmeste:
Cleanup needed before removing the data file

Fixes #1904

### DIFF
--- a/cmd/session-migrate.go
+++ b/cmd/session-migrate.go
@@ -37,6 +37,9 @@ func migrateSessionV7ToV8() {
 			fatalIf(err.Trace(sid), "Unable to load version `7`. Migration failed please report this issue at https://github.com/minio/mc/issues.")
 		}
 
+		// Close underlying session data file.
+		sV7.DataFP.Close()
+
 		sessionVersion, e := strconv.Atoi(sV7.Header.Version)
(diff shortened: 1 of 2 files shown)


cmd/session-migrate.go
```diff
@@ -37,6 +37,9 @@ func migrateSessionV7ToV8() {
 			fatalIf(err.Trace(sid), "Unable to load version `7`. Migration failed please report this issue at https://github.com/minio/mc/issues.")
 		}
 
+		// Close underlying session data file.
+		sV7.DataFP.Close()
```
krisis: if we are closing the session data file right after returning from `loadSessionV7`, why can't we close it at the end of `loadSessionV7`?

vadmeste: > if we are closing the session data file right after returning from loadSessionV7, why can't we close it at the end of loadSessionV7?

session has already its own Close() method which can be executed after loading session so closing should be done outside loadSessionV7.

harshavardhana: :+1:

deekoder:
Can I go ahead and merge this PR.

harshavardhana:
@krisis needs to review, once more.

Issue #2134, opened 2017-04-19
Please add "lowercase" into error message for mb creation


i4o:
## Expected behaviour

mc mb do/GoProExports
<ERROR> Unable to make bucket ‘do/GoProExports’. Bucket name can contain LOWERCASE alphabet, '-' and numbers, but first character should be an alphabet or number

## Actual behaviour

>mc mb do/GoProExports

<ERROR> Unable to make bucket ‘do/GoProExports’. Bucket name can contain alphabet, '-' and numbers, but first character should be an alphabet or number

When I try this in GUI I get correct error message. Took me a while to find out I have to use lowercase.

vadmeste:
@i4o, thanks for reporting. Will do :)

badscooter23:
Hang on a second... this does not make complete sense to me (perhaps I am missing something)...

On Minio... with mc... the mb command will force the bucket name to be all lowercase characters, a `-`, or a number

On S3... with aws (cli) ... the mb command allows mixed case

Scotts-MacBook-Pro-2:mc scott$ aws s3 mb s3://scott-GoPro
make_bucket: scott-GoPro

Seems to me that the error is not with the error message but with the routine (isValidBucketName) that is doing the validation... 

I think this is a different bug!

badscooter23:
Bottom line, this bug will actually show up in multiple commands - any mc command that takes a bucket name in the path... 

It's not just an error message problem with `mc mb`

badscooter23:
Further example:

`mc mb GoPro` 

will succeed in creating a "bucket" (which is really a directory) in the local file system. The problem is with the routine that is attempting to validate the bucket name in the `cmd/client-s3.go`

badscooter23:
Here is the problem... 
The minio code is refers to this Amazon S3 documentation
http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html

which in turn points at this page
http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html

The second page states:

- Bucketnames must be at least 3 and no more than 63 characters long.

- Bucket names must be a series of one or more labels. Adjacent labels are separated by a single period (.). Bucket names can contain lowercase letters, numbers, and hyphens. Each label must start and end with a lowercase letter or a number.

- Bucket names must not be formatted as an IP address (e.g., 192.168.5.4).

[...]

PR #2135, opened 2017-04-19
Improved Error Msg For Bucket Names that Contain non-LOWERCASE chars (#2134)

badscooter23:
Improved the error message for various `mc` commands where the user specifies a path or url and the bucket name in that path contains non-lowercase characters.

More specifically, this applies when the path refers to a backend that is an S3 object store and is going through the S3-client. If the backend is a normal/local file system that follows a different path.  

The change is in the function isValidBucketName in cmd/client-s3.go. This routine is invoked when validating the path for `mb` command (as noted in the problem description) but is also used to validate the bucket name for other `mc` commands (eg: `mc events`). 

Examples: 
Scotts-MacBook-Pro-2:mc scott$ mc mb myminio/GoPro
mc: <ERROR> Unable to make bucket `myminio/GoPro`. Bucket names can contain lowercase alpha characters `a-z`, numbers '0-9', and '-'. First character cannot be a '-'
Scotts-MacBook-Pro-2:mc scott$ mc events list myminio/GoPro
mc: <ERROR> Cannot enable notification on the specified bucket. Bucket names can contain lowercase alpha characters `a-z`, numbers '0-9', and '-'. First character cannot be a '-'
Scotts-MacBook-Pro-2:mc scott$ 

Note: some `mc` commands (such as `ls` and `share`) will report other errors in this same case... We may need to look at these cases as well for consistency.

Scotts-MacBook-Pro-2:mc scott$ mc ls myminio/GoPro
mc: <ERROR> Unable to stat `myminio/GoPro`. Bucket name contains invalid characters.
Scotts-MacBook-Pro-2:mc scott$ mc share download myminio/GoPro
mc: <ERROR> Unable to stat `myminio/GoPro`. Bucket name contains invalid characters.
Scotts-MacBook-Pro-2:mc scott$

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -746,7 +746,7 @@ func isValidBucketName(bucketName string) *probe.Error {
 		return probe.NewError(errors.New("Bucket name should be more than 3 characters and less than 64 characters"))
 	}
 	if !validBucketName.MatchString(bucketName) {
-		return probe.NewError(errors.New("Bucket name can contain alphabet, '-' and numbers, but first character should be an alphabet or number"))
+		return probe.NewError(errors.New("Bucket names can only contain lowercase alpha characters `a-z`, numbers '0-9', or '-'. First/last character cannot be a '-'"))
 	}
 	return nil
 }


badscooter23:
@harshavardhana Why did my PR fail the `codecov/patch` test? 
Is there something I need to do to fix this and complete the merge?

harshavardhana:
codecov/patch failure can be ignored for `mc` for now. We need to up the coverage on each PR.  What it is says that you haven't added anything to cover the changes which were submitted in this PR.

badscooter23:
Ah... ok...

But I didn't add any paths either... just changed a message...  :)

harshavardhana:
> But I didn't add any paths either... just changed a message...  :)

If you want full green add a test to check for the change in the message.

cmd/client-s3.go
```diff
@@ -746,7 +746,7 @@ func isValidBucketName(bucketName string) *probe.Error {
 		return probe.NewError(errors.New("Bucket name should be more than 3 characters and less than 64 characters"))
 	}
 	if !validBucketName.MatchString(bucketName) {
-		return probe.NewError(errors.New("Bucket name can contain alphabet, '-' and numbers, but first character should be an alphabet or number"))
+		return probe.NewError(errors.New("Bucket names can contain lowercase alpha characters `a-z`, numbers '0-9', and '-'. First character cannot be a '-'"))
```
vadmeste: Isn't `Bucket names can *only* contain` more firm ?

badscooter23: I agree with this wording change... will do it now! :)

PR #2139, opened 2017-04-22
mb: Update pkg to retry mb with unmatched region

### LINKED ISSUE #2107: errors: [feature] Some errors in `mc` can guide users
## Expected behaviour

For example error rather than just saying `Region does not match` could have improved upon and stated that

```
mc: <ERROR> Unable to make bucket ‘dev/platform’. Region does not match may be you can try `
```

## Actual behaviour

```
mc: <ERROR> Unable to make bucket ‘dev/platform’. Region does not match.
```

Any `mc` command. 

## mc version
master

vadmeste:
Fixes #2107

### DIFF
--- a/vendor/github.com/minio/minio-go/api-error-response.go
+++ b/vendor/github.com/minio/minio-go/api-error-response.go
@@ -1,5 +1,5 @@
 /*
- * Minio Go Library for Amazon S3 Compatible Cloud Storage (C) 2015 Minio, Inc.
+ * Minio Go Library for Amazon S3 Compatible Cloud Storage (C) 2015, 2016, 2017 Minio, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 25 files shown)


cmd/client-s3.go
```diff
@@ -762,6 +763,11 @@ func (c *s3Client) MakeBucket(region string) *probe.Error {
 	}
 	e := c.api.MakeBucket(bucket, region)
 	if e != nil {
+		// Enhance error message when invalid region error is received
+		if errResp, ok := e.(minio.ErrorResponse); ok && errResp.Code == "InvalidRegion" {
+			errResp.Message = fmt.Sprintf("The server rejected `%s` region. Kindly pass the correct region value using `--region` flag.", region)
```
harshavardhana: We actually retry with the Region returned in the headers. not sure if this is needed..

harshavardhana:
This PR is waiting on - https://github.com/minio/minio-go/pull/665

PR #2140, opened 2017-04-23
cmd: Add encrypt to enable enc/decrypt objects

### LINKED ISSUE #2074: cp/mirror: [feature] bring mc transparent encryption
## Expected behaviour

```json
"encryption": {                                                                                                                                   
   "AES": {                                                                                                                                       
      "enable": true,                                                                                                                             
      "key": ""                                                                                                                                   
   },                                                                                                                                             
   "RSA": {                                                                                                                                       
      "enable": false,                                                                                                                            
   },
}
```

```
mc encrypt <alias> on
mc encrypt <alias> off
```

```
```

```
```

## Actual behaviour

## mc version
master 

## System information
All platforms

vadmeste:
Fixes #2074

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -35,6 +35,7 @@ import (
 
 	"github.com/minio/mc/pkg/httptracer"
 	"github.com/minio/minio-go"
+	"github.com/minio/minio-go/pkg/encrypt"
 	"github.com/minio/minio-go/pkg/policy"
 	"github.com/minio/minio-go/pkg/s3utils"
 	"github.com/minio/minio/pkg/probe"
@@ -42,10 +43,11 @@ import (
 
 // S3 client
 type s3Client struct {
-	mutex        *sync.Mutex
-	targetURL    *clientURL
-	api          *minio.Client
-	virtualStyle bool
+	mutex            *sync.Mutex
+	targetURL        *clientURL
+	api              *minio.Client
+	encryptMaterials encrypt.Materials
+	virtualStyle     bool
 }
 
 const (
@@ -163,6 +165,8 @@ func newFactory() func(config *Config) (Client, *probe.Error) {
 
 		// Store the new api object.
 		s3Clnt.api = api
+
+		s3Clnt.encryptMaterials = config.encryptMaterials
 
 		return s3Clnt, nil
 	}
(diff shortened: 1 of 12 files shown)


harshavardhana:
Is this still a WIP @vadmeste ?

vadmeste:
Yes, I made more simple changes, this PR can be reviewed now.

harshavardhana:
> Also I need to think if we should add an API in minio-go which returns the real size of an object after being decrypted.

why is this needed?

harshavardhana:
Keeping this waiting on @abperiasamy - its blocked to finalize the approach here.

vadmeste:
The size of an object in encrypted form is different from the original size. mc compares the size of the downloaded data (data are clear because minio-go decrypts the stream on the fly) and finds that the size of the downloaded stream is different to the size of the same object in the server (using client-fs.Stat() primitive)

bollig:
Any updates on this? Encrypt is a feature we're excited to leverage.

harshavardhana:
@vadmeste after discussing with @abperiasamy - we decided that we don't need to do any config changes and it was simplified further. 

For the time being we will only support AES256 encryption on the client side which will also work with aes256-server when the server side implementation comes in.  Only commands which require this change are listed below.

```
- mc cp --encrypt-key=<key> --encrypt-type=aes256-server /etc/passwd myminio/bucket/
- mc cp -r --encrypt-key=<key> --encrypt-type=aes256 s3/bucket1/encrypted-objects myminio/bucket/encrypted-objects
- mc mirror --encrypt-key|-e=<key> --encrypt-type=aes256-server myminio/ s3/
- mc cat, mc pipe 
```
```
`-e` or `--encrypt-key` if has no key then we need to prompt on the command line. 
```

With this we can also support now FS to FS, FS to S3, S3 to FS decryption and encryption transparently.

harshavardhana:
Unblocked this PR since we have finalized the approach.

vadmeste:
mc mirror with encryption capability is weird because sometimes the size of encrypted object size is different to the size of plain representation. I'll investigate more.

fwessels:
Have seen stuff like this before, has to do with working on blocks of eg 32 or 64 byte size (depending on algorithm) whereby the final block is likely to be padded with zeros. (And so the encrypted stream may be a little longer).

vadmeste:
> Have seen stuff like this before, has to do with working on blocks of eg 32 or 64 byte size (depending on algorithm) whereby the final block is likely to be padded with zeros. (And so the encrypted stream may be a little longer).

Yes. But you can't know the exact length of the plain data from the length of encrypted data (you can with the reverse way). Though as you said padding will make the size of the encrypted data little longer (the maximum is the original size + block size).

I'll make the first step first then will see how to make this perfect.

vadmeste:
Closing this in favor of https://github.com/minio/mc/pull/2197

PR #2142, opened 2017-04-27
Add --newer-than & --older-than flags

### LINKED ISSUE #2025: Implement options --newer-than and --older-than for cp, rm, mirror
Hello, 
is it possible to implement timestamp difference in mirror and diff function.
mirror only with size different is unusable for us. A timestamp sync will be very useful.
Regards, 
Nicolas

vadmeste:
Use --newer-than & --older-than to select newer or older files
in copy, remove and mirroring operations.

Fixes #2025

### DIFF
--- a/cmd/cp-main.go
+++ b/cmd/cp-main.go
@@ -41,6 +41,14 @@ var (
 			Name:  "recursive, r",
 			Usage: "Copy recursively.",
 		},
+		cli.IntFlag{
+			Name:  "older-than",
+			Usage: "Copy objects older than N days",
+		},
+		cli.IntFlag{
+			Name:  "newer-than",
+			Usage: "Copy objects newer than N days",
+		},
 	}
 )
 
@@ -74,10 +82,16 @@ EXAMPLES:
    4. Copy a bucket recursively from aliased Amazon S3 cloud storage to local filesystem on Windows.
       $ {{.HelpName}} --recursive s3\documents\2014\ C:\Backups\2014
 
-   5. Copy an object with name containing unicode characters to Amazon S3 cloud storage.
+   5. Copy files older than 7 days from Minio cloud storage to Amazon S3 cloud storage.
+      $ {{.HelpName}} --older-than 7 play/mybucket/burningman2011/ s3/mybucket/
+
+   6. Copy files newer than 7 days from Minio cloud storage to a local path.
+      $ {{.HelpName}} --newer-than 7 play/mybucket/burningman2011/ ~/latest/
+
+   7. Copy an object with name containing unicode characters to Amazon S3 cloud storage.
       $ {{.HelpName}} 本語 s3/andoria/
 
-   6. Copy a local folder with space separated characters to Amazon S3 cloud storage.
+   8. Copy a local folder with space separated characters to Amazon S3 cloud storage.
       $ {{.HelpName}} --recursive 'workdir/documents/May 2014/' s3/miniocloud
 
 `,
@@ -190,6 +204,9 @@ func doPrepareCopyURLs(session *sessionV8, trapCh <-chan bool, cancelCopy contex
 	// Access recursive flag inside the session header.
 	isRecursive := session.Header.CommandBoolFlags["recursive"]
 
+	olderThan := session.Header.CommandIntFlags["older-than"]
+	newerThan := session.Header.CommandIntFlags["newer-than"]
+
 	// Create a session data file to store the processed URLs.
 	dataFP := session.NewDataWriter()
 
@@ -225,6 +242,17 @@ func doPrepareCopyURLs(session *sessionV8, trapCh <-chan bool, cancelCopy contex
 				session.Delete()
 				fatalIf(probe.NewError(e), "Unable to prepare URL for copying. Error in JSON marshaling.")
 			}
+
+			// Skip objects older than --older-than parameter if specified
+			if olderThan > 0 && isOlder(cpURLs.SourceContent, olderThan) {
+				continue
+			}
+
+			// Skip objects newer than --newer-than parameter if specified
+			if newerThan > 0 && isNewer(cpURLs.SourceContent, newerThan) {
+				continue
+			}
+
 			fmt.Fprintln(dataFP, string(jsonData))
 			if !globalQuiet && !globalJSON {
 				scanBar(cpURLs.SourceContent.URL.String())
@@ -376,9 +404,15 @@ func mainCopy(ctx *cli.Context) error {
 	// Additional command speific theme customization.
 	console.SetColor("Copy", color.New(color.FgGreen, color.Bold))
 
+	recursive := ctx.Bool("recursive")
+	olderThan := ctx.Int("older-than")
+	newerThan := ctx.Int("newer-than")
+
 	session := newSessionV8()
 	session.Header.CommandType = "cp"
-	session.Header.CommandBoolFlags["recursive"] = ctx.Bool("recursive")
+	session.Header.CommandBoolFlags["recursive"] = recursive
+	session.Header.CommandIntFlags["older-than"] = olderThan
+	session.Header.CommandIntFlags["newer-than"] = newerThan
 
 	var e error
 	if session.Header.RootPath, e = os.Getwd(); e != nil {
--- a/cmd/mirror-main.go
+++ b/cmd/mirror-main.go
@@ -70,6 +70,14 @@ var (
 			Name:  "exclude",
 			Usage: "Exclude file/object that matches the passed file name pattern.",
 		},
+		cli.IntFlag{
+			Name:  "older-than",
+			Usage: "Select objects older than N days",
+		},
+		cli.IntFlag{
+			Name:  "newer-than",
+			Usage: "Select objects newer than N days",
+		},
 	}
 )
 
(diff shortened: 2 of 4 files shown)


vadmeste:
Ready for review

cmd/cp-main.go
```diff
@@ -220,6 +223,15 @@ func doPrepareCopyURLs(session *sessionV8, trapCh <-chan bool) {
 				session.Delete()
 				fatalIf(probe.NewError(e), "Unable to prepare URL for copying. Error in JSON marshaling.")
 			}
+
```
harshavardhana: Comment..

cmd/cp-main.go
```diff
@@ -220,6 +223,15 @@ func doPrepareCopyURLs(session *sessionV8, trapCh <-chan bool) {
...
+			if olderThan > 0 && isOlder(cpURLs.SourceContent, olderThan) {
+				continue
+			}
+
```
harshavardhana: Add a comment..

cmd/mirror-main.go
```diff
@@ -429,7 +430,16 @@ func (mj *mirrorJob) startMirror() {
 				}
 				continue
 			}
+
 			if sURLs.SourceContent != nil {
+				if mj.olderThan > 0 && isOlder(sURLs.SourceContent, mj.olderThan) {
```
harshavardhana: Comment.. for each if.

cmd/mirror-main.go
```diff
@@ -542,12 +554,23 @@ func copyBucketPolicies(srcClt, dstClt Client, isForce bool) *probe.Error {
 // runMirror - mirrors all buckets to another S3 server
 func runMirror(srcURL, dstURL string, ctx *cli.Context) *probe.Error {
 
+	isFake := ctx.Bool("fake")
+	isRemove := ctx.Bool("remove")
+	isWatch := ctx.Bool("watch")
+	isForce := ctx.Bool("force")
+
+	olderThan := ctx.Int("older-than")
```
harshavardhana: Combine this code everywhere.. i see that parsing older-than and newer-than can be put into a single function.

cmd/rm-main.go
```diff
@@ -275,7 +268,16 @@ func mainRm(ctx *cli.Context) error {
 	isRecursive := ctx.Bool("recursive")
 	isFake := ctx.Bool("fake")
 	isStdin := ctx.Bool("stdin")
-	older := ctx.Int("older-than")
+
+	olderThan := ctx.GlobalInt("older-than")
```
harshavardhana: Same as above..

cmd/utils.go
```diff
@@ -130,3 +130,23 @@ func buildS3Config(alias, urlStr string) (*Config, *probe.Error) {
...
+}
+
+// isNewer returns true if the passed object is newer than newerRef
+func isNewer(c *clientContent, newerRef int) bool {
+	now := time.Now().UTC()
+	objectAge := now.Sub(c.Time)
+	if objectAge > (time.Duration(newerRef) * Day) {
+		return true
+	}
+	return false
```
harshavardhana: Also document `--older-than` and `--newer-than` for each commands..

cmd/flags.go
```diff
@@ -54,6 +54,14 @@ var globalFlags = []cli.Flag{
 		Name:  "insecure",
 		Usage: "Disable SSL certificate verification.",
 	},
+	cli.IntFlag{
+		Name:  "older-than",
+		Usage: "Select objects older than N days",
+	},
+	cli.IntFlag{
+		Name:  "newer-than",
```
harshavardhana: If these are global flags shouldn't they be not displayed with other commands which don't support this? like `mc share`  `mc ls` `mc events` ... 

We should only add these flags for the relevant commands which do support it.

vadmeste: > If these are global flags shouldn't they be not displayed with other commands which don't support this? like mc share mc ls mc events ...
> 
> We should only add these flags for the relevant commands which do support it.

Yes, as I remember that was @abperiasamy proposal. I also don't recommend having it as a global options because it causes more confusion for the user and we don't have any benefits in the other side.

harshavardhana:
Keeping this waiting on @abperiasamy - its blocked to finalize the approach here.

cmd/utils.go
```diff
@@ -130,3 +130,23 @@ func buildS3Config(alias, urlStr string) (*Config, *probe.Error) {
 
 	return s3Config, nil
 }
+
+// isOlder returns true if the passed object is older than olderRef
+func isOlder(c *clientContent, olderRef int) bool {
+	now := time.Now().UTC()
+	objectAge := now.Sub(c.Time)
+	if objectAge < (time.Duration(olderRef) * Day) {
```
balamurugana: The `if` is not required. you could just do `return objectAge < (time.Duration(olderRef) * Day)`

cmd/utils.go
```diff
@@ -130,3 +130,23 @@ func buildS3Config(alias, urlStr string) (*Config, *probe.Error) {
 
 	return s3Config, nil
 }
+
+// isOlder returns true if the passed object is older than olderRef
+func isOlder(c *clientContent, olderRef int) bool {
+	now := time.Now().UTC()
+	objectAge := now.Sub(c.Time)
```
balamurugana: It can be one liner like `objectAge := time.Now().UTC().Sub(c.Time)`

cmd/utils.go
```diff
@@ -130,3 +130,23 @@ func buildS3Config(alias, urlStr string) (*Config, *probe.Error) {
...
+		return true
+	}
+	return false
+}
+
+// isNewer returns true if the passed object is newer than newerRef
+func isNewer(c *clientContent, newerRef int) bool {
+	now := time.Now().UTC()
+	objectAge := now.Sub(c.Time)
+	if objectAge > (time.Duration(newerRef) * Day) {
```
balamurugana: same as comments in `isOlder`

cmd/rm-main.go
```diff
@@ -209,14 +214,14 @@ func removeRecursive(url string, isIncomplete bool, isFake bool, older int) erro
...
-				// time difference of info.Time with current time is less than older time.
-				continue
-			}
+		// Skip objects older than --older-than parameter if specified
+		if olderThan > 0 && isOlder(content, olderThan) {
+			continue
+		}
+
+		// Skip objects newer than --newer-than parameter if specified
+		if newerThan > 0 && isNewer(content, newerThan) {
```
balamurugana: I see we could use `--older-than` only, `newer-than` only or `older-than` and `newer-than` together.  We would need to add proper examples `CustomHelpTemplate` in for those combinations for applicable commands.

Similarly I couldn't able to understand `older-than` and `newer-than` combination use case.

harshavardhana:
Is this PR still needed @vadmeste ?

kannappanr:
ping @vadmeste

harshavardhana:
Do you need to discuss this again @vadmeste ?

cmd/cp-main.go
vadmeste: > Comment..

Done

vadmeste:
@harshavardhana @balamurugana PR rebased, please review again.

vadmeste:
ping @harshavardhana @balamurugana

cmd/mirror-main.go
```diff
@@ -70,6 +70,14 @@ var (
 			Name:  "exclude",
 			Usage: "Exclude file/object that matches the passed file name pattern.",
 		},
+		cli.IntFlag{
+			Name:  "older-than",
+			Usage: "Select objects older than N days",
+		},
+		cli.IntFlag{
+			Name:  "newer-than",
+			Usage: "Select objects newer than N days",
```
harshavardhana: Should we have a common place for these flag definitions and reuse it everywhere?

vadmeste: > Should we have a common place for these flag definitions and reuse it everywhere?

In general, I prefer having a specific documentation for each case: mirroring, copying and removing.. for a better documentation clarity. I won't worry too much about duplication here.

vadmeste:
ping @balamurugana

cmd/utils.go
```diff
@@ -132,3 +132,21 @@ func lineTrunc(content string, maxLen int) string {
 	sndPart := string(runes[rlen-halfLen:])
 	return fstPart + "…" + sndPart
 }
+
+// isOlder returns true if the passed object is older than olderRef
+func isOlder(c *clientContent, olderRef int) bool {
+	objectAge := UTCNow().Sub(c.Time)
+	if objectAge < (time.Duration(olderRef) * Day) {
```
balamurugana: you could just return the condition i.e. `if` is not needed.

cmd/utils.go
```diff
@@ -132,3 +132,21 @@ func lineTrunc(content string, maxLen int) string {
...
+		return true
+	}
+	return false
+}
+
+// isNewer returns true if the passed object is newer than newerRef
+func isNewer(c *clientContent, newerRef int) bool {
+	objectAge := UTCNow().Sub(c.Time)
+	if objectAge > (time.Duration(newerRef) * Day) {
```
balamurugana: same as above

cmd/rm-main.go
```diff
@@ -307,7 +316,8 @@ func mainRm(ctx *cli.Context) error {
 	isRecursive := ctx.Bool("recursive")
 	isFake := ctx.Bool("fake")
 	isStdin := ctx.Bool("stdin")
-	older := ctx.Int("older-than")
+	olderThan := ctx.Int("older-than")
+	newerThan := ctx.Int("newer-than")
```
balamurugana: Is it allowed to use `older-than` and `newer-than` flags together? If true, we would need to update help documents with examples, else we need to error out properly here.  This comment is applicable for all mc commands here.

cmd/utils.go
vadmeste: > you could just return the condition i.e. if is not needed.

Fixed

cmd/utils.go
vadmeste: > same as above

Fixed

cmd/rm-main.go
vadmeste: > Is it allowed to use older-than and newer-than flags together?

Yeah, it is allowed, for example if a user wants to avoid selecting objects belonging to a known time frame, like avoid mirroring all objects wrongly updated in the whole last week.

> If true, we would need to update help documents with examples,

I am the one who don't prefer adding too much number of examples, I already added for --older-than & --newer-than.

balamurugana: I see `older-than` and `-newer-than` as separate examples.  Could you point me where both are in used?

vadmeste: sorry, I mean I already add two separate examples for --older-than and --newer-than and I just think we don't need to add another one to keep the number of examples reduced

balamurugana: IMO at least one example needs to be added to avoid question from users about using both options.

However please update https://github.com/minio/mc/blob/master/docs/minio-client-complete-guide.md for these options.

vadmeste: > However please update https://github.com/minio/mc/blob/master/docs/minio-client-complete-guide.md for these options.

oups yeah, we didn't update that doc for a while now, I see at least mb, mirror, stat commands not updated, let's do that in another PR to review and fix everything at one, I'll open a new issue for that, what do you think ?

balamurugana: OK

vadmeste:
@balamurugana, so can you approve this PR if no more comments?

PR #2143, opened 2017-04-28
mb: Add -p flag to ignore existing bucket/dir

### LINKED ISSUE #2115: Please add an 'ignore existing' flag for 'mb' sub-command
When provisioning an environment (either development, or production-ish) it's often useful to be able to say "create a new bucket in minio (or s3, etc) if it doesn't exist" - essentially the same as the `-p` flag for `mkdir`.

I'd suggest **not** using `-p`, as this has different semantics in `mkdir`. Perhaps `--ignore-existing` is a better option?

vadmeste:
If a remote directory/bucket exists, -p will still make mb cmd
happy if the target already exists.

Fixes #2115

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -942,7 +942,10 @@ func (f *fsClient) listRecursiveInRoutine(contentCh chan *clientContent) {
 }
 
 // MakeBucket - create a new bucket.
-func (f *fsClient) MakeBucket(region string) *probe.Error {
+func (f *fsClient) MakeBucket(region string, ignoreExisting bool) *probe.Error {
+	// TODO: ignoreExisting has no effect currently. In the future, we want
+	// to call os.Mkdir() when ignoredExisting is disabled and os.MkdirAll()
+	// otherwise.
 	e := os.MkdirAll(f.PathURL.Path, 0777)
 	if e != nil {
 		return probe.NewError(e)
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -752,7 +752,7 @@ func isValidBucketName(bucketName string) *probe.Error {
 }
 
 // MakeBucket - make a new bucket.
-func (c *s3Client) MakeBucket(region string) *probe.Error {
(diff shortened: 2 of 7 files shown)


cmd/client-s3.go
```diff
@@ -762,6 +762,10 @@ func (c *s3Client) MakeBucket(region string) *probe.Error {
 	}
 	e := c.api.MakeBucket(bucket, region)
 	if e != nil {
+		// Ignore bucket already existing error when ignoreExisting flag is enabled
+		if ignoreExisting && minio.ToErrorResponse(e).Code == "BucketAlreadyOwnedByYou" {
+			return nil
```
harshavardhana: There is one more Code to look for. `BucketAlreadyExists`  its reported by some regions.

vadmeste: > There is one more Code to look for. BucketAlreadyExists its reported by some regions.

Thanks, done.

cmd/mb-main.go
```diff
@@ -32,6 +32,10 @@ var (
 			Value: "us-east-1",
 			Usage: "Specify bucket region. Defaults to `us-east-1`.",
 		},
+		cli.BoolFlag{
+			Name:  "ignore-existing, p",
+			Usage: "Ignore if remote bucket/directory exists",
```
harshavardhana: remove "remote" from the sentecne. We don't need to say that.

cmd/client-fs.go
```diff
@@ -942,8 +942,13 @@ func (f *fsClient) listRecursiveInRoutine(contentCh chan *clientContent) {
 }
 
 // MakeBucket - create a new bucket.
-func (f *fsClient) MakeBucket(region string) *probe.Error {
-	e := os.MkdirAll(f.PathURL.Path, 0777)
+func (f *fsClient) MakeBucket(region string, ignoreExisting bool) *probe.Error {
+	var e error
+	if ignoreExisting {
+		e = os.MkdirAll(f.PathURL.Path, 0777)
```
harshavardhana: This changes backward compatibility... Lets do 'MkdirAll` anyways until some release in future. Add a TODO to move it back to `Mkdir` in future.

PR #2155, opened 2017-05-14
Add UTCNow() helper function

harshavardhana:
This patch adds UTCNow() function which returns current UTC time.

This is equivalent of UTCNow() == time.Now().UTC()

Refer https://github.com/minio/minio/pull/3931


harshavardhana:
This is ready for review @vadmeste

PR #2166, opened 2017-05-31
cat: Ignore Stat() error and pass to Get()

### LINKED ISSUE #2156: cat: tries to do listing
Is list operation needed for cat command?

```
krishna@escape:~$ mc --debug cat myminio/test/passwd
mc: <DEBUG> GET /test/?location= HTTP/1.1
Host: localhost:9000
User-Agent: Minio (linux; amd64) minio-go/2.0.4 mc/DEVELOPMENT.GOGET
Authorization: AWS4-HMAC-SHA256 Credential=**REDACTED**/20170516/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=**REDACTED**
X-Amz-Content-Sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
X-Amz-Date: 20170516T185647Z
Accept-Encoding: gzip

mc: <DEBUG> HTTP/1.1 403 Forbidden
Transfer-Encoding: chunked
Accept-Ranges: bytes
Content-Type: application/xml
Date: Tue, 16 May 2017 18:56:47 GMT
Server: Minio/DEVELOPMENT.GOGET (linux; amd64)
Vary: Origin
X-Amz-Bucket-Region: us-east-1
X-Amz-Request-Id: 14BF2B1D84DCC1FF

e1
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Access Denied.</Message><Key></Key><BucketName></BucketName><Resource>/test/</Resource><RequestId>3L137</RequestId><HostId>3L137</HostId></Error>
0

mc: <DEBUG> Response Time:  2.075711ms

mc: <DEBUG> GET /test/?delimiter=%2F&max-keys=1000&prefix=passwd HTTP/1.1
Host: localhost:9000
User-Agent: Minio (linux; amd64) minio-go/2.0.4 mc/DEVELOPMENT.GOGET
Authorization: AWS4-HMAC-SHA256 Credential=**REDACTED**/20170516/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=**REDACTED**
X-Amz-Content-Sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
X-Amz-Date: 20170516T185647Z
Accept-Encoding: gzip

mc: <DEBUG> HTTP/1.1 403 Forbidden
Transfer-Encoding: chunked
Accept-Ranges: bytes
Content-Type: application/xml
Date: Tue, 16 May 2017 18:56:47 GMT
Server: Minio/DEVELOPMENT.GOGET (linux; amd64)
Vary: Origin
X-Amz-Bucket-Region: us-east-1
X-Amz-Request-Id: 14BF2B1D84FAE241

e1
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Access Denied.</Message><Key></Key><BucketName></BucketName><Resource>/test/</Resource><RequestId>3L137</RequestId><HostId>3L137</HostId></Error>
0

mc: <DEBUG> Response Time:  2.703904ms

mc: <ERROR> Unable to read from `myminio/test/passwd`. Access Denied.
 (3) cat-main.go:168 cmd.mainCat(..) Tags: [myminio/test/passwd]
 (2) cat-main.go:88 cmd.catURL(..) Tags: [myminio/test/passwd]
 (1) client-url.go:181 cmd.url2Stat(..) Tags: [myminio/test/passwd]
 (0) client-s3.go:880 cmd.(*s3Client).Stat(..)
 Release-Tag:DEVELOPMENT.GOGET | Commit:DEVELOPMENT. | Host:escape | OS:linux | Arch:amd64 | Lang:go1.7.5 | Mem:1.0MB/8.4MB | Heap:1.0MB/4.8MB
krishna@escape:~$ 

```

vadmeste:
client-s3.Get() calls Stat() method because Get() is supposed to return
object metadata. This PR makes a change in client interface so metadata
is returned in Stat() method. This is beneficial for us to avoid calling
HEAD method in S3 when run `mc cat`

Fixes #2156

### DIFF
--- a/cmd/cat-main.go
+++ b/cmd/cat-main.go
@@ -92,7 +92,7 @@ func catURL(sourceURL string) *probe.Error {
 		if client.GetURL().Type == objectStorage {
 			size = content.Size
 		}
-		if reader, err = getSourceStream(sourceURL); err != nil {
+		if reader, err = getSourceStreamFromURL(sourceURL); err != nil {
 			return err.Trace(sourceURL)
 		}
 	}
--- a/cmd/cat-main.go
+++ b/cmd/cat-main.go
@@ -83,13 +83,13 @@ func catURL(sourceURL string) *probe.Error {
 		reader = os.Stdin
 	default:
 		var err *probe.Error
+		// Try to stat the object, the purpose is to extract the
+		// size of S3 object so we can check if the size of the
+		// downloaded object is equal to the original one. FS files
+		// are ignored since some of them have zero size though they
+		// have contents like files under /proc.
 		client, content, err := url2Stat(sourceURL)
-		if err != nil {
-			return err.Trace(sourceURL)
-		}
-		// Ignore size for filesystem objects since os.Stat() would not
-		// return proper size all the time, for example with /proc files.
-		if client.GetURL().Type == objectStorage {
+		if err == nil && client.GetURL().Type == objectStorage {
 			size = content.Size
 		}
 		if reader, err = getSourceStreamFromURL(sourceURL); err != nil {
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -449,9 +449,8 @@ func (f *fsClient) Copy(source string, size int64, progress io.Reader) *probe.Er
 	return nil
 }
 
-// get - get wrapper returning reader and additional metadata if any.
-// currently only returns metadata.
-func (f *fsClient) get() (io.Reader, map[string][]string, *probe.Error) {
+// get - get wrapper returning object reader.
+func (f *fsClient) get() (io.Reader, *probe.Error) {
 	tmppath := f.PathURL.Path
 	// Golang strips trailing / if you clean(..) or
 	// EvalSymlinks(..). Adding '.' prevents it from doing so.
@@ -463,21 +462,18 @@ func (f *fsClient) get() (io.Reader, map[string][]string, *probe.Error) {
 	_, e := filepath.EvalSymlinks(tmppath)
 	if e != nil {
 		err := f.toClientError(e, f.PathURL.Path)
-		return nil, nil, err.Trace(f.PathURL.Path)
+		return nil, err.Trace(f.PathURL.Path)
 	}
 	fileData, e := os.Open(f.PathURL.Path)
 	if e != nil {
 		err := f.toClientError(e, f.PathURL.Path)
-		return nil, nil, err.Trace(f.PathURL.Path)
+		return nil, err.Trace(f.PathURL.Path)
 	}
-	metadata := map[string][]string{
-		"Content-Type": {guessURLContentType(f.PathURL.Path)},
-	}
-	return fileData, metadata, nil
+	return fileData, nil
 }
 
 // Get returns reader and any additional metadata.
-func (f *fsClient) Get() (io.Reader, map[string][]string, *probe.Error) {
+func (f *fsClient) Get() (io.Reader, *probe.Error) {
 	return f.get()
 }
 
@@ -1021,6 +1017,9 @@ func (f *fsClient) Stat(isIncomplete bool) (content *clientContent, err *probe.E
 	content.Size = st.Size()
 	content.Time = st.ModTime()
 	content.Type = st.Mode()
+	content.Metadata = map[string][]string{
+		"Content-Type": {guessURLContentType(f.PathURL.Path)},
+	}
 	return content, nil
 }
 
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -497,40 +497,27 @@ func (c *s3Client) Watch(params watchParams) (*watchObject, *probe.Error) {
 }
 
 // Get - get object with metadata.
-func (c *s3Client) Get() (io.Reader, map[string][]string, *probe.Error) {
+func (c *s3Client) Get() (io.Reader, *probe.Error) {
 	bucket, object := c.url2BucketAndObject()
 	reader, e := c.api.GetObject(bucket, object)
 	if e != nil {
 		errResponse := minio.ToErrorResponse(e)
 		if errResponse.Code == "NoSuchBucket" {
-			return nil, nil, probe.NewError(BucketDoesNotExist{
+			return nil, probe.NewError(BucketDoesNotExist{
 				Bucket: bucket,
 			})
 		}
 		if errResponse.Code == "InvalidBucketName" {
-			return nil, nil, probe.NewError(BucketInvalid{
+			return nil, probe.NewError(BucketInvalid{
 				Bucket: bucket,
 			})
 		}
 		if errResponse.Code == "NoSuchKey" || errResponse.Code == "InvalidArgument" {
-			return nil, nil, probe.NewError(ObjectMissing{})
+			return nil, probe.NewError(ObjectMissing{})
 		}
-		return nil, nil, probe.NewError(e)
+		return nil, probe.NewError(e)
 	}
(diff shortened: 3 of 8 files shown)


vadmeste:
Not WIP anymore. Ready for testing and reviewing and discussing design change of client interface.

cmd/client-fs.go
```diff
@@ -451,7 +451,7 @@ func (f *fsClient) Copy(source string, size int64, progress io.Reader) *probe.Er
 
 // get - get wrapper returning reader and additional metadata if any.
```
krisis: The comment on additional metadata is superfluous with this change, please remove it.

cmd/client-s3.go
```diff
@@ -864,71 +851,75 @@ func (c *s3Client) Stat(isIncomplete bool) (*clientContent, *probe.Error) {
...
-		return nil, probe.NewError(e)
+		return nil, nil, probe.NewError(e)
 	}
 	objectMetadata.URL = *c.targetURL
 	objectMetadata.Time = objectStat.LastModified
 	objectMetadata.Size = objectStat.Size
 	objectMetadata.Type = os.FileMode(0664)
-	return objectMetadata, nil
+
+	metadata := objectStat.Metadata
```
krisis: If `metadata` can be constructed from `*clientContent` then why should we return it separately? Also, we ignore it in most (other) places.

cmd/client-fs.go
vadmeste: > The comment on additional metadata is superfluous with this change, please remove it.

Done

cmd/client-s3.go
vadmeste: > If metadata can be constructed from *clientContent

Not sure what do you mean, `clientContent` is defined in mc and `metadata` is returned from minio-go. so metadata is not constructed from clientContent.

krisis: Isn't​ it possible to construct a `map[string]string` (metadata) from `objectMetadata` even outside this function?

vadmeste: Are you suggesting that we should merge the returned metadata with |objectMetadata` instead of returning it separately ?

krisis: Yes. IIUC, we don't use metadata everywhere. So, I am suggesting we construct `map[string] string` wherever required with corresponding values found in `objectMetadata` from the Stat method. Does that make sense?

vadmeste: > Yes. IIUC, we don't use metadata everywhere. So, I am suggesting we construct map[string] string wherever required with corresponding values found in objectMetadata from the Stat method. Does that make sense?

Yes, I agree.

So the client interface will be changed from:
```
 40 type Client interface {
 41         // Common operations
 42         Stat(isIncomplete bool) (content *clientContent, metadata map[string][]string, err *probe.Error)
```
To
```
 40 type Client interface {
 41         // Common operations
 42         Stat(isIncomplete bool) (content *clientContent, err *probe.Error)
```
and clientContent will have new Metadata additional field:
```
 75 type clientContent struct {
 76         URL  clientURL
 77         Time time.Time
 78         Size int64
 79         Type os.FileMode
            Metadata map[string][]string
 80         Err  *probe.Error
 81 }
```

@abperiasamy, @harshavardhana what do you think ?

harshavardhana: Yes clientContent can have the metadata which is completely fine @vadmeste and it is the right place IMHO.

vadmeste: > Yes clientContent can have the metadata which is completely fine @vadmeste and it is the right place IMHO.

Okay, will do.

vadmeste:
I made further changes, can you review again ? this PR was partly fixing the issue that I wanted to address. But now cat command is also changed.

@deekoder, can you block this PR? because it cann't be fully tested unless Minio cache gateway feature gets merged.

harshavardhana:
ping @krisis anything else from your end on this PR?

cmd/cat-main.go
```diff
@@ -83,16 +83,16 @@ func catURL(sourceURL string) *probe.Error {
...
+		// are ignored since some of them have zero size though they
+		// have contents like files under /proc.
 		client, content, err := url2Stat(sourceURL)
-		if err != nil {
-			return err.Trace(sourceURL)
-		}
-		// Ignore size for filesystem objects since os.Stat() would not
-		// return proper size all the time, for example with /proc files.
-		if client.GetURL().Type == objectStorage {
+		if err == nil && client.GetURL().Type == objectStorage {
```
krisis: Previously, we return immediately if `err != nil` from `url2Stat`. Why are we ignoring the error in this change?

vadmeste: > Previously, we return immediately if err != nil from url2Stat. Why are we ignoring the error in this change?

In the first place, url2Stat() is only needed to fetch the size of the remote object so we can check that we downloaded all the object data from the server. So actually, testing on url2Stat() is not crucial since we call getSourceStreamFromURL() later. 

But why we ignore that error now? it is because we need to enhance support of mc with the new minio cache gateway. In some cases, with the cache gateway, a HEAD object returns 404 but GET can return 200 with object data. Refer to #2156 or ping me online.

krisis: @vadmeste, OK. I am withholding approval of PR pending test. Otherwise, the change looks good to me.

cmd/client-s3.go
```diff
@@ -894,12 +884,14 @@ func (c *s3Client) Stat(isIncomplete bool) (*clientContent, *probe.Error) {
 			objectMetadata.Time = objectStat.LastModified
 			objectMetadata.Size = objectStat.Size
 			objectMetadata.Type = os.FileMode(0664)
+			objectMetadata.Metadata = map[string][]string{}
```
krisis: why should we try to find the given object by listing the objects in the bucket for fetching stat information? We can directly call `c.api.StatObject(bucket, object)`, right?

vadmeste: @krisis, Stat() in client-s3 is supposed to be able to Stat() a directory but StatObject() of minio-go cannot.

vadmeste:
Can we allow this to be merged? reviewers can mark this PR reviewed and not tested

cmd/client-s3.go
krisis: @vadmeste Ah, didn't see that. Thanks for the explanation.

deekoder:
I am not sure if I should merge it if its not tested.

vadmeste:
> I am not sure if I should merge it if its not tested.

This PR was open for a long time, we can't test because Minio cache is not ready yet, also by time we will have to rebase it many times. 

Also I'll have less opened issues in my buzz list if we merge this :)

krisis:
@deekoder Tested locally (using `mc cat --debug`) and works as expected. Like @vadmeste mentioned, this change can't be tested in the context of edge cache.

harshavardhana:
> @deekoder Tested locally (using mc cat --debug) and works as expected. Like @vadmeste mentioned, this change can't be tested in the context of edge cache.

IMO its a good change to have anycase.

deekoder:
Since its approved by the reviewers, can i merge it? Its not clear whether testability was an issue?

vadmeste:
> Since its approved by the reviewers, can i merge it?

Yes you can.


> Its not clear whether testability was an issue?

Because Minio cache is not ready yet.. reviewers checked debug log to confirm it is working as expected

PR #2177, opened 2017-06-07
Updated build instructions from source.

CarterMcClellan:
Updated build instructions to match harshavardhana's request

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,10 @@ If you do not have a working Golang environment, please follow [How to install G
 
 ```sh
 go get -u github.com/minio/mc
+cd github.com/minio/mc
+make
 ```
+
 
 ## Add a Cloud Storage Service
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -101,8 +101,11 @@ If you do not have a working Golang environment, please follow [How to install G
 
 ```sh
 go get -u github.com/minio/mc
+cd github.com/minio/mc
```
harshavardhana: Lets change these instructions like this 

```sh
go get -d github.com/minio/mc
cd ${GOPATH}/src/github.com/minio/mc
make
```

`go get -u` in-fact downloads, updates and compiles. We only wish to download in this situation so we can just use `go get -d` which indicates the go tool to download the source only not compile it.  Since compilation is achieved by doing `make` instead.

PR #2179, opened 2017-06-09
config: default S3v2 for google endpoints

CarterMcClellan:
config: Force S3v2 for google endpoints

### DIFF
--- a/cmd/config-host-add.go
+++ b/cmd/config-host-add.go
@@ -17,6 +17,8 @@
 package cmd
 
 import (
+	urlpkg "net/url"
+
 	"github.com/fatih/color"
 	"github.com/minio/cli"
 	"github.com/minio/mc/pkg/console"
@@ -126,6 +128,16 @@ func mainConfigHostAdd(ctx *cli.Context) error {
 	accessKey := args.Get(2)
 	secretKey := args.Get(3)
 	api := args.Get(4)
+
+	urlHost, _ := urlpkg.Parse(url)
+
+	if isGoogle(urlHost.Host) {
+		if api == "S3v4" {
+			fatalIf(errInvalidArgument().Trace(api), "Unsupported API signature for url. Please use `mc config host add "+alias+" "+url+" "+accessKey+" "+secretKey+" S3v2` instead")
+		}
+		api = "S3v2"
+	}
+
 	if api == "" {
 		api = "S3v4"
 	}
--- a/cmd/config-host-add.go
+++ b/cmd/config-host-add.go
@@ -129,11 +129,11 @@ func mainConfigHostAdd(ctx *cli.Context) error {
 	secretKey := args.Get(3)
 	api := args.Get(4)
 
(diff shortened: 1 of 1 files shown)


cmd/config-host-add.go
```diff
@@ -128,6 +128,9 @@ func mainConfigHostAdd(ctx *cli.Context) error {
 	api := args.Get(4)
 	if api == "" {
 		api = "S3v4"
+		if url == "https://storage.googleapis.com" {
```
harshavardhana: We have to do this little elaborate check to cater for these conditions. 

- URL is `storage.googleapis.com`  and then no signature type was specified default to `s3v2` which this patch addresses. 
- URL is `storage.googleapis.com` and then signature type was specified by it was specified wrong as `s3v4` where we need to error out indicating that user should use `s3v2` instead in the error message. 

To check if the url is `google` we should just use the helper 

```go
if isGoogle(url) {
    api = "S3v2"
}
```

cmd/client-s3.go
```diff
@@ -940,7 +940,11 @@ func isAmazonAccelerated(host string) bool {
 }
 
 func isGoogle(host string) bool {
-	return s3utils.IsGoogleEndpoint(url.URL{Host: host})
+	url, err := url.Parse(host)
```
vadmeste: @CarterMcClellan thanks for this PR.

`isGoogle()` is used to receive only a host address without scheme and path. 

Now the problem is that, if we go with this change and then move to go 1.8, we will have the following (unexpected, I know) result:

`isGoogle("storage.googleapis.com") = false`

That's because with go 1.8, `url.Parse("storage.googleapis.com")` returns `url.Host = ""` and `url.Path = "storage.googleapis.com"`

So I really suggest moving url.Parse() code in config-host-add.go with the change that you already did.

cmd/config-host-add.go
```diff
@@ -126,6 +128,16 @@ func mainConfigHostAdd(ctx *cli.Context) error {
 	accessKey := args.Get(2)
 	secretKey := args.Get(3)
 	api := args.Get(4)
+
+	urlHost, _ := urlpkg.Parse(url)
```
vadmeste: Could you rename `urlHost` to `parsedURL` ? that would be more meaningful.

cmd/config-host-add.go
```diff
@@ -126,6 +128,16 @@ func mainConfigHostAdd(ctx *cli.Context) error {
...
+
+	if isGoogle(urlHost.Host) {
+		if api == "S3v4" {
+			fatalIf(errInvalidArgument().Trace(api), "Unsupported API signature for url. Please use `mc config host add "+alias+" "+url+" "+accessKey+" "+secretKey+" S3v2` instead")
```
vadmeste: A missing dot at the end of the error message.

vadmeste:
One final comment is, can you reword your commit message like:

`config: Force S3v4 for google endpoints`

PR #2180, opened 2017-06-10
config: Hosts listing cuts fields and adds '...'

vadmeste:
Fields's contents are cutted when the user lists
the configured aliases (mc config host list) 

New format:

  New host list format when listing several hosts
```
  myminiopublic :    http://localhost:9000
  myminios      :    https://minio1.duckdns.org:...  Q3AM3UQ867SPQQA43P2F  zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG  S3v4
  testazure     :    http://172.17.0.7:9000          myazureac             Gh5543wekitnifILbZam1KYY3TGlswRminio1...  S3v4
```
  
  New host list format when listing one host:
```
  azure :  http://localhost:9000  myazureac  Gh5543wekitnifILbZam1KYY3TGlswRminio1u7BJ86wekitnifILbZam1K==  S3v4

```

### DIFF
--- a/cmd/config-host-list.go
+++ b/cmd/config-host-list.go
@@ -94,7 +94,12 @@ func printHosts(hosts ...hostMessage) {
 	}
 	for _, host := range hosts {
 		// Format properly for alignment based on alias length.
-		host.Alias = fmt.Sprintf("%-*.*s", maxAlias, maxAlias, host.Alias)
+		host.Alias = fmt.Sprintf("%-*.*s:", maxAlias, maxAlias, host.Alias)
+		if host.AccessKey == "" || host.SecretKey == "" {
+			host.AccessKey = ""
+			host.SecretKey = ""
+			host.API = ""
+		}
 		printMsg(host)
 	}
 }
@@ -115,12 +120,13 @@ func listHosts(alias string) {
 	if alias != "" {
 		if v, ok := conf.Hosts[alias]; ok {
 			printHosts(hostMessage{
-				op:        "list",
-				Alias:     alias,
-				URL:       v.URL,
-				AccessKey: v.AccessKey,
-				SecretKey: v.SecretKey,
-				API:       v.API,
+				op:          "list",
+				prettyPrint: false,
+				Alias:       alias,
+				URL:         v.URL,
+				AccessKey:   v.AccessKey,
+				SecretKey:   v.SecretKey,
+				API:         v.API,
 			})
 			return
 		}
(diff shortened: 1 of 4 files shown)


harshavardhana:
> play : https://play.minio.io:9000 | Q3AM3UQ867SPQQA43P2F | zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG | S3v4

The reason a pipe was avoided was that its easy to copy paste for sharing the keys, this makes it harder.

vadmeste:
> The reason a pipe was avoided was that its easy to copy paste for sharing the keys, this makes it harder.

I usually double click when I want to copy, also I am not sure what you suggest ?

harshavardhana:
> I usually double click when I want to copy, also I am not sure what you suggest ?

It is like this currently this is how it is printed 

```
mc config host list  play
play: https://play.minio.io:9000      Q3AM3UQ867SPQQA43P2F  zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG  S3v4
```

What i generally do is to share i just copy paste the the entire line after `play:` we can't do that with adding special separators like `|` 

Lets do it like this specified command without any alias

```
mc config host list
``` 

Should list the longer strings as `...`  with a limited character length of 40 by default. 

If alias is indeed specified like 

```
mc config host list play
``` 

Then we can display the entire access and secret keys without `...` - it helps keep the current behavior and also keep the output cleaner.

vadmeste:
okay nice idea @harshavardhana.

vadmeste:
Done @harshavardhana

cmd/config-host.go
```diff
@@ -48,26 +48,72 @@ func mainConfigHost(ctx *cli.Context) error {
...
+
+		// Format fields and construct message
+		message := ""
+		for _, f := range fields {
+			// Skip empty field
+			// if len(f.content) == 0 {
+			//	continue
+			// }
+			// Default field format without pretty effect
+			fieldFormat := "%s  "
```
harshavardhana: It would be nice to convert this to a proper function perhaps? generalize it and add a unit test. ?

cmd/config-host-list.go
```diff
@@ -94,7 +94,12 @@ func printHosts(hosts ...hostMessage) {
 	}
 	for _, host := range hosts {
 		// Format properly for alignment based on alias length.
-		host.Alias = fmt.Sprintf("%-*.*s", maxAlias, maxAlias, host.Alias)
+		host.Alias = fmt.Sprintf("%-*.*s :", maxAlias, maxAlias, host.Alias)
+		if host.AccessKey == "" || host.SecretKey == "" {
+			host.AccessKey = ""
+			host.SecretKey = ""
+			host.API = ""
```
harshavardhana: Perhaps we can save empty API string when the `config host add` was called first time? instead of doing this later?

vadmeste: > Perhaps we can save empty API string when the config host add was called first time? instead of doing this later?

I think it is better if we do that here since users can modify config.json manually.. we can't prevent them.

harshavardhana: Ah yes. this is also needed..

vadmeste:
Comments addressed.

cmd/pretty-table.go
```diff
@@ -0,0 +1,84 @@
...
+		} else {
+			fieldContent = contents[i]
+		}
+
+		// Add separator if this is not the last column
+		if i < totalColumns-1 {
+			fieldFormat += t.separator
+		}
+
+		// Add the field to the resulted message
```
poornas: change // Add the field to the resulted message to 
// Add the field to the message

PR #2181, opened 2017-06-12
Fix travis and some code formatting

vadmeste:


vadmeste:
Looks like .travis was not checking source code formatting, can you check @harshavardhana ?

harshavardhana:
> Looks like .travis was not checking source code formatting, can you check @harshavardhana ?

It does check @vadmeste you need to add the code that i mentioned from minio repo for it to fail.

vadmeste:
> It does check @vadmeste you need to add the code that i mentioned from minio repo for it to fail.

yes, I meant there was a double problem, you can check the change in .travis

harshavardhana:
@vadmeste but you already fixed the erroneous files :-)

PR #2182, opened 2017-06-14
Add Docker release files for mc

NitishT:
These Dockerfile(s) will be used for `mc` release images for various platforms. 

Statically built binary is downloaded from dl.minio.io website and used in the docker image

### DIFF
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -1,0 +1,9 @@
+FROM alpine:3.5
+
+RUN \
+    apk add --no-cache ca-certificates && \
+    apk add --no-cache --virtual .build-deps curl && \
+    curl https://dl.minio.io/client/mc/release/linux-amd64/mc > /usr/bin/mc && \
+    chmod +x /usr/bin/mc && apk del .build-deps
+
+ENTRYPOINT ["mc"]
--- a/Dockerfile.release.aarch64
+++ b/Dockerfile.release.aarch64
@@ -1,0 +1,9 @@
+FROM resin/aarch64-alpine:3.5
+
+RUN \
+    apk add --no-cache ca-certificates && \
+    apk add --no-cache --virtual .build-deps curl && \
+    curl https://dl.minio.io/client/mc/release/linux-arm64/mc > /usr/bin/mc && \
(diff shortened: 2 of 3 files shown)


NitishT:
Built locally using https://blog.hypriot.com/post/setup-simple-ci-pipeline-for-arm-images/ works fine.

Dockerfile.release
```diff
@@ -0,0 +1,12 @@
+FROM alpine:3.5
+
+WORKDIR /go/src/app
```
harshavardhana: You don't need this.. @NitishT also you can follow Dockerfile.release 

```
     apk add --no-cache ca-certificates && \
     apk add --no-cache --virtual .build-deps curl && \
```

We need ca-certificates as well. 
```
     echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf && \
```

Dockerfile.release
```diff
@@ -0,0 +1,12 @@
...
+
+COPY . /go/src/app
+
+RUN \
+    apk add --no-cache curl && \
+    curl https://dl.minio.io/client/mc/release/linux-amd64/mc > /usr/bin/mc && \
+    chmod +x /usr/bin/mc
+
+ENTRYPOINT ["mc"]
```
harshavardhana: Odd newline here.

Dockerfile.release.aarch64
```diff
@@ -0,0 +1,12 @@
+FROM resin/aarch64-alpine:3.5
+
+WORKDIR /go/src/app
+
+COPY . /go/src/app
```
harshavardhana: Same as above.

Dockerfile.release.armhf
```diff
@@ -0,0 +1,12 @@
+FROM resin/armhf-alpine:3.5
+
+WORKDIR /go/src/app
+
```
harshavardhana: Same as above.

NitishT:
Updated as per your comments @harshavardhana 

Also as discussed, did not use
 
```
echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf && \
```

PR #2188, opened 2017-06-21
mirror: Add timestamp check similar to aws s3 sync

### LINKED ISSUE #2187: Minio fails to mirror file after changes if file's size matches
## Expected behaviour

`mc mirror` should always mirror files with changes

## Actual behaviour

If the file size is the same, but the content differs then `mc` ignores the change.

Failing use-cases:

* File contents changed from "test" to "tset"
* File metadata changed (from changed yesterday to changed today)

## mc version

I've tried with the latest `mc` version from June.

## System information

Tested clients - Mac or Windows 2012.

Server versions tested (via Docker image) 

```
Version: 2017-04-29T00:40:27Z
Release-Tag: RELEASE.2017-04-29T00-40-27Z
Commit-ID: eb50175ad911d496bf583a599de89547f0c9be89
```

```
Version: 2017-06-13T19:01:01Z
Release-Tag: RELEASE.2017-06-13T19-01-01Z
Commit-ID: 353f2d3a6e30f595a721abcd9cf1b108b041ee21
```

harshavardhana:
Currently `mc mirror` requires source to be
uploaded only if the size is different than the
object on remote server. This PR adds support to
check for last modified time of the source object
as well such that we upload if there are timestamp
differences as well.

Fixes #2187

### DIFF
--- a/cmd/diff-main.go
+++ b/cmd/diff-main.go
@@ -91,6 +91,8 @@ func (d diffMessage) String() string {
 		msg = console.Colorize("DiffType", "! "+d.SecondURL)
 	case differInSize:
 		msg = console.Colorize("DiffSize", "! "+d.SecondURL)
+	case differInTime:
+		msg = console.Colorize("DiffTime", "! "+d.SecondURL)
 	default:
 		fatalIf(errDummy().Trace(d.FirstURL, d.SecondURL),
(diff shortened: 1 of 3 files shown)


harshavardhana:
This is approved by @abperiasamy - taking it in.

PR #2189, opened 2017-06-21
mc: Inherit global flags for initialization messages.

### LINKED ISSUE #2089: mc: Global flags do not affect initialization messages
## Expected behaviour
Global flags `--no-color`, `--quiet`, `--json`, and `--debug` should be used for all output
All output should be de-colorized/supressed/json/verbose as specified
## Actual behaviour
The initialization messages are output with color in plaintext regardless of the specified options
```
$ mc --no-color --quiet --json --config-folder=/tmp config host list
mc: Configuration written to ‘/tmp/config.json’. Please update your access credentials.
mc: Successfully created ‘/tmp/share’.
mc: Initialized share uploads ‘/tmp/share/uploads.json’ file.
mc: Initialized share downloads ‘/tmp/share/downloads.json’ file.
{"status":"success","alias":"gcs  ","URL":"https://storage.googleapis.com","accessKey":"YOUR-ACCESS-KEY-HERE","secretKey":"YOUR-SECRET-KEY-HERE","api":"S3v2"}
```
image

1. Run `mc` with one of the above options with a `config-folder` that has not been previously used

## mc version
```
Version: 2017-02-06T20:16:19Z
Release-tag: RELEASE.2017-02-06T20-16-19Z
Commit-id: 2c8115de4edc5612525488b1b3b804689d336d01
```
## System information
Bash shell on Ubuntu

harshavardhana:
Fixes #2089

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -181,7 +181,9 @@ func initMC() {
 		err := saveMcConfig(newMcConfig())
 		fatalIf(err.Trace(), "Unable to save new mc config.")
 
-		console.Infoln("Configuration written to `" + mustGetMcConfigPath() + "`. Please update your access credentials.")
+		if !globalQuiet && !globalJSON {
+			console.Infoln("Configuration written to `" + mustGetMcConfigPath() + "`. Please update your access credentials.")
+		}
 	}
 
 	// Check if mc session folder exists.
@@ -218,11 +220,11 @@ func registerBefore(ctx *cli.Context) error {
 	// Migrate any old version of config / state files to newer format.
 	migrate()
 
-	// Initialize default config files.
-	initMC()
-
 	// Set global flags.
 	setGlobalsFromContext(ctx)
+
+	// Initialize default config files.
+	initMC()
 
 	// Check if config can be read.
 	checkConfig()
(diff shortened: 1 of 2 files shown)


cmd/main.go
```diff
@@ -174,14 +175,31 @@ func getSystemData() map[string]string {
 	}
 }
 
+type initMessage struct {
```
vadmeste: Can you add a comment here?

cmd/main.go
```diff
@@ -174,14 +175,31 @@ func getSystemData() map[string]string {
...
+	Status  string `json:"status"`
+	Message string `json:"msg"`
+}
+
+func (m initMessage) String() string {
+	return m.Message
```
vadmeste: Missing some colors here (it was green)

harshavardhana:
Fixed all the comments @vadmeste

harshavardhana:
> With this output, since we don't have a type field it will be tough for the parsers to know if it is an init message or a config-list message. Better not to intermix messages of different types? If we want to intermix then the message should contain a type field for the parsers to differentiate.

Json is free from those problems if the relevant fields don't exist they just are nil. So it is not a big deal this is just for consistency. Ideally this case is never hit and it is usually never parsed in this manner.

vadmeste:
I see it is better if we don't print init related messages when json is activated. It doesn't seem that someone will need that. Besides as @krishnasrinivas said, it intermixes json structure.

harshavardhana:
> I see it is better if we don't print init related messages when json is activated. It doesn't seem that someone will need that. Besides as @krishnasrinivas said, it intermixes json structure.

Made the change check again @vadmeste @krishnasrinivas - please!

PR #2200, opened 2017-07-02
mirror/diff: Ignore errors during objectDifference.

### LINKED ISSUE #2199: Handling filesystem stat errors
Is there a way to not fail an entire mirror operation or simply skip errors, such as stat related ones? I'm attempting to mirror a relatively large directory structure with some potentially invalid symlinks. These seem to throw a stat error (i presume when trying to follow them) and the entire mirror operation fails. 

I'd like to skip any error'd files and mirror the rest if at all possible.

## Expected behaviour
Skip (or allow by option) errors with specific files during a mirror operation.

## Actual behaviour
Mirror operation fails with an error as follows:
mc: <ERROR> Unable to mirror objects. stat /mnt/test/file: bad message
 (1) difference.go:95 cmd.difference.func1(..) 
 (0) client-fs.go:929 cmd.(*fsClient).listRecursiveInRoutine(..)

Attempt to mirror a directory structure with an invalid symbolic link or potentially a file without read access.

## mc version
Version: 2017-06-15T03:38:43Z
Release-tag: RELEASE.2017-06-15T03-38-43Z
Commit-id: c32b1ea48b651125e33015ea0d1596cb377685a4

## System information
CentOS Linux release 7.3.1611 (Core) 
3.10.0-514.26.1.el7.x86_64

harshavardhana:
Files which are broken in symlink, not accessible
and unsupported should be ignored. We should proceed
to next object.

Fixes #2199

### DIFF
--- a/cmd/diff-main.go
+++ b/cmd/diff-main.go
@@ -180,7 +180,8 @@ func doDiffMain(firstURL, secondURL string) error {
 	for diffMsg := range objectDifference(firstClient, secondClient, firstURL, secondURL) {
 		if diffMsg.Error != nil {
 			errorIf(diffMsg.Error, "Unable to calculate objects difference.")
-			break
+			// Ignore error and proceed to next object.
+			continue
 		}
 		printMsg(diffMsg)
 	}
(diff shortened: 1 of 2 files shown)


harshavardhana:
why is this blocked @deekoder ?

harshavardhana:
Any reviews?

PR #2217, opened 2017-07-16
cat: Print ^? for unprintable characters 

### LINKED ISSUE #2185: [feature] cat: Binary dump on terminal by default can lead to terminal crash.
## Expected behaviour
Avoid binary file to be written to terminal if the data is in binary format. Borrowing ideas from https://github.com/curl/curl/commit/5385450afd61328e7d24b50eeffc2b1571cd9e2f

## Actual behaviour
`mc cat` on a binary file such as video or images can in-fact crash the terminal, some times leads to data loss. 

`mc cat` a video or a image file on a gnome terminal.

## mc version
All versions

## System information
All Linux terminals.

vadmeste:
In terminals, cat avoids directly printing some binary characters, such
as control characters or non printable characters but prints ^? instead,
that way, cat will avoid corrupting terminals

Fixes #2185

### DIFF
--- a/cmd/cat-main.go
+++ b/cmd/cat-main.go
@@ -17,12 +17,17 @@
 package cmd
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
 	"os"
 	"strings"
 	"syscall"
+	"unicode"
+	"unicode/utf8"
+
+	"golang.org/x/crypto/ssh/terminal"
 
 	"github.com/minio/cli"
 	"github.com/minio/mc/pkg/probe"
@@ -59,6 +64,55 @@ EXAMPLES:
       $ {{.HelpName}} part.* > complete.img
 
 `,
+}
+
+// prettyStdout replaces some non printable characters
+// with <hex> format to be better viewable by the user
+type prettyStdout struct {
+	// Internal data to pretty-print
+	writer io.Writer
+	// Internal buffer which contains pretty printed
+	// form of binary (no printable) characters
+	buffer *bytes.Buffer
+}
+
+// newPrettyStdout returns an initialized prettyStdout struct
+func newPrettyStdout(w io.Writer) *prettyStdout {
+	return &prettyStdout{
+		writer: w,
+		buffer: bytes.NewBuffer([]byte{}),
+	}
+}
+
+// Read() returns pretty printed binary characters
+func (s prettyStdout) Write(input []byte) (int, error) {
+	inputLen := len(input)
+
+	// Convert no printable characters to '^?'
+	// and fill into s.buffer
+	for len(input) > 0 {
+		r, size := utf8.DecodeRune(input)
+		if unicode.IsPrint(r) || unicode.IsSpace(r) {
+			s.buffer.WriteRune(r)
+		} else {
+			s.buffer.WriteString("^?")
+		}
+		input = input[size:]
+	}
+
+	bufLen := s.buffer.Len()
+
+	// Copy all buffer content to the writer (stdout)
+	n, err := s.buffer.WriteTo(s.writer)
+	if err != nil {
+		return 0, err
+	}
+
+	if int(n) != bufLen {
+		return 0, errors.New("error when writing to stdout")
+	}
+
+	return inputLen, nil
 }
 
 // checkCatSyntax performs command-line input validation for cat command.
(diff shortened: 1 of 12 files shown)


cmd/cat-main.go
```diff
@@ -105,6 +162,13 @@ func catOut(r io.Reader, size int64) *probe.Error {
 	var n int64
 	var e error
 
+	// In case of a user showing the object content in a terminal,
+	// avoid printing control and other bad characters to avoid
+	// terminal session corruption
+	if terminal.IsTerminal(int(os.Stdout.Fd())) {
```
balamurugana: IMO we would need to introduce an option and work accordingly exactly like `cat -A` option

We shouldn't display control chars in hex automagically.

vadmeste: > IMO we would need to introduce an option and work accordingly exactly like cat -A option

Looks like I can make an output similar to cat -A but I am not too much convinced how much would that be useful.

However, I don't think we need to add any new option flag: If the output is a terminal, it means the user wants to visually see that data, so I guess we can make the output prettier. But if the output is not a terminal, like `mc cat myminio/testbucket/object >/tmp/file` then we won't modify the output.

What do you think @balamurugana  ?

balamurugana: @vadmeste I used `cat -A` several times to view text files to know DOS format and tab character.  I have no experience in binary files with that.  We could intelligently print binary files with hex or other encoding.  If we don't go with `cat -A` behavior, someone can use it like `mc cat ... | cat -A`

I see `less` and `cat -A` show same output.

harshavardhana:
what was finally decided on this? @vadmeste @balamurugana @krisis

vadmeste:
The purpose of this PR is to protect the terminal from random contents coming from `mc cat`, it is a common case that the terminal gets corrupted after running a cat on a binary file, however, @balamurugana thinks that we also need to add a flag to `mc cat` to be able to see \r\n characters, though we can do that using `mc cat | cat -A`

We probably need to vote. :)

balamurugana:
@vadmeste after my comment, we discussed more on slack and came to the conclusion that
* no extra argument is required (mc cat -A is not needed)
* this PR would encode binary content like output of `cat -A` or `less` command

vadmeste:
For the sake of simplification, this PR only prints printable characters and all whitespaces, other not printable will be printed as '^?'. This would avoid terminal corrupted.

Tested using the following commands:
```sh
$ dd if=/dev/urandom of=/tmp/random.data bs=1024 count=1024
$ mc cat /tmp/random.data # no more terminal corruption
```

@balamurugana, @krisis, can you review again ?

vadmeste:
> @vadmeste can you add some unit tests here?

Done

krisis:
@vadmeste I tested this PR with the following steps,
```
dd if=/dev/urandom of=/tmp/non-print bs=10K count=1
mc cp /tmp/non-print s1/test/non-print
mc cat s1/test/non-print
```
The last command returns the following error,
```mc: <ERROR> Unable to read from `s1/test/non-print`. Input reader closed pre-maturely. Expected `10240` bytes, but only received `20188` bytes.```

**Comment by vadmeste** [comment]
@krisis, this is fixed. But I had to change nearly everthing, can you review again?

File: cmd/cat-main.go [hunk 2]
```diff
@@ -61,6 +66,55 @@ EXAMPLES:
...
+
+// Read() returns pretty printed binary characters
+func (s prettyStdout) Write(input []byte) (int, error) {
+	inputLen := len(input)
+
+	// Convert no printable characters to '^?'
+	// and fill into s.buffer
+	for len(input) > 0 {
+		r, size := utf8.DecodeRune(input)
+		if unicode.IsPrint(r) || unicode.IsSpace(r) {
```
krisis: we could simplify this to `unicode.IsPrint(r)` since `' '` is a printable character.

cmd/cat_test.go
```diff
@@ -0,0 +1,64 @@
...
+	testCases := []struct {
+		originText string
+		prettyText string
+	}{
+		{"", ""},
+		{"text", "text"},
+		// Check with new lines
+		{"text\r\n", "text\r\n"},
+		{"\ttext\n", "\ttext\n"},
+		// Print unicode and check it is not alterd
```
krisis: typo: s/alterd/altered

cmd/cat-main.go
vadmeste: > we could simplify this to unicode.IsPrint(r) since ' ' is a printable character.

yeah, true for space, but we need IsSpace() for other characters like '\n', '\r', '\t'

```
unicode.IsPrint('\n') => false
unicode.IsSpace('\n') => true
```

cmd/cat_test.go
vadmeste: > typo: s/alterd/altered

Done

PR #2219, opened 2017-07-17
Correct typos in mc admin heal warning message.

deekoder:


cmd/admin-main.go
```diff
@@ -43,7 +43,7 @@ var adminCmd = cli.Command{
 }
 
 func mainAdminBefore(ctx *cli.Context) error {
-	color.Yellow("\t *** Warning:Experiment feature. Not Ready for Production ***")
+	color.Yellow("\t *** mc admin heal is EXPERIMENTAL ***")
```
vadmeste: This message will be shown in all admin commands and not only with heal. I am not sure we want for all admin commands or only for heal.

deekoder: My understanding is we want this when they type mc admin itself. We do have to fix this error message asap. Right now the typo is egregious.

harshavardhana:
@vadmeste can you take a look again.. `mc admin` cli full change is good.

deekoder:
Had a discussion with @vadmeste who feels this message should not show up when user types `mc admin info` if we only want to say heal is experimental. But @deekoder wants to fix this asap as its egregious. Merging this and sending PR to move message to heal subcommand in new issue.

Issue #2212, opened 2017-07-12
mc diff help output is inconsistent


harshavardhana:
## Expected behaviour
mc diff should have the same output as our other commands. DESCRIPTION shouldn't be added or printed in this manner.

## Actual behaviour
Current behavior is we print DESCRIPTION in the help text which is not correct and is inconsistent with other commands in `mc`.
```
$ mc diff
NAME:
  mc diff - Show differences between two folders or buckets.

USAGE:
  mc diff [FLAGS] FIRST SECOND

FLAGS:
  --config-folder value, -C value  Path to configuration folder. (default: "/home/harsha/.mc")
  --quiet, -q                      Suppress chatty console output.
  --no-color                       Disable color theme.
  --json                           Enable JSON formatted output.
  --debug                          Enable debug output.
  --insecure                       Disable SSL certificate verification.
  --help, -h                       Show help.
  
DESCRIPTION:
  Diff only lists missing objects or objects with size differences. It *DOES NOT* compare contents. i.e. Objects of same name and size, but differ in contents are not noticed.

OUTPUT:
  Differences between source and destination are represented by mark notations with following meaning:
    > - indicates file should be copied.
    < - indicates file should be deleted.
    ! - indicates file differs in size or type.

EXAMPLES:
  1. Compare a local folder with a folder on Amazon S3 cloud storage.
     $ mc diff ~/Photos s3/MyBucket/Photos

  2. Compare two different folders on a local filesystem.
     $ mc diff ~/Photos /Media/Backup/Photos
```

## Steps to reproduce the behaviour
Type `mc diff`

## mc version
master

## System information

All operating systems.

PR #2222, opened 2017-07-18
standardize diff help message

poornas:
Fixes issue #2212 to standardize diff help message to be similar to other mc commands.

### DIFF
--- a/cmd/diff-main.go
+++ b/cmd/diff-main.go
@@ -35,14 +35,13 @@ var (
 // Compute differences between two files or folders.
 var diffCmd = cli.Command{
 	Name:        "diff",
-	Usage:       "Show differences between two folders or buckets.",
-	Description: "Diff only lists missing objects or objects with size differences. It *DOES NOT* compare contents. i.e. Objects of same name and size, but differ in contents are not noticed.",
+	Usage:       "List objects with size difference or missing between two folders or buckets.",
+	Description: "Diff only lists missing objects or objects with size differences. It *DOES NOT* compare contents. i.e. objects of same name and size, but differing in content are not noticed.",
 	Action:      mainDiff,
 	Before:      setGlobalsFromContext,
 	Flags:       append(diffFlags, globalFlags...),
 	CustomHelpTemplate: `NAME:
   {{.HelpName}} - {{.Usage}}
-
 USAGE:
   {{.HelpName}} [FLAGS] FIRST SECOND
 
@@ -50,9 +49,7 @@ FLAGS:
   {{range .VisibleFlags}}{{.}}
   {{end}}
 DESCRIPTION:
-  {{.Description}}
-
-OUTPUT:
+	{{.Description}}
   Differences between source and destination are represented by mark notations with following meaning:
     > - indicates file should be copied.
     < - indicates file should be deleted.


cmd/diff-main.go
```diff
@@ -42,16 +42,12 @@ var diffCmd = cli.Command{
...
   {{.HelpName}} - {{.Usage}}
-
 USAGE:
   {{.HelpName}} [FLAGS] FIRST SECOND
 
 FLAGS:
   {{range .VisibleFlags}}{{.}}
   {{end}}
-DESCRIPTION:
-  {{.Description}}
```
vadmeste: @poornas, by removing description, we will lose this text:
`  Diff only lists missing objects or objects with size differences. It *DOES NOT* compare contents. i.e. Objects of same name and size, but differ in contents are not noticed.` which is kind of important to know.

Can you you manage to add this information in USAGE ?

poornas: done.

cmd/diff-main.go
```diff
@@ -41,17 +41,13 @@ var diffCmd = cli.Command{
 	Before:      setGlobalsFromContext,
 	Flags:       append(diffFlags, globalFlags...),
 	CustomHelpTemplate: `NAME:
-  {{.HelpName}} - {{.Usage}}
-
+  {{.HelpName}} - {{.Usage}}{{.Description}}
```
vadmeste: I think it is better to remove Description and modify Usage

cmd/diff-main.go
```diff
@@ -34,24 +34,19 @@ var (
...
 // Compute differences between two files or folders.
 var diffCmd = cli.Command{
-	Name:        "diff",
-	Usage:       "Show differences between two folders or buckets.",
-	Description: "Diff only lists missing objects or objects with size differences. It *DOES NOT* compare contents. i.e. Objects of same name and size, but differ in contents are not noticed.",
-	Action:      mainDiff,
-	Before:      setGlobalsFromContext,
-	Flags:       append(diffFlags, globalFlags...),
+	Name:   "diff",
+	Usage:  "Show differences between two folders or buckets. Diff only lists missing objects or objects with size differences. It *DOES NOT* compare contents. i.e. Objects of same name and size, but differ in contents are not noticed.",
```
balamurugana: I think below `Usage` is better
```
List objects with size difference or missing between two folders or buckets
```

In `Description` you could have what algo it does like
```
Diff only lists missing objects or objects with size differences between given folders or buckets. It *DOES NOT* check for content difference of objects those having same name and size.
```

I can even think of removing `OUTPUT` section and put the info into `Description` like
```
Diff only lists missing objects or objects with size differences between given folders or buckets. It *DOES NOT* check for content difference of objects those having same name and size.

Differences are marked in output by:
    > - indicates object missing in destination.
    < - indicates object missing in the source.
    ! - indicates object size differed in source and destination.
```

poornas: @balamurugana, Description section is being removed here to make it consistent with other mc commands.

cmd/diff-main.go
```diff
@@ -34,25 +34,21 @@ var (
...
   {{.HelpName}} [FLAGS] FIRST SECOND
 
 FLAGS:
   {{range .VisibleFlags}}{{.}}
   {{end}}
 DESCRIPTION:
-  {{.Description}}
-
-OUTPUT:
+  Diff only lists missing objects or objects with size differences. It *DOES NOT* compare contents. i.e. objects of same name and size, but differing in content are not noticed.
```
harshavardhana: We can take this verbatim description from the `diffCmd` initializer and keep the help template simpler and clean.

poornas: done

poornas:
ping..needs review

deekoder:
@vadmeste  do you approve?

Issue #2211, opened 2017-07-12
session: Reimplement `mc session` sub-commands to use cli sub-command style


harshavardhana:
## Expected behaviour
mc session when typed should print in the same manner like `mc admin` or `mc config host` prints their sub-command and each sub-command should document itself. 

## Actual behaviour

Currently it can seen that it is harder to document our help text below which provides confusing documentation.

```
$ mc session 
NAME:
  mc session - Manage saved sessions for cp and mirror commands.

USAGE:
  mc session [FLAGS] OPERATION [SESSION-ID]

OPERATION:
  resume   Resume a previously saved session.
  clear    Clear a previously saved session.
  list     List all previously saved sessions.

SESSION-ID:
  SESSION - Session can either be $SESSION-ID or "all".

FLAGS:
  --config-folder value, -C value  Path to configuration folder. (default: "/home/harsha/.mc")
  --quiet, -q                      Suppress chatty console output.
  --no-color                       Disable color theme.
  --json                           Enable JSON formatted output.
  --debug                          Enable debug output.
  --insecure                       Disable SSL certificate verification.
  --help, -h                       Show help.
  
EXAMPLES:
   1. List sessions.
      $ mc session list

   2. Resume session.
      $ mc session resume ygVIpSJs

   3. Clear session.
      $ mc session clear ygVIpSJs

   4. Clear all sessions.
      $ mc session clear all
```

## Steps to reproduce the behaviour
Just type `mc session`

## mc version
master

## System information
All operating systems

awwalker:
@deekoder @harshavardhana if no one is actively working on this I can give it a go...

harshavardhana:
Fixed by #2225

PR #2225, opened 2017-07-21
session: Reimplement using sub-commands

awwalker:
Fix for: #2211

### DIFF
--- a/cmd/session-clear.go
+++ b/cmd/session-clear.go
@@ -1,0 +1,157 @@
+/*
+ * Minio Client (C) 2017 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"encoding/json"
+
+	"github.com/fatih/color"
+	"github.com/minio/cli"
+	"github.com/minio/mc/pkg/console"
+	"github.com/minio/mc/pkg/probe"
+)
+
+var sessionClearFlags = []cli.Flag{
+	cli.BoolFlag{
+		Name:  "force",
+		Usage: "Force a dangerous clear operation.",
+	},
+}
+
+var sessionClear = cli.Command{
+	Name:            "clear",
+	Usage:           "Clear a previously saved session.",
+	Action:          mainSessionClear,
+	Before:          setGlobalsFromContext,
+	Flags:           append(sessionClearFlags, globalFlags...),
+	HideHelpCommand: true,
+	CustomHelpTemplate: `NAME:
+  {{.HelpName}} - {{.Usage}}
+  
+USAGE:
+  {{.HelpName}} SESSION-ID|all
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}
+EXAMPLES:
+  1. Clear session.
+	 $ {{.HelpName}} ygVIpSJs
+  2. Clear all sessions.
+	 $ {{.HelpName}} all
+  3. Forcefully clear an obsolete session.
+	 $ {{.HelpName}} ygVIpSJs --force
+`,
+}
+
+// clearSessionMessage container for clearing session messages.
+type clearSessionMessage struct {
+	Status    string `json:"success"`
+	SessionID string `json:"sessionId"`
+}
+
+// String colorized clear session message.
+func (c clearSessionMessage) String() string {
+	msg := "Session `" + c.SessionID + "`"
+	var colorizedMsg string
+	switch c.Status {
+	case "success":
+		colorizedMsg = console.Colorize("ClearSession", msg+" cleared successfully.")
+	case "forced":
+		colorizedMsg = console.Colorize("ClearSession", msg+" cleared forcefully.")
+	}
+	return colorizedMsg
+}
+
+// JSON jsonified clear session message.
+func (c clearSessionMessage) JSON() string {
+	clearSessionJSONBytes, e := json.Marshal(c)
+	fatalIf(probe.NewError(e), "Unable to marshal into JSON.")
+
+	return string(clearSessionJSONBytes)
+}
+
+// forceClear - Remove a saved session.
+// Used if --force flag is applied.
+func forceClear(sid string, session *sessionV8) {
+	if session != nil {
+		if err := session.Delete().Trace(sid); err == nil {
+			// Force unnecesseray removal successful.
+			printMsg(clearSessionMessage{Status: "success", SessionID: sid})
+			return
+		}
+	}
+	// Remove obsolete session files.
+	removeSessionFile(sid)
+	removeSessionDataFile(sid)
+	printMsg(clearSessionMessage{Status: "forced", SessionID: sid})
+}
+
+// clearSession clear sessions.
+func clearSession(sid string, isForce bool) {
+	var toRemove []string
+	if sid == "all" {
+		toRemove = getSessionIDs()
+	} else {
+		toRemove = append(toRemove, sid)
+	}
+	for _, sid := range toRemove {
+		session, err := loadSessionV8(sid)
+		if !isForce {
+			fatalIf(err.Trace(sid), "Unable to load session `"+sid+"`. Use --force flag to remove obsolete session files.")
+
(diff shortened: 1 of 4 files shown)


cmd/session-clear.go
```diff
@@ -0,0 +1,126 @@
+package cmd
```
harshavardhana: Missing license header..

cmd/session-clear.go
```diff
@@ -0,0 +1,126 @@
...
+	Action:          mainSessionClear,
+	Before:          setGlobalsFromContext,
+	Flags:           globalFlags,
+	HideHelpCommand: true,
+	CustomHelpTemplate: `NAME:
+  {{.HelpName}} - {{.Usage}}
+  
+USAGE:
+  {{.HelpName}} [SESSION-ID/all]
+
```
harshavardhana: This extra space can be removed..

harshavardhana: Also session id is not optional .. so we should avoid [] and to indicate options you can use | instead of /

cmd/session-clear.go
```diff
@@ -0,0 +1,126 @@
...
+	session, err := loadSessionV8(sid)
+	if err != nil {
+		// `mc session clear <broken-session-id>` assumes that user is aware that the session is unuseful
+		// and wants the associated session files to be removed
+		removeSessionFile(sid)
+		removeSessionDataFile(sid)
+		printMsg(clearSessionMessage{Status: "forced", SessionID: sid})
+		return
+	}
+
```
harshavardhana: We are okay to crash here if session is nil . Defensive coding in this manner can mask bugs avoid it.

cmd/session-clear.go
```diff
@@ -0,0 +1,126 @@
...
+	if !isSessionExists(sid) {
+		fatalIf(errDummy().Trace(sid), "Session `"+sid+"` not found.")
+	}
+
+	session, err := loadSessionV8(sid)
+	if err != nil {
+		// `mc session clear <broken-session-id>` assumes that user is aware that the session is unuseful
+		// and wants the associated session files to be removed
+		removeSessionFile(sid)
+		removeSessionDataFile(sid)
```
harshavardhana: This assumption is not correct.. I would suggest we don't do anything forcefully if not stated by the user. We report it to use for manual recovery..provide an option --force instead to provide this behavior.

cmd/session-list.go
```diff
@@ -0,0 +1,73 @@
+package cmd
```
harshavardhana: License header..

cmd/session-main.go
```diff
@@ -36,237 +26,22 @@ var (
...
-	var closestSessions []string
-	for _, value := range sessionsTree.PrefixMatch(session) {
-		closestSessions = append(closestSessions, value.(string))
-	}
-	sort.Strings(closestSessions)
-	return closestSessions
-}
-
+	Name:            "session",
+	Usage:           "Manage saved sessions for cp and mirror commands.",
```
harshavardhana: Mirror doesn't have session any more we should remove it from here.

cmd/session-resume.go
```diff
@@ -0,0 +1,126 @@
+package cmd
```
harshavardhana: License header..

cmd/session-resume.go
```diff
@@ -0,0 +1,126 @@
...
+	// Additional command specific theme customization.
+	console.SetColor("Command", color.New(color.FgWhite, color.Bold))
+	console.SetColor("SessionID", color.New(color.FgYellow, color.Bold))
+	console.SetColor("SessionTime", color.New(color.FgGreen))
+
+	if !isSessionDirExists() {
+		fatalIf(createSessionDir().Trace(), "Unable to create session folder.")
+	}
+
+	args := ctx.Args()
```
harshavardhana: I see args is not used anywhere you could simply do ctx.Args().Get(0)

cmd/session-resume.go
```diff
@@ -0,0 +1,126 @@
...
+
+USAGE:
+  {{.HelpName}} [SESSION-ID/all]
+
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}
+EXAMPLES:
+  1. Resume session. 
+	 $ {{.HelpName}} ygVIpSJs
```
harshavardhana: Should we also provide `mc session resume all` ?

harshavardhana: Can you open a bug if possible to support resume all we don't need it in this PR.

cmd/session-clear.go
awwalker: Done.

cmd/session-list.go
awwalker: Done.

cmd/session-resume.go
awwalker: Done.

cmd/session-resume.go
awwalker: I like the idea of being consistent. I can open a bug...and even send the PR later

cmd/session-clear.go
```diff
@@ -0,0 +1,167 @@
...
+  {{.HelpName}} SESSION-ID|all
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}
+EXAMPLES:
+  1. Clear session.
+	 $ {{.HelpName}} clear ygVIpSJs
+  2. Clear all sessions.
+	 $ {{.HelpName}} clear all
+  3. Forcefully clear a useless session.
```
vadmeste: "useless session" meaning is not clear even for me. I am going to continue to read the code to understand it :)

vadmeste: @awwalker, Do you mean "outdated" session files here ?

cmd/session-clear.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	// Remove useless session files.
+	removeSessionFile(sid)
+	removeSessionDataFile(sid)
+	printMsg(clearSessionMessage{Status: "forced", SessionID: sid})
+}
+
+// clearSession clear sessions.
+func clearSession(sid string, isForce bool) {
+	if sid == "all" {
```
vadmeste: @awwalker, to avoid repetitive code here, you can do the following:

```
var sidsToRemove []string
if sid == "all" {
   sidsToRemove = getSessionIDs()
} else {
   sidsToRemove = append(sidsToRemove, sid)
}

// Now remove all sids in sids To remove

```

awwalker: Good idea I'll refactor that

cmd/session-clear.go
awwalker: @vadmeste Yes outdated might be a better word for them. The purpose of clear is to remove any active sessions but sometimes the command fatally errors out. A forced clear ignores any of these and again attempts to remove the session file and session data file. Does this make sense?

harshavardhana:
@vadmeste anything pending here?

vadmeste:
> @vadmeste anything pending here?

@awwalker, I think it would be better if we rename "useless" to "obsolete", that's more close to the real meaning, what do you think?

cmd/session-resume.go
```diff
@@ -0,0 +1,141 @@
...
+			errorMsg += fmt.Sprintf("\n\nDid you mean?\n")
+			for _, session := range closestSessions {
+				errorMsg += fmt.Sprintf("        `mc resume session %s`", session)
+				// break on the first one, it is good enough.
+				break
+			}
+		}
+		fatalIf(errDummy().Trace(sessionID), errorMsg)
+	}
+	s, err := loadSessionV8(sessionID)
```
vadmeste: @awwalker, can you move the remaining code to a new function ? like resumeSession()

awwalker:
@vadmeste I agree about the comment and reworded it. Do you want me to rebase and squash this to get appveyor running again?

cmd/session-clear.go
```diff
@@ -0,0 +1,157 @@
...
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}
+EXAMPLES:
+  1. Clear session.
+	 $ {{.HelpName}} clear ygVIpSJs
+  2. Clear all sessions.
+	 $ {{.HelpName}} clear all
+  3. Forcefully clear an obsolete session.
+	 $ {{.HelpName}} --force
```
vadmeste: I am not sure if this is correct?

`mc session clear --force`, what this is supposed to do ? shouldn't we give a session id here? also I noticed that
fi I provide a random session id, all sessions will be removed. Can you check that @awwalker  ?

awwalker: @vadmeste yes you are right this needs a session name. --force flag was added so that user is aware that files are obsolete. Before a clear would transparently delete these files without telling user they were no longer useful. 

What do you mean with providing a random session id?

vadmeste: > What do you mean with providing a random session id?

I mean that if you run 'mc session clear --force any_id', all sessions will be removed.

awwalker: Yes I see the issue I'll fix it.

awwalker:
@vadmeste rebased/squashed. ready for one more check over.

PR #2234, opened 2017-08-03
set region from s3 endpoint if present

### LINKED ISSUE #2224: 'mc ls' gives error with s3 endpoint other than s3.amazonaws.com
## Expected behaviour
I expect it to list the buckets, and also to be able to list individual buckets

## Actual behaviour
I can list the buckets if I use s3.amazonaws.com, but not individual buckets.
If I specify the ireland endpoint in my host config, then I can't even list buckets.

Error:
```
mc: <ERROR> Unable to stat `ai`. Amazon S3 endpoint should be 's3.amazonaws.com'.
```

1. `mc config host add ai https://s3-eu-west-1.amazonaws.com ...`
1. `mc ls ai`

## mc version
```
Version: 2017-06-15T03:38:43Z
Release-tag: RELEASE.2017-06-15T03-38-43Z
Commit-id: c32b1ea48b651125e33015ea0d1596cb377685a4
```

## System information
WSL
`Linux DESKTOP-2RF28S0 4.4.0-43-Microsoft #1-Microsoft Wed Dec 31 14:42:53 PST 2014 x86_64 x86_64 x86_64 GNU/Linux`

poornas:
Fixes #2224 

Currently list bucket fails on mc if the endpoint entered in mc config has a region specific endpoint such as s3.eu-west-1.amazonaws.com.  This PR parses the region from endpoint and instantiates the client with region if specified.

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -25,6 +25,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
@@ -33,6 +34,7 @@ import (
 
 	"github.com/minio/mc/pkg/httptracer"
 	"github.com/minio/minio-go"
+	"github.com/minio/minio-go/pkg/credentials"
 	"github.com/minio/minio-go/pkg/policy"
 	"github.com/minio/minio-go/pkg/s3utils"
 	"github.com/minio/minio/pkg/probe"
@@ -44,6 +46,7 @@ type s3Client struct {
 	targetURL    *clientURL
 	api          *minio.Client
 	virtualStyle bool
+	region       string
 }
 
 const (
@@ -91,7 +94,6 @@ func newFactory() func(config *Config) (Client, *probe.Error) {
 				hostName = googleHostName
 			}
 		}
-
 		// Generate a hash out of s3Conf.
 		confHash := fnv.New32a()
 		confHash.Write([]byte(hostName + config.AccessKey + config.SecretKey))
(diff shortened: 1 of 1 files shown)


cmd/client-s3.go
```diff
@@ -923,6 +925,23 @@ func isVirtualHostStyle(host string) bool {
 	return isAmazon(host) && !isAmazonChina(host) || isGoogle(host) || isAmazonAccelerated(host)
 }
 
+// getRegionFromHost - parse region from host name if present.
+func getRegionFromHost(host string) (region string) {
+	region = ""
+	if isGoogle(host) {
```
harshavardhana: You can use s3utils.IsGoogleCloudEndpoint() - perhaps combine below if and Google in a switch case and return?

cmd/mb-main.go
```diff
@@ -29,7 +29,7 @@ var (
 	mbFlags = []cli.Flag{
 		cli.StringFlag{
 			Name:  "region",
-			Value: "us-east-1",
+			Value: "",
```
harshavardhana: Leave this as is.. since we don't expect users to set this value.. if its empty its used as `us-east-1` in MakeBucket().  Keeping it explicit serves as a way to document..

cmd/mb-main.go
```diff
@@ -124,7 +122,11 @@ func mainMakeBucket(ctx *cli.Context) error {
 			cErr = exitStatus(globalErrorExitStatus)
 			continue
 		}
-
+		// Save region.
```
harshavardhana: This comment can be better.. `save region` doesn't mean anything here.

cmd/mb-main.go
poornas: Using empty string deliberately here - minio-go overrides region specified in client constructor only if location string is empty.  If  mb call passes us-east-1 as region, all mc mb calls without region will fail with header mismatch and confusing error message that the location should not be us-east-1.

cmd/client-s3.go
poornas: Done

cmd/mb-main.go
poornas: @harshavardhana, overriding region inside client-s3 so that default is not changed

PR #2235, opened 2017-08-11
Enable ubuntu snaps for mc

harshavardhana:
This PR will enable edge channel only.


vadmeste:
Working with Debian, not possible to build the package but I can test it, is that possible to test it now @harshavardhana ?

harshavardhana:
@vadmeste your reviews please..

PR #2236, opened 2017-08-11
Move pkg/probe from vendor dir to package dir

donatello:
The probe package is no longer available in the Minio server
repo. This change just moves probe from the vendor directory to its
own dir under `pkg`.

This is a maintenance fix.


balamurugana:
With the context, why do we need probe package for mc?

donatello:
> With the context, why do we need probe package for mc?
@balamurugana because it wont compile if we remove it? This is just a maintenance fix, and I didn't go into the question of replacing it with something else.

donatello:
@vadmeste could i have your review?

PR #2243, opened 2017-08-24
Rename the snap to minio-client

elopio:
In linux, mc is midnight commander.
In the traditional linux distro world, name conflicts are forbidden. Because we are trying to make this decentralized and self-serving, we need less strict rules. For this, we added the concept of aliases. Then the user can decide what application owns the mc name.

You could make a claim for the mc name, and keep it. But I think it would be less confusing for existing linux users to rename the snap and use mc as an alias.

I would love your feedback about this feature. Please let me know what do you think. Also, this needs somebody from your org to register the new name in build.snapcraft.io to get the continuous delivery working.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,4 +1,4 @@
-name: mc
+name: minio-client
 version: master
 summary: ls, cp, mkdir, mirror commands for filesystems and object storage.
 description: |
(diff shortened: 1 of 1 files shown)


harshavardhana:
> I would love your feedback about this feature. Please let me know what do you think. Also, this needs somebody from your org to register the new name in build.snapcraft.io to get the continuous delivery working.

Yes this change is fine since alias approach sounds easier.

elopio:
Thanks for the merge :)

Now you need to register the new name. You can do it from https://dashboard.snapcraft.io, or with:

    snapcraft register minio-client

PR #2246, opened 2017-08-31
Vendorize new changes from minio-go

harshavardhana:
This PR brings in changes for

- PutObjectStream with no content-length is heavily
  optimized so `mc pipe` will not run out of memory
  for large uploads.
- Fixes a bug in Signature v2 for HTTP requests.
- Support us-gov-west-1 region as well.
- Auto region handling is inside minio-go now.


harshavardhana:
@vadmeste can you review?

PR #2248, opened 2017-09-05
find: fix parse error for --newer and --older flags

poornas:
mc find fails with parseError when --newer and --older flags are used. This PR fixes the parsing.

### DIFF
--- a/cmd/find.go
+++ b/cmd/find.go
@@ -432,14 +432,16 @@ func TimeHelper(pattern string) (int, error) {
 		"m": 30,
 		"y": 365,
 	}
-	i, err = strconv.Atoi(pattern)
-	if err != nil {
-		t = pattern[len(pattern)-2:]
-		i, err = strconv.Atoi(pattern[:len(pattern)-2])
-
+	if len(pattern) > 1 {
+		i, err = strconv.Atoi(pattern[:len(pattern)-1])
 		if err != nil {
-			return 0, err
+			t = pattern[len(pattern)-2:]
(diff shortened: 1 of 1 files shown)


CarterMcClellan:
\''' 
func TimeHelper(pattern string) (int, error) {
        var i int 
        var t string
        var err error
        conversion := map[string]int{
                "d": 1,
                "w": 7,
                "m": 30, 
                "y": 365,
        }   
        
	// parse
	i, err = strconv.Atoi(pattern)
	
	// if the user input a lone integer we interpret it as a number of days
	if err == nil {
		return i, err
	}
	
	// if we couldnt parse, we try checking to see
	// if the input contains a conversion multiplier
        t = pattern[len(pattern)-1:]

	// we parse all but that last character
	// (which we treat as a conversion multiplier)
        i, err = strconv.Atoi(pattern[:len(pattern)-1])
	
	// if we still cant parse, the user has input a wonky string
	// older/newer flags will tell users that the input string is not valid
	// when call to TimeHelper is made	
        if err != nil {
           return 0, err 

        }
	
	// if an error is not caught and we reach this statement below,
	// then we want to return the time as days multiplied by the correct constant      
        return i * conversion[strings.ToLower(t)], nil 
}
\'''

CarterMcClellan:
^^^ Couldn't we change the time helper to be like this, so that time related flags will accept arguments of length 1.

poornas:
agree @CarterMcClellan, how about I close this PR, and you can send in a fresh one?

poornas:
closing this in favor of #2249

PR #2249, opened 2017-09-06
Fix parsing error with mc find --older, and --newer

CarterMcClellan:
Previously mc find older and newer flags were meant to have time conversion params which were 2 characters long,  that was updated to one character, time helper was not, so I updated the indexing and addressed the corner case, where a user inputs an integer by itself with no "time conversion suffix"

This bug was originally found by @poornas discussion being found in pr #2248.

All code changes made in this pr are the same as commented in pr #2248 .

### DIFF
--- a/cmd/find.go
+++ b/cmd/find.go
@@ -432,17 +432,25 @@ func TimeHelper(pattern string) (int, error) {
 		"m": 30,
 		"y": 365,
 	}
+
+	// parse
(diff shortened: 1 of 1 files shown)


poornas:
@CarterMcClellan , there are conflicts in cmd/find.go

harshavardhana:
ping @donatello waiting on your reviews.

PR #2253, opened 2017-09-17
add functional tests

balamurugana:
Functional tests are automatically run when `make test` is executed.


deekoder:
This PR precedes https://github.com/minio/mint/pull/139

functional-tests.sh
```diff
@@ -0,0 +1,478 @@
...
+    shift
+
+    err=$("$@")
+    rv=$?
+    if [ "$rv" -ne 0 ]; then
+        if [ -n "$MINT_MODE" ]; then
+            err=$(python -c 'import sys,json; print(json.dumps(sys.stdin.read()))' <<<"$err")
+            printf '{"name": "mc", "duration": "%d", "function": "%s", "status": "FAIL", "error": "%s"}\n' "$(get_duration "$start_time")" "$func_name" "$err"
+        else
+            echo "mc: $func_name: $err"
```
nitisht: Why not print same `err` in both cases

balamurugana: Because one is json friendly string, not required in another case

balamurugana:
> We also need to tie functional tests with mc releases, just they way we do it for other minio-sdks. This makes it easy to track and debug which version of mc tests Mint is running.

Version based function test is mint problem not for mc repo

nitisht:
> Version based function test is mint problem not for mc repo

But tests are in mc repo, so versioning needs to be done here. All other minio-sdks also follow the same pattern. `mint` just refers a version of the test and pulls it locally to test, but the versions are maintained by the SDK. I guess we can just create a tag and then pull that tag from `mint`.

PR #2254, opened 2017-09-20
Allow multipart threads to be configured via environment variable

### LINKED ISSUE #1974: cp/mirror: [feature] More threads for multipart?
## Expected behaviour

Faster upload of large file multipart transfers, similar to aws s3 command.

## Actual behaviour

With process explorer, mc is using 4 parallel connections to s3 compared to 10 connections for aws cli s3

mc.exe cp large file

## mc version

Version: 2016-12-09T18:23:19Z
Release-tag: RELEASE.2016-12-09T18-23-19Z
Commit-id: 64166319eeaed70cdb0e1a24ea97416a9600052e

## System information

Windows

I was looking to see if there was a flag for specifying the number of threads to use for upload, but I did not see anything.   Does the multipart upload break down the file in 5MB parts?

In practice, I am seeing ~1Mbps upload with mc.   I get 3-5Mbps with aws cli.   Using process explorer, I see there are four connections to aws when using mc.  aws cli is using 10.

I don't know if that would be the only difference in the performance.

poornas:
Fixes #1974.

Allow number of threads used by mc/cp operations for large uploads to be configurable by allowing a new environment variable MC_MULTIPART_THREADS that can be set by the user before starting a large upload operation.


cmd/client-s3.go
```diff
@@ -54,6 +56,19 @@ const (
 	googleHostName = "storage.googleapis.com"
 )
 
+// getNumMultiPartThreads gets the number of multipart threads to be used for mc cp/mirror functions
+func getNumMultiPartThreads() uint {
+	multipartThreadStr := os.Getenv("DEFAULT_MULTIPART_THREADS")
+	if multipartThreadStr != "" {
+		numThreads, err := strconv.ParseUint(multipartThreadStr, 10, 32)
+		if err != nil {
```
harshavardhana: We should also perhaps restrict users from not setting this value too high i.e not beyond their runtime.NCPU() range.. also the env should have MC_ prefix. 

MC_THREADS or MC_MULTIPART_THREADS

cmd/client-s3_test.go
```diff
@@ -219,8 +219,8 @@ func (s *TestSuite) TestObjectOperations(c *C) {
 
 	var reader io.Reader
 	reader = bytes.NewReader(object.data)
-	n, err := s3c.Put(reader, int64(len(object.data)), map[string][]string{
-		"Content-Type": {"application/octet-stream"},
+	n, err := s3c.Put(reader, int64(len(object.data)), map[string]string{
```
harshavardhana: We should open another bug to introduce context in `mc`

cmd/client-s3.go
```diff
@@ -54,6 +56,19 @@ const (
...
+			log.Fatal("DEFAULT_MULTIPART_THREADS must be a positive integer.")
+		}
+		return uint(numThreads)
+	}
+	return 0
```
harshavardhana: Set this to 4 as default.

poornas: I thought the default should be decided by minio-go as opposed to hardcoding the default in mc as well. When the opts has 0, minio-go sdk automatically overrides it to the default of 4

harshavardhana: mc can have its own default as well, just for clarity. minio-go can have its own behavior we don't need to tie this with `minio-go`

poornas: done.

cmd/client-s3.go
poornas: done.

cmd/client-s3_test.go
poornas: not sure how context would be useful since the user interacts with mc only via command line.

cmd/client-s3.go
poornas: ncpu() is not exposed by runtime package. NumCPU() isn't relevant. Could you please clarify?

poornas:
@harshavardhana , ok will mark this as blocked until then.

cmd/client-s3.go
```diff
@@ -54,6 +56,20 @@ const (
...
+func getNumMultiPartThreads() uint {
+	multipartThreadStr := os.Getenv("MC_MULTIPART_THREADS")
+	if multipartThreadStr != "" {
+		numThreads, err := strconv.ParseUint(multipartThreadStr, 10, 32)
+		if err != nil {
+			log.Fatal("MC_MULTIPART_THREADS must be a positive integer.")
+		}
+		return uint(numThreads)
+	}
+	return 4
```
vadmeste: Can you add a new global constant definition ? probably `const defaultMultipartThreadsNum = 4`

cmd/common-methods.go
```diff
@@ -84,19 +84,23 @@ func getSourceStream(alias string, urlStr string, fetchStat bool) (reader io.Rea
...
-	metadata = map[string][]string{}
+	metadata = map[string]string{}
 	if fetchStat {
 		st, err := sourceClnt.Stat(false)
 		if err != nil {
 			return nil, nil, err.Trace(alias, urlStr)
 		}
-		metadata = st.Metadata
+		for k, v := range st.Metadata {
+			if len(v) > 0 {
```
vadmeste: HTTP header value can be empty, I see it sometimes in some HTTP responses but I am not sure how it can be useful. 

However, I think we don't have to bother ourselves to check if header value is empty or not at this stage. Any reason for this check ?

poornas: guarding against errors since the value is a list, and the length could be 0?

cmd/client-s3.go
poornas: done.

cmd/common-methods.go
vadmeste: > guarding against errors since the value is a list, and the length could be 0?

yeah you are right, then I think we should do treat the case of len(v) == 0 and we set an empty string to metadata[k]. What do you think?

harshavardhana:
Can you re-vendorize with the latest master on minio-go @poornas ?

poornas:
done.

poornas:
@harshavardhana and @vadmeste , can we review this now?

harshavardhana:
> @harshavardhana and @vadmeste , can we review this now?

Can you re-vendorize? @poornas since we made a few changes..

poornas:
@harshavardhana, done

PR #2258, opened 2017-09-23
Support copy in parallel

chibiegg:
Add `--parallel` flag to `cp` subcommand to copy in parallel.

# For example

```
mc cp --recursive --parallel 10 play/mybucket/burningman2011/ s3/mybucket/
```


harshavardhana:
ping @donatello pending your review.

chibiegg:
Thanks a lot!

PR #2264, opened 2017-09-29
refactor find command to be simpler

harshavardhana:


cmd/find.go
```diff
@@ -73,21 +68,60 @@ func pathMatch(path, pattern string) bool {
...
+
+	now := UTCNow()
+
+	// Find all time in which the time in which the object was just created is after the current time
+	t := time.Date(now.Year(), now.Month(), now.Day()-i, now.Hour(), now.Minute(), 0, 0, time.UTC)
+	return createTime.Before(t)
+}
+
+// newerMatch checks to see if the given object was created before the given threshold
+func newerMatch(createTime time.Time, pattern string) bool {
```
ebozduman: `pattern` : better naming can be used here... maybe `formattedTime`

cmd/find.go
```diff
@@ -73,21 +68,60 @@ func pathMatch(path, pattern string) bool {
...
+	i, err := parseTime(pattern)
+	fatalIf(err.Trace(pattern), "Error parsing string passed to flag newer")
+
+	now := UTCNow()
+
+	t := time.Date(now.Year(), now.Month(), now.Day()-i, now.Hour(), now.Minute(), 0, 0, time.UTC)
+	return createTime.After(t) || createTime.Equal(t)
+}
+
+// largerSizeMatch checks to see if the given object is larger than the given threshold
```
ebozduman: `threshold` =? `pattern`
Not clear!

cmd/find.go
```diff
@@ -73,21 +68,60 @@ func pathMatch(path, pattern string) bool {
...
+	i, e := humanize.ParseBytes(pattern)
+	fatalIf(probe.NewError(e), "Error parsing string passed to flag larger")
+
+	return int64(i) < size
+}
+
+// smallerSizeMatch checks to see if the given object is smaller than the given threshold
+func smallerSizeMatch(size int64, pattern string) bool {
+	i, e := humanize.ParseBytes(pattern)
+	fatalIf(probe.NewError(e), "Error parsing string passed to flag smaller")
```
ebozduman: We can have a better error message here

cmd/find.go
```diff
@@ -73,21 +68,60 @@ func pathMatch(path, pattern string) bool {
...
-	printString := SubArgsHelper(ctx.String("print"), path, fileContent)
-	printMsg(findMSG{
-		Path: printString,
-	})
+func doFindPrint(ctx *cli.Context, path string, fileContent contentMessage) {
+	fileContent.Key = stringFindReplace(ctx.String("print"), path, fileContent)
+	printMsg(findMSG{fileContent})
 }
 
 // doFindExec passes the users input along to the command line, also dealing with substitution arguments
```
ebozduman: Comment is not helpful.

cmd/find.go
```diff
@@ -375,7 +379,9 @@ func watchEvents(ctx *cli.Context, clnt Client, params watchParams, alias string
 
 // fileContentMatch is used to take the params passed to find, in addition to the current
```
ebozduman: Comment needs to be reworded.

cmd/find.go
```diff
@@ -375,7 +379,9 @@ func watchEvents(ctx *cli.Context, clnt Client, params watchParams, alias string
...
 // file and call the appropriate "pattern matching methods"
-func fileContentMatch(fileContent contentMessage, orOp bool, ctx *cli.Context) bool {
+func fileContentMatch(ctx *cli.Context, fileContent contentMessage) bool {
+	orOp := ctx.Bool("or")
+
 	match := true
 
 	if (match && !orOp) && ctx.String("ignore") != "" {
```
ebozduman: Since both of the operations are `&&`, there is no need to have parenthesis around `(match && !orOp)`.

cmd/find.go
```diff
@@ -384,70 +390,60 @@ func fileContentMatch(fileContent contentMessage, orOp bool, ctx *cli.Context) b
...
-		match = doFindSmallerSize(fileContent.Size, ctx.String("smaller"))
+		match = smallerSizeMatch(fileContent.Size, ctx.String("smaller"))
 	}
 
 	return match
 }
 
-// TimeHelper is used in conjunction with the Newer and Older flags to convert the input
+// parseTime is used in conjunction with the Newer and Older flags to convert the input
 // into an interperable integer (in days)
```
ebozduman: Typo in the comment: `interperable`
The message is not clear.

cmd/find.go
```diff
@@ -40,47 +39,100 @@ import (
...
-	return regexp.MatchString(pattern, path)
+// regexMatch reports whether path matches the regex pattern.
+func regexMatch(pattern, path string) bool {
+	matched, e := regexp.MatchString(pattern, path)
+	errorIf(probe.NewError(e).Trace(pattern), "Unable to regex match with input pattern")
+	return matched
+}
+
+// olderMatch matches whether if the createTime is older than the allowed threshold.
+func olderMatch(threshold string, createTime time.Time) bool {
```
kannappanr: Will olderThan be a better name?

harshavardhana: Yes agreed.

cmd/find.go
```diff
@@ -40,47 +39,100 @@ import (
...
+	t, err := parseTime(threshold)
+	fatalIf(err.Trace(threshold), "Unable to parse input threshold value into time.Time")
+	return createTime.Before(t)
+}
+
+// newerMatch matches whether if the createTime is newer than the allowed threshold.
+func newerMatch(threshold string, createTime time.Time) bool {
```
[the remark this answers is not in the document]
harshavardhana: Yes

cmd/find.go
```diff
@@ -191,31 +200,31 @@ func doFind(clnt Client, ctx *cli.Context) {
...
+			s := ""
+
+			// We are going to be parsing the path by x amounts
+			pathParts := strings.SplitAfter(filePath, separator)
 
-			// handle invalid params
-			// ex. user specifies:
-			// maxdepth 2, but the given object only has a maxdepth of 1
+			// Handle invalid params eg: user specifies
+			// Maxdepth 2, but the given object only
```
kannappanr: Is the comment correct, here? given object having maxdepth less than 2 in this case is valid right?

cmd/find.go
```diff
@@ -191,31 +200,31 @@ func doFind(clnt Client, ctx *cli.Context) {
...
+			// has a maxdepth of 1
 			if (len(pathParts)-1) < i || i < 0 {
 
-				// -1 is meant to handle each the array being 0 indexed, but the size not being 0 indexed
+				// -1 is meant to handle each the array being 0 indexed,
```
kannappanr: Comment needs to more clear.

cmd/find.go
harshavardhana: It can be more clear,.. let me add.

cmd/find.go
harshavardhana: Sure

cmd/find-main.go
```diff
@@ -110,6 +109,19 @@ UNITS
    rate of conversion being 7 days in 1 week, 30 days in 1 month, and 365
    days in one year.
 
+FORMAT
```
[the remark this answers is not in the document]
harshavardhana: done

cmd/find.go
```diff
@@ -385,126 +381,95 @@ func watchEvents(ctx *cli.Context, clnt Client, params watchParams, alias string
...
+	objectURL := content.URL.String()
+	newClnt, err := newClientFromAlias(targetAlias, objectURL)
+	fatalIf(err.Trace(targetAlias, objectURL), "Unable to initialize new client from alias.")
 
-		// set default expiry for each url (point of no longer valid), to be 7 days
-		expiry := time.Duration(604800) * time.Second
-		shareURL, err := newClnt.ShareDownload(expiry)
-		fatalIf(err, "Error with ShareDownloa")
+	// Set default expiry for each url (point of no longer valid), to be 7 days
+	shareURL, err := newClnt.ShareDownload(time.Duration(604800) * time.Second)
```
kannappanr: This can be
shareURL, err := newClnt.ShareDownload(humanize.Week.Seconds())

harshavardhana: Actually we can make this a constant.

harshavardhana:
> shareURL, err := newClnt.ShareDownload(humanize.Week.Seconds())

There is no such function as humanize.Week.Seconds(), i will stick to the existing approach but move it like a constant.

cmd/find.go
```diff
@@ -190,142 +199,129 @@ func doFind(clnt Client, ctx *cli.Context) {
...
+		if !ctx.Bool("watch") && match && ctx.String("exec") != "" {
+			doFindExec(ctx, fileContent)
+		} else if match && ctx.String("print") != "" {
+			doFindPrint(ctx, fileContent)
+		} else if match {
+			printMsg(findMessage{fileContent})
+		}
 		if ctx.Bool("watch") {
-			doFindWatch(ctx.Args().Get(0), ctx)
+			doFindWatch(ctx, ctx.Args().Get(0))
```
kannappanr: can I pass watch and exec as part of the same command as given in the example 7?
```
 mc find s3/buck --name "*foo" --watch --exec "mc cp {} play/test"

```
If I can, does doFindWatch do both watch and exec internally?

cmd/find.go
```diff
@@ -190,142 +199,129 @@ func doFind(clnt Client, ctx *cli.Context) {
...
-			}
+		// Maxdepth can modify the filepath to end as a directory prefix
+		// to be consistent with the find behavior, we wont list directories
+		// so any paths which end with a separator are ignored.
+		if strings.HasSuffix(fileContent.Key, separator) {
+			continue
 		}
 
+		match := doFindMatch(ctx, fileContent)
+		if !ctx.Bool("watch") && match && ctx.String("exec") != "" {
```
kannappanr: Why is this re-arranged ? Is it like if exec gets called, then print will not get called.

harshavardhana: Yes exec and print cannot work together.

cmd/find.go
harshavardhana: Yes that is intended.... watch does exec internally.

cmd/find.go
```diff
@@ -190,142 +199,129 @@ func doFind(clnt Client, ctx *cli.Context) {
...
 
-			// we are going to be parsing the path by x amounts
-			pathParts := strings.SplitAfter(filePath, "/")
+			newPath := ""
 
-			// handle invalid params
-			// ex. user specifies:
-			// maxdepth 2, but the given object only has a maxdepth of 1
-			if (len(pathParts)-1) < i || i < 0 {
+			// We are going to be parsing the path by x amounts
```
ebozduman: What is `x amounts`?

cmd/find.go
```diff
@@ -190,142 +199,129 @@ func doFind(clnt Client, ctx *cli.Context) {
...
+			pathParts := strings.SplitAfter(fileContent.Key, separator)
 
-				// -1 is meant to handle each the array being 0 indexed, but the size not being 0 indexed
-				i = len(pathParts) - 1
+			// Check if max-depth is 2, but the given object only
```
ebozduman: `but the given object only has a maximum depth of 1 use that instead.`
=>
`but if the given object only has a maximum depth of 1, then use that instead.`

cmd/find.go
```diff
@@ -190,142 +199,129 @@ func doFind(clnt Client, ctx *cli.Context) {
...
+			// has a maximum depth of 1 use that instead.
+			if (len(pathParts)-1) < maxDepth || maxDepth < 0 {
+				// -1 is meant to handle each the array being 0 indexed,
```
ebozduman: Do we really need this comment  here?
Needs to be removed or reworded.

harshavardhana: We don't need it.

cmd/find.go
harshavardhana: Done

cmd/find.go
```diff
@@ -190,142 +199,127 @@ func doFind(clnt Client, ctx *cli.Context) {
...
-		s := humanize.IBytes(uint64(fileContent.Size))
-		str = strings.Replace(str, "{size}", s, -1)
+		str = strings.Replace(str, "{size}", humanize.IBytes(uint64(fileContent.Size)), -1)
 	}
 
-	if strings.Contains(str, "{url}") {
-		s := GetPurl(path)
-		str = strings.Replace(str, "{url}", s, -1)
+	// replace all instances of {"size"}
+	if strings.Contains(str, `{"size"}`) {
```
ebozduman: This check has been already done above, line#285. It looks like we don't need the above if block.

cmd/find.go
```diff
@@ -190,142 +199,127 @@ func doFind(clnt Client, ctx *cli.Context) {
...
-	if strings.Contains(str, "{\""+"base"+"\"}") {
-		str = strings.Replace(str, "{\""+"base"+"\"}", strconv.Quote(filepath.Base(path)), -1)
+		str = strings.Replace(str, "{time}", fileContent.Time.Format(printDate), -1)
 	}
 
-	// replace all instances of {"dir"}
-	if strings.Contains(str, "{\""+"dir"+"\"}") {
-		str = strings.Replace(str, "{\""+"dir"+"\"}", strconv.Quote(filepath.Dir(path)), -1)
+	// replace all instances of {"time"}
+	if strings.Contains(str, `{"time"}`) {
```
ebozduman: {"time"} check is also duplicated. The if block above at line#296 needs to be removed.

cmd/find.go
```diff
@@ -40,47 +39,100 @@ import (
...
+	fatalIf(probe.NewError(e).Trace(bytesFmt), "Unable to parse input threshold value into size bytes")
+
+	return int64(i) < size
+}
+
+// smallerSizeMatch matches whether if the input size bytes is smaller than the allowed threshold.
+func smallerSizeMatch(bytesFmt string, size int64) bool {
+	i, e := humanize.ParseBytes(bytesFmt)
+	fatalIf(probe.NewError(e).Trace(bytesFmt), "Unable to parse input threshold value into size bytes")
+
```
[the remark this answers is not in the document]
harshavardhana: It's on purpose for readability.

cmd/find.go
```diff
@@ -40,47 +39,100 @@ import (
...
+	t, err := parseTime(threshold)
+	fatalIf(err.Trace(threshold), "Unable to parse input threshold value into time.Time")
+	return createTime.After(t) || createTime.Equal(t)
+}
+
+// largerSizeMatch matches whether if the input size bytes is larger than the allowed threshold.
+func largerSizeMatch(bytesFmt string, size int64) bool {
+	i, e := humanize.ParseBytes(bytesFmt)
+	fatalIf(probe.NewError(e).Trace(bytesFmt), "Unable to parse input threshold value into size bytes")
+
```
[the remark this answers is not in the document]
harshavardhana: It's on purpose for readability.

cmd/find.go
harshavardhana: Needs to be fixed.

cmd/find.go
```diff
@@ -40,47 +39,100 @@ import (
...
-	printMsg(findMSG{fileContent})
+func doFindPrint(ctx *cli.Context, fileContent contentMessage) {
+	fileContent.Key = stringsReplace(ctx.String("print"), fileContent)
+	printMsg(findMessage{fileContent})
 }
 
-// doFindExec passes the users input along to the command line, also dealing with substitution arguments
-func doFindExec(ctx *cli.Context, path string, fileContent contentMessage) {
-	commandString := SubArgsHelper(ctx.String("exec"), path, fileContent)
+// doFindExec executes the input command line, additionally also formats input
```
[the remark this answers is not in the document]
harshavardhana: Surw

deekoder:
Approving this provided the rewrite will address all the scenarios that are still broken.

PR #2265, opened 2017-09-29
mc rm  to support  entire namespace removal

### LINKED ISSUE #2250: mc rm --recursive --force option should work on sitename
## Expected behaviour
`mc rm --recursive --force --older-than=90 play 
` should succeed and recursively delete buckets older than 90 days
## Actual behaviour
```
➜  mc git:(master) mc rm --recursive --force  --older-than=90  play
mc: <ERROR> Failed to remove `play` recursively. Bucket `` does not exist.
➜  mc git:(master)
```
## mc version
Version: 2017-09-06T16:33:47Z
Release-tag: DEVELOPMENT.2017-09-06T16-33-47Z
Commit-id: b3955cc7e8dad50f5f13d072ec71a9e7dd52d38f
## System information
ubuntu 17.04

poornas:
Fixes #2250 
Currently, mc rm --recursive --force works on buckets but not on the entire site. This PR adds support for rm --recursive to delete entire site.

To distinguish full site removal from removing a bucket and provide appropriate warning to the user, the user needs to supply --dangerous flag in addition to the --recursive and --force flags for full site rm.

The rm --recursive --force option will continue to work as before at the bucket level or on the file system.

### DIFF
--- a/cmd/rm-main.go
+++ b/cmd/rm-main.go
@@ -21,6 +21,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/fatih/color"
@@ -41,7 +43,11 @@ var (
 		},
 		cli.BoolFlag{
 			Name:  "force",
-			Usage: "Force a dangerous remove operation.",
+			Usage: "Allow a recursive remove operation.",
+		},
+		cli.BoolFlag{
+			Name:  "dangerous",
+			Usage: "Allow site-wide removal of buckets and objects.",
 		},
 		cli.BoolFlag{
 			Name:  "incomplete, I",
@@ -82,16 +88,22 @@ EXAMPLES:
    1. Remove a file.
       $ {{.HelpName}} 1999/old-backup.tgz
 
-   2. Remove all objects recursively.
+   2. Remove all objects recursively from bucket 'jazz-songs' matching 'louis' prefix.
       $ {{.HelpName}} --recursive s3/jazz-songs/louis/
 
-   3. Remove all objects older than '90' days.
+   3. Remove all objects older than '90' days recursively from bucket 'jazz-songs' that match 'louis' prefix.
       $ {{.HelpName}} --recursive --older-than=90 s3/jazz-songs/louis/
 
    4. Remove all objects read from STDIN.
       $ {{.HelpName}} --force --stdin
 
-   5. Drop all incomplete uploads on 'jazz-songs' bucket.
+   5. Remove all buckets and objects recursively from S3 host
+      $ {{.HelpName}} --recursive --dangerous s3
+
+   6. Remove all buckets and objects older than '90' days recursively from host
+      $ {{.HelpName}} --recursive --dangerous --older-than=90 s3
+
+   7. Drop all incomplete uploads on 'jazz-songs' bucket.
       $ {{.HelpName}} --incomplete --recursive s3/jazz-songs/
 
 `,
@@ -121,7 +133,19 @@ func checkRmSyntax(ctx *cli.Context) {
 	isForce := ctx.Bool("force")
 	isRecursive := ctx.Bool("recursive")
 	isStdin := ctx.Bool("stdin")
-
+	isDangerous := ctx.Bool("dangerous")
+	isNamespaceRemoval := false
+	for _, url := range ctx.Args() {
+		// clean path for aliases like s3/.
+		//Note: UNC path using / works properly in go 1.9.2 even though it breaks the UNC specification.
+		url = filepath.ToSlash(filepath.Clean(url))
+		// namespace removal applies only for non FS. So filter out if passed url represents a directory
+		if !isAliasURLDir(url) {
+			_, path := url2Alias(url)
+			isNamespaceRemoval = (path == "")
+			break
+		}
+	}
 	if !ctx.Args().Present() && !isStdin {
 		exitCode := 1
 		cli.ShowCommandHelpAndExit(ctx, "rm", exitCode)
@@ -129,8 +153,16 @@ func checkRmSyntax(ctx *cli.Context) {
 
 	// For all recursive operations make sure to check for 'force' flag.
 	if (isRecursive || isStdin) && !isForce {
+		if isNamespaceRemoval {
+			fatalIf(errDummy().Trace(),
+				"This operation results in site-wide removal of buckets and objects. If you are really sure, retry this command with ‘--dangerous’ and ‘--force’ flags.")
+		}
 		fatalIf(errDummy().Trace(),
 			"Removal requires --force option. This operation is *IRREVERSIBLE*. Please review carefully before performing this *DANGEROUS* operation.")
+	}
+	if (isRecursive || isStdin) && isNamespaceRemoval && !isDangerous {
+		fatalIf(errDummy().Trace(),
+			"This operation results in site-wide removal of buckets and objects. If you are really sure, retry this command with ‘--dangerous’ and ‘--force’ flags.")
 	}
 }
 
@@ -190,7 +222,6 @@ func removeRecursive(url string, isIncomplete bool, isFake bool, older int) erro
 		errorIf(pErr.Trace(url), "Failed to remove `"+url+"` recursively.")
 		return exitStatus(globalErrorExitStatus) // End of journey.
 	}
-
 	contentCh := make(chan *clientContent)
 	errorCh := clnt.Remove(isIncomplete, contentCh)
 
@@ -208,6 +239,7 @@ func removeRecursive(url string, isIncomplete bool, isFake bool, older int) erro
 			close(contentCh)
 			return exitStatus(globalErrorExitStatus)
 		}
+		urlString := content.URL.Path
 
 		if older > 0 {
 			// Check whether object is created older than given time.
@@ -218,13 +250,14 @@ func removeRecursive(url string, isIncomplete bool, isFake bool, older int) erro
 				continue
 			}
 		}
-
-		urlString := content.URL.Path
-		printMsg(rmMessage{
-			Key:  targetAlias + urlString,
-			Size: content.Size,
-		})
-
+		// list internally mimics recursive directory listing of object prefixes for s3 similar to FS.
+		// The rmMessage needs to be printed only for actual objects being deleted and not prefixes.
+		if !(content.Time.IsZero() && strings.HasSuffix(urlString, string(filepath.Separator))) {
+			printMsg(rmMessage{
+				Key:  targetAlias + urlString,
+				Size: content.Size,
+			})
+		}
 		if !isFake {
 			sent := false
 			for !sent {
@@ -238,7 +271,6 @@ func removeRecursive(url string, isIncomplete bool, isFake bool, older int) erro
 						// Ignore Permission error.
 						continue
 					}
-
 					close(contentCh)
 					return exitStatus(globalErrorExitStatus)
 				}
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -659,91 +659,101 @@ func (c *s3Client) removeIncompleteObjects(bucket string, objectsCh <-chan strin
 	return removeObjectErrorCh
 }
 
-// Remove - remove object or bucket.
+// Remove - remove object or bucket(s).
 func (c *s3Client) Remove(isIncomplete bool, contentCh <-chan *clientContent) <-chan *probe.Error {
-	bucket, _ := c.url2BucketAndObject()
-
 	errorCh := make(chan *probe.Error)
-	var bucketContent *clientContent
 
-	// Goroutine
-	// 1. calls removeIncompleteObjects() for incomplete uploads
-	//    or minio-go.RemoveObjects().
-	// 2. executes another Goroutine to copy contentCh to objectsCh.
-	// 3. reads statusCh and copies to errorCh.
+	prevBucket := ""
+	// maintain cContentCh, objectsCh, statusCh for each bucket
+	var cContentCh chan *clientContent
+	var objectsCh chan string
+	var statusCh <-chan minio.RemoveObjectError
+
+	isRemoveBucket := false
+
 	go func() {
 		defer close(errorCh)
-
-		objectsCh := make(chan string)
-		var statusCh <-chan minio.RemoveObjectError
-		if isIncomplete {
-			statusCh = c.removeIncompleteObjects(bucket, objectsCh)
-		} else {
-			statusCh = c.api.RemoveObjects(bucket, objectsCh)
-		}
-
-		// doneCh to control below Goroutine.
-		doneCh := make(chan struct{})
-		defer close(doneCh)
-
-		// Goroutine reads contentCh and copies to objectsCh.
-		go func() {
-			defer close(objectsCh)
-
-			for {
-				// Read from contentCh or doneCh.  If doneCh is read, exit the function.
-				var content *clientContent
-				var ok bool
-				select {
-				case content, ok = <-contentCh:
-					if !ok {
-						// Closed channel.
-						return
-					}
-				case <-doneCh:
-					return
-				}
-
-				// Convert content.URL.Path to objectName for objectsCh.
-				_, objectName := c.splitPath(content.URL.Path)
-
-				// Currently only supported hosts with virtual style
-				// are Amazon S3 and Google Cloud Storage.
-				// which also support objects with "/" as delimiter.
-				// Skip trimming "/" and let the server reply error
-				// if any.
-				if !c.virtualStyle {
-					objectName = strings.TrimSuffix(objectName, string(c.targetURL.Separator))
-				}
-
-				// As object name is empty, we need to remove the bucket as well.
-				if objectName == "" {
-					bucketContent = content
-					continue
-				}
-
-				// Write to objectsCh or read doneCh. If doneCh is read, exit the function.
-				select {
-				case objectsCh <- objectName:
-				case <-doneCh:
-					return
+		for content := range contentCh {
+			// Convert content.URL.Path to objectName for objectsCh.
+			bucket, objectName := c.splitPath(content.URL.Path)
+			// Currently only supported hosts with virtual style
+			// are Amazon S3 and Google Cloud Storage.
+			// which also support objects with "/" as delimiter.
+			// Skip trimming "/" and let the server reply error
+			// if any.
+			if !c.virtualStyle {
+				objectName = strings.TrimSuffix(objectName, string(c.targetURL.Separator))
+			}
+			// Init cContentCh channel and objectsCh the first time.
+			if prevBucket == "" {
+				cContentCh = make(chan *clientContent)
+				objectsCh = make(chan string)
+				prevBucket = bucket
+				if isIncomplete {
+					statusCh = c.removeIncompleteObjects(bucket, objectsCh)
+				} else {
+					statusCh = c.api.RemoveObjects(bucket, objectsCh)
 				}
 			}
-		}()
 
-		// Read statusCh and write to errorCh.
-		for removeStatus := range statusCh {
-			errorCh <- probe.NewError(removeStatus.Err)
+			if prevBucket != bucket {
+				// Close cContentCh when bucket changes. Remove bucket if
+				// it qualifies.
+				close(cContentCh)
+				if objectsCh != nil {
+					close(objectsCh)
+				}
+				for removeStatus := range statusCh {
+					errorCh <- probe.NewError(removeStatus.Err)
+				}
+				if isRemoveBucket && !isIncomplete {
+					if err := c.api.RemoveBucket(prevBucket); err != nil {
+						errorCh <- probe.NewError(err)
+					}
+				}
+				// re-init cContentCh and objectsCh for next bucket
+				isRemoveBucket = false
+				cContentCh = make(chan *clientContent)
+				objectsCh = make(chan string)
+				if isIncomplete {
+					statusCh = c.removeIncompleteObjects(bucket, objectsCh)
+				} else {
+					statusCh = c.api.RemoveObjects(bucket, objectsCh)
+				}
+				prevBucket = bucket
+			}
+
+			if objectName != "" {
+				objectsCh <- objectName
+			} else {
+				// end of bucket - close the objectsCh
+				isRemoveBucket = true
+				if objectsCh != nil {
+					close(objectsCh)
+				}
+				objectsCh = nil
+			}
 		}
-
-		// Remove bucket for regular objects.
-		if bucketContent != nil && !isIncomplete {
-			if err := c.api.RemoveBucket(bucket); err != nil {
+		// close cContentCh and objectsCh at end of contentCh
+		if cContentCh != nil {
+			close(cContentCh)
+		}
+		if objectsCh != nil {
+			close(objectsCh)
+		}
+		// write remove objects status to errorCh
+		if statusCh != nil {
+			for removeStatus := range statusCh {
+				errorCh <- probe.NewError(removeStatus.Err)
+			}
+		}
+		// Remove last bucket if it qualifies.
+		if isRemoveBucket && !isIncomplete {
+			if err := c.api.RemoveBucket(prevBucket); err != nil {
 				errorCh <- probe.NewError(err)
 			}
 		}
 	}()
-
 	return errorCh
 }
 
@@ -1208,8 +1218,7 @@ func (c *s3Client) listIncompleteRecursiveInRoutine(contentCh chan *clientConten
 }
 
 // Convert objectMultipartInfo to clientContent
-func (c *s3Client) objectMultipartInfo2ClientContent(entry minio.ObjectMultipartInfo) clientContent {
-	bucket, _ := c.url2BucketAndObject()
+func (c *s3Client) objectMultipartInfo2ClientContent(bucket string, entry minio.ObjectMultipartInfo) clientContent {
 
 	content := clientContent{}
 	url := *c.targetURL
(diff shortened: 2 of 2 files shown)


cmd/rm-main.go
```diff
@@ -82,16 +87,22 @@ EXAMPLES:
...
 
    4. Remove all objects read from STDIN.
       $ {{.HelpName}} --force --stdin
 
-   5. Drop all incomplete uploads on 'jazz-songs' bucket.
+	 5. Remove all buckets and objects recursively from entire namespace
+      $ {{.HelpName}} --recursive --namespace s3
+
+   6. Remove all buckets and objects older than '90' days from entire namespace
+			$ {{.HelpName}} --recursive --namespace --older-than=90 s3
```
harshavardhana: Tabbing is off.

cmd/rm-main.go
```diff
@@ -121,14 +132,24 @@ func checkRmSyntax(ctx *cli.Context) {
 	isForce := ctx.Bool("force")
 	isRecursive := ctx.Bool("recursive")
 	isStdin := ctx.Bool("stdin")
-
+	isNamespace := ctx.Bool("namespace")
+	isNamespaceRemoval := false
+	for _, url := range ctx.Args() {
+		if isHostName(url) {
```
balamurugana: This method does not work for removing a directory with non-aliases.  For example `mc rm mylocaldir` would fail.  Basically we need to differentiate alias and local dir

cmd/rm-main.go
```diff
@@ -121,14 +132,24 @@ func checkRmSyntax(ctx *cli.Context) {
...
+		}
+	}
 	if !ctx.Args().Present() && !isStdin {
 		exitCode := 1
 		cli.ShowCommandHelpAndExit(ctx, "rm", exitCode)
 	}
-
+	if (isRecursive || isStdin) && isNamespaceRemoval && !isNamespace {
+		fatalIf(errDummy().Trace(),
+			"Full namespace removal requires --namespace option. This operation removes all the buckets on your namespace. Please review carefully before performing this *DANGEROUS* operation.")
```
balamurugana: This check is not interesting.  removing local file system, we don't need any restriction like that.

cmd/rm-main.go
```diff
@@ -220,27 +241,35 @@ func removeRecursive(url string, isIncomplete bool, isFake bool, older int) erro
...
 
 		urlString := content.URL.Path
-		printMsg(rmMessage{
-			Key:  targetAlias + urlString,
-			Size: content.Size,
-		})
+		// print the rmMessage only if content is a bucket or object. clnt.List lists
+		// prefixes as well which unifies the S3 listing with folder listing - however
+		// these prefixes are not actual objects on s3 and the content.Time defaults to
+		// zero. Filter these out to prevent erroneous rmMessage being printed to console.
```
balamurugana: You can skip entries those are ending with `/` and having zero time for completeness.

cmd/rm-main.go
```diff
@@ -220,27 +241,35 @@ func removeRecursive(url string, isIncomplete bool, isFake bool, older int) erro
...
 			for !sent {
 				select {
 				case contentCh <- content:
 					sent = true
-				case pErr := <-errorCh:
+				case pErr, ok := <-errorCh:
+					if !ok {
+						close(contentCh)
+						return exitStatus(globalErrorExitStatus)
+					}
```
balamurugana: We would need to exit on first error here.  This check leads to continuation.  Below is the right logic
```go
case pErr, ok := <-errorCh:
	if ok {
		errorIf(pErr.Trace(urlString), "Failed to remove `"+urlString+"`.")
		switch pErr.ToGoError().(type) {
		case PathInsufficientPermission:
			// Ignore Permission error.
			continue
		}
	}
	close(contentCh)
	return exitStatus(globalErrorExitStatus)
```

IMO, having `ok` is not required because `pErr` will be `nil` on closed channel and we exit anyway.

cmd/rm-main.go
```diff
@@ -264,6 +293,43 @@ func removeRecursive(url string, isIncomplete bool, isFake bool, older int) erro
...
+func isHostName(url string) bool {
+	parts := strings.Split(url, "/")
+	return len(parts) == 1
+}
+
+// removes all the buckets recursively.
+func removeSiteRecursive(url string, isIncomplete bool, isFake bool, older int) (err error) {
+	clnt, pErr := newClient(url)
+	if pErr != nil {
+		errorIf(pErr.Trace(url), "Failed to remove `"+url+"` recursively.")
```
balamurugana: This is not appropriate error message here.

cmd/rm-main.go
```diff
@@ -285,7 +351,13 @@ func mainRm(ctx *cli.Context) error {
 	// Support multiple targets.
 	for _, url := range ctx.Args() {
 		if isRecursive {
-			err = removeRecursive(url, isIncomplete, isFake, older)
+			if isHostName(url) {
+				if isNamespace {
+					err = removeSiteRecursive(url, isIncomplete, isFake, older)
```
balamurugana: Unable to understand the need of this function.

Current code fails for below commands
```bash
$ mc rm --recursive --fake --force play
mc: <ERROR> Failed to remove `play` recursively. Bucket `` does not exist.
```

What I am able to visualize with `--namespace` flag is by running `mc rm --recursive --fake --force play`, it could get list of buckets from `play` alias and keep calling `removeRecursive()` with each bucket URL.

cmd/rm-main.go
```diff
@@ -264,6 +293,43 @@ func removeRecursive(url string, isIncomplete bool, isFake bool, older int) erro
 	return nil
 }
 
+func isHostName(url string) bool {
```
harshavardhana: Add unit tests?

cmd/rm-main.go
poornas: done.

cmd/rm-main.go
poornas: changed this to incorporate feedback

cmd/rm-main.go
poornas: removed this function.

cmd/rm-main.go
poornas: why?

balamurugana: This is fine because of `pErr.Trace(url)`.  Sorry for the confusion.

cmd/rm-main.go
```diff
@@ -245,7 +246,7 @@ func removeRecursive(url string, isIncomplete bool, isFake bool, older int) erro
 		// prefixes as well which unifies the S3 listing with folder listing - however
 		// these prefixes are not actual objects on s3 and the content.Time defaults to
 		// zero. Filter these out to prevent erroneous rmMessage being printed to console.
-		if !content.Time.IsZero() {
+		if !(content.Time.IsZero() && strings.HasSuffix(urlString, string(filepath.Separator))) {
```
balamurugana: S3 service sends `/` as separator whereas on windows `\`. We would need to check whether this works on windows or not.

cmd/rm-main.go
```diff
@@ -351,7 +344,7 @@ func mainRm(ctx *cli.Context) error {
 	// Support multiple targets.
 	for _, url := range ctx.Args() {
 		if isRecursive {
-			if isHostName(url) {
+			if !isAliasURLDir(url) && (strings.Count(url, string(filepath.Separator)) == 0) {
```
balamurugana: same as above

poornas: isAliasURLDir is used internally by the checkCopySyntax function. Here the intent is to make sure only the alias is entered, and it is at the root level not a subdirectory.

poornas:
@harshavardhana and @balamurugana , updated the PR with a different approach - recursive removal is handled agnostic to fs/s3.

harshavardhana:
> To distinguish full site removal from removing a bucket and provide appropriate warning to the user, the user needs to supply --dangerous flag in addition to the --recursive and --force flags for full site rm.

Tested the PR `mc rm -r --force play/`  seems to delete everything, but strangely `--dangerous` flag is removed was that intended?

poornas:
that was an intermediate commit - updated the PR to fix messaging  as well.

harshavardhana:
Can you sqaush the commits here,. Will test once again after they are squashed.

poornas:
done.

cmd/rm-main.go
```diff
@@ -82,16 +88,22 @@ EXAMPLES:
    1. Remove a file.
       $ {{.HelpName}} 1999/old-backup.tgz
 
-   2. Remove all objects recursively.
+   2. Remove all objects recursively from a bucket.
```
harshavardhana: `Remove all objects recursively from bucket 'jazz-songs' matching 'louis' prefix.`

cmd/rm-main.go
```diff
@@ -82,16 +88,22 @@ EXAMPLES:
...
       $ {{.HelpName}} --recursive s3/jazz-songs/louis/
 
-   3. Remove all objects older than '90' days.
+   3. Remove all objects older than '90' days from a bucket.
```
harshavardhana: Same as above but with 'older than 90 days'

cmd/rm-main.go
```diff
@@ -82,16 +88,22 @@ EXAMPLES:
...
       $ {{.HelpName}} --recursive --older-than=90 s3/jazz-songs/louis/
 
    4. Remove all objects read from STDIN.
       $ {{.HelpName}} --force --stdin
 
-   5. Drop all incomplete uploads on 'jazz-songs' bucket.
+   5. Remove all buckets and objects recursively from host
```
harshavardhana: `Remove all buckets and objects recursively from S3 host`

poornas:
@harshavardhana  and @balamurugana , review comments were addressed. needs a revisit.

cmd/rm-main.go
```diff
@@ -121,17 +133,32 @@ func checkRmSyntax(ctx *cli.Context) {
 	isForce := ctx.Bool("force")
 	isRecursive := ctx.Bool("recursive")
 	isStdin := ctx.Bool("stdin")
-
+	isDangerous := ctx.Bool("dangerous")
+	isNamespaceRemoval := false
+	for _, url := range ctx.Args() {
+		if !isAliasURLDir(url) && (strings.Count(url, string(filepath.Separator)) == 0) {
```
balamurugana: 1. Please add a comment about meaning of this logic.
2. `()` around `strings.` is not required.

cmd/rm-main.go
```diff
@@ -219,12 +247,12 @@ func removeRecursive(url string, isIncomplete bool, isFake bool, older int) erro
 			}
 		}
 
-		urlString := content.URL.Path
-		printMsg(rmMessage{
-			Key:  targetAlias + urlString,
-			Size: content.Size,
-		})
-
+		if !(content.Time.IsZero() && strings.HasSuffix(urlString, string(filepath.Separator))) {
```
balamurugana: Please add a comment about this check

cmd/rm-main.go
poornas: done

cmd/rm-main.go
poornas: done.

cmd/rm-main.go
```diff
@@ -121,17 +133,34 @@ func checkRmSyntax(ctx *cli.Context) {
 	isForce := ctx.Bool("force")
 	isRecursive := ctx.Bool("recursive")
 	isStdin := ctx.Bool("stdin")
-
+	isDangerous := ctx.Bool("dangerous")
+	isNamespaceRemoval := false
+	for _, url := range ctx.Args() {
+		// check if url represents just the namespace without bucket or object prefix specified
+		// in the url
+		if !isAliasURLDir(url) && strings.Count(url, string(filepath.Separator)) == 0 {
```
balamurugana: A valid input `s3/` and unusual input `s3//` wouldn't work here.  We need to take care properly i.e. we shouldn't use `strings.Count()` inside use url to alias conversion.

If mentioned input is not supported, we have to error out properly.

poornas: you are correct @balamurugana , added additional validation so that user input errors are caught right away.

cmd/rm-main.go
```diff
@@ -121,17 +133,54 @@ func checkRmSyntax(ctx *cli.Context) {
...
+			fatalIf(err.Trace(aliasedURL), "Failed to remove `"+aliasedURL+"`.")
+		}
+		// validate extraneus delimiters such as play// which won't be caught by previous check
+		if path != "" && bucket == "" {
+			fatalIf(err.Trace(aliasedURL), "Failed to remove `"+aliasedURL+"`.")
 
+		}
+		if alias != "" && path == "" && isBucketEmpty {
+			isNamespaceRemoval = true
+			break
```
balamurugana: The whole checker logic is confusing to me.  In `rm`  we should not check for bucket or object which is completely abstracted by client interface.  I see below logic is good enough

```go
url = filepath.ToSlash(url)
if !isAliasURLDir(url) {
    alias, path := url2Alias(url)
    isNamespaceRemoval = (path == "")
}
```

poornas: without the check on lines 157-158, errors like mc rm play// wont be caught. Earlier it was being caught in s3 api because bucket name was not permitted to be empty. Now by allowing namespace removal, bucket name can be empty and there is no way to validate cases like multiple consecutive / delimiters being entered as input. Lines 160-163 can be replaced by the logic you proposed.

balamurugana: I see `url = filepath.ToSlash(url)` is wrong, you should use `url = filepath.Clean(url)`

poornas: done.

cmd/rm-main.go
```diff
@@ -121,17 +133,34 @@ func checkRmSyntax(ctx *cli.Context) {
 	isForce := ctx.Bool("force")
 	isRecursive := ctx.Bool("recursive")
 	isStdin := ctx.Bool("stdin")
-
+	isDangerous := ctx.Bool("dangerous")
+	isNamespaceRemoval := false
+	for _, url := range ctx.Args() {
+		url = filepath.Clean(url)
```
balamurugana: you could use `path.Clean()` to work with MS Windows.

poornas: doesn't filepath.Clean do the same?

balamurugana: no

harshavardhana: path.Clean() doesn't do anything on windows @balamurugana @poornas - filepath.Clean() should be used for cross platform friendly behavior.

balamurugana: See below.  This had been discussed long back

```go
package main

import (
	"fmt"
	"path"
	"path/filepath"
)

func main() {
	fmt.Println("filepath.Clean():")
	fmt.Println(filepath.Clean("/s3"))
	fmt.Println(filepath.Clean("s3/"))
	fmt.Println(filepath.Clean("s3/mybucket//object//"))
	fmt.Println(filepath.Clean(`\s3`))
	fmt.Println(filepath.Clean(`s3\`))
	fmt.Println(filepath.Clean(`s3\mybucket\\object\\`))

	fmt.Println("path.Clean():")
	fmt.Println(path.Clean("/s3"))
	fmt.Println(path.Clean("s3/"))
	fmt.Println(path.Clean("s3/mybucket//object//"))
	fmt.Println(path.Clean(`\s3`))
	fmt.Println(path.Clean(`s3\`))
	fmt.Println(path.Clean(`s3\mybucket\\object\\`))
}
```

On MS Windows
```
filepath.Clean():
\s3
s3
s3\mybucket\object
\s3
s3
s3\mybucket\object
path.Clean():
/s3
s3
s3/mybucket/object
\s3
s3\
s3\mybucket\\object\\
```

On GNU/Linux

```
filepath.Clean():
/s3
s3
s3/mybucket/object
\s3
s3\
s3\mybucket\\object\\
path.Clean():
/s3
s3
s3/mybucket/object
\s3
s3\
s3\mybucket\\object\\
```

harshavardhana: i don't think so you understand the problem of using path.Clean() , path.Clean() doesn't work on windows since it never cleans paths such as `\\` or `C:\mypath\...\otherdir\` . path.Clean() doesn't understand C:\ etc. so it would be wrong to use path.Clean() for cross platform reasons. 

For `mc` we do support such paths and need to behave properly for windows natively once we do filepath.Clean() if we want all the `\` to `/` then we should use filepath.ToSlash() to convert the paths properly.

Using code like in this manner `filepath.ToSlash(filepath.Clean(p))` would work for all operating systems.

balamurugana: How is it possible to store object name containing `\`?

harshavardhana: filepath.ToSlash is used to do proper Windows translation, filepath.ToSlash has no effect on Unixes. Now if you are arguing that if there are files which have `\` on Linux then they are stored fine with this mechanism as well.  Now on windows there is no filename which explicitly contains `\` since its a separator. 

Here is how the behavior is 

- filepath.ToSlash converts all `\` on Windows to `/` 
- filepath.ToSlash converts all `/` to '/` on Unixes. 

filepath.ToSlash is not ubiquitously converting `\` to `/` - https://play.golang.org/p/fFPdEZoAEf

harshavardhana:
@poornas can you squash your changes? would it possible to bring in some unit tests as well if possible?

cmd/rm-main.go
```diff
@@ -121,17 +134,34 @@ func checkRmSyntax(ctx *cli.Context) {
 	isForce := ctx.Bool("force")
 	isRecursive := ctx.Bool("recursive")
 	isStdin := ctx.Bool("stdin")
-
+	isDangerous := ctx.Bool("dangerous")
+	isNamespaceRemoval := false
+	for _, url := range ctx.Args() {
+		url = path.Clean(url)
```
harshavardhana: It is required to perhaps bring change like this instead `filepath.ToSlash(filepath.Clean(url))` here, for windows paths such as `C:\` path.Clean will not work.

Also question for me why is this path.Clean needed? is it for isAliasURLDir() ? A comment would be needed i think here.

poornas: @y4m4,  on further thinking, the path.Clean and isAliasURLDir() check are superfluous and can be dispensed with altogether.

harshavardhana: Okay that would be better.

poornas:
@harshavardhana , have not introduced stand alone functions that can be unit tested here. It might be worthwhile to add a mint cleanup function in the functional tests that exercises this code - but that will cleanup play server and may not be entirely good.

harshavardhana:
> @harshavardhana , have not introduced stand alone functions that can be unit tested here. It might be worthwhile to add a mint cleanup function in the functional tests that exercises this code - but that will cleanup play server and may not be entirely good.

Okay that makes sense.

cmd/rm-main.go
```diff
@@ -121,17 +133,31 @@ func checkRmSyntax(ctx *cli.Context) {
 	isForce := ctx.Bool("force")
 	isRecursive := ctx.Bool("recursive")
 	isStdin := ctx.Bool("stdin")
-
+	isDangerous := ctx.Bool("dangerous")
+	isNamespaceRemoval := false
+	for _, url := range ctx.Args() {
+		_, path := url2Alias(url)
```
balamurugana: What are all expected behaviors for below command line arguments?
```
1. mc rm mylocaldir
2. mc rm mylocaldir/
3. mc rm mylocalfile
4. mc rm c:
5. mc rm c:\
```

poornas: ➜  mc git:(rr2) mkdir mylocaldir
➜  mc git:(rr2) mc rm mylocaldir 
Removing `mylocaldir`.
➜  mc git:(rr2) mkdir mylocaldir
➜  mc git:(rr2) touch mylocaldir/testfile
➜  mc git:(rr2) ✗ mc rm mylocaldir 
Removing `mylocaldir`.
mc: <ERROR> Failed to remove `mylocaldir`. remove mylocaldir: directory not empty
➜  mc git:(rr2) ✗ mc rm mylocaldir 
Removing `mylocaldir`.
mc: <ERROR> Failed to remove `mylocaldir`. remove mylocaldir: directory not empty

poornas: however, removing the isAliasURLDir() check will interfere with mc rm messaging in recursive removal. Which is why the check was introduced in the first place - for filtering out fs.
```➜  mc git:(rr2) ✗ mc rm --r --force mylocaldir 
mc: <ERROR> Full site removal requires --dangerous option. This operation removes all the buckets on your namespace. Please review carefully before performing this *DESTRUCTIVE* operation. ```

Comment (poornas) [reply cmd/rm-main.go]: @harshavardhana and @balamurugana , added the isAliasURLDir() check back and a comment on why it is needed.

File: cmd/rm-main.go [hunk 21]
```diff
@@ -121,17 +133,34 @@ func checkRmSyntax(ctx *cli.Context) {
 	isForce := ctx.Bool("force")
 	isRecursive := ctx.Bool("recursive")
 	isStdin := ctx.Bool("stdin")
-
+	isDangerous := ctx.Bool("dangerous")
+	isNamespaceRemoval := false
+	for _, url := range ctx.Args() {
+		// namespace removal applies only for non FS. So filter out if passed url represents a directory
+		if !isAliasURLDir(url) {
+			_, path := url2Alias(url)
```
balamurugana: Does `url2Alias()` take care of returning clean path including windows UNC path?

poornas: yes, see test output below..
```
C:\go\bin> cd ..\src\github.com\minio\mc
C:\go\src\github.com\minio\mc [rr2 ≡]> mc rm --r --force \\myserver\myfolder\myfile
mc.exe: <ERROR> Failed to remove `\\myserver\myfolder\myfile` recursively. open \\myserver\myfolder\myfile: The network name cannot be found.
C:\go\src\github.com\minio\mc [rr2 ≡]> mc rm --r --force C:
Removing `C:.git\FETCH_HEAD`.

C:\go\src\github.com\minio\mc> mc rm C:\
Removing `C:\`.
mc.exe: <ERROR> Failed to remove `C:\`. Insufficient permissions to access this file `C:\`
C:\go\src\github.com\minio\mc>
C:\go\src\github.com\minio\mc> mc rm C:
Removing `C:`.
mc.exe: <ERROR> Failed to remove `C:`. remove C:: The process cannot access the file because it is being used by a
ther process.
C:\go\src\github.com\minio\mc>

C:\go\src\github.com\minio\mc> mc rm --r --force  C:
Removing `C:cmd\mc_test.go`.

~> mc mb a/b/c/d
~> mc rm a
Removing `a`.
mc.exe: <ERROR> Failed to remove `a`. remove a: The directory is not empty.
~> mc rm \\TMP\ss\
mc.exe: <ERROR> Failed to remove `\\TMP\ss\`. GetFileAttributesEx \\TMP\ss: The network name cannot be
~> mc rm C:\Users\vagrant\a\\\
Removing `C:\Users\vagrant\a\\\`.
mc.exe: <ERROR> Failed to remove `C:\Users\vagrant\a\\\`. remove C:\Users\vagrant\a\\\: The directory
~> mc rm C:\Users\vagrant\a\\\b\\\c\\\d
Removing `C:\Users\vagrant\a\\\b\\\c\\\d`.
~>
```

balamurugana: We used `path.Clean()` to take care aliases like `s3` and `s3/`.  Hope return `path` from `url2Alias()` is clean.

poornas: yes, added a comment for that as well.

cmd/rm-main.go
```diff
@@ -121,17 +133,36 @@ func checkRmSyntax(ctx *cli.Context) {
 	isForce := ctx.Bool("force")
 	isRecursive := ctx.Bool("recursive")
 	isStdin := ctx.Bool("stdin")
-
+	isDangerous := ctx.Bool("dangerous")
+	isNamespaceRemoval := false
+	for _, url := range ctx.Args() {
+		// clean path for aliases like s3/
+		url = filepath.ToSlash(filepath.Clean(url))
```
balamurugana: interestingly UNC path using  `/` is working in go1.9.2 by breaking the UNC path specification.

harshavardhana: It is not breaking @balamurugana they automatically convert it internally properly. Before any Win32 calls are made `os` package converts them to necessary values. 

FWIW Win32 calls support `/` path separators as well.

balamurugana: Check on windows how it behaves UNC path with `/`

poornas:
```
C:\GO\src\github.com\minio\mc [(0f4e134...) +0 ~1 -0 !]> mc mb a/b/c/d/e
Bucket created successfully `a/b/c/d/e`.
C:\GO\src\github.com\minio\mc [(0f4e134...) +0 ~1 -0 !]> tree a
Folder PATH listing for volume Windows 10
Volume serial number is 00000200 4286:2744
C:\GO\SRC\GITHUB.COM\MINIO\MC\A
└───b
    └───c
        └───d
            └───e
C:\GO\src\github.com\minio\mc [(0f4e134...) +0 ~1 -0 !]> mc rm a/b/c/d/e
Removing `a/b/c/d/e`.
C:\GO\src\github.com\minio\mc [(0f4e134...) +0 ~1 -0 !]>
C:\GO\src\github.com\minio\mc [(0f4e134...) +0 ~1 -0 !]> mc rm //network1/blah/blah
mc.exe: <ERROR> Failed to remove `//network1/blah/blah`. Requested file `/blah/blah` not found
C:\GO\src\github.com\minio\mc [(0f4e134...) +0 ~1 -0 !]> mc rm --r --force //network1/blah/blah
mc.exe: <ERROR> Failed to remove `//network1/blah/blah` recursively. open /blah/blah: The system canno
h specified.
```

balamurugana: No surprise go applications are working cleanly.  See my above comment.  Just add a comment what exactly i mentioned.

poornas: done.

harshavardhana:
Moving this to blocked as requested by @poornas  until @abperiasamy gives final review on the changes and UI.

poornas:
after discussion with @abperiasamy , changed messaging for the  --force and --dangerous flags and error message if flags are not supplied for site wide rm command. This is unblocked now

PR #2267, opened 2017-10-01
Ignore test_watch_object() for mint in functional test

balamurugana:


functional-tests.sh
```diff
@@ -437,7 +437,9 @@ function run_test()
     test_presigned_get_object
     test_cat_object
     test_mirror_list_objects
-    test_watch_object
+    if [ -z "$MINT_MODE" ]; then
```
vadmeste: Can you also add a documentation of MINT_MODE in the beginning of this file? the purpose of MINT_MODE is not fully clear for me.

balamurugana:
> Changes look good. Some context why is this needed will be useful here

Notification support is not there in case of all gateways.  Its better to avoid this case in mint environment where as travis-ci does the test.

PR #2269, opened 2017-10-04
Add exclude option to mirror command

### LINKED ISSUE #1903: mirror: [feature] mc mirror --exclude option to exclude unwanted system files like .Ds_Store and/or to only mirror only certain extensions
## Expected behaviour

1. `mc mirror --exclude .* --exclude *.temp` excludes all .* files and *.temp files
2. `mc mirror --only *.a`, mirrors only *.a files

Remark : would not mind for a `mc mirror --watch --only` if an watch event was created for non compliant files, as long as they are not mirrored.
## Actual behaviour
na
na
## mc version
na

kannappanr:
Added exclude options to exclude unwanted system files in an ```mc mirror``` command.

Exclude option excludes the source files/objects that match the passed shell file name pattern.

Examples
```
mc mirror --exclude ".*" --exclude "*.temp" play/test ~/test
```
excludes all .* files and *.temp files

```
mc --quiet mirror --exclude ".*" ~/data/ play/test
```
excludes all .* files from being mirrored from the source to the target

Fixes #1903

### DIFF
--- a/cmd/difference.go
+++ b/cmd/difference.go
@@ -20,6 +20,8 @@ import (
 	"strings"
 	"unicode/utf8"
 
+	// golang does not support flat keys for path matching, find does
+	"github.com/minio/minio/pkg/wildcard"
 	"golang.org/x/text/unicode/norm"
 )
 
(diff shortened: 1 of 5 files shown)


vadmeste:
@kannappanr, can you update mirror help message and add an example or two of --exclude usage ?

vadmeste:
@kannappanr can you also update this PR description with a brief specification of --exclude arg ?

cmd/difference.go
```diff
@@ -133,6 +151,12 @@ func difference(sourceClnt, targetClnt Client, sourceURL, targetURL string, isRe
 
 			current := urlJoinPath(targetURL, srcSuffix)
 			expected := urlJoinPath(targetURL, tgtSuffix)
+
```
ebozduman: We can move the `matchExcludeOptions` if block from here to above before line#134, so that `matchExcludeOptions` if block inside `if tgtEOF {` on line#134 can be eliminated. 
Line number #149 also needs to be moved up for the above change to work.

kannappanr: Done.

kannappanr:
@vadmeste ,  I have updated the PR with a description of Exclude option and also added an example to the mc mirror help command.

cmd/difference.go
```diff
@@ -53,17 +54,29 @@ func (d differType) String() string {
...
 
 func dirDifference(sourceClnt, targetClnt Client, sourceURL, targetURL string) (diffCh chan diffMessage) {
-	return difference(sourceClnt, targetClnt, sourceURL, targetURL, false, true, DirFirst)
+	return difference(sourceClnt, targetClnt, sourceURL, targetURL, false, true, DirFirst, nil)
+}
+
+func matchExcludeOptions(excludeOptions []string, srcSuffix string) bool {
+	if excludeOptions != nil {
+		for _, pattern := range excludeOptions {
+			r, err := filepath.Match(pattern, srcSuffix)
```
harshavardhana: We should add a testCase for this , also perhaps we should use wildcard package from server used by find.go instead than filepath.Match()

An example case is here - https://play.golang.org/p/Ega9qgD4Qz - Since we deal with flat-namespace for both filesystem and object storage we should use this instead.

kannappanr: done

cmd/difference_test.go
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package cmd
+
+import (
+	"testing"
+)
+
+//TestFind is the structure used to contain params pertinent to find related tests
+type TestExclude struct {
```
harshavardhana: We don't need a type to be outside we can simply use 

```
testCases := []struct{}
``` 

As opaque structs, this is to avoid namespace because `TestExclude` as a struct name is just too generic.  You can move the code entirely inside the test function itself. 

Look at the master `find_test.go` for more examples on this.

PR #2273, opened 2017-10-06
Fixes find command and bring it to a workable condition.

harshavardhana:
- Watch now avoids an unnecessary waitgroup and simplifies overall flow.
- Watch now honors maxDepth parameter as well.
- Parsing logic is fully outside in the mainFind() function.
- Enhance and fix behavior of nameMatch to comply with unix
  find behavior.


harshavardhana:
Still working on the unit tests, but it is ready for the early eyes.

harshavardhana:
> Still working on the unit tests, but it is ready for the early eyes.

Done, ready for review.

cmd/find-main.go
```diff
@@ -163,24 +162,44 @@ EXAMPLES:
...
-	// verify that there are no empty arguments
 	for _, arg := range args {
 		if strings.TrimSpace(arg) == "" {
 			fatalIf(errInvalidArgument().Trace(args...), "Unable to validate empty argument.")
 		}
 	}
 
+	// Extract input URLs.
+	URLs := ctx.Args()
+	for _, url := range URLs {
```
kannappanr: comment like "Validate the extracted URLs. " will be good.

cmd/find-main.go
```diff
@@ -193,11 +212,52 @@ func mainFind(ctx *cli.Context) error {
...
+	if ctx.String("older") != "" {
+		olderThan, err = parseTime(ctx.String("older"))
+		fatalIf(err.Trace(ctx.String("older")), "Unable to parse input threshold value into time.Time")
+	}
+	if ctx.String("newer") != "" {
+		newerThan, err = parseTime(ctx.String("newer"))
+		fatalIf(err.Trace(ctx.String("newer")), "Unable to parse input threshold value into time.Time")
+	}
+
+	var e error
```
kannappanr: Why can't we use already declared 'err' variable?

harshavardhana: they both have different type.

cmd/find-main.go
harshavardhana: Sure

cmd/find.go
```diff
@@ -74,6 +73,14 @@ func (f findMessage) JSON() string {
 func nameMatch(pattern, path string) bool {
 	matched, e := filepath.Match(pattern, filepath.Base(path))
 	errorIf(probe.NewError(e).Trace(pattern, path), "Unable to match with input pattern")
+	if !matched {
```
kannappanr: When will this happen?

cmd/find.go
```diff
@@ -145,30 +115,95 @@ func doFindExec(ctx *cli.Context, fileContent contentMessage) {
 	console.Println(string(out.Bytes()))
 }
 
-// doFindWatch - enables listening on the input path, listens for all file/object
+// watchFind - enables listening on the input path, listens for all file/object
```
kannappanr: Line 113 above, looks like unreachable code. If it is unreachable, we need to remove it.

cmd/find-main.go
```diff
@@ -67,17 +70,13 @@ var (
 			Name:  "smaller",
 			Usage: "Match all objects smaller than specified size in units (see UNITS)",
 		},
-		cli.StringFlag{
+		cli.UintFlag{
```
ebozduman: Can we have the flags alphabetically sorted  and in the usage, for the end-user's readability, below, line#97?
Can we have the flags alphabetically sorted in the usage below, line#97, for end-user readability purposes, and here just for easy debugging purposes?

cmd/find.go
harshavardhana: It is a compatibility code.. it will happen when the filepath.Match() didn't match say for example if user said `mc find pkg --name console`

cmd/find-main.go
```diff
@@ -163,24 +162,54 @@ EXAMPLES:
...
-		cli.ShowCommandHelpAndExit(ctx, "find", 1)
-	}
-
-	if ctx.Bool("watch") && !strings.Contains(args[0], "/") {
-		console.Println("Users must specify a bucket name for watch")
-		console.Fatalln()
+	if !ctx.Args().Present() {
+		args = []string{"./"} // No args just default to present directory.
+	} else if ctx.Args().Get(0) == "." {
+		ctx.Args()[0] = "./" // If the arg is '.' treat it as './'.
```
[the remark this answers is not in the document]
harshavardhana: No reason.

cmd/find-main.go
```diff
@@ -163,24 +162,54 @@ EXAMPLES:
...
-
-	// help message on [mc][find]
-	if !args.Present() {
-		cli.ShowCommandHelpAndExit(ctx, "find", 1)
-	}
-
-	if ctx.Bool("watch") && !strings.Contains(args[0], "/") {
-		console.Println("Users must specify a bucket name for watch")
-		console.Fatalln()
+	if !ctx.Args().Present() {
```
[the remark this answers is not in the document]
harshavardhana: No reason.

cmd/find-main.go
```diff
@@ -191,13 +220,63 @@ func mainFind(ctx *cli.Context) error {
 	checkFindSyntax(ctx)
 
 	args := ctx.Args()
-
-	if !ctx.Args().Present() {
-		args = []string{"."}
+	if !args.Present() {
```
ebozduman: Checks and codes between line#222 and line#227 are already covered as part of the call to `checkFindSyntax(ctx)`. So, they are unnecessary and can be removed.

cmd/find-main.go
```diff
@@ -191,13 +220,63 @@ func mainFind(ctx *cli.Context) error {
...
+		args = []string{"./"} // Not args present default to present directory.
+	} else if args.Get(0) == "." {
+		args[0] = "./" // If the arg is '.' treat it as './'.
 	}
 
 	clnt, err := newClient(args[0])
-	fatalIf(err.Trace(args...), "Unable to initialize `"+args[0]+"`")
+	fatalIf(err.Trace(args...), "Unable to initialize `"+args[0]+"`.")
+
+	var (
```
ebozduman: It could be declared like `var olderThan, newerThan time.Time`

cmd/find.go
```diff
@@ -93,159 +98,180 @@ func regexMatch(pattern, path string) bool {
...
+// Descend at most (a non-negative integer) levels of files
+// below the starting-prefix and trims the suffix. This function
+// returns path as is without manipulation if the maxDepth is 0
+// i.e (not set).
+func trimSuffixAtMaxDepth(startPrefix, path, separator string, maxDepth uint) string {
+	if maxDepth == 0 {
+		return path
+	}
+	// Remove the requested prefix from consideration, maxDepth is
+	// only considered for all other levels excluding th starting prefix.
```
ebozduman: `...excluding th? starting prefix.` ?

cmd/find.go
```diff
@@ -316,89 +342,26 @@ func stringsReplace(args string, fileContent contentMessage) string {
...
-		match = olderThanMatch(ctx.String("older"), fileContent.Time)
-	} else if ctx.String("newer") != "" {
-		match = newerThanMatch(ctx.String("newer"), fileContent.Time)
-	} else if ctx.String("larger") != "" {
-		match = largerSizeMatch(ctx.String("larger"), fileContent.Size)
-	} else if ctx.String("smaller") != "" {
-		match = smallerSizeMatch(ctx.String("smaller"), fileContent.Size)
+	if ctx.ignorePattern != "" {
+		match = !pathMatch(ctx.ignorePattern, fileContent.Key)
+	} else if ctx.namePattern != "" {
```
ebozduman: To be able to combine choices with `AND`, these elses need to be removed.

cmd/find-main.go
harshavardhana: This is needed for argument adjustment.

cmd/find-main.go
```diff
@@ -163,24 +162,54 @@ EXAMPLES:
...
-	// verify that there are no empty arguments
 	for _, arg := range args {
 		if strings.TrimSpace(arg) == "" {
 			fatalIf(errInvalidArgument().Trace(args...), "Unable to validate empty argument.")
 		}
 	}
 
+	// Extract input URLs and validate.
+	URLs := ctx.Args()
+	for _, url := range URLs {
```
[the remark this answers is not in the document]
harshavardhana: we can.

cmd/find.go
```diff
@@ -316,89 +342,33 @@ func stringsReplace(args string, fileContent contentMessage) string {
...
-		match = regexMatch(ctx.String("regex"), fileContent.Key)
-	} else if ctx.String("older") != "" {
-		match = olderThanMatch(ctx.String("older"), fileContent.Time)
-	} else if ctx.String("newer") != "" {
-		match = newerThanMatch(ctx.String("newer"), fileContent.Time)
-	} else if ctx.String("larger") != "" {
-		match = largerSizeMatch(ctx.String("larger"), fileContent.Size)
-	} else if ctx.String("smaller") != "" {
-		match = smallerSizeMatch(ctx.String("smaller"), fileContent.Size)
+	if match && ctx.ignorePattern != "" {
```
kannappanr: match here is unncessary

harshavardhana: It is for completeness.. otherwise logically it is hard.

kannappanr: OK

PR #2274, opened 2017-10-06
Allow disabling multipart uploads

Furtchet:
Added MC_MULTIPART_DISABLE environment variable. This allows for cp and mirroring to S3 repositories that don't support a multipart upload, like https://github.com/joyent/s3-manta-bridge

### DIFF
--- a/cmd/cat-main.go
+++ b/cmd/cat-main.go
@@ -17,12 +17,17 @@
 package cmd
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
 	"os"
 	"strings"
 	"syscall"
+	"unicode"
+	"unicode/utf8"
+
+	"golang.org/x/crypto/ssh/terminal"
 
 	"github.com/minio/cli"
 	"github.com/minio/mc/pkg/probe"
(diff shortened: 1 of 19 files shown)


Furtchet:
I tried running `git rebase -i` to fix the .*main.go conflicts from https://github.com/minio/mc/blob/master/CONTRIBUTING.md but I might just have made it worse. Let me know if I should destroy my branch and commit again.

harshavardhana:
Please rebase the PR also lets discuss this first in a Github issue @Furtchet , intentions are not clear to add this just for Manta Object Storage.

Furtchet:
Thank you harshavardhana, I was unaware of the significance of the vendor directory. Also, next time I will start with an issue instead of a PR.

Furtchet:
@balamurugana,

This was useful because the software I was doing a `mc mirror` to was https://github.com/joyent/s3-manta-bridge and it didn't support multipart uploads. Yes I know the better solution would be to improve s3-manta-bridge to support multipart but that was more difficult because manta doesn't support multipart upload (currently https://github.com/joyent/rfd/blob/master/rfd/0065/README.md). And maybe someone else would like this feature too.

PR #2275, opened 2017-10-07
Add new stat sub command

### LINKED ISSUE #2251: Retrieve metadata from objects
## Expected behaviour

Metadata is returned as `X-Amz-Meta-*` HTTP request headers, or as metadata keys in JSON response. Perhaps a new cmdline flag (ex: `mc ls --head mybucket/object.txt`) or in DEBUG output (ex: `mc --debug ls mybucket/object.txt`).

The metadata appears when using the python `awscli` cmdline tool.

## Actual behaviour

User-defined metadata is never returned.

```
mc --json ls mybucket/object.txt`
{"status":"success","type":"file","lastModified":"2017-09-12T03:55:50.209Z","size":86016,"key":"object.txt"}

aws --endpoint-url http://localhost:9000 s3api head-object --bucket mybucket --key object.txt
{
    "AcceptRanges": "bytes", 
    "ContentType": "text/plain; charset=utf-8", 
    "LastModified": "Tue, 12 Sep 2017 03:55:50 GMT", 
    "ContentLength": 86016, 
    "ETag": "\"b181c8c08967f88003bf81b97e07470a\"", 
    "Metadata": {
        "username": "tester", 
        "location": "office", 
        "email": "tester@test.com", 
        "sha256": "1ab503a3f4e48431aba92bc8769b2b3becb4865437b5165990a42b7fbbb2dee5"
    }
}
```

## mc version

```
Version: 2017-09-10T05:49:25Z
Release-tag: DEVELOPMENT.2017-09-10T05-49-25Z
Commit-id: 631f7fc194fe0ad8d26cb16ca264ca6665fcd151
```

## System information

Linux w/kernel 3.16.38 x86_64

EDIT: wrong kernel version

poornas:
This PR adds the ability to stat contents of object or objects within a bucket.
e.g. `mc stat play/bucket/object and mc stat --recursive
play/bucket/prefix`

Fixes #2251 

This is wip because the format of output to console still needs to be finalized.

### DIFF
--- a/cmd/client-fs_darwin.go
+++ b/cmd/client-fs_darwin.go
@@ -1,7 +1,7 @@
-// +build linux
+// +build darwin
 
 /*
- * Minio Client (C) 2015, 2016, 2017 Minio, Inc.
+ * Minio Client (C) 2017 Minio, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this fs except in compliance with the License.
@@ -19,31 +19,21 @@
 package cmd
 
 import (
-	"encoding/hex"
-	"strings"
-
 	"github.com/pkg/xattr"
 	"github.com/rjeczalik/notify"
-
-	"unicode/utf8"
 )
 
 var (
-	// EventTypePut contains the notify events that will cause a put (write)
-	EventTypePut = []notify.Event{notify.InCloseWrite | notify.InMovedTo}
+	// EventTypePut contains the notify events that will cause a put (writer)
+	EventTypePut = []notify.Event{notify.Create, notify.Write, notify.Rename}
 	// EventTypeDelete contains the notify events that will cause a delete (remove)
-	EventTypeDelete = []notify.Event{notify.InDelete | notify.InDeleteSelf | notify.InMovedFrom}
+	EventTypeDelete = []notify.Event{notify.Remove}
 	// EventTypeGet contains the notify events that will cause a get (read)
-	EventTypeGet = []notify.Event{notify.InAccess | notify.InOpen}
+	EventTypeGet = []notify.Event{} // On macOS, FreeBSD, Solaris this is not available.
 )
 
 // IsGetEvent checks if the event return is a get event.
 func IsGetEvent(event notify.Event) bool {
-	for _, ev := range EventTypeGet {
-		if event&ev != 0 {
-			return true
-		}
-	}
 	return false
 }
 
@@ -59,12 +49,7 @@ func IsPutEvent(event notify.Event) bool {
 
 // IsDeleteEvent checks if the event returned is a delete event
 func IsDeleteEvent(event notify.Event) bool {
-	for _, ev := range EventTypeDelete {
-		if event&ev != 0 {
-			return true
-		}
-	}
-	return false
+	return event&notify.Remove != 0
 }
 
 // getXAttr fetches the extended attribute for a particular key on
@@ -74,10 +59,7 @@ func getXAttr(path, key string) (string, error) {
 	if e != nil {
 		return "", e
 	}
-	if utf8.ValidString(string(data)) {
-		return string(data), nil
-	}
-	return hex.EncodeToString(data), nil
+	return string(data), nil
 }
 
 // getAllXattrs returns the extended attributes for a file if supported
@@ -92,14 +74,11 @@ func getAllXattrs(path string) (map[string]string, error) {
 		return nil, e
 	}
 	for _, key := range list {
-		// filter out system specific xattr
-		if strings.HasPrefix(key, "system.") {
-			continue
-		}
 		xMetadata[key], e = getXAttr(path, key)
 		if e != nil {
 			return nil, e
 		}
+
 	}
 	return xMetadata, nil
 }
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -839,11 +839,8 @@ func (c *s3Client) Stat(isIncomplete, isFetchMeta bool) (*clientContent, *probe.
 		if !exists {
 			return nil, probe.NewError(BucketDoesNotExist{Bucket: bucket})
 		}
-		bucketMetadata := &clientContent{}
-		bucketMetadata.URL = *c.targetURL
-		bucketMetadata.Type = os.ModeDir
-		bucketMetadata.Metadata = map[string]string{}
-		return bucketMetadata, nil
+		bucketMetadata := c.bucketStat()
+		return &bucketMetadata, nil
 	}
 
 	// Remove trailing slashes needed for the following ListObjects call.
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -54,8 +54,16 @@ const (
 	amazonHostName            = "s3.amazonaws.com"
 	amazonHostNameAccelerated = "s3-accelerate.amazonaws.com"
 
-	googleHostName = "storage.googleapis.com"
+	googleHostName            = "storage.googleapis.com"
+	serverEncryptionKeyPrefix = "x-amz-server-side-encryption"
 )
+
+// cseHeaders is list of client side encryption headers
+var cseHeaders = []string{
+	"X-Amz-Meta-X-Amz-Iv",
+	"X-Amz-Meta-X-Amz-Key",
+	"X-Amz-Meta-X-Amz-Matdesc",
+}
 
 // getNumMultiPartThreads gets the number of multipart threads to be used for mc cp/mirror functions
 // If not set, default to 4.
@@ -814,7 +822,7 @@ func (c *s3Client) listObjectWrapper(bucket, object string, isRecursive bool, do
 }
 
 // Stat - send a 'HEAD' on a bucket or object to fetch its metadata.
-func (c *s3Client) Stat(isIncomplete bool) (*clientContent, *probe.Error) {
+func (c *s3Client) Stat(isIncomplete, isFetchMeta bool) (*clientContent, *probe.Error) {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 	bucket, object := c.url2BucketAndObject()
@@ -834,8 +842,7 @@ func (c *s3Client) Stat(isIncomplete bool) (*clientContent, *probe.Error) {
 		bucketMetadata := &clientContent{}
 		bucketMetadata.URL = *c.targetURL
 		bucketMetadata.Type = os.ModeDir
-		bucketMetadata.Metadata = map[string][]string{}
-
+		bucketMetadata.Metadata = map[string]string{}
 		return bucketMetadata, nil
 	}
 
@@ -858,14 +865,17 @@ func (c *s3Client) Stat(isIncomplete bool) (*clientContent, *probe.Error) {
 				objectMetadata.Time = objectMultipartInfo.Initiated
 				objectMetadata.Size = objectMultipartInfo.Size
 				objectMetadata.Type = os.FileMode(0664)
-				objectMetadata.Metadata = map[string][]string{}
+				objectMetadata.Metadata = map[string]string{}
+				objectMetadata.EncryptionHeaders = map[string]string{}
 				return objectMetadata, nil
 			}
 
 			if strings.HasSuffix(objectMultipartInfo.Key, string(c.targetURL.Separator)) {
 				objectMetadata.URL = *c.targetURL
 				objectMetadata.Type = os.ModeDir
-				objectMetadata.Metadata = map[string][]string{}
+				objectMetadata.Metadata = map[string]string{}
+				objectMetadata.EncryptionHeaders = map[string]string{}
+
 				return objectMetadata, nil
 			}
 		}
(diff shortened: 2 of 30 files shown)


vadmeste:
Actually I don't feel that someone will need metadata of a set of objects. I would suggest to avoid adding a new command 'stat', and showing metadata when `mc ls` returns only one file.

poornas:
@vadmeste, will discuss this with @abperiasamy  tomorrow.

harshavardhana:
> @vadmeste, will discuss this with @abperiasamy tomorrow.

what was the conclusion of this discussion? @poornas

poornas:
@vadmeste - checked with @abperiasamy  - he envisions stat subcommand as doing way more than just listing metadata for an object. stat command could also list whether the object is encrypted,ETag and so forth. For the file system, the stat command would give extended attributes returned by attr command. The --recursive flag will be handy to list objects in bucket and filter on metadata using jq.

vadmeste:
@poornas, got it.. Is this still WIP ?

poornas:
yes- will update the PR and ping you.

poornas:
@vadmeste, updated the PR.

poornas:
@harshavardhana , build passes now.

cmd/client-fs_linux.go
```diff
@@ -60,3 +61,20 @@ func IsDeleteEvent(event notify.Event) bool {
 	}
 	return false
 }
+
+// GetXattr returns the extended attributes for a file if supported
+// by the OS
+func GetXattr(path string) map[string][]string {
+	xMetadata := map[string][]string{}
+	if xattr.Supported(path) {
```
harshavardhana: We shouldn't ignore the errors in this one, other than syscall.ENOTSUP, something like this would work. 

```go
func isSysErrNotSupported(e error) bool {
	xerr, ok := e.(*xattr.Error)
	if ok {
		return xerr.Err == syscall.ENOTSUP
	}
	return false
}

func getXAttr(path, key string) (string, error) {
	data, e := xattr.Get(path, key)
	if e != nil && !isSysErrNotSupported(e) {
		return "", e
	}
	return string(data), nil
}

func getAllXattrs(path string) (map[string]string, error) {
	xMetadata := make(map[string]string)
	list, e := xattr.List(path)
	if e != nil && !isSysErrNotSupported(e) {
		return nil, e

	}
	for _, key := range list {
		xMetadata[key], e = getXAttr(path, key)
		if e != nil {
			return nil, e
		}

	}
	return xMetadata, nil
}
```

Also `map[string][]string`  can be simply `map[string]string` it is harder to manage slice `[]string`, perhaps client-s3.go API can be changed to address this as well.  We should also remove the `clientContent.Metadata` to have `map[string][]string` type. getSourceStream() seems to have a weird code like following..

```
        metadata = map[string]string{}
        if fetchStat {
                st, err := sourceClnt.Stat(false)
                if err != nil {
                        return nil, nil, err.Trace(alias, urlStr)
                }
                for k, v := range st.Metadata {
                        if len(v) > 0 {
                                metadata[k] = v[0]
                        }
                }
        }
        return reader, metadata, nil
```

This type of code can be avoided.

cmd/client-fs_linux.go
```diff
@@ -60,3 +61,20 @@ func IsDeleteEvent(event notify.Event) bool {
...
+		list, err := xattr.List(path)
+		if err == nil {
```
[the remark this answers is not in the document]
poornas: done.

cmd/client-fs_linux.go
poornas: ok,fixed this.

cmd/client-fs_darwin.go
```diff
@@ -0,0 +1,91 @@
...
+	data, e := xattr.Get(path, key)
+	if e != nil && !isSysErrNotSupported(e) {
+		return "", e
+	}
+	return string(data), nil
+}
+
+// getAllXattrs returns the extended attributes for a file if supported
+// by the OS
+func getAllXattrs(path string) (map[string]string, error) {
```
vadmeste: Is `getAllXattrs()` a recursive `getXAttr` ?

cmd/client-fs_darwin.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// getXAttr fetches the extended attribute for a particular key on
+// file
+func getXAttr(path, key string) (string, error) {
+	data, e := xattr.Get(path, key)
+	if e != nil && !isSysErrNotSupported(e) {
+		return "", e
+	}
+	return string(data), nil
```
vadmeste: I think for a better clearness of the code, we shouldn't rely on 'data' variable when e represents system not supported error.

cmd/client-fs_darwin.go
```diff
@@ -0,0 +1,91 @@
...
+	return false
+}
+
+// IsDeleteEvent checks if the event returned is a delete event
+func IsDeleteEvent(event notify.Event) bool {
+	return event&notify.Remove != 0
+}
+
+// isSysErrNotSupported returns true if OS does not support call.
+func isSysErrNotSupported(e error) bool {
```
vadmeste: I think it is better if we rename this function to isXAttrNotSupportedErr()

cmd/client-s3.go
```diff
@@ -874,24 +884,48 @@ func (c *s3Client) Stat(isIncomplete bool) (*clientContent, *probe.Error) {
...
+				objectMetadata.Metadata = stat.Metadata
+				objectMetadata.EncryptionHeaders = stat.EncryptionHeaders
+			}
 			return objectMetadata, nil
 		}
 
 		if strings.HasSuffix(objectStat.Key, string(c.targetURL.Separator)) {
 			objectMetadata.URL = *c.targetURL
 			objectMetadata.Type = os.ModeDir
-			objectMetadata.Metadata = map[string][]string{}
```
vadmeste: This is a directory (os.ModeDir), I don't think we need to fetch for Metadata here.

cmd/main.go
```diff
@@ -287,6 +287,8 @@ func registerApp() *cli.App {
 	registerCmd(configCmd)  // Configure minio client.
 	registerCmd(updateCmd)  // Check for new software updates.
 	registerCmd(versionCmd) // Print version.
+	registerCmd(findCmd)    // Find specific String patterns
```
vadmeste: findCmd is already there, we don't need to add it again

cmd/stat-main.go
```diff
@@ -0,0 +1,153 @@
...
+	for _, targetURL := range args {
+		var clnt Client
+		clnt, err := newClient(targetURL)
+		fatalIf(err.Trace(targetURL), "Unable to initialize target `"+targetURL+"`.")
+		if isAliasURLDir(targetURL) && !isRecursive {
+			fatalIf(errInvalidArgument(), "Use --recursive option to stat by prefix")
+
+		}
+		if !isRecursive || !isAliasURLDir(targetURL) {
+			var st *clientContent
```
vadmeste: I recommend that you move this block in doStat() itself so it will be more easier to read

cmd/client-fs_darwin.go
poornas: this is not a recursive function. Just extracts xattr for the list of xattrs available for the path

cmd/main.go
poornas: removed.

cmd/client-fs_darwin.go
poornas: @vadmeste , removed the isSysErrNotSupported() function because the xattr package already provides a Supported() method that implements this.

cmd/client-fs_darwin.go
poornas: removed this function.

cmd/stat-main.go
poornas: done.

poornas:
@vadmeste and @harshavardhana , can this be reviewed again? your feedback was incorporated into the PR.

cmd/stat.go
```diff
@@ -0,0 +1,223 @@
...
+func printStats(stats ...statMessage) {
+	var maxKey = 0
+	for _, stat := range stats {
+		if len(stat.Key) > maxKey {
+			maxKey = len(stat.Key)
+		}
+	}
+
+	for _, stat := range stats {
+		stat.Key = fmt.Sprintf("%-10s: %-*.*s", "Name", maxKey, maxKey, stat.Key)
```
vadmeste: In general, it is better to avoid modifying passed arguments especially in function called 'printStats()'. In this case, I think it is totally okay to pass maxKey to printStat() as an additional argument so printStat() can know how to pretty print.

Besides this leads to a bug when we pass --json flag, name field will be "Name: Key"

poornas: fixed this.

vadmeste:
@poornas, one last thing, `mc stat` on a fs file already configured with an acl attribute doesn't show a clean output

```
13:30 $ sudo $(which mc) stat /tmp/rootdir/myfile
Name      : /tmp/rootdir/myfile
Date      : 2017-10-24 13:21:14 CET 
Size      : 11B    
Type      : file 
Metadata  :
  system.posix_acl_access: ������������� ���� 
```

harshavardhana:
> @poornas, one last thing, mc stat on a fs file already configured with an acl attribute doesn't show a clean output
> 
> 13:30 $ sudo $(which mc) stat /tmp/rootdir/myfile
> Name      : /tmp/rootdir/myfile
> Date      : 2017-10-24 13:21:14 CET 
> Size      : 11B    
> Type      : file 
> Metadata  :
>   system.posix_acl_access: ����������������������� �����

system.posix_acl_access has byte arrays perhaps we should translate them properly.

poornas:
@vadmeste, updated the PR to filter out system level extended attributes for linux because the built-in attr tool seems to filter  system level attributes from the  syscall.listxattr() output  whereas the pkg/xattr implementation used in this PR does not.

cmd/stat.go
```diff
@@ -0,0 +1,222 @@
...
+
+	return string(jsonMessageBytes)
+}
+
+// parseContent parse client Content container into printer struct.
+func parseStat(c *clientContent) statMessage {
+	content := statMessage{}
+	content.Date = c.Time.Local()
+	// guess file type.
+	content.Type = func() string {
```
harshavardhana: This translation is achieved by `ls.go` already do we need to do this?

cmd/stat.go
```diff
@@ -0,0 +1,222 @@
...
+	content.ETag = c.ETag
+	content.EncryptionHeaders = c.EncryptionHeaders
+	return content
+}
+func getKey(c *clientContent) string {
+	switch {
+	// for windows make sure to print in 'windows' specific style.
+	case runtime.GOOS == "windows":
+		c.URL.Path = strings.Replace(c.URL.Path, "/", "\\", -1)
+		c.URL.Path = strings.TrimSuffix(c.URL.Path, "\\")
```
harshavardhana: Also a repeated code, we can have a common translation function for both ls and stat here.

cmd/stat.go
```diff
@@ -0,0 +1,222 @@
...
+		prefixPath = filepath.ToSlash(prefixPath)
+		// Trim prefix path from the content path.
+		contentURL = strings.TrimPrefix(contentURL, prefixPath)
+		content.URL.Path = contentURL
+		//parsedContent := parseStat(content)
+		_, stat, err := url2StatWithMetadata(targetURL+getKey(content), true)
+		if err != nil {
+			continue
+		}
+		stats = append(stats, parseStat(stat))
```
harshavardhana: We cannot build the list like this we should simply printStat() one by one as the list occurs.

cmd/stat.go
```diff
@@ -0,0 +1,222 @@
...
+	console.Println()
+}
+
+// Prints all the stats.
+func printStats(stats ...statMessage) {
+	var maxKey = 0
+	for _, stat := range stats {
+		if len(stat.Key) > maxKey {
+			maxKey = len(stat.Key)
+		}
```
harshavardhana: I see why you wanted this so that the alignment seems to be correct for all objects, i think it is better to do it more efficiently than have a tabular column adjustment.  You can pick a maximum value and add `...` at the beginning so that they are ellipsided and we can display the file name instead of a full absolute path. 

This way we can be efficient and also never loose information.. for an expanded view user can always do `mc stat --json`.  Terminal UI can choose sane defaults for these things.

cmd/stat-main.go
```diff
@@ -0,0 +1,124 @@
...
+	 {{.HelpName}} - {{.Usage}}
+ 
+ USAGE:
+	 {{.HelpName}} [FLAGS] TARGET [TARGET ...]
+ 
+ FLAGS:
+	 {{range .VisibleFlags}}{{.}}
+	 {{end}}
+ EXAMPLES:
+		1. Stat all contents of mybucket on Amazon S3 cloud storage.
```
harshavardhana: Wait too many tabs here.. you should simply copy from ls-main.go and edit it,

cmd/stat.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	return c.URL.Path
+}
+
+// doStat - list all entities inside a folder.
+func doStat(clnt Client, isRecursive bool, targetURL string) error {
+	if !isRecursive || !isAliasURLDir(targetURL) {
+		isIncomplete := false
+		isFetchMeta := true
```
harshavardhana: We don't need isFetchMeta , it can be simply `fetchMeta` the type is boolean so `is` is sort of implicit. This applies to `isRecursive, isIncomplete` as well but we can do this later - since `is` should be generally used for a function not a variable name.

cmd/stat.go
poornas: done.

cmd/stat.go
poornas: ok, renamed, tho imo isX or xFlag is more intuitive without having to refer to variable type.

cmd/stat.go
poornas: good point , this increased the speed of stat command.

poornas:
@harshavardhana , updated PR.

deekoder:
Are the codecov failures spurious or relevant? @poornas  @vadmeste  @harshavardhana

harshavardhana:
> Are the codecov failures spurious or relevant? @poornas @vadmeste @harshavardhana

We can add tests for parseStat function implemented for `stat.go`

poornas:
added test for parseStat

PR #2276, opened 2017-10-09
Enable mc build tests on go1.9

harshavardhana:


.travis.yml
```diff
@@ -9,7 +9,8 @@ env:
 - ARCH=i686
```
balamurugana: Is 32 bit required?

.travis.yml
```diff
@@ -9,7 +9,8 @@ env:
...
 
 go:
-- 1.8
+- 1.8.x
```
balamurugana: Do we need to backward compatible?

.travis.yml
harshavardhana: mc is not decided yet @balamurugana as a client perhaps its okay so we kept it.

harshavardhana: Actually you are right we don't need.. removing.

.travis.yml
harshavardhana: We do not have to be, but mc has been allowed to be built since pretty much 1.7

PR #2282, opened 2017-10-11
Fix ENV handling in mc

harshavardhana:
Now ENV supports following style and doesn't require
a pre-requisite to have an entry in `config.json`

MC_HOSTS_myminio=http://minio:minio123@localhost:9000/

Notice signature style is auto-probed if env is set.

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -1,5 +1,5 @@
 /*
- * Minio Client (C) 2015 Minio, Inc.
+ * Minio Client (C) 2015, 2016, 2017 Minio, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,11 +21,13 @@ import (
 	"fmt"
 	"io"
 	"math/rand"
+	"net/url"
 	"os"
 	"runtime"
 	"strings"
 	"time"
 
+	"github.com/minio/cli"
 	"github.com/minio/mc/pkg/console"
 	"github.com/minio/mc/pkg/probe"
 )
@@ -95,9 +97,68 @@ func splitStr(path, sep string, n int) []string {
 	return splits
 }
 
+// parse url usually obtained from env.
+func parseEnvURL(envURL string) (*url.URL, string, string, *probe.Error) {
+	u, e := url.Parse(envURL)
+	if e != nil {
+		return nil, "", "", probe.NewError(e).Trace(envURL)
+	}
+
+	var accessKey, secretKey string
+	// Check if username:password is provided in URL, with no
+	// access keys or secret we proceed and perform anonymous
+	// requests.
+	if u.User != nil {
+		accessKey = u.User.Username()
+		secretKey, _ = u.User.Password()
+	}
+
+	// Look for if URL has invalid values and return error.
+	if !((u.Scheme == "http" || u.Scheme == "https") &&
+		(u.Path == "/" || u.Path == "") && u.Opaque == "" &&
+		u.ForceQuery == false && u.RawQuery == "" && u.Fragment == "") {
+		return nil, "", "", errInvalidArgument().Trace(u.String())
+	}
+
+	// Now that we have validated the URL to be in expected style.
+	u.User = nil
+
+	return u, accessKey, secretKey, nil
+}
+
+func buildS3ConfigFromEnv(envURL string) (*Config, *probe.Error) {
+	u, accessKey, secretKey, err := parseEnvURL(envURL)
+	if err != nil {
+		return nil, err.Trace(envURL)
+	}
+
+	s3Config, err := newS3Config(cli.Args{u.String(), accessKey, secretKey})
+	if err != nil {
+		return nil, err.Trace(envURL)
+	}
+
+	s3Config.AppName = "mc"
+	s3Config.AppVersion = Version
+	s3Config.AppComments = []string{os.Args[0], runtime.GOOS, runtime.GOARCH}
+	s3Config.Debug = globalDebug
+	s3Config.Insecure = globalInsecure
+
+	return s3Config, nil
+}
+
+const mcEnvHostsPrefix = "MC_HOSTS_"
+
 // buildS3Config fetches config related to the specified alias
 // to create a new config structure
 func buildS3Config(alias, urlStr string) (*Config, *probe.Error) {
+	// Check if we can get the values from ENV, if not use from `config.json`.
+	if alias == "" {
+		alias, _ = url2Alias(urlStr)
+	}
+	if envConfig, ok := os.LookupEnv(mcEnvHostsPrefix + alias); ok {
+		return buildS3ConfigFromEnv(envConfig)
+	}
+
 	hostCfg := mustGetHostConfig(alias)
 	if hostCfg == nil {
 		return nil, probe.NewError(fmt.Errorf("The specified alias: %s not found", urlStr))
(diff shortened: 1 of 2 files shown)


cmd/utils.go
```diff
@@ -95,9 +95,55 @@ func splitStr(path, sep string, n int) []string {
...
+			"are found in your environment but not suitable for use. "+
+			"Falling back to the standard config.", envConfig, alias))
+	}
+	if !isValidHostURL(hostURL) {
+		return nil, errInvalidURL(hostURL)
+	}
+	if signatureType == "" {
+		signatureType = "s3v4"
+	}
+	if signatureType != "" && !isValidAPI(signatureType) {
```
vadmeste: signatureType is never empty at this stage, we don't need to do this check `signatureType != ""`

cmd/utils.go
```diff
@@ -95,9 +95,55 @@ func splitStr(path, sep string, n int) []string {
...
+	// We have a valid alias in env. We populate the
+	// credentials from env.
+	s3Config := new(Config)
+
+	// Fetch keys from the environement, otherwise, get them from the config file
+	keys := splitStr(envConfig, ":", 4)
+	var (
+		accessKey     = keys[0]
+		secretKey     = keys[1]
+		hostURL       = keys[2]
```
vadmeste: Isn't hostURL == urlStr in this case ?

harshavardhana: Not always @vadmeste we are overriding it.

vadmeste:
@harshavardhana, it looks like there is a problem with your last change, can you test again ?

`MC_SECRET_myminio=minio:minio123:http://localhost:9000  mc ls myminio` gives unexpected result.

harshavardhana:
> MC_SECRET_myminio=minio:minio123:http://localhost:9000 mc ls myminio gives unexpected result.

Will check @vadmeste

cmd/utils.go
```diff
@@ -95,9 +96,92 @@ func splitStr(path, sep string, n int) []string {
...
+func getAlias(aliasURL string) string {
+	alias, _ := url2Alias(aliasURL)
+	return alias
+}
+
 // buildS3Config fetches config related to the specified alias
 // to create a new config structure
 func buildS3Config(alias, urlStr string) (*Config, *probe.Error) {
+	// Check if we can get the values from ENV, if not use from `config.json`.
+	if envConfig, ok := os.LookupEnv("MC_TARGET_" + getAlias(urlStr)); ok {
```
vadmeste: I am not sure, isn't it easier to use 'alias' variable instead of 'getAlias(urlStr)' ?

cmd/utils.go
```diff
@@ -95,9 +96,92 @@ func splitStr(path, sep string, n int) []string {
...
+	}
+	if !isValidAPI(signatureType) {
+		return "", "", errInvalidArgument().Trace(signatureType, scheme)
+	}
+
+	return tokens[0], signatureType, nil
+}
+
+func buildS3ConfigFromEnv(envConfig, urlStr string) (*Config, *probe.Error) {
+	_, p := url2Alias(urlStr)
```
vadmeste: Can you move this line just above where 'p' variable is used ?

cmd/utils.go
```diff
@@ -95,9 +96,92 @@ func splitStr(path, sep string, n int) []string {
 	return splits
 }
 
+// Parse scheme format, supported formats are
+// - "http"
+// - "http+s3v2"
+// - "http+s3v4"
+func parseScheme(scheme string) (string, string, *probe.Error) {
```
vadmeste: Can you lower string scheme ? so scheme can be case insensitive

cmd/utils.go
harshavardhana: alias variable is empty @vadmeste if alias not present. So there is a catch 22 situation :-)

cmd/utils.go
harshavardhana: Good idea will add a unit test.

cmd/utils.go
harshavardhana: Sure

cmd/utils.go
```diff
@@ -95,9 +96,103 @@ func splitStr(path, sep string, n int) []string {
 	return splits
 }
 
+// Parse scheme format, supported formats are
+// - "http"
+// - "http+s3v2"
+// - "http+s3v4"
```
balamurugana: I think we can remove `http+s3v4` i.e. we can treat it as default.  This way we have four schemes `http`, `http+s3v2`,` https` and `https+s3v2` only

harshavardhana: this is just an example @balamurugana

balamurugana: If you say `http+s3v4` is not a valid scheme, then we would need to remove from example as well.  If supported, my recommendation is to remove the support.

harshavardhana: No we don't need to parseScheme is generic @balamurugana

harshavardhana:
Changed the PR to now auto-probe signature from ENV instead of `http+s3v4` type of form.

cmd/utils.go
```diff
@@ -95,9 +97,62 @@ func splitStr(path, sep string, n int) []string {
...
+	s3Config.AppComments = []string{os.Args[0], runtime.GOOS, runtime.GOARCH}
+	s3Config.Debug = globalDebug
+	s3Config.Insecure = globalInsecure
+
+	return s3Config, nil
+}
+
+func getAliasFromEnv(envs []string) string {
+	for _, k := range envs {
+		if strings.HasPrefix(k, "MC_TARGET_") {
```
vadmeste: Can you add a constant `const mcEnvPrefix = "MC_TARGET_"` ?

cmd/utils.go
```diff
@@ -95,9 +97,62 @@ func splitStr(path, sep string, n int) []string {
...
+}
+
 // buildS3Config fetches config related to the specified alias
 // to create a new config structure
 func buildS3Config(alias, urlStr string) (*Config, *probe.Error) {
+	// Check if we can get the values from ENV, if not use from `config.json`.
+	// NOTE: getAliasFromEnv() we will use alias from env if available, we will
+	// override ENV values with config entries as well.
+	if aliasEnv := getAliasFromEnv(os.Environ()); aliasEnv != "" {
+		alias = aliasEnv
```
vadmeste: alias will hold `myminio=http://minio:minio123@localhost:9000`, you need further processing here.

Besides, one problem with this approach is that we can't support more than one alias, and that's the reason we have this env format MC_TARGET_alias.

I can approve this and I'll send a separate PR to address alias issue.

harshavardhana: oh yes you are right @vadmeste i need to filter it out.  Yes that is a limitation right now, let me see if it can be addressed properly.

cmd/utils.go
harshavardhana: :+1:

harshavardhana:
Now the ENVs support multiple aliases @vadmeste - please review again.

harshavardhana:
@balamurugana PR is updated

cmd/utils.go
```diff
@@ -95,9 +97,65 @@ func splitStr(path, sep string, n int) []string {
...
+	u, e := url.Parse(envURL)
+	if e != nil {
+		return nil, "", "", probe.NewError(e).Trace(envURL)
+	}
+
+	var accessKey, secretKey string
+	if u.User != nil { // Check if username:password is provided in URL.
+		accessKey = u.User.Username()
+		secretKey, _ = u.User.Password()
+	} // Empty access and secret keys are okay, we just perform anonymous requests.
```
balamurugana: You could have this comment above the `if` block for better understanding

harshavardhana: Let me add` // else`

balamurugana: A comment before `if` would help reading comments before the code.   If you see a code itself is clear enough, you can remove comments.

harshavardhana:
@balamurugana you might have to approve again.

harshavardhana:
So got approved from @abperiasamy - @deekoder we can merge this only change done in this PR now is we moved from MC_TARGET_ to MC_HOSTS_

PR #2289, opened 2017-10-15
Implement parallel cp/mirror tasks

### LINKED ISSUE #2196: Copy Objects to minio is unusually slow
Copy objects to minio is unusual slow

## Expected Behavior

Faster?

## Current Behavior
```
./mc cp --recursive folder/ local/bucket/ is running around 4.5 MB/s
```
```
dd if=/dev/zero of=bigfile bs=1m count=1024
```
```
./mc cp bigfile minio/test is running 128MB/s
```
## Context

The read speed is around 50-70 MB/s . Files under the bucket are something like `BucketName/MediaID/Quality/MovieName/Files`

## Your Environment

Version: 2017-05-05T01:14:51Z
Supermicro SuperStorage 6028R-E1CR12L
CPU : model name : Intel(R) Xeon(R) CPU E5-2630 v4 @ 2.20GHz
CORE : 40
Memory : 512 GB
Disks : 12 * 4TB
Ubuntu 16.04

vadmeste:
After the calculation of objects that need to be copied/removed, mc
sends tasks (upload/remove) to a parallel manager which monitors the
progress of tasks and decide to allocate more threads to it to
accelerate the overall work.

This is very useful when we copy/mirror many relatively small files.

e.g. copy of 2000 files of 1kb gets much faster (here S3 is slow because US East region is not convenient for my current place)

Without parallel upload:
```
17:25 $ time mc mirror source s3/vadmeste/
...
real	8m7.110s
user	0m4.759s
sys	0m2.531s
```

With parallel upload:
```
17:34 $ time mc mirror source s3/vadmeste/
...
real	1m41.369s
user	0m4.275s
sys	0m1.441s
```

Fixes #2196

### DIFF
--- a/cmd/accounting-reader.go
+++ b/cmd/accounting-reader.go
@@ -122,6 +122,11 @@ func (a *accounter) Set(n int64) *accounter {
 	return a
 }
 
+// Get gets current value atomically
+func (a *accounter) Get() int64 {
+	return atomic.LoadInt64(&a.current)
+}
+
 // Add add to current value atomically.
 func (a *accounter) Add(n int64) int64 {
 	return atomic.AddInt64(&a.current, n)
--- a/cmd/cp-main.go
+++ b/cmd/cp-main.go
@@ -25,7 +25,6 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"sync"
 	"syscall"
 
 	"github.com/cheggaaa/pb"
@@ -41,11 +40,6 @@ var (
 		cli.BoolFlag{
 			Name:  "recursive, r",
 			Usage: "Copy recursively.",
-		},
-		cli.UintFlag{
-			Name:  "parallel, p",
-			Usage: "Number of copies in parallel",
-			Value: 1,
 		},
 	}
 )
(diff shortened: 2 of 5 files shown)


harshavardhana:
Is this PR ready? @vadmeste also please resolve the conflicts.

vadmeste:
Ready to be reviewed.

vvaidy:
This change caught me by surprise since it broke backward compatibility when it took away the `-p <N>` flag. Is there some value in preserving the `-p` option to act as a manual override of the number of concurrent threads?

harshavardhana:
Yes it was meant to be broken @vvaidy  we added `-p` by mistake as our general intention with `mc` is to automatically handle these scenarios.

Feel free to open a new issue if you wish this to be additionally configurable as well.

harshavardhana:
@systemlevel yes it is used in both cp and mirror, but we do need to investigate the issue you have mentioned.

vadmeste:
> @systemlevel yes it is used in both cp and mirror, but we do need to investigate the issue you have mentioned.

Well, this is not rsync, mc will make a lot of web requests when there are too many small files. But I guess we need to do a concurrent upload using minio-go and check if mc is not doing any unnecessary overhead.

cc @kannappanr, anything to add ?

PR #2290, opened 2017-10-16
Fix command line docs and update complete-guide

harshavardhana:
Also re-ordered snap below binary download for windows in priority.

### DIFF
--- a/cmd/session-clear.go
+++ b/cmd/session-clear.go
@@ -44,16 +44,20 @@ var sessionClear = cli.Command{
   
 USAGE:
   {{.HelpName}} SESSION-ID|all
+
+SESSION-ID:
+  SESSION - Session can either be $SESSION-ID or all
+
 FLAGS:
   {{range .VisibleFlags}}{{.}}
   {{end}}
 EXAMPLES:
   1. Clear session.
-	 $ {{.HelpName}} ygVIpSJs
+     $ {{.HelpName}} ygVIpSJs
   2. Clear all sessions.
-	 $ {{.HelpName}} all
+     $ {{.HelpName}} all
   3. Forcefully clear an obsolete session.
-	 $ {{.HelpName}} ygVIpSJs --force
+     $ {{.HelpName}} ygVIpSJs --force
 `,
 }
 
--- a/cmd/session-resume.go
+++ b/cmd/session-resume.go
@@ -38,14 +38,17 @@ var sessionResume = cli.Command{
   {{.HelpName}} - {{.Usage}}
 
 USAGE:
-  {{.HelpName}} [SESSION-ID/all]
+  {{.HelpName}} SESSION-ID
+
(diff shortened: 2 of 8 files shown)


cmd/session-clear.go
```diff
@@ -44,16 +44,20 @@ var sessionClear = cli.Command{
   
 USAGE:
   {{.HelpName}} SESSION-ID|all
+
+SESSION-ID:
+  SESSION - Session can either be $SESSION-ID
```
krisis: this seems incomplete. Shouldn't this be `$SESSION-ID` or `all`?

cmd/session-resume.go
```diff
@@ -38,14 +38,17 @@ var sessionResume = cli.Command{
   {{.HelpName}} - {{.Usage}}
 
 USAGE:
-  {{.HelpName}} [SESSION-ID/all]
+  {{.HelpName}} SESSION-ID
+
+SESSION-ID:
```
krisis: this seems incomplete. Shouldn't this be `$SESSION-ID` or `all`?

docs/minio-client-complete-guide.md
```diff
@@ -23,23 +23,64 @@ version  Print version info.
...
+
+```sh
+chmod +x mc
+./mc --help
+```
+
+### Binary Download (Microsoft Windows)
+| Platform | Architecture | URL |
+| ---------- | -------- |------|
+|Microsoft Windows|64-bit|https://dl.minio.io/client/mc/release/windows-amd64/mc.exe |
```
krisis: why are we calling this 64-bit and not 64-bit Intel?

cmd/session-resume.go
```diff
@@ -38,14 +38,17 @@ var sessionResume = cli.Command{
...
+  SESSION - Session can either be $SESSION-ID
```
poornas: incomplete like @krisis pointed in previous instances

cmd/session-clear.go
harshavardhana: yes it should be `all`

cmd/session-resume.go
harshavardhana: No resume is not all @krisis

docs/minio-client-complete-guide.md
harshavardhana: Old baggage..

cmd/session-resume.go
harshavardhana: Not incomplete `resume` is purposefully not supporting `all`

krisis: @harshavardhana then the description shouldn't be `SESSION - Session can either be ...`, right?

harshavardhana: yes removing @krisis

harshavardhana:
can this be reviewed again @krisis @poornas ?

Issue #2255, opened 2017-09-21
add context in mc


poornas:
## Expected behaviour
cp/mirror calls should use context in Put Api call.
## Actual behaviour
Currently cp/mirror doesn't use request context in the Put Api call. With context availability in minio-go, mc client should take advantage of this and use context to cancel Put calls gracefully.
## mc version
- (paste output of `mc version`)

PR #2291, opened 2017-10-16
Add context in Put api call for cp/mirror

poornas:
Addresses issue #2255. This PR passes context in mc Put api calls so that request cancellation with context

### DIFF
--- a/cmd/common-methods.go
+++ b/cmd/common-methods.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"context"
 	"io"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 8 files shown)


cmd/common-methods.go
```diff
@@ -105,7 +106,9 @@ func putTargetStream(alias string, urlStr string, reader io.Reader, size int64,
 	if err != nil {
 		return 0, err.Trace(alias, urlStr)
 	}
-	n, err := targetClnt.Put(reader, size, metadata, progress)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
harshavardhana: We can combine this `cancel` with `trapCh` so that if we do get a signal like CTRL-C `<-trapCh` we cleanly exit.

poornas: done.

PR #2292, opened 2017-10-17
Add newS3Config helper to auto-probe signature type.

harshavardhana:


cmd/config-host-add.go
```diff
@@ -116,36 +123,72 @@ func addHost(alias string, hostCfgV8 hostConfigV8) {
...
-	parsedURL, _ := urlpkg.Parse(url)
+	s3Client, err := s3New(s3Config)
+	if err != nil {
+		return nil, err.Trace(args...)
+	}
 
-	if isGoogle(parsedURL.Host) {
-		if api == "S3v4" {
-			fatalIf(errInvalidArgument().Trace(api), "Unsupported API signature for url. Please use `mc config host add "+alias+" "+url+" "+accessKey+" "+secretKey+" S3v2` instead.")
+	for content := range s3Client.List(false, false, DirNone) {
```
kannappanr: Can we call BucketExists on a random bucketname and not worry about whether it exists or not? but concentrate more or if it returns an error or not.

harshavardhana: ListBuckets() is better because of multiple reasons.

- There is no requirement on the right error message sent by the server (BucketExists) might bring that, since `mc` needs to work on many clients they might not send the right error.
- ListBuckets() doesn't have any region requirement so doesn't involve any multiple calls underneath.

It would have been nicer if the S3 REST API had a a signature probe call, but we do not have that.  I prefer to keep it ListBuckets().

harshavardhana:
Code coverage will fail here because we are making network operations and we cannot add a unit tests, we have a functional tests here already though.

PR #2294, opened 2017-10-22
fix JSON format error in functional test

balamurugana:
Previously `error` field is not populated properly due to double
formatting error message into JSON string.  This patch fixes the issue
by avoid second time formatting in `printf` command.


balamurugana:
@nitisht 
> Some context on what exactly this PR fixes will be helpful here.

updated

PR #2318, opened 2017-11-28
mirror: Deprecate --force instead add --overwrite

### LINKED ISSUE #2317: Add new flag --force-remove to mirror command
## Expected Behavior

I have no idea why `Delete is not allowed ...` appears but it doesn't matter.

It should have option to force remove but don't overwrite files.

## Current Behavior

When running `mirror --remove` there are warnings `{Delete,Overwide} not allowed for <filename>. Use --force to override this behavior.`

Using `--force` will allow to delete files as expected but it will also overwrite all files.

## Possible Solution

Flag `--force-remove` would leave existing files would not overwrite existing files and remove extraneous files.

## Context

Mirror files from local directory to the bucket.

## mc version
- 2017-10-14T00:51:16Z

## System information
Fedora 27, Alpine 3.6, Ubuntu 17.10

Moved from https://github.com/minio/minio/issues/5144

harshavardhana:
Fixes #2317

### DIFF
--- a/cmd/mirror-main.go
+++ b/cmd/mirror-main.go
@@ -37,8 +37,13 @@ import (
 var (
 	mirrorFlags = []cli.Flag{
 		cli.BoolFlag{
-			Name:  "force",
-			Usage: "Force overwrite of an existing target(s).",
+			Name:   "force",
+			Usage:  "Force allows forced overwrite or removal of file(s) on target(s).",
+			Hidden: true, // Hidden since this option is deprecated.
+		},
+		cli.BoolFlag{
+			Name:  "overwrite",
+			Usage: "Overwrite file(s) on target(s).",
 		},
 		cli.BoolFlag{
 			Name:  "fake",
@@ -50,7 +55,7 @@ var (
 		},
 		cli.BoolFlag{
 			Name:  "remove",
-			Usage: "Remove extraneous file(s) on target.",
+			Usage: "Remove extraneous file(s) on target(s).",
 		},
 		cli.StringFlag{
 			Name:  "region",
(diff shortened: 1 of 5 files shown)


harshavardhana:
Command line change is approved by @abperiasamy already.

cmd/mirror-main.go
```diff
@@ -608,7 +620,7 @@ func runMirror(srcURL, dstURL string, ctx *cli.Context) *probe.Error {
 				}
 				// Copy policy rules from source to dest if flag is activated
 				if ctx.Bool("a") {
-					err := copyBucketPolicies(srcClt, dstClt, ctx.Bool("force"))
+					err := copyBucketPolicies(srcClt, dstClt, ctx.Bool("overwrite"))
```
vadmeste: We should use `isOverwrite` here, not `ctx.Bool("overwrite")`

cmd/mirror-url.go
```diff
@@ -141,8 +149,7 @@ func deltaSourceTarget(sourceURL string, targetURL string, isForce bool, isFake
 			}
 		case differInSecond:
 			if isRemove {
-				// todo(nl5887): I'd all force and fake checks to the the actual mirror / harvest
-				if !isForce && !isFake {
+				if !isFake {
```
vadmeste: Something weird here,

`isRemove` is set and we are throwing an error errDeleteNotAllowed if `isFake` is not enabled ?

cmd/mirror-url.go
```diff
@@ -110,21 +116,23 @@ func deltaSourceTarget(sourceURL string, targetURL string, isForce bool, isFake
...
-			targetContent := &clientContent{URL: *newClientURL(targetPath)}
-			URLsCh <- URLs{
-				SourceAlias:   sourceAlias,
-				SourceContent: sourceContent,
-				TargetAlias:   targetAlias,
-				TargetContent: targetContent,
+			if isOverwrite {
+				if !isFake {
+					// Size differs and force not set
+					URLsCh <- URLs{Error: errOverWriteNotAllowed(diffMsg.SecondURL)}
```
vadmeste: Here, `isOverwrite` is enabled, we should not throw an error here.

cmd/mirror-main.go
harshavardhana: why would we do that? the code is kept similar to before wherever `force` was used now uses `overwrite`

cmd/mirror-url.go
harshavardhana: Logic is not correct.

cmd/mirror-url.go
harshavardhana: Again logic error.

cmd/mirror-main.go
harshavardhana: i see now.

vadmeste: > why would we do that? the code is kept similar to before wherever force was used now uses overwrite

Just to have full backward compatibility, isOverwrite is just calculated above.

harshavardhana:
@vadmeste @kannappanr can you take a look?

cmd/mirror-url.go
```diff
@@ -140,33 +143,28 @@ func deltaSourceTarget(sourceURL string, targetURL string, isForce bool, isFake
...
-						Error: errDeleteNotAllowed(diffMsg.SecondURL),
-					}
-					continue
-				}
+			if !isRemove && !isFake {
+				// Object removal not allowed if --remove is not set.
 				URLsCh <- URLs{
-					TargetAlias:   targetAlias,
-					TargetContent: diffMsg.secondContent,
+					Error: errDeleteNotAllowed(diffMsg.SecondURL),
```
vadmeste: This practically means shows an error when --remove is not passed but an object exists only in target.

vadmeste: I mean we should not show an error in this case, we should continue.

harshavardhana: We should @vadmeste otherwise users will not know that they can pass `--remove` it is complicated to not allow that. 

Without this warning message users have no way to know that delete is disallowed or that we found extraneous files on destination.

harshavardhana: IMO we should throw this warning message such that we can guide the users.

vadmeste: I've just tested it.. it is already a WARNING.. not an error.

By the way, errOverWriteNotAllowed and errDeleteNotAllowed errors in typed-errors.go still have --force flag.. that's also need to be changed.

vadmeste: LGTM otherwise.

harshavardhana: Yes will change that @vadmeste

harshavardhana:
@vadmeste @kannappanr updated.

PR #2322, opened 2017-12-02
Rearrange exclude code to filter source and target URLs

### LINKED ISSUE #2319: Mirror command with --watch doesn't honor --exclude option
## Expected behaviour
mirror command with --watch should honor --exclude option

## Actual behaviour
mirror command with --watch doesn't honor --exclude option

1. Create 2 minio instance on different node, and make a test bucket on both instance
2. Config the mc host, let's say local and target
3. ./mc mirror --force --remove --watch --exclude ".*" --exclude "\*.temp" local/test target/test
4. Create another shell, execute: ./mc cp test.temp local/test
5. You can see mirror command sync the test.temp to target incorrectly
6. ./mc ls target/test/ can see the test.temp file

## mc version
Version: 2017-10-14T00:51:16Z
Release-tag: RELEASE.2017-10-14T00-51-16Z
Commit-id: 785e14a725357b39e22b74483cd202e7effa6195

## System information
Ubuntu 16.04

kannappanr:
Move exclude code handler to mirror-url and mirror-main

Fixes #2319

### DIFF
--- a/cmd/mirror-url.go
+++ b/cmd/mirror-url.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"github.com/minio/cli"
+	"github.com/minio/minio/pkg/wildcard"
 )
 
 //
@@ -72,6 +73,15 @@ func checkMirrorSyntax(ctx *cli.Context) {
 	}
 }
 
+func matchExcludeOptions(excludeOptions []string, srcSuffix string) bool {
+	for _, pattern := range excludeOptions {
+		if wildcard.Match(pattern, srcSuffix) {
+			return true
+		}
+	}
+	return false
+}
+
 func deltaSourceTarget(sourceURL, targetURL string, isFake, isOverwrite, isRemove bool, excludeOptions []string, URLsCh chan<- URLs) {
 	// source and targets are always directories
 	sourceSeparator := string(newClientURL(sourceURL).Separator)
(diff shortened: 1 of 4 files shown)


cmd/mirror-url.go
```diff
@@ -102,12 +112,25 @@ func deltaSourceTarget(sourceURL, targetURL string, isFake, isOverwrite, isRemov
...
 			continue
 		}
+
+		srcSuffix := strings.TrimPrefix(diffMsg.FirstURL, sourceURL)
+		//Skip the source object if it matches the Exclude options provided
+		if matchExcludeOptions(excludeOptions, srcSuffix) {
+			continue
+		}
+
+		tgtSuffix := strings.TrimPrefix(diffMsg.SecondURL, targetURL)
```
vadmeste: It is easy to understand: 'please exclude some files in source when mirroring'.. but it is not easy to understand 'please exclude some files in target when mirroring`.. Is there any useful use case for the latter scenario ?

kannappanr: @vadmeste , if --remove flags is passed, the files in the target will be removed. So, exclude options should be applied to both source and target.

vadmeste: > @vadmeste , if --remove flags is passed, the files in the target will be removed. So, exclude options should be applied to both source and target.

okay, then you need to update --exclude help message.

kannappanr: done

PR #2329, opened 2017-12-10
Refactor {build,new}S3Config to remove a duplicate code

### LINKED ISSUE #2314: mc --insecure flag not effective with config host add cmd
## Expected behaviour
mc `--insecure` flag should allow self signed certificates.

## Actual behaviour
I created a self signed certificate with domain - `domain.abc`. The certificate is configured with Nginx proxy, supposed to pass on the request to a Minio server on the same host. 

Minio browser is accessible via `https://domain.abc/minio` (after ignoring the security warning). When I try to set up an alias with mc, I get the below error

```sh
$ mc --insecure config host add myminio https://domain.abc minio minio123
mc: <ERROR> Unable to initialize new config from the provided credentials Get https://domain.abc/probe-bucket-sign/?location=: x509: certificate signed by unknown authority
```
- Generate self signed certificate 

```sh
$ openssl genrsa -out private.key 2048
$ openssl req -new -x509 -days 3650 -key private.key -out public.crt -subj "/C=US/ST=state/L=location/O=organization/CN=domain.abc"
```

- Add certs to dir `/etc/nginx/ssl/domain.abc/`
- Configure Nginx with below config

```sh
upstream minio_servers {
    server 127.0.0.1:9000;
}

server {
    listen 443 ssl;
    server_name domain.abc www.domain.abc;

    ssl on;
    ssl_certificate         /etc/nginx/ssl/domain.abc/public.crt;
    ssl_certificate_key     /etc/nginx/ssl/domain.abc/private.key;

    location / {
        proxy_set_header Host $http_host;
        proxy_pass       http://minio_servers;
    }
}
```
- Try adding mc alias

## mc version
- 2017-10-31T10:13:16Z

vadmeste:
S3 config struct initialization is duplicated. Refactor it.

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -18,7 +18,6 @@ package cmd
 
 import (
 	"crypto/tls"
-	"fmt"
 	"io"
 	"math/rand"
 	"os"
@@ -95,28 +94,27 @@ func splitStr(path, sep string, n int) []string {
 	return splits
 }
 
-// buildS3Config fetches config related to the specified alias
-// to create a new config structure
-func buildS3Config(alias, urlStr string, hostCfg *hostConfigV8) (*Config, *probe.Error) {
-	if hostCfg == nil {
-		return nil, probe.NewError(fmt.Errorf("The specified alias: %s not found", urlStr))
-	}
-
+// newS3Config simply creates a new Config struct using the passed
+// parameters.
+func newS3Config(urlStr string, hostCfg *hostConfigV8) *Config {
 	// We have a valid alias and hostConfig. We populate the
 	// credentials from the match found in the config file.
 	s3Config := new(Config)
 
-	s3Config.HostURL = urlStr
-	s3Config.AccessKey = hostCfg.AccessKey
-	s3Config.SecretKey = hostCfg.SecretKey
-	s3Config.Signature = hostCfg.API
 	s3Config.AppName = "mc"
 	s3Config.AppVersion = Version
 	s3Config.AppComments = []string{os.Args[0], runtime.GOOS, runtime.GOARCH}
 	s3Config.Debug = globalDebug
 	s3Config.Insecure = globalInsecure
 
-	return s3Config, nil
+	s3Config.HostURL = urlStr
+	if hostCfg != nil {
+		s3Config.AccessKey = hostCfg.AccessKey
+		s3Config.SecretKey = hostCfg.SecretKey
+		s3Config.Signature = hostCfg.API
+	}
+
+	return s3Config
 }
 
 // lineTrunc - truncates a string to the given maximum length by
--- a/cmd/client-admin.go
+++ b/cmd/client-admin.go
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"crypto/tls"
+	"fmt"
 	"hash/fnv"
 	"net"
 	"net/http"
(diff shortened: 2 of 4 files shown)


harshavardhana:
I am not sure why this refactor is needed ? @vadmeste  afaik the issue is simply not being able to set globalIsInsecure properly.

vadmeste:
@harshavardhana, there was some code duplication, this PR just removes the duplication, then it will automatically fix the bug.

harshavardhana:
> @harshavardhana, there was some code duplication, this PR just removes the duplication, then it will automatically fix the bug.

Shouldn't a refactor be done in a separate PR @vadmeste , i would ask that we fix the bug. Are you saying that refactor is required to fix this issue?

vadmeste:
> Shouldn't a refactor be done in a separate PR @vadmeste , 

I agree it is easier to send two PRs to make reviewing simpler, I will do

> Are you saying that refactor is required to fix this issue?

It is not required.. but it fixes the bug, it just removes the buggy duplicated code and use the correct code.

harshavardhana:
Is this still a WIP? also looks like there are conflicts now.

vadmeste:
No WIP anymore, please review @harshavardhana & @poornas

cmd/utils.go
```diff
@@ -126,6 +126,72 @@ func parseEnvURL(envURL string) (*url.URL, string, string, *probe.Error) {
 	return u, accessKey, secretKey, nil
 }
```
harshavardhana: Add a comment here?

cmd/utils.go
```diff
@@ -126,6 +126,72 @@ func parseEnvURL(envURL string) (*url.URL, string, string, *probe.Error) {
...
+	// Test s3 connection for API auto probe
+	s3Config.Signature = "s3v4"
+	s3Config.HostURL = urlJoinPath(url, "probe-bucket-sign")
+
+	s3Client, err := s3New(s3Config)
+	if err != nil {
+		return nil, err.Trace(args...)
+	}
+
+	if _, err = s3Client.Stat(false, false); err != nil {
```
harshavardhana: Can you convert this into like a `probeSignature`  function which returns either "s3v4" or "s3v2" along with error.?

vadmeste:
done @harshavardhana

cmd/utils.go
```diff
@@ -126,6 +126,92 @@ func parseEnvURL(envURL string) (*url.URL, string, string, *probe.Error) {
...
+				return "", err
+			}
+			if _, err = s3Client.Stat(false, false); err != nil {
+				switch err.ToGoError().(type) {
+				case BucketDoesNotExist:
+					// Bucket doesn't exist, means signature probing worked with V2.
+				default:
+					return "", err
+				}
+			}
```
poornas: In the event that "probe-bucket-sign" exists and server supports only s3v2, the api would erroneously use s3v4 in the config. Need to set to s3v2 if second Stat call doesn't return an error.

cmd/utils.go
```diff
@@ -126,6 +126,92 @@ func parseEnvURL(envURL string) (*url.URL, string, string, *probe.Error) {
 	return u, accessKey, secretKey, nil
 }
 
+// probeS3Signature - auto probe S3 server signature: issue a Stat call
+// using v4 signature then v2 in case of failure.
+func probeS3Signature(accessKey, secretKey, url string) (string, *probe.Error) {
+	// Test s3 connection for API auto probe
+	s3Config := &Config{
+		// S3 connection parameters
```
harshavardhana: may be passing the Debug as well so that `--debug` works if this fails.

cmd/utils.go
harshavardhana: why would that happen? s3Config.Signature is set to "s3v2"

poornas: you are right. missed that.

vadmeste: > In the event that "probe-bucket-sign" exists and server supports only s3v2, the api would erroneously use s3v4 in the config.

s3v2 signature is already set here: https://github.com/minio/mc/pull/2329/commits/2358b3bf6ae919e15409a2d5cd605f457fd9f7a8#diff-b4695c7b6902e27273329af6261b0c73R153

harshavardhana:
This is not needed @vadmeste i think since #2365

vadmeste:
@harshavardhana this change is still needed even after your recent merge, please review again.

Issue #2339, opened 2017-12-19
fix mc stat output 


poornas:
## Expected behaviour

1. mc stat should error for site level stat ``` mc stat s3```
2. ``` mc --r stat s3 ``` should list stat for all buckets
3. ```mc stat s3/bucket``` should show bucket in output
4. ```mc stat s3/bucket --r ``` should show bucket in output
5. ```mc stat s3/bucket/prefix/``` shows stat of objects in bucket with prefix prefix/
## Actual behaviour
mc stat command currently lists bucket info stat for buckets without -r flag, and metadata, stat info for s3/bucket/obj that actually resolves to an object with -r flag. Expected behavior here is to mimic mc ls listing.
## mc version
- (paste output of `mc version`)

PR #2340, opened 2017-12-20
stat: fix output to show bucket in stat listing

poornas:
Fixes : #2339 
Show stat of all buckets at site level, and mimic
mc ls behavior in stat output.


poornas:
@krisis , PTAL.

PR #2345, opened 2018-01-03
mirror with --watch should copy and exit for s3 endpoints

harshavardhana:
Current code wouldn't exit unless we press CTRL+C
even if the endpoints do not support --watch, this PR
fixes this behavior.

### DIFF
--- a/cmd/mirror-main.go
+++ b/cmd/mirror-main.go
@@ -130,16 +130,12 @@ type mirrorJob struct {
 	// channel for errors
 	errorCh chan *probe.Error
 
-	// Contains if watcher is currently running.
-	watcherRunning bool
-
 	// the global watcher object, which receives notifications of created
 	// and deleted files
 	watcher *Watcher
 
 	// Hold operation status information
-	status  Status
-	scanBar scanBarFunc
+	status Status
 
 	// waitgroup for status goroutine, waits till all status
 	// messages have been written and received
@@ -148,11 +144,6 @@ type mirrorJob struct {
 	queueCh chan func() URLs
 	// channel for status messages
 	statusCh chan URLs
-
-	// Store last watch error
-	watchErr *probe.Error
-	// Store last mirror error
-	mirrorErr *probe.Error
 
 	TotalObjects int64
 	TotalBytes   int64
@@ -252,8 +243,6 @@ func (mj *mirrorJob) startStatus() {
 
 		for sURLs := range mj.statusCh {
 			if sURLs.Error != nil {
-				// Save last mirror error
-				mj.mirrorErr = sURLs.Error
 				// Print in new line and adjust to top so that we
 				// don't print over the ongoing progress bar.
 				if sURLs.SourceContent != nil {
@@ -284,12 +273,12 @@ func (mj *mirrorJob) stopStatus() {
 }
 
 // this goroutine will watch for notifications, and add modified objects to the queue
-func (mj *mirrorJob) watchMirror(ctx context.Context, cancelMirror context.CancelFunc) {
-
+func (mj *mirrorJob) watchMirror(ctx context.Context, cancelMirror context.CancelFunc, errCh chan<- *probe.Error) {
 	for {
 		select {
 		case event, ok := <-mj.watcher.Events():
 			if !ok {
+				errCh <- nil
 				// channel closed
 				return
 			}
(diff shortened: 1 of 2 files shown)


cmd/parallel-manager.go
```diff
@@ -73,8 +73,8 @@ func (p *ParallelManager) addWorker() {
 	go func() {
 		for {
 			// Wait for jobs
-			fn := <-p.queueCh
-			if fn == nil {
+			fn, ok := <-p.queueCh
+			if fn == nil || !ok {
```
krisis: why are you using `fn` before checking `ok`?

cmd/mirror-main.go
```diff
@@ -541,13 +536,12 @@ func newMirrorJob(srcURL, dstURL string, isFake, isRemove, isOverwrite, isWatch
...
-		statusCh:       make(chan URLs),
-		queueCh:        make(chan func() URLs),
-		wgStatus:       new(sync.WaitGroup),
-		watcherRunning: true,
-		watcher:        NewWatcher(UTCNow()),
+		status:   status,
+		scanBar:  func(s string) {},
+		statusCh: make(chan URLs),
+		queueCh:  make(chan func() URLs),
+		wgStatus: new(sync.WaitGroup),
```
krisis: is `watcherRunning` member needed elsewhere or can it be removed from `mirrorJob`?

harshavardhana: It can be removed not needed.

cmd/parallel-manager.go
harshavardhana: I can use only one..

harshavardhana:
@vadmeste PTAL

harshavardhana:
@vadmeste can you take a look?

cmd/mirror-main.go
```diff
@@ -493,25 +488,15 @@ exit:
...
-		go mj.watchMirror(ctx, cancelMirror)
-	}
-
 	// Start mirroring.
 	mj.startMirror(ctx, cancelMirror)
 
+	// Starts watcher loop for watching for new events.
 	if mj.isWatch {
-		<-mj.trapCh
+		mj.watchMirror(ctx, cancelMirror)
```
vadmeste: We need to check the error message returned by this error, and I guess only print it as a warning.

cmd/mirror-main.go
```diff
@@ -493,25 +488,15 @@ exit:
...
-		// Enable progress bar reader only during default mode
-		mj.scanBar = scanBarFactory()
-	}
-
 	// start the status go routine
 	mj.startStatus()
 
-	// Starts additional watcher thread for watching for new events.
-	if mj.isWatch {
-		go mj.watchMirror(ctx, cancelMirror)
```
vadmeste: It is not commented in the code, but the reason why we have this in a go-routine is to avoid missing new changes in source such as removing files, adding new files.. this is very useful especially when mirroring takes some times. Any reason why you are removing this @harshavardhana ?

cmd/mirror-main.go
harshavardhana: Requires some restructuring let me do that.

cmd/mirror-main.go
harshavardhana: I know the reason @vadmeste , i generally thought it is not a big deal. Since they will be caught up later but i see now that it might be not a good idea and it is possible that we do not exit properly.

harshavardhana:
Simplified the code further @vadmeste - avoiding racy updates to provide struct members. Remove `watchErr`, `mirrorErr`

cmd/mirror-main.go
```diff
@@ -407,15 +396,17 @@ func (mj *mirrorJob) watchMirror(ctx context.Context, cancelMirror context.Cance
...
 		case err := <-mj.watcher.Errors():
-			mj.watchErr = err
 			switch err.ToGoError().(type) {
 			case APINotImplemented:
-				// Ignore error if API is not implemented.
-				mj.watcherRunning = false
-				return
-			default:
-				errorIf(err, "Unexpected error during monitoring.")
+				errorIf(err.Trace(), "Watch is not implemented ignoring.")
```
vadmeste: I think this message is better, 'Unable to watch source path'.

harshavardhana:
fixed @vadmeste PTAL

PR #2347, opened 2018-01-06
Copy metadata during mirror/cp operation

poornas:
Fixes https://github.com/minio/minio/issues/5324
Fixes https://github.com/minio/minio-go/issues/885

metadata is not being copied during mirror and copy operation. Turn this back on,
while also filtering out custom xattr metadata set on file system that do not 
respect http header convention. 

minio-go converts this metadata to http headers by prepending the key with x-amz-meta. file system xattributes which will fail http header convention need to be filtered out.


harshavardhana:
Its not just mirror this would enable for cp as well.

kannappanr:
@poornas Travis and Appveyor failed, PTAL

poornas:
fixed build failures - Can be reviewed.

poornas:
@donatello, PTAL.

PR #2348, opened 2018-01-06
fix missing metadata in stat cmd

poornas:
mc stat without --r is missing metadata for stat on objectname.

### DIFF
--- a/cmd/stat.go
+++ b/cmd/stat.go
@@ -144,9 +144,6 @@ func doStat(clnt Client, isRecursive bool, targetAlias, targetURL string) error 
 			continue
 		}
 		url := targetAlias + getKey(content)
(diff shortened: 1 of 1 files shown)


cmd/stat.go
```diff
@@ -144,9 +144,6 @@ func doStat(clnt Client, isRecursive bool, targetAlias, targetURL string) error
 			continue
 		}
 		url := targetAlias + getKey(content)
-		if !isAliasURLDir(targetURL) {
-			url = content.URL.Path
-		}
```
krishnasrinivas: `doStat(clnt Client, isRecursive bool, targetAlias, targetURL string)` looks like targetURL is not used ... you can remove this as well.

poornas: done.

PR #2350, opened 2018-01-09
New Admin API client-side implementation

donatello:
This is a companion PR to minio/minio#5351 - please refer to it for the context. It includes client side changes for using the new Admin APIs.

### DIFF
--- a/cmd/admin-heal-result-item.go
+++ b/cmd/admin-heal-result-item.go
@@ -1,0 +1,117 @@
+/*
+ * Minio Client (C) 2018 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/minio/minio/pkg/madmin"
+)
+
+type hri struct {
+	*madmin.HealResultItem
+}
+
+func newHRI(i *madmin.HealResultItem) *hri {
+	return &hri{i}
+}
+
+// getObjectHCCChange - returns before and after color change for
+// objects
+func (h hri) getObjectHCCChange() (b, a hCol, err error) {
+	parityShards := h.ParityBlocks
+	dataShards := h.DataBlocks
+
+	onlineBefore, onlineAfter := h.GetOnlineCounts()
+	surplusShardsBeforeHeal := onlineBefore - dataShards
+	surplusShardsAfterHeal := onlineAfter - dataShards
+
+	b, err = getHColCode(surplusShardsBeforeHeal, parityShards)
+	if err != nil {
+		return
+	}
+	a, err = getHColCode(surplusShardsAfterHeal, parityShards)
+	return
+
+}
+
+// getReplicatedFileHCCChange - fetches health color code for metadata
+// files that are replicated.
+func (h hri) getReplicatedFileHCCChange() (b, a hCol, err error) {
+	getColCode := func(numAvail int) (c hCol, err error) {
+		// calculate color code for replicated object similar
+		// to erasure coded objects
+		quorum := h.DiskCount/2 + 1
+		surplus := numAvail - quorum
+		parity := h.DiskCount - quorum
+		c, err = getHColCode(surplus, parity)
+		return
+	}
+
+	onlineBefore, onlineAfter := h.GetOnlineCounts()
+	b, err = getColCode(onlineBefore)
+	if err != nil {
+		return
+	}
+	a, err = getColCode(onlineAfter)
+	return
+}
+
+func (h hri) makeHealEntityString() string {
+	switch h.Type {
+	case madmin.HealItemObject:
+		return h.Bucket + "/" + h.Object
+	case madmin.HealItemBucket:
+		return h.Bucket
+	case madmin.HealItemMetadata:
+		return "[disk-format]"
+	case madmin.HealItemBucketMetadata:
+		return fmt.Sprintf("[bucket-metadata]%s/%s", h.Bucket, h.Object)
+	}
+	return "** unexpected **"
+}
+
+func (h hri) getHRTypeAndName() (typ, name string) {
+	name = fmt.Sprintf("%s/%s", h.Bucket, h.Object)
+	switch h.Type {
+	case madmin.HealItemMetadata:
+		typ = "system"
+		name = h.Detail
+	case madmin.HealItemBucketMetadata:
+		typ = "system"
+		name = "bucket-metadata:" + name
+	case madmin.HealItemBucket:
+		typ = "bucket"
+	case madmin.HealItemObject:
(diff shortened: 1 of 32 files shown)


cmd/admin-heal.go
```diff
@@ -19,37 +19,41 @@ package cmd
...
+		Name:  "dry-run, n",
+		Usage: "Only inspect data, but do not mutate",
+	},
+	cli.BoolFlag{
+		Name:  "incomplete, I",
+		Usage: "Heal uploads recursively",
+	},
+	cli.BoolFlag{
+		Name:  "force-start, f",
+		Usage: "Force start a heal sequence",
```
vadmeste: I think this is better 'Force start a new heal sequence'

cmd/admin-heal.go
```diff
@@ -72,304 +76,556 @@ EXAMPLES:
...
+	colRow := hColTable[parityShards]
+	for index, val := range colRow {
+		if val != -1 && surplusShards <= val {
+			return hColOrder[index], err
+		}
+	}
+	return c, fmt.Errorf("this will not be returned")
+}
+
+type hri struct {
```
vadmeste: Any reason why we declare this struct ? wouldn't be enough if we use madmin.HealResultItem ?

vadmeste: Can you comment this struct ?

cmd/admin-heal.go
```diff
@@ -72,304 +76,556 @@ EXAMPLES:
...
+	if surplusShards < 0 {
+		return hColGrey, err
 	}
+	colRow := hColTable[parityShards]
+	for index, val := range colRow {
+		if val != -1 && surplusShards <= val {
+			return hColOrder[index], err
+		}
+	}
+	return c, fmt.Errorf("this will not be returned")
```
vadmeste: I think this is a better error msg: 'cannot get a heal color code'

donatello:
> We need to provide proper errors when run against an FS mode, this error is not useful.
> 
> ```
> $ mc admin heal myminio
> mc: <ERROR> Failed to start heal sequence A header you provided implies functionality that is not implemented
> mc: <ERROR> Unable to display follow heal status A header you provided implies functionality that is not implemented
> 
> ```

The above is being addressed by adding a new server error via https://github.com/minio/minio/pull/5383

> 
> Are we still supporting?
> ```
> 
>   4. Heal all objects including uploads under 'dir' prefix
>        $ mc admin heal --incomplete --recursive play/testbucket/dir/
> 
> ```

Yes, adding `--incomplete` flag heals incomplete multipart uploads too.

> ```
> $ mc admin info myminio
> ●  localhost:9000
>    Uptime : online since 6 minutes ago
>   Version : 2018-01-10T17:54:10Z
>    Region : 
>  SQS ARNs : arn:minio:sqs::test2:redis arn:minio:sqs::test1:redis 
>   Network : Incoming 10KiB, Outgoing 7.8KiB
>   Storage : Total 156GiB, Free 75GiB
> 
> ```
> localhost:9000 is wrong in this.

It is the current behaviour (i.e. same without this PR). What should it be if not localhost:9000? If it is not related to this PR, I can fix it in existing mc/server.

> Bucket is not valid is perhaps not the correct message here, we need to guide the users here.
> 
> ```
> mc admin lock list myminio
> mc: <ERROR> Cannot get lock status. The specified bucket is not valid.
> 
> ```

The problem here is that the bucket is not specified. I will send a separate PR adding validation to fix this.

cmd/admin-heal.go
```diff
@@ -72,304 +76,556 @@ EXAMPLES:
...
+	colRow := hColTable[parityShards]
+	for index, val := range colRow {
+		if val != -1 && surplusShards <= val {
+			return hColOrder[index], err
+		}
+	}
+	return c, fmt.Errorf("cannot get a heal color code")
+}
+
+type hri struct {
```
vadmeste: Can you move HRI related code to a new source file ?

cmd/admin-heal.go
```diff
@@ -72,304 +76,556 @@ EXAMPLES:
...
-	case madmin.CanHeal:
-		msg += "can be healed"
-	case madmin.Corrupted:
-		msg += "cannot be healed"
-	case madmin.QuorumUnavailable:
-		msg += "quorum not available for healing"
-	}
-	msg += ".\n"
-	return console.Colorize("Heal", msg)
+type uiData struct {
```
vadmeste: Can you move uiData code to a new source file ?

donatello: Done.

cmd/admin-heal.go
donatello: Done.

cmd/admin-heal.go
```diff
@@ -19,37 +19,41 @@ package cmd
...
+	cli.BoolFlag{
+		Name:  "recursive, r",
+		Usage: "Heal recursively",
+	},
+	cli.BoolFlag{
+		Name:  "dry-run, n",
+		Usage: "Only inspect data, but do not mutate",
+	},
+	cli.BoolFlag{
+		Name:  "incomplete, I",
```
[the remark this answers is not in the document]
donatello: Done.

cmd/admin-service-stop.go
```diff
@@ -0,0 +1,99 @@
+/*
+ * Minio Client (C) 2017 Minio, Inc.
```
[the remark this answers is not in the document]
donatello: Done.

vadmeste:
@donatello, I think you can still move getHPrintCol, getHColCode and lineTrunc to another file, the one or more which depend on these functions.

cmd/admin-heal.go
```diff
@@ -29,23 +27,24 @@ import (
...
-		},
-	}
-)
+var adminHealFlags = []cli.Flag{
+	cli.BoolFlag{
+		Name:  "recursive, r",
+		Usage: "Heal recursively",
+	},
+	cli.BoolFlag{
+		Name:  "dry-run, n",
```
vadmeste: Examples are not updated, --fake is still being used.

vadmeste:
```
18:08 $ mc admin heal myminio/testbucket/1mb.bin
 ◓  testbucket/1mb.bin
    0/1 objects; 1024 KiB in 1s
    ┌────────┬───┬─────────────────────┐
    │ Green  │ 3 │ 100.0% ████████████ │
    │ Yellow │ 0 │   0.0%              │
    │ Red    │ 0 │   0.0%              │
    │ Grey   │ 0 │   0.0%              │
    └────────┴───┴─────────────────────┘
```

I am not sure I understand exactly '0/1 objects', or it has to be '1/1 objects' ?

cmd/admin-heal.go
donatello: Fixed.

cmd/admin-heal-result-item.go
```diff
@@ -0,0 +1,117 @@
...
+		surplus := numAvail - quorum
+		parity := h.DiskCount - quorum
+		c, err = getHColCode(surplus, parity)
+		return
+	}
+
+	onlineBefore, onlineAfter := h.GetOnlineCounts()
+	b, err = getColCode(onlineBefore)
+	if err != nil {
+		return
```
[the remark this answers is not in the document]
donatello: These are system metadata files that are healed such as format.json, and bucket metadata files such as policy.json, notifications.xml, etc.

donatello:
@vadmeste 
>  I think you can still move getHPrintCol, getHColCode and lineTrunc to another file, the one or more which depend on these functions.

Done.

> I am not sure I understand exactly '0/1 objects', or it has to be '1/1 objects' ?

The phrase "1/3 objects" implies that 1 object was healed out of 3 objects scanned. See the `-q` and `--json` options too.

donatello:
This PR is building again.

PR #2353, opened 2018-01-10
Improve lock error message when bucket is not specified

donatello:
Existing output:

```
$ /tmp/mc admin lock list myminio
mc: <ERROR> Cannot get lock status. The specified bucket is not valid.
$ /tmp/mc admin lock clear --force play
mc: <ERROR> Cannot clear the specified locks. The specified method is not allowed against this resource.

```

They are errors because no bucket is specified.

Modified output (due to this PR):
```
$ mc admin lock list play
mc: <ERROR> Cannot query locks. This operation requires a bucket to be specified.
$ mc admin lock clear --force play
mc: <ERROR> Cannot clear locks. This operation requires a bucket to be specified.
```

The problem error message was reported [here](https://github.com/minio/mc/pull/2350#pullrequestreview-87930846).

### DIFF
--- a/cmd/admin-lock-clear.go
+++ b/cmd/admin-lock-clear.go
@@ -105,6 +105,13 @@ func checkAdminLockClearSyntax(ctx *cli.Context) {
 		cli.ShowCommandHelpAndExit(ctx, "clear", 1)
 	}
 
+	// Check if a bucket is specified.
+	aliasedURL := filepath.ToSlash(ctx.Args().Get(0))
+	splits := splitStr(aliasedURL, "/", 3)
+	if splits[1] == "" {
+		fatalIf(errBucketNotSpecified().Trace(aliasedURL), "Cannot clear locks.")
+	}
+
 	if isForce := ctx.Bool("force"); isForce {
 		return
 	}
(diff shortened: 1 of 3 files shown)


cmd/admin-lock-clear.go
```diff
@@ -135,6 +135,10 @@ func mainAdminLockClear(ctx *cli.Context) error {
 
 	splits := splitStr(aliasedURL, "/", 3)
 
+	if splits[1] == "" {
+		fatalIf(errBucketNotSpecified().Trace(aliasedURL), "Cannot clear locks.")
```
harshavardhana: Perhaps you can add this in `checkSyntax` for this command

cmd/admin-lock-list.go
```diff
@@ -122,6 +122,10 @@ func mainAdminLockList(ctx *cli.Context) error {
 
 	splits := splitStr(aliasedURL, "/", 3)
 
+	if splits[1] == "" {
+		fatalIf(errBucketNotSpecified().Trace(aliasedURL), "Cannot query locks.")
```
harshavardhana: Perhaps you can add this in `checkSyntax` for this command

donatello:
@harshavardhana Done.

donatello:
@krishnasrinivas ping for review.

PR #2372, opened 2018-01-27
Large bucket support admin API Changes

harshavardhana:


harshavardhana:
There is one more pending change related to `mc admin info` which is currently WIP, once done will remove WIP

harshavardhana:
updated the pr with `mc admin info` changes as well.

harshavardhana:
Removed the support `admin info` UI changes, after discussing with @abperiasamy we will bring changes to `admin info` and other command line changes in a separate PR.

harshavardhana:
PTAL @vadmeste @donatello

vadmeste:
LGTM.. However @harshavardhana  can you rebase large bucket PR in Minio server so we can fully test this PR ?

PR #2374, opened 2018-01-30
Fixes mc diff's inconsistent behavior

### LINKED ISSUE #2349: One case of wrong diff calcuation
## Expected behaviour

`mc diff src/ target/` and `mc diff target/ src/` should have the same calculation result (with little direction difference)

## Actual behaviour

I have one case where `mc diff src/ target/` doesn't show any output and `mc diff target/ src/` shows some diff.

```
$ mkdir dir/; touch dir/file
$ mc mb s3/mybucket
$ mc mirror dir/ s3/mybucket/
$ mc diff dir/ s3/mybucket/
(no output => no diff)
$ mc diff s3/mybucket/ dir/
! dir/file
```

## mc version
Master version

## System information
All

ebozduman:
Fixes #2349

### DIFF
--- a/cmd/mirror-url.go
+++ b/cmd/mirror-url.go
@@ -136,7 +136,9 @@ func deltaSourceTarget(sourceURL, targetURL string, isFake, isOverwrite, isRemov
 			// No difference, continue.
 		case differInType:
 			URLsCh <- URLs{Error: errInvalidTarget(diffMsg.SecondURL)}
-		case differInSize, differInSrcTime, differInTgtTime:
+		case differInTgtTime:
+			// do nothing
+		case differInSize, differInSrcTime:
 			if !isOverwrite && !isFake {
 				// Size or time differs but --overwrite not set.
 				URLsCh <- URLs{Error: errOverWriteNotAllowed(diffMsg.SecondURL)}
--- a/cmd/mirror-url.go
+++ b/cmd/mirror-url.go
@@ -136,7 +136,7 @@ func deltaSourceTarget(sourceURL, targetURL string, isFake, isOverwrite, isRemov
 			// No difference, continue.
 		case differInType:
 			URLsCh <- URLs{Error: errInvalidTarget(diffMsg.SecondURL)}
-		case differInSize, differInTime:
+		case differInSize, differInSrcTime, differInTgtTime:
 			if !isOverwrite && !isFake {
 				// Size or time differs but --overwrite not set.
 				URLsCh <- URLs{Error: errOverWriteNotAllowed(diffMsg.SecondURL)}
--- a/cmd/diff-main.go
+++ b/cmd/diff-main.go
@@ -55,7 +55,7 @@ DESCRIPTION:
   Differences between source and destination are represented by mark notations with following meaning:
     > - indicates file should be copied.
     < - indicates file should be deleted.
-    ! - indicates file differs in size or type.
+    ! - indicates file differs in size or type (yellow), or modification time (cyan).
 
 EXAMPLES:
   1. Compare a local folder with a folder on Amazon S3 cloud storage.
@@ -90,7 +90,9 @@ func (d diffMessage) String() string {
 		msg = console.Colorize("DiffType", "! "+d.SecondURL)
 	case differInSize:
 		msg = console.Colorize("DiffSize", "! "+d.SecondURL)
-	case differInTime:
+	case differInSrcTime:
+		msg = console.Colorize("DiffTime", "! "+d.FirstURL)
+	case differInTgtTime:
 		msg = console.Colorize("DiffTime", "! "+d.SecondURL)
 	default:
 		fatalIf(errDummy().Trace(d.FirstURL, d.SecondURL),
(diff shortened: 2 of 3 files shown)


cmd/mirror-url.go
```diff
@@ -136,7 +136,7 @@ func deltaSourceTarget(sourceURL, targetURL string, isFake, isOverwrite, isRemov
 			// No difference, continue.
 		case differInType:
 			URLsCh <- URLs{Error: errInvalidTarget(diffMsg.SecondURL)}
-		case differInSize, differInTime:
+		case differInSize, differInSrcTime, differInTgtTime:
```
harshavardhana: We only need to handle differInSrcTime - we simply ignore differInTgtTime.

ebozduman: We need to take care of `differInTgtTime` for the case when `tgtTime` is newer than `srcTime`.
See below code, between line numbers 204~213 in difference.go file:

```
:
:
} else if tgtTime.After(srcTime) {
	// Regular files differing in timestamp.
	// Target timestamp is newer.
	diffCh <- diffMessage{
		FirstURL:      srcCtnt.URL.String(),
		SecondURL:     tgtCtnt.URL.String(),
		Diff:          differInTgtTime,
		firstContent:  srcCtnt,
		secondContent: tgtCtnt,
	}
}
:
:
```

Otherwise, we fall into `case default:` where we see the following error message:

```
$ ./mc mirror myminio/dir/ myminio/mybucketnew/
 0 B / ? ┃░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░▓┃
srcTime 2018-01-30 21:41:00.906 +0000 UTC
tgtTime 2018-01-30 22:02:36.11 +0000 UTC
mc: <ERROR> Unable to prepare URL for copying. Unrecognized diffType: time provided.
```

harshavardhana: We don't need to do that target is assumed to be always same as source, if target is modified we shouldn't do anything about it @ebozduman  and it is treated as unsupported. 

NOTE: `mc mirror` is uni-directional i.e all decisions made to copy data are done only source to target. What we need to do is simply ignore the case of `differInTgtTime`  not even go to `default` case.

vadmeste:
@ebozduman can you elaborate more how this PR fixes the issue ?

AFAIK, comparing mod time is not a good idea.. because mod time will be always different.

For example, if you run mirror at 10:00:00, the mirrored object can have 10:00:01 modtime when it reaches the server.

harshavardhana:
@vadmeste it is assumed that target server modTime is newer than source since that is the direction of our copy operation. It is implemented in accordance with AWS CLI. On the other hand `mc diff` will always show target to be newer as expected.

ebozduman:
That's right @vadmeste; After running `mc mirror` command, the modification times will always be different, since target file modification time will be newer. 

`mc mirror` command depends on `mc diff` command's output to detect modification time difference. That's why we have to have the modification time check in `mc diff` command's implementation.
`mc diff` reports modification time difference when source file is newer than the target file, and `mc mirror` command goes ahead and updates the target file. 

In the opposite case, `mc diff` command stays quite, which is the reason for inconsistency, and the reason for this issue. In other words :

`mc diff sourceFile targetFile` output is not the same with `mc diff targetFile sourceFile` output.

With the newly introduced fix, `mc diff` command reports the same difference in both cases and hence have a consistent output. However, when target file is newer than the source file, `--overwrite` option is requested, instead of mirroring right away like in the other case.

cmd/mirror-url.go
ebozduman: How about if user still would like to overwrite when target file is newer than the source file?

ebozduman: I talked to @harshavardhana .
`"mc mirror"` is not going to do anything when source file modification time is behind target file modification time, which should never happen and if it does, it is all user's responsibility how to handle it. This is exactly the same behavior with s3 sync command.
"mc mirror" output in this case will be like:
```
$ ./mc mirror myminio/mybucket myminio/dir --overwrite
 0 B / ? ┃░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░▓┃ 0s
```

vadmeste:
@ebozduman if I understood well, this PR doesn't change functionality, only code enhancement, right ?

ebozduman:
Yes @vadmeste,

It doesn't change functionality. It just fixes the inconsistent output, `mc diff` returns when the order of the command's arguments is changed.

Since, `mc mirror` depends on `mc diff`'s output and it decides whether to copy objects based on `mc diff`'s output, there is also an adaptation change on the `mc mirror` side.

vadmeste:
@ebozduman, I am still confused, let's talk privately whenever you have time.

ebozduman:
@vadmeste,

@abperiasamy, @harshavardhana and I've discussed the subject and it turned out to be your concerns are valid and the problem you filed the issue for is actually by design. So, there is nothing to be fixed and the this change/PR needs to be reverted back.

The reasoning/explanation is that `mc diff` is expected to work exactly the same way as `mc mirror`, hence `mc diff` not reporting anything when target object's modification time is ahead of the source object is correct as there won't be any mirroring in this scenario.

I'll revert back this PR and close the issue with the label `"working as intended"`

vadmeste:
@ebozduman yeah makes sense, thanks!

ebozduman:
Closing it since everything works as required. The behavior is by design.

PR #2377, opened 2018-02-02
Mirror watch removes tgt content

### LINKED ISSUE #2373: mc mirror --remove --watch delete all target instance's file when it can't monitor source instance
## Expected behaviour
mc mirror --remove --watch should do nothing when it can't monitor source instance

## Actual behaviour
mc mirror --remove --watch delete all target instance's file when it can't monitor source instance

1. create 2 standalone minio instance(source and target), and create test bucket
2. on source node, execute: `mc mirror --remove --watch source/test target/test`
3. cp some files to source, mirror working properly
4. do something(I don't know how) and restart mc(with the same config), mc pop the following error(3 times):
```
{"status":"error","error":{"message":"Failed to start monitoring.","cause":{"message":"Get http://127.0.0.1:9000/: dial tcp 127.0.0.1:9000: connect: cannot assign requested address","error":{"Op":"Get","URL":"http://127.0.0.1:9000/","Err":{"Op":"dial","Net":"tcp","Source":null,"Addr":{"IP":"127.0.0.1","Port":9000,"Zone":""},"Err":{"Syscall":"connect","Err":99}}}},"type":"fatal","sysinfo":{"host.arch":"amd64","host.cpus":"8","host.lang":"go1.9.2","host.name":"172-168-0-16","host.os":"linux","mem.heap.total":"2.4MB","mem.heap.used":"1.4MB","mem.total":"4.4MB","mem.used":"1.4MB"}}}
```
5. mc starts to delete all the target files
## mc version
- (Version: 2017-12-12T01:08:02Z
Release-tag: RELEASE.2017-12-12T01-08-02Z
Commit-id: 4cb6e3534286819d3e11db7c3659164655020bd3)

ebozduman:
Fixes #2373 

When there is a network connection error on the server side where source files are, we stop `mc mirror` command with connection error. If `mc mirror` is restarted with `--remove` and `--watch` options, the list returned from the source server turns out as empty, and `mc mirror` thinks the existing files in target bucket need to be removed and cleans them up. 
This PR stops the `mc mirror` process, when the source server is unresponsive or when there is network connection problems, like `connection refused`.

### DIFF
--- a/cmd/mirror-url.go
+++ b/cmd/mirror-url.go
@@ -64,9 +64,9 @@ func checkMirrorSyntax(ctx *cli.Context) {
 		fatalIf(errInvalidArgument().Trace(), "Invalid target arguments to mirror command.")
 	}
 
-	url := newClientURL(tgtURL)
-	if url.Host != "" {
-		if url.Path == string(url.Separator) {
+	clientURL := newClientURL(tgtURL)
+	if clientURL.Host != "" {
+		if clientURL.Path == string(clientURL.Separator) {
 			fatalIf(errInvalidArgument().Trace(tgtURL),
 				fmt.Sprintf("Target `%s` does not contain bucket name.", tgtURL))
 		}
@@ -114,8 +114,8 @@ func deltaSourceTarget(sourceURL, targetURL string, isFake, isOverwrite, isRemov
 	// List both source and target, compare and return values through channel.
 	for diffMsg := range objectDifference(sourceClnt, targetClnt, sourceURL, targetURL) {
 		if diffMsg.Error != nil {
-			errorIf(diffMsg.Error, "Unable to mirror objects.")
-			// Ignore error and proceed to next object.
+			// Send all errors through the channel
+			URLsCh <- URLs{Error: diffMsg.Error}
 			continue
 		}
 
--- a/cmd/mirror-main.go
+++ b/cmd/mirror-main.go
@@ -460,11 +460,7 @@ func (mj *mirrorJob) startMirror(ctx context.Context, cancelMirror context.Cance
 				return
 			}
 			if sURLs.Error != nil {
-				if strings.Contains(sURLs.Error.ToGoError().Error(), " is a folder.") {
-					mj.status.errorIf(sURLs.Error.Trace(), "Folder cannot be copied. Please use `...` suffix.")
-				} else {
-					mj.status.errorIf(sURLs.Error.Trace(), "Unable to prepare URL for copying.")
-				}
+				errCh <- sURLs.Error
 				continue
 			}
 
@@ -526,20 +522,35 @@ func (mj *mirrorJob) mirror(ctx context.Context, cancelMirror context.CancelFunc
 	// Start mirroring.
 	go mj.startMirror(ctx, cancelMirror, mirrorErrCh)
 
-	if err := <-mirrorErrCh; err != nil {
-		return err.Trace()
-	}
+	// Wait until all progress bar updates are actually shown and quit.
+	defer mj.stopStatus()
 
-	if mj.isWatch {
-		if err := <-watchErrCh; err != nil {
+	var err *probe.Error
+	var ok bool
+	for {
+		select {
+		case err, ok = <-mirrorErrCh:
+			if !ok || err == nil {
+				if mj.isWatch {
+					continue
+				}
+				return nil
+			}
+		case err, ok = <-watchErrCh:
+			if !ok || err == nil {
+				return nil
+			}
+		}
+		if err != nil {
+			switch err.ToGoError().(type) {
+			case BrokenSymlink, TooManyLevelsSymlink, PathNotFound,
+				PathInsufficientPermission, ObjectOnGlacier:
+				continue
+			}
 			return err.Trace()
 		}
+
 	}
-
-	// Wait until all progress bar updates are actually shown and quit.
-	mj.stopStatus()
-
-	return nil
 }
 
 func newMirrorJob(srcURL, dstURL string, isFake, isRemove, isOverwrite, isWatch bool, excludeOptions []string, olderThan, newerThan int) *mirrorJob {


ebozduman:
I need to clean up this PR since it also included changes from another PR which is still active.

cmd/mirror-url.go
```diff
@@ -113,9 +114,15 @@ func deltaSourceTarget(sourceURL, targetURL string, isFake, isOverwrite, isRemov
 
 	// List both source and target, compare and return values through channel.
 	for diffMsg := range objectDifference(sourceClnt, targetClnt, sourceURL, targetURL) {
+
 		if diffMsg.Error != nil {
+			// End the process if connection is refused,
+			// otherwise ignore the error and proceed to next object.
+			if _, ok := diffMsg.Error.ToGoError().(*url.Error); ok {
```
vadmeste: Good catch @ebozduman !

However, I don't imagine why we are only erroring out network connection issues and why we *were* ignoring errors in the first place.

objectDifference() is listing objects in source & target and then calculating the difference, I don't see any possible errors that can be ignored in this case.

I think we should return for all errors and fix later if users eventually complain.

ebozduman: I think that is about how watch logic works with `mc mirror`.

The requirement/design is such that we don't stop when there is an error and keep continue with the rest of the source file list until the list is exhausted. Like, if there are permission problems associated with the source or target files, and in error conditions like `TooManySymLinks`, `PathNotFound`, etc. we want to continue to complete going through the full source list. 
However, there are fatal errors we should not ignore.
So, when there is a fatal error like network connection problem with the source server, we kept going instead of stopping the mirroring process. Any other error condition, in which we can communicate with both source and target, we should keep continue.

poornas:
@ebozduman , there is a travis failure that needs to be fixed.

ebozduman:
Fixed Travis error and squashed the commits.

cmd/mirror-url.go
vadmeste: yeah, you are right, I forgot about these FS errors in listing.

But also, S3 listing can return some errors. S3 returns list of objects 1000 by 1000, if S3 decides to return 500 status code or SlowDown, etc.. then we will have the same problem, and objects can be deleted without the user's consensus.

I really recommend that we do the reverse way, we decide (not necessarily now) which errors we can safely ignore and return all other errors. What do you think @ebozduman ?

ebozduman: Sure, that might be possible.
Let me also discuss this with @harshavardhana

ebozduman: @vadmeste,

I think what you suggest depends on which error condition list, ignore list or error out list, we can come up fully with. In other words, we need to know the exact list of errors we'll either ignore and remove files or display and halt the action when hit. 

After discussing with @harshavardhana this morning about how we can decide to error out when `remove` action is requested, he came up with a different suggestion, which I'll share with you to get your opinion and then also discuss with @abperiasamy.

Coming out with a **full** list of error messages in both cases is not easy, so it is most likely that the file list we choose to come up with is not going to be complete. Besides, the after affect for `copy`/`overwrite`, and `remove` transactions are drastically different, when we hit an error condition, and the returned file list is erratic. That is; if we make the wrong decision and remove some files from the target by mistake, the target file/data is lost, while when some files are NOT copied over or overwritten, we'll still be fine.

With that in mind, @harshavardhana suggested to always stop when there is any kind of errors regardless of its type with the claim that the produced file list cannot be trusted when there is an error and hence we should not proceed with any action. Since we'll display the error message in this scenario, user will have an idea about the reason why mirroring did not happen. 
With this design change, we'll be safe in `remove` since no action has been taken.  With `copy/overwrite`, mirroring will just be delayed.

Let me know what you think about the above suggestion, that is; if we always error out instead of trying to make intelligent decisions to be able to selectively choose among error conditions.
Please also let me know if there is a need for more clarification.

vadmeste: @ebozduman yeah I agree with this solution as a first step. If someday, some users complain and say that mc  should ignore some of them we can take a look again.

vadmeste: I expect that we would be to construct a list of errors that we can safely ignore. But let's start by making mirroring stops for every error.

harshavardhana: Actually my mistake we need to bring in errors that we need to ignore. Because some objects can be on `Glacier` and we need to skip them.

ebozduman: @harshavardhana, @vadmeste,
We currently ignore the following errors:
1. BrokenSymlink, 
2. TooManyLevelsSymlink, 
3. PathNotFound,
4. PathInsufficientPermission, 
5. ObjectOnGlacier

We stop on all other errors.

I've tested this fix in the following scenarios:
1. CTRL-C'ed source server and saw the following error message. Then restarted `mc mirror --remove --watch` and confirmed that the target bucket content is not removed as  it was happening before this fix.
2. I've also tested the fix by having a file with no permissions and mirroring did not stop when it hit the permission problem.

If you think there are more errors we need to be aware of and ignore, and if you can think of other tests that can be run to verify the fix, please let me know.

cmd/mirror-url.go
```diff
@@ -113,10 +113,11 @@ func deltaSourceTarget(sourceURL, targetURL string, isFake, isOverwrite, isRemov
...
 	// List both source and target, compare and return values through channel.
 	for diffMsg := range objectDifference(sourceClnt, targetClnt, sourceURL, targetURL) {
+
 		if diffMsg.Error != nil {
-			errorIf(diffMsg.Error, "Unable to mirror objects.")
-			// Ignore error and proceed to next object.
-			continue
+			// Send all errors through the channel
+			URLsCh <- URLs{Error: diffMsg.Error}
+			return
```
vadmeste: @ebozduman if we are returning here, it means we are not doing any objectDifference() anymore. So if one of these errors occur `BrokenSymlink, TooManyLevelsSymlink, PathNotFound, PathInsufficientPermission, ObjectOnGlacier`.. the code is not ignoring errors, it just stops without showing any error to the user.

What do you think ?

poornas: agree with @vadmeste - shouldn't this be a continue instead of return statement?

ebozduman: Will replace `return` with `continue`.

ebozduman: Rebased.

PR #2379, opened 2018-02-03
Sort help commands alphabetically

### LINKED ISSUE #2378: Output of --help is not alphabetically sorted
## Expected behaviour
Output of --help is alphabetically sorted

## Actual behaviour
Output of --help is not alphabetically sorted

mc --help

## mc version
Release-tag: RELEASE.2018-01-18T21-18-56Z
Commit-id: d9663e5d96d5f32725867167de2a0403704ae537

## System information
Ubuntu 16.04 x64

ricardoseriani:
Closes #2378

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -35,24 +35,24 @@ var globalFlags = []cli.Flag{
 		Usage: "Path to configuration folder.",
 	},
 	cli.BoolFlag{
-		Name:  "quiet, q",
-		Usage: "Disable progress bar display.",
-	},
-	cli.BoolFlag{
-		Name:  "no-color",
-		Usage: "Disable color theme.",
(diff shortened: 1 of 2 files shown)


harshavardhana:
There is good reason to not provide sorted list of commands and it's intentional. Because to convey the priority/importance of commands based on their likely usage.

Sorting changes this behavior and avoids any control over which command needs to be projected to provide first class attention to the user of `mc` to avoid overwhelming and soul searching with a list of lot of commands.

This PR as such is not needed since we do intentionally know why we didn't sort them in the place.

kannappanr:
@ricardoseriani Thanks for sending us a PR. But, the order of commands is already displayed the way we want it to be.

PR #2383, opened 2018-02-06
Revises 'mc diff' command help menu

### LINKED ISSUE #2382: mc diff command help menu needs to be revised
## mc version

ebozduman:
Fixes #2382

### DIFF
--- a/cmd/diff-main.go
+++ b/cmd/diff-main.go
@@ -35,7 +35,7 @@ var (
 // Compute differences between two files or folders.
 var diffCmd = cli.Command{
 	Name:   "diff",
-	Usage:  "List objects with size difference or missing between two folders or buckets.",
+	Usage:  "List differences in object name, size, and date between folders.",
 	Action: mainDiff,
 	Before: setGlobalsFromContext,
 	Flags:  append(diffFlags, globalFlags...),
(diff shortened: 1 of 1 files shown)


cmd/diff-main.go
```diff
@@ -35,7 +35,7 @@ var (
 // Compute differences between two files or folders.
 var diffCmd = cli.Command{
 	Name:   "diff",
-	Usage:  "List objects with size difference or missing between two folders or buckets.",
+	Usage:  "List differences in object name, size, and date between folders.",
```
vadmeste: I think 'two folders' will still be more clear.

ebozduman: @vadmeste,
`between` in English is inherently involves 2 objects/things, unlike `among`, which is used when many or more than 2 items/units are involved in the action. That's why, there is no need to say here `two folders` explicitly.

vadmeste: > @vadmeste,
> between in English is inherently involves 2 objects/things, unlike among, which is used when many or more than 2 items/units are involved in the action. That's why, there is no need to say here two folders explicitly.

:+1:

cmd/diff-main.go
```diff
@@ -49,21 +49,20 @@ FLAGS:
...
+  compare the differences in object content. 
 
-  Differences between source and destination are represented by mark notations with following meaning:
-    > - indicates file should be copied.
-    < - indicates file should be deleted.
-    ! - indicates file differs in size or type.
+LEGEND:
+    > - object is only in source.
+    < - object is only in destination.
+    ! - object is newer in source.
```
nitisht: > object is newer in source.

how about `newer object is in source.`

cmd/diff-main.go
```diff
@@ -49,21 +49,20 @@ FLAGS:
   {{range .VisibleFlags}}{{.}}
   {{end}}
 DESCRIPTION:
-  Diff only lists missing objects or objects with size differences. It *DOES NOT* compare contents. 
-  i.e. objects of same name and size, but differing in content are not noticed.
+  Diff only calculates differences in object name, size and time. It *DOES NOT*
+  compare the differences in object content.
```
nitisht: > Diff only calculates differences in object name, size and time. It *DOES NOT* compare the differences in object content.

How about `Diff only calculates differences in object name, size and last modified time. It *DOES NOT* compare  object content.`

ebozduman: I've also suggested the same, but @abperiasamy thinks `time` is good enough. 
I'll make the other change (`It *DOES NOT* compare object content.`).

cmd/diff-main.go
ebozduman: OK.

PR #2390, opened 2018-02-13
Remove response timeout for admin operations

harshavardhana:


cmd/client-admin.go
```diff
@@ -84,10 +84,16 @@ func newAdminFactory() func(config *Config) (*madmin.AdminClient, *probe.Error)
...
 				TLSHandshakeTimeout:   10 * time.Second,
 				ExpectContinueTimeout: 1 * time.Second,
 				TLSClientConfig:       tlsConfig,
+				// Set this value so that the underlying transport round-tripper
+				// doesn't try to auto decode the body of objects with
+				// content-encoding set to `gzip`.
+				//
+				// Refer:
+				//    https://golang.org/src/net/http/transport.go?h=roundTrip#L1843
+				DisableCompression: true,
```
ebozduman: @harshavardhana 
Why is `DisableCompression: True` is part of this change?

harshavardhana: Must be a copy paste issue.

PR #2400, opened 2018-02-24
add encrypt option to cp, mirror and cat commands

### LINKED ISSUE #2074: cp/mirror: [feature] bring mc transparent encryption

poornas:
This fixes #2074
 Add --encrypt-key flags for cp, mirror, cat,pipe and stat commands to pass server side encryption keys. Alternately, encryption keys can be specified by MC_ENCRYPT_KEY env variable.

The  --enctrypt-key flag accepts a list of alias/object-prefix=sse-key pairs delimited by spaces. When present, sse headers are added to the request for server side encryption/decryption as the case may be.

e.g. 
```
export MC_ENCRYPT_KEY="play/images=32byteslongsecretkeymustbegiven1 s3/testbucket=32byteslongsecretkeymustbegiven2"
```

### DIFF
--- a/cmd/cat-main.go
+++ b/cmd/cat-main.go
@@ -34,7 +34,12 @@ import (
 )
 
 var (
-	catFlags = []cli.Flag{}
+	catFlags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "encrypt-key",
+			Usage: "Decrypt object (using server-side encryption)",
+		},
+	}
 )
 
 // Display contents of a file.
@@ -53,6 +58,9 @@ USAGE:
 FLAGS:
   {{range .VisibleFlags}}{{.}}
   {{end}}
+
+ENVIRONMENT VARIABLES:
+   MC_ENCRYPT_KEY:  List of comma delimited prefix=secret values
 EXAMPLES:
    1. Stream an object from Amazon S3 cloud storage to mplayer standard input.
       $ {{.HelpName}} s3/ferenginar/klingon_opera_aktuh_maylotah.ogg | mplayer -
@@ -62,7 +70,9 @@ EXAMPLES:
 
    3. Concatenate multiple files to one.
       $ {{.HelpName}} part.* > complete.img
-
+   
+   4. Stream a server encrypted object from Amazon S3 cloud storage to standard output.
+      $ {{.HelpName}} --encrypt-key 's3/ferenginar=32byteslongsecretkeymustbegiven1' s3/ferenginar/klingon_opera_aktuh_maylotah.ogg
 `,
 }
 
@@ -129,7 +139,7 @@ func checkCatSyntax(ctx *cli.Context) {
 }
 
 // catURL displays contents of a URL to stdout.
-func catURL(sourceURL string) *probe.Error {
+func catURL(sourceURL string, encKeyDB map[string][]prefixSSEPair) *probe.Error {
 	var reader io.Reader
 	size := int64(-1)
 	switch sourceURL {
@@ -146,7 +156,7 @@ func catURL(sourceURL string) *probe.Error {
 		if err == nil && client.GetURL().Type == objectStorage {
 			size = content.Size
 		}
-		if reader, err = getSourceStreamFromURL(sourceURL); err != nil {
+		if reader, err = getSourceStreamFromURL(sourceURL, encKeyDB); err != nil {
 			return err.Trace(sourceURL)
 		}
 	}
@@ -227,10 +237,16 @@ func mainCat(ctx *cli.Context) error {
 			}
 		}
 	}
+	sseKeys := os.Getenv("MC_ENCRYPT_KEY")
+	if key := ctx.String("encrypt-key"); key != "" {
+		sseKeys = key
+	}
 
+	encKeyDB, err := parseEncryptionKeys(sseKeys)
+	fatalIf(err, "Unable to parse encryption keys")
 	// Convert arguments to URLs: expand alias, fix format.
 	for _, url := range args {
-		fatalIf(catURL(url).Trace(url), "Unable to read from `"+url+"`.")
+		fatalIf(catURL(url, encKeyDB).Trace(url), "Unable to read from `"+url+"`.")
 	}
 	return nil
 }
--- a/cmd/mirror-main.go
+++ b/cmd/mirror-main.go
@@ -82,6 +82,10 @@ var (
 			Name:  "storage-class, sc",
 			Usage: "Set storage class for object",
 		},
+		cli.StringFlag{
+			Name:  "encrypt-key",
+			Usage: "Encrypt/Decrypt objects (using server-side encryption)",
+		},
 	}
 )
 
@@ -101,6 +105,9 @@ USAGE:
 FLAGS:
   {{range .VisibleFlags}}{{.}}
   {{end}}
+
+ENVIRONMENT VARIABLES:
+   MC_ENCRYPT_KEY: List of comma delimited prefix=secret values
 
 EXAMPLES:
    1. Mirror a bucket recursively from Minio cloud storage to a bucket on Amazon S3 cloud storage.
@@ -135,6 +142,9 @@ EXAMPLES:
 
   10. Mirror a bucket older than 30 days from Amazon S3 bucket test to a local folder.
       $ {{.HelpName}} --older-than=30 s3/test ~/test
+	
+  11. Mirror server encrypted objects from Minio cloud storage to a bucket on Amazon S3 cloud storage 
+      $ {{.HelpName}} --encrypt-key "minio/photos=32byteslongsecretkeymustbegiven1,s3/archive=32byteslongsecretkeymustbegiven2" minio/photos/ s3/archive
 
 `,
 }
@@ -177,6 +187,7 @@ type mirrorJob struct {
 	storageClass                           string
 
 	excludeOptions []string
+	encKeyDB       map[string][]prefixSSEPair
 }
 
 // mirrorMessage container for file mirror messages
@@ -216,7 +227,7 @@ func (mj *mirrorJob) doRemove(sURLs URLs) URLs {
 	targetWithAlias := filepath.Join(sURLs.TargetAlias, sURLs.TargetContent.URL.Path)
 
 	// Remove extraneous file on target.
-	err := probe.NewError(removeSingle(targetWithAlias, isIncomplete, mj.isFake, 0, 0))
+	err := probe.NewError(removeSingle(targetWithAlias, isIncomplete, mj.isFake, 0, 0, sURLs.encKeyDB))
 	return sURLs.WithError(err)
 }
 
@@ -343,6 +354,9 @@ func (mj *mirrorJob) watchMirror(ctx context.Context, cancelMirror context.Cance
 			// newClient needs the unexpanded  path, newCLientURL needs the expanded path
 			targetAlias, expandedTargetPath, _ := mustExpandAlias(targetPath)
 			targetURL := newClientURL(expandedTargetPath)
+			sourcePath := filepath.ToSlash(filepath.Join(sourceAlias, sourceURL.Path))
+			srcSSEKey := getSSEKey(sourcePath, mj.encKeyDB[sourceAlias])
+			tgtSSEKey := getSSEKey(targetPath, mj.encKeyDB[targetAlias])
 
 			if event.Type == EventCreate {
 				// we are checking if a destination file exists now, and if we only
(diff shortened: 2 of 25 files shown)


poornas:
@harshavardhana , PTAL.

poornas:
rebased to fix conflicts

cmd/cat-main.go
```diff
@@ -63,6 +72,9 @@ EXAMPLES:
    3. Concatenate multiple files to one.
       $ {{.HelpName}} part.* > complete.img
 
+   4. Stream a server encrypted object from Amazon S3 cloud storage to standard output.
+      $ {{.HelpName}} --encrypt 'customerspecifiedencryptky32bits' s3/ferenginar/klingon_opera_aktuh_maylotah.ogg
```
kannappanr: `...32bits` is a bit misleading, can this end with `...32byte`

poornas: good idea. This is back to wip for a different approach of taking encryption keys at command line.

cmd/cat-main.go
```diff
@@ -62,7 +70,9 @@ EXAMPLES:
 
    3. Concatenate multiple files to one.
       $ {{.HelpName}} part.* > complete.img
-
+   
+   4. Stream a server encrypted object from Amazon S3 cloud storage to standard output.
+      $ {{.HelpName}} --encrypt-key 's3/ferenginar=32byteslongsecretkeymustbegiven1' s3/ferenginar/klingon_opera_aktuh_maylotah.ogg
```
vadmeste: @poornas, I wonder what happens if '=' is present in the prefix or in the password. The same problem is what happens if the password contains a space.

I think you can use cli.StringSliceFlag here and probably oblige the user to supply base64 format of the key field, base64 doesn't seem to generate ':' character so the final result can be like that:
`mc cat --encrypt-key 's3/whatever:myplainpassword' 's3/another:mysecondplainkey' --encrypt-b64key 's3/whaterver:CBwbGVhc3VyZS4=' 'myminio/anotherwhatever:dxfwbGVhc3VyZS4' `

What do you think ?

poornas: @vadmeste, the ui has already been decided to not allow base64 key but to specifically ask for a 32 byte key. Your other points on avoiding '=' or spaces is valid. Will update the pr with more robust parsing

poornas: @vadmeste , updated the PR , PTAL.

poornas:
added functional tests - can be reviewed.

cmd/utils.go
```diff
@@ -157,3 +162,69 @@ func getLookupType(l string) minio.BucketLookupType {
...
+// parse list of space separated alias/prefix=sse key values entered on command line and
+// construct a map of alias to prefix and sse pairs
+func parseEncryptionKeys(ssekeys string) (encMap map[string][]prefixSSEPair, err *probe.Error) {
+	encMap = make(map[string][]prefixSSEPair)
+	if ssekeys == "" {
+		return
+	}
+	fields := strings.Fields(ssekeys)
+	for _, field := range fields {
+		r := regexp.MustCompile("([^=\\s]+)=([^\\s]+)")
```
vadmeste: This can be initialized only one time outside the for loop

cmd/cat-main.go
vadmeste: > @vadmeste , updated the PR , PTAL.

Cool, I saw your change.

S3 encryption key can support a space in the key, I've just checked. I think we really need to do array style instead of space style in this case. Array as argument is supported in cli package.

cmd/utils.go
poornas: done.

cmd/cat-main.go
poornas: @vadmeste, since base64 encoding doesn't seem to use "," we can switch to that as field separator. using string slices seems to entail specifying --encrypt-key alias/bucket=key1 --encrypt-key alias/bucket/prefix=key2 which would become cumbersome for the user. What do you think?

poornas:
@kannappanr , travis failure is on multipart sse put test. Since the test works locally on latest master - wondering whether play server is on release version/ behind master.

cmd/cat-main.go
vadmeste: > @vadmeste, since base64 encoding doesn't seem to use "," we can switch to that as field separator. using string slices seems to entail specifying --encrypt-key alias/bucket=key1 --encrypt-key alias/bucket/prefix=key2 which would become cumbersome for the user. What do you think?

In this case, I think both ',' and space would work.

But if we will be using base64, then we shouldn't use = as a field separator since = exists in base64 result. So I guess ':' can be useful in this case, like:
`--encrypt-key 'myminio/:plainpasswordetc s3/:bXlsb25nc2VjcmV0a2V5Cg=='`

One last thing, what about if we support both plain and base64 password at the same time ?

`--encrypt-key "myminio/:plain-32-bytes-long-secret-key"` and `--encrypt-key "myminio/:base64,bXlsb25nc2VjcmV0a2V5Cg==`

Otherwise your PR is all good for me.

harshavardhana:
It's behind master. Will fix it.

cmd/cat-main.go
poornas: @vadmeste, the current regex will recognize the first = and not parse the = in the key, so that is safe.Updated PR  to use , as delimiter for k-v pairs. Think that should take care of your concerns.

As far as accepting plain and base64 keys - think it will provide less intuitive user experience than just sticking to one input option.

cmd/utils.go
```diff
@@ -157,3 +162,70 @@ func getLookupType(l string) minio.BucketLookupType {
...
+		prefix := strings.TrimSpace(key)
+		if _, ok := encMap[alias]; !ok {
+			encMap[alias] = make([]prefixSSEPair, 0)
+		}
+		ps := prefixSSEPair{prefix: prefix, sseKey: value}
+		encMap[alias] = append(encMap[alias], ps)
+	}
+	// sort encryption keys in descending order of prefix length
+	for _, encKeys := range encMap {
+		sort.Sort(byPrefixLength(encKeys))
```
harshavardhana: you don't want to save them back into encMap after sorting?

functional-tests.sh
```diff
@@ -77,6 +77,8 @@ if [ "$ENABLE_HTTPS" != "1" ]; then
 fi
 
 SERVER_ALIAS="myminio"
+SERVER_ALIAS_1="myminio-ssl"
```
harshavardhana: Name this as `SERVER_ALIAS_TLS` instead which can provide more meaning.

functional-tests.sh
```diff
@@ -574,6 +576,204 @@ function test_config_host_add_error()
     log_success "$start_time" "${FUNCNAME[0]}"
 }
 
+function test_put_object_with_sse()
+{
+    show "${FUNCNAME[0]}"
+    start_time=$(get_time)
+    object_name="mc-test-object-$RANDOM"
+    cli_flag="${SERVER_ALIAS}/${BUCKET_NAME}=32byteslongsecretkeymustbegiven1"
+    #put encrypted object; then delete with correct secret key
```
harshavardhana: A comment should follow the bash style of `#<space><my_comment>`

functional-tests.sh
```diff
@@ -574,6 +576,204 @@ function test_config_host_add_error()
...
+    log_success "$start_time" "${FUNCNAME[0]}"
+}
+
+function test_put_object_with_sse_error()
+{
+    show "${FUNCNAME[0]}"
+    start_time=$(get_time)
+    object_name="mc-test-object-$RANDOM"
+    cli_flag="${SERVER_ALIAS}/${BUCKET_NAME}=32byteslongsecretkeymustbegiven"
+    #put object with invalid encryption key; should fail
```
harshavardhana: A comment should follow the bash style of `#<space><my_comment>`

functional-tests.sh
```diff
@@ -574,6 +576,204 @@ function test_config_host_add_error()
...
+    log_success "$start_time" "${FUNCNAME[0]}"
+}
+
+function test_cat_object_with_sse()
+{
+    show "${FUNCNAME[0]}"
+    start_time=$(get_time)
+    object_name="mc-test-object-$RANDOM"
+    cli_flag="${SERVER_ALIAS}/${BUCKET_NAME}=32byteslongsecretkeymustbegiven1"
+    #put encrypted object; then cat object correct secret key
```
harshavardhana: A comment should follow the bash style of `#<space><my_comment>`

functional-tests.sh
```diff
@@ -574,6 +576,204 @@ function test_config_host_add_error()
...
+    # now do a server side copy and store it unencrypted.
+    assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd cp --encrypt-key "${cli_flag}" "${SERVER_ALIAS}/${BUCKET_NAME}/${prefix}/${object_name}" "${SERVER_ALIAS}/${BUCKET_NAME}/${object_name}"
+    # cat the unencrypted destination object. should return data without any error
+    assert_success "${MC_CMD[@]}" cat "${SERVER_ALIAS}/${BUCKET_NAME}/${object_name}" > "${object_name}.downloaded"
+    assert_success "$start_time" "${FUNCNAME[0]}" show_on_failure $? "unable to download object using 'mc cat'"
+    assert_success "$start_time" "${FUNCNAME[0]}" check_md5sum "$FILE_1_MB_MD5SUM" "${object_name}.downloaded"
+    assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd rm "${object_name}.downloaded" "${SERVER_ALIAS}/${BUCKET_NAME}/${object_name}"
+    #mc rm on encrypted object without encryption key should fail
+    assert_failure "$start_time" "${FUNCNAME[0]}" mc_cmd rm "${SERVER_ALIAS}/${BUCKET_NAME}/${prefix}/${object_name}"
+     #mc rm on encrypted object with encryption key should pass
```
harshavardhana: indentation problem

cmd/mirror-main.go
```diff
@@ -135,6 +142,9 @@ EXAMPLES:
 
   10. Mirror a bucket older than 30 days from Amazon S3 bucket test to a local folder.
       $ {{.HelpName}} --older-than=30 s3/test ~/test
+	
+  11. Mirror server encrypted objects from Minio cloud storage to a bucket on Amazon S3 cloud storage 
+      $ {{.HelpName}} --encrypt-key "minio/photos=32byteslongsecretkeymustbegiven1 , s3/archive=32byteslongsecretkeymustbegiven2" minio/photos/ s3/archive
```
harshavardhana: is the space needed before `,` and after? we can simplify the examples without space to avoid copy-paste mistakes.

cmd/cat-main.go
```diff
@@ -129,7 +139,7 @@ func checkCatSyntax(ctx *cli.Context) {
 }
 
 // catURL displays contents of a URL to stdout.
-func catURL(sourceURL string) *probe.Error {
+func catURL(sourceURL string, encKeydb map[string][]prefixSSEPair) *probe.Error {
```
harshavardhana: nit: variable `encKeydb` should be usually called `encKeyDB`

cmd/utils.go
poornas: In place sorting - tested to double check.
```
➜  mc git:(encrypt) ✗ mc cat myminio-ssl/testbucket/obj --insecure --encrypt-key "myminio/testbucket=32byteslongsecretkeymustgenerate, myminio/testbucket/a/b=32byteslongsecretkeymustgenerat1,myminio/testbucket/a/b/c=32byteslongsecretkeymustgenerat2,myminio/testbucket/a/b/c/d=32byteslongsecretkeymustgenerat4"
encryption keys after parse map[myminio:[{myminio/testbucket/a/b/c/d 32byteslongsecretkeymustgenerat4} {myminio/testbucket/a/b/c 32byteslongsecretkeymustgenerat2} {myminio/testbucket/a/b 32byteslongsecretkeymustgenerat1} {myminio/testbucket 32byteslongsecretkeymustgenerate}]]
```

functional-tests.sh
poornas: done

functional-tests.sh
poornas: # done

cmd/mirror-main.go
poornas: done.

PR #2408, opened 2018-03-08
change encryption key parsing to allow spaces in sse-c key

poornas:
The base64 encryption key passed in command line could have spaces. Earlier implementation was splitting on spaces, which could lead to erroneous parsing.


cmd/cat-main.go
```diff
@@ -243,7 +243,7 @@ func mainCat(ctx *cli.Context) error {
 	}
 
 	encKeyDB, err := parseEncryptionKeys(sseKeys)
-	fatalIf(err, "Unable to parse encryption keys")
+	fatalIf(err, "Unable to parse encryption keys:")
```
harshavardhana: You don't need to add ":" we do not follow that syntax. any particular reason for this change?

poornas: looked more readable. can revert

cmd/utils.go
```diff
@@ -178,37 +176,46 @@ type prefixSSEPair struct {
...
-		matches := r.FindStringSubmatch(field)
-		if len(matches) != 3 {
-			return nil, probe.NewError(errors.New("sse-c prefix should be of the form prefix=key"))
+	prefix := ""
+	ssekey := ""
+	index := 0 // start index of prefix
+	vs := 0    // start index of sse-c key
+	sseKeyLen := 32
+	delim := 1
+	l := len(sseKeys)
```
krisis: suggestion: can we use a different single letter variable? `l` can be easily mistaken for `1`.

cmd/utils.go
```diff
@@ -178,37 +176,46 @@ type prefixSSEPair struct {
...
+		if (vs+sseKeyLen < l) && sseKeys[vs+sseKeyLen] != ',' {
+			return nil, probe.NewError(errors.New("sse-c prefix=secret should be delimited by , and secret should be 32 bytes long"))
+		}
 		if _, ok := encMap[alias]; !ok {
 			encMap[alias] = make([]prefixSSEPair, 0)
 		}
-		ps := prefixSSEPair{prefix: prefix, sseKey: value}
+		ps := prefixSSEPair{prefix: prefix, sseKey: ssekey}
 		encMap[alias] = append(encMap[alias], ps)
+		// advance index 33 bytes for the next key start
```
krisis: the comment should be,
`// advance index by (sseKeyLen + delim) bytes for the next key start index`

krisis:
We can use https://golang.org/pkg/encoding/csv/ which supports comma-separated values with spaces in them. This will simplify our encryption keys parsing code.

poornas:
@krisis, the encryption key can come in with "," as part of the values - which is why the current implementation has been used to reliably take the first 32 bytes from encountering the first '=' sign as the encryption key.

cmd/utils.go
poornas: done.

cmd/utils.go
poornas: done

poornas:
@krisis, added unit tests - PTAL.

cmd/client-s3.go
```diff
@@ -511,10 +511,8 @@ func (c *s3Client) Get(sseKey string) (io.Reader, *probe.Error) {
 	bucket, object := c.url2BucketAndObject()
 	var opts minio.GetObjectOptions
 	if sseKey != "" {
-		key := minio.NewSSEInfo([]byte(sseKey), "AES256")
-		for k, v := range key.GetSSEHeaders() {
-			opts.Set(k, v)
-		}
+		key := encrypt.DefaultPBKDF([]byte(sseKey), []byte(bucket+object))
```
harshavardhana: this is not correct @poornas DefaultPBKDF means we are taking password not sseKey as input. This would mean that 32byte long string is no longer a requirement. So you can simplify the code even further if you wish to use password-based style.  If not you should use `encrypt.NewSSEC` for now and we need to think about the UI again to support both key and password styles with @abperiasamy

appveyor.yml
```diff
@@ -17,6 +17,7 @@ install:
   - go version
   - go env
   - python --version
+  - go get -u golang.org/x/crypto/argon2
```
harshavardhana: this is not needed if you have vendorized it already what is the error that you see? @poornas

poornas: removed this.

cmd/client-s3.go
poornas: fixed. will discuss password-based style with @abperiasamy

PR #2413, opened 2018-03-16
Simplify parallel manager implementation

vadmeste:
Each 5 seconds, calculate current bandwidth and add new workers as
bandwidth gets increased. Stop adding workers if it is not useful
for increasing bandwidth.


vadmeste:
@harshavardhana there was a bug in the old implementation as it was increasing workers indefinitely sometimes.. This implementation does the same thing but with shorter code.

PR #2431, opened 2018-04-07
Simplified build system and move to go1.10.1

harshavardhana:
minio server simplified its build system, this
change borrows the same technique to `mc` as well.

This PR fixes build messages and overall behavior
of our builds on travis and appveyor.


harshavardhana:
@kannappanr @nitisht PTAL

PR #2438, opened 2018-04-25
Support server side copy object with progress

harshavardhana:
Support server side copy object with progress
and also uses a more optimized version to perform
multipart server side copy operation.

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -1,5 +1,5 @@
 /*
- * Minio Client (C) 2015, 2016, 2017 Minio, Inc.
+ * Minio Client (C) 2015, 2016, 2017, 2018 Minio, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,7 @@ package cmd
 import (
 	"context"
 	"crypto/tls"
+	"encoding/json"
 	"hash/fnv"
 	"io"
 	"net"
@@ -30,8 +31,6 @@ import (
 	"strings"
 	"sync"
 	"time"
-
-	"io/ioutil"
 
 	"github.com/minio/mc/pkg/httptracer"
 	"github.com/minio/mc/pkg/probe"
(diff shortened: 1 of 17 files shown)


cmd/client-s3.go
```diff
@@ -833,10 +829,18 @@ func (c *s3Client) GetAccessRules() (map[string]string, *probe.Error) {
...
+	policyStr, e := c.api.GetBucketPolicy(bucket)
+	if e != nil {
+		return nil, probe.NewError(e)
+	}
+	if policyStr == "" {
+		return policies, nil
+	}
+	var p policy.BucketAccessPolicy
+	if e = json.Unmarshal([]byte(policyStr), &p); e != nil {
+		return nil, probe.NewError(e)
```
poornas: @harshavardhana , the changes to GetAccessRules() and SetAccess() seems unconnected to this PR. Is this in error?

harshavardhana: It since you will have to update the minio-go anyways.

poornas: ok,got it

cmd/client-s3.go
```diff
@@ -863,10 +874,24 @@ func (c *s3Client) SetAccess(bucketPolicy string) *probe.Error {
...
+		if e = json.Unmarshal([]byte(policyStr), &p); e != nil {
+			return probe.NewError(e)
+		}
+	}
+	p.Statements = policy.SetPolicy(p.Statements, policy.BucketPolicy(bucketPolicy), bucket, object)
+	policyB, e := json.Marshal(p)
+	if e != nil {
+		return probe.NewError(e)
+	}
+	if e = c.api.SetBucketPolicy(bucket, string(policyB)); e != nil {
```
vadmeste: One problem here when setting 'none' policy with mc.

It can be easily reproduced:
```
mc policy download s3/vadmeste/
mc policy none s3/vadmeste
```

harshavardhana: FIxed

cmd/client-s3.go
harshavardhana: Fixed

Issue #2443, opened 2018-05-04
mc mirror doesn't work properly on OSX


thecodeassassin:
## Expected behaviour

Files should be copied the same as on linux

## Actual behaviour

Files are copied with their full path instead of the directory being trimmed.

## Steps to reproduce the behaviour

$  mc mirror -w /tmp/some_directory play/test/some_directory
$ echo "sadasdsad" > /tmp/some_directory/___file___

sync event:
/private/tmp/some_directory/___file___:    10 B / 10

$ mc ls play/test/some_directory
[2018-05-04 23:21:51 CEST]     0B private/

## mc version
```
Version: 2018-04-28T00:08:20Z
Release-tag: RELEASE.2018-04-28T00-08-20Z
Commit-id: 1b794a4ae1af46bdeb86fbdb7ba16bbd31366d83
```

## System information
OSX 10.13.3
MacBook Pro (15-inch, 2017)
3,1 GHz Intel Core i7

PR #2445, opened 2018-05-10
mc mirror watch fix in darwin

Praveenrajmani:
The event.Path had few prefixes appended in it (only in darwin).
Instead of taking the trimmed directory path, It takes the full path, say '/private/...".
This disturbs the target URL. Which indeed inserts it as a whole '/private/{target_path}',
So trimming the prefixes in the event.path solved it.

Solves:#2443

### DIFF
--- a/cmd/mirror-main.go
+++ b/cmd/mirror-main.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"sync"
 	"syscall"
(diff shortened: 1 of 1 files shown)


cmd/mirror-main.go
```diff
@@ -337,13 +337,17 @@ func (mj *mirrorJob) watchMirror(ctx context.Context, cancelMirror context.Cance
 					sourceURLFull = tmpSrcURL
 				}
 			}
+			eventPath := event.Path
 
-			sourceURL := newClientURL(event.Path)
-			aliasedPath := strings.Replace(event.Path, sourceURLFull, mj.sourceURL, -1)
+			// Strip the prefixes in the event path. Happens in darwin OS only
+			eventPath = eventPath[strings.Index(eventPath, sourceURLFull):len(eventPath)]
```
harshavardhana: If this happens only on darwin, we can limit the code to darwin only i.e perhaps bring an `if runtime.GOOS == darwin` check so that we do not disturb the existing code on all other platforms.

Praveenrajmani: @harshavardhana , The change wont disturb the other flows . will generate and check the other binaries as well !

Praveenrajmani: Updated the change

PR #2446, opened 2018-05-14
fix for documentation issue

### LINKED ISSUE #2430: Documentation update: create sample bucket
## Summary
Could you add a sample usage right after the `Add a Cloud Storage Service` section? Something like `mc mb <ALIAS>/<BUCKET>` and `mc ls <ALIAS>/<BUCKET>`? It was unclear to me that I needed <ALIAS>. 

I filled out the rest of your template even though it felt unrelated to this issue/request. Thanks for this client! 

## Expected behaviour
An example of using mc in readme

## Actual behaviour
No example of using mc in reame

 - be unfamiliar with mc
 - install locally
 - configure to talk to remote by using `mc config host add $hostName $host $accessKey 
 - $accessSecret`
 - wonder how to make a bucket and file
 - read the very top of the readme's quick start guide, see the `mb` and `cp` commands.
 - run: `mc mb $bucket` and `mc ls` and see $bucket and think you are good to go. You are not.
 - later learn that it needs to be `mc mb $hostName/$bucket`. 

## mc version
Version: 2018-03-25T01:22:22Z
Release-tag: RELEASE.2018-03-25T01-22-22Z
Commit-id: da5c19848d1e82a24eddb453b01e83d4a0660de4

## System information
MacOSX

aarushiarya:
Fixes #2430

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ mc ls play
 ```
 
 Make a bucket
-`mb` command creates a new bucket on object storage.
+`mb` command creates a new bucket.
 
 *Example:*
 ```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -147,6 +147,28 @@ mc ls play
 [2016-01-28 17:23:11 PST]     0B newbucket/
 [2016-03-20 09:08:36 PDT]     0B s3git-test/
 ```
+
+Make a bucket
+`mb` command creates a new bucket on object storage.
```
deekoder: we can remove object storage in this line

PR #2448, opened 2018-05-23
Properly shutdown parallel copy in copy

vadmeste:


vadmeste:
ping @krisis, please take a look.

kannappanr:
@vadmeste Travis build is failing PTAL

vadmeste:
@poornas @krisis, this PR has a little bug and for that I rewrote this PR. Can you review again?

cmd/cp-main.go
```diff
@@ -367,7 +365,7 @@ loop:
 	for {
 		select {
 		case <-trapCh:
-			stopParallel()
+			ctrlC = true
```
harshavardhana: This type of unsafe variable update is not correct @vadmeste a go-routine shouldn't read from a variable which can be edited externally. While it may work is still a racy code. I would expect a channel to be used here instead. which basically kills urlscanning.

vadmeste: > This type of unsafe variable update is not correct @vadmeste a go-routine shouldn't read from a variable which can be edited externally. While it may work is still a racy code. I would expect a channel to be used here instead. which basically kills urlscanning.

You are right, fixed.

vadmeste:
@poornas, can you take a look again ?

PR #2454, opened 2018-05-29
Remove deprecated snap mc package

harshavardhana:


README_zh_CN.md
```diff
@@ -75,13 +75,6 @@ chmod +x mc
...
 ```
 
-### 快照
-您可以安装最新的`minio-client`[快照版](https://snapcraft.io)，并协助我们验证在master分支的最新修改，这是[所有支持的linux发行版](https://snapcraft.io/docs/core/install) ，以下命令可安装快照版：
-
-```sh
-sudo snap install minio-client --edge --devmode
-```
-
 当一个新的版本发布时，你都会自动更新。
```
ebozduman: Just by comparing what has been changed above in English  version of the same text without understanding Chinese characters, I bet this is the corresponding line for `Every time a new version is pushed to the store, you will get it updated automatically.` which has been deleted above. Hence this line also has to be deleted.

Issue #2428, opened 2018-04-05
Documentation is not available for "mc admin"


kmkanagaraj:
## Expected behaviour
Documentation should be available for "mc admin" commands

## Actual behaviour
Documentation is not available for "mc admin" commands

## Steps to reproduce the behaviour
1. Goto https://docs.minio.io/docs/minio-client-complete-guide
2. Find no references to "mc admin"

## mc version
- (paste output of `mc version`)
master

PR #2456, opened 2018-05-30
Add 'mc admin' complete guide

harshavardhana:


docs/minio-admin-complete-guide.md
```diff
@@ -0,0 +1,311 @@
+# Minio Admin Complete Guide [![Slack](https://slack.minio.io/slack?type=svg)](https://slack.minio.io)
```
nitisht: Shouldn't this be `MC Admin Complete Guide`

docs/minio-admin-complete-guide.md
```diff
@@ -0,0 +1,311 @@
...
+
+## 5. Everyday Use
+You may add shell aliases for info, healing.
+
+```sh
+alias minfo='mc admin info'
+alias mheal='mc admin heal'
+```
+
+## 6. Global Options
```
nitisht: IMHO steps 1 through 5 are repetitive from https://github.com/minio/mc/blob/master/docs/minio-client-complete-guide.md . We can probably refer the `mc complete guide` doc in the beginning here and start from step 6 in this doc.

harshavardhana: It is repeated on purpose since minio client complete guide repeats it from README.md

docs/minio-admin-complete-guide.md
harshavardhana: I am not sure `MC` didn't sound right `Minio Client Admin` also doesn't sound right. So `Minio Admin` was only that made since it is `Minio Admin` more so than `mc admin`

docs/minio-admin-complete-guide.md
nitisht: okay, in that case we can probably take this and do a cleanup to avoid repetition in the `complete guides`.

docs/minio-admin-complete-guide.md
harshavardhana: The reason to avoid `mc` is that it doesn't make sense to call this 'client admin' since it is simply `admin`. I will prefer to keep it `Minio Admin` specifically because it is a `Minio Admin` guide not a client guide.

docs/minio-admin-complete-guide.md
```diff
@@ -0,0 +1,303 @@
...
+    "version": "2018-05-28T04:31:38Z",
+    "commitID": "7d8c5ffb13334f4aec20a35bd2575bd7c740fb7a",
+    "region": "",
+    "sqsARN": []
+  }
+}
+```
+
+### Option [--no-color]
+This option disables the color theme. It useful for dumb terminals.
```
ebozduman: `It useful for dumb terminals.`
 =>
`It`**`is`**`useful for dumb terminals.`

nitisht:
Fixes https://github.com/minio/mc/issues/2428

PR #2459, opened 2018-06-05
Fix joinPath function to handle directories properly

harshavardhana:


cmd/client-s3.go
```diff
@@ -1428,21 +1428,12 @@ func (c *s3Client) listIncompleteRecursiveInRoutineDirOpt(contentCh chan *client
...
-
-		if !strings.HasSuffix(retPath, pathSep) {
-			retPath += pathSep
-		}
-
-		retPath += segment
+func (c *s3Client) joinPath(bucket string, objects ...string) string {
+	p := string(c.targetURL.Separator) + bucket
+	for _, o := range objects {
+		p = p + string(c.targetURL.Separator) + o
```
ebozduman: `p = p + string(c.targetURL.Separator) + o`
 =>
`p += string(c.targetURL.Separator) + o`

harshavardhana: I kept it explicit on purpose to avoid confusing with multiple "+"

Issue #2467, opened 2018-06-18
Add support for ppc64le


rpsene:
When trying to use mc on Linux/ppc64le I noticed that this platform is not supported. I tried to build it from source code and it worked perfectly.

root@rpsene-development1:~/go/src/github.com/minio/mc# uname -m
ppc64le

root@rpsene-development1:~/go/src/github.com/minio/mc# make test
Installing golint
Installing gocyclo
Installing deadcode
Installing misspell
Installing ineffassign
Running vet
Running fmt
Running lint
Running cyclo
Running deadcode
Checking dependencies
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = (unset),
	LC_CTYPE = "UTF-8",
	LANG = "en_US.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to a fallback locale ("en_US.UTF-8").
Checking for project in GOPATH
Building minio binary to './mc'
Running unit tests
?   	github.com/minio/mc	[no test files]
ok  	github.com/minio/mc/cmd	(cached)
ok  	github.com/minio/mc/pkg/console	(cached)
ok  	github.com/minio/mc/pkg/hookreader	(cached)
ok  	github.com/minio/mc/pkg/httptracer	(cached)
ok  	github.com/minio/mc/pkg/ioutils	(cached)
ok  	github.com/minio/mc/pkg/probe	(cached)
Running functional tests
Running test_make_bucket()
Running test_make_bucket_error()
Running test_put_object()
Running test_put_object_error()
Running test_put_object_with_storage_class()
Running test_put_object_with_storage_class_error()
Running test_put_object_multipart()
Running test_get_object()
Running test_get_object_multipart()
Running test_presigned_post_policy_error()
Running test_presigned_put_object()
Running test_presigned_get_object()
Running test_cat_object()
Running test_mirror_list_objects()
Running test_mirror_list_objects_storage_class()
Running test_find()
Running test_find_empty()
Running test_watch_object()
Running test_put_object_with_sse()
Running test_put_object_with_sse_error()
Running test_put_object_multipart_sse()
Running test_get_object_with_sse()
Running test_cat_object_with_sse()
Running test_cat_object_with_sse_error()
Running test_copy_object_with_sse()
Running test_copy_object_with_sse2()
Running test_sse_key_rotation()
Running test_mirror_with_sse()
Running test_rm_object_with_sse()
Running test_config_host_add()
[...]

PR #2468, opened 2018-06-18
Add support for ppc64le

rpsene:
This commit adds support for ppc64le, fixing the issue #2467 

Signed-off-by: Rafael Peria de Sene <rpsene@br.ibm.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,28 @@
 sudo: false
 language: go
 
-os:
-- linux
-
-env:
-- ARCH=x86_64
-
-go:
-- 1.10.1
-
-script:
-- make test GOFLAGS="-race"
-- make coverage
-- diff -au <(gofmt -d *.go) <(printf "")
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,21 +1,28 @@
...
-go:
-- 1.10.1
-
-script:
-- make test GOFLAGS="-race"
-- make coverage
-- diff -au <(gofmt -d *.go) <(printf "")
-- diff -au <(gofmt -d cmd) <(printf "")
-- diff -au <(gofmt -d pkg) <(printf "")
+matrix:
```
vadmeste: If I correctly understood the documentation here https://docs.travis-ci.com/user/multi-os/, it is enough to set os with two architectures:
 
```
os: 
   - linux
   - linux-ppc64le
```

rpsene: I switched to matrix given that the - make test GOFLAGS="-race" is not supported on Power.

PR #2471, opened 2018-06-20
Allow creating directories from S3 to filesystem

### LINKED ISSUE #2460: Show error during recursive cp data from s3
The source s3 path has a structure like this:

```
.
├── 111.png
└── train
    └── test
        └── 111\\.png
```

The command is this:

```
$ mc cp -r s3/mybucket/data/ data/
```

## Expected behaviour

No Error at all, and all data downloaded into `data/` folder.

## Actual behaviour

```
mc: <ERROR> Failed to copy `xxxx/train/`. Object does not exist
mc: <ERROR> Failed to copy `xxxx/train/test/`. Object does not exist
```

Create a s3 path look like below.

## mc version

2018-05-25T20:20:15Z

## System information

Docker, Mac OS 10.13.4

harshavardhana:
Fixes #2460

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -212,21 +212,27 @@ func (f *fsClient) put(reader io.Reader, size int64, metadata map[string][]strin
 	// For filesystem this is a redundant information.
 
 	// Extract dir name.
-	objectDir, _ := filepath.Split(f.PathURL.Path)
-	objectPath := f.PathURL.Path
+	objectDir, objectName := filepath.Split(f.PathURL.Path)
 
-	avoidResumeUpload := isStreamFile(objectPath)
-	// Write to a temporary file "object.part.minio" before commit.
-	objectPartPath := objectPath + partSuffix
-	if avoidResumeUpload {
-		objectPartPath = objectPath
-	}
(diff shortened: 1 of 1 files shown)


cmd/client-fs.go
```diff
@@ -212,21 +212,31 @@ func (f *fsClient) put(reader io.Reader, size int64, metadata map[string][]strin
...
 		// Create any missing top level directories.
 		if e := os.MkdirAll(objectDir, 0777); e != nil {
 			err := f.toClientError(e, f.PathURL.Path)
 			return 0, err.Trace(f.PathURL.Path)
 		}
+		// Check if object name is empty, must be an empty directory
+		if objectName == "" {
+			reader = hookreader.NewHook(reader, progress)
+			// Discard bytes until currentOffset.
+			if _, e := io.CopyN(ioutil.Discard, reader, 0); e != nil {
```
vadmeste: I am not sure that io.CopyN(.., .., 0) doesn't something tangible at all, is this really needed to fix the bug ?

harshavardhana: It is need to show progress bar @vadmeste

PR #2473, opened 2018-06-20
Vendorize minio-go inside mc

kannappanr:
Fixes 2465


harshavardhana:
@kannappanr you need to vendorize new dependencies as well.

PR #2476, opened 2018-06-28
 Fix 'mc mb' to guide the user when bucket is not passed

harshavardhana:
Additionally this PR also allows for creating
prefixes on S3 compatible object storage backends
in the following form

```
mc mb s3/harshavardhana/object/prefix/
```


harshavardhana:
ping @vadmeste

PR #2499, opened 2018-07-31
Implement select command for MC

Colhodm:
MC support for the S3 Select command
Example command: mc select --expression "myQuery" s3/mybucket/my.csv
Does not currently support configurable query options through command line(e.g compression type, no headers ..), only default options are utilized right now
Requires the version of minio-go from this pull request:https://github.com/minio/minio-go/pull/1013


harshavardhana:
Will be re-implemented..

Issue #2470, opened 2018-06-20
mc mirror complains about extra file in destination


nitisht:
** Not using template on purpose ** 
as reported by @abperiasamy 

mc: <ERROR> Unable to perform a mirror action. Delete not allowed for `/home/ab/Videos/MD5SUM`. Use `--remove` to override this behavior.

mc mirror complains about extra file in the destination. Having extra files is not an error. (--remove is optional).

PR #2500, opened 2018-08-01
Remove error statement in mc mirror

sinhaashish:
mc mirror complains about extra file in destination.
Having extra files is not an error so error statement is removed
Fixes: #2470


poornas:
@sinhaashish , pls check build failure

ebozduman:
@sinhaashish,
This is why the build failed: when you removed the call for `errDeleteNotAllowed(diffMsg.SecondURL)` from the error block, the function became obsolete since no one else calls it. 
Please remove the function definition also.

sinhaashish:
Done. Please review

PR #2503, opened 2018-08-02
Remove/deprecate admin lock commands

harshavardhana:


harshavardhana:
@krishnasrinivas @donatello this PR has to be merged and released along with minio server as well due to breaking changes on server.

PR #2506, opened 2018-08-07
Updating minio-go library

vadmeste:
This PR includes latest minio-go fixes.

### DIFF
--- a/vendor/github.com/minio/minio-go/api-put-bucket.go
+++ b/vendor/github.com/minio/minio-go/api-put-bucket.go
@@ -178,6 +178,87 @@ func (c Client) removeBucketPolicy(bucketName string) error {
 	return nil
 }
 
+// SetBucketLifecycle set the lifecycle on an existing bucket.
+func (c Client) SetBucketLifecycle(bucketName, lifecycle string) error {
+	// Input validation.
+	if err := s3utils.CheckValidBucketName(bucketName); err != nil {
+		return err
+	}
+
+	// If lifecycle is empty then delete it.
+	if lifecycle == "" {
+		return c.removeBucketLifecycle(bucketName)
+	}
+
+	// Save the updated lifecycle.
+	return c.putBucketLifecycle(bucketName, lifecycle)
+}
+
+// Saves a new bucket lifecycle.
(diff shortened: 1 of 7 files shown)


vendor/github.com/minio/minio-go/retry.go
```diff
@@ -85,36 +85,32 @@ func (c Client) newRetryTimer(maxRetry int, unit time.Duration, cap time.Duratio
...
-				return true
-			} else if strings.Contains(err.Error(), "connection timed out") {
-				// If err is a net.Dial timeout, retry.
-				return true
-			} else if strings.Contains(err.Error(), "net/http: HTTP/1.x transport connection broken") {
-				// If error is transport connection broken, retry.
-				return true
-			}
+		}
+		if strings.Contains(err.Error(), "Connection closed by foreign host") {
```
ebozduman: These `if` statements are `unreachable` since they are put inside the `switch` command, but not enclosed with a known `case` or `default` keyword, which are the only way to execute a code inside a `switch` command.
The problem will go away if the `if` statements are put inside a `default` block as it is in the old code.

vendor/github.com/minio/minio-go/utils.go
```diff
@@ -223,6 +223,7 @@ var supportedHeaders = []string{
 	"content-disposition",
 	"content-language",
 	"x-amz-website-redirect-location",
+	"expires",
```
ebozduman: How is this change related to this PR?

vendor/github.com/minio/minio-go/api-put-bucket.go
```diff
@@ -178,6 +178,87 @@ func (c Client) removeBucketPolicy(bucketName string) error {
...
+		contentMD5Base64: sumMD5Base64(b),
+	}
+
+	// Execute PUT to upload a new bucket lifecycle.
+	resp, err := c.executeMethod(context.Background(), "PUT", reqMetadata)
+	defer closeResponse(resp)
+	if err != nil {
+		return err
+	}
+	if resp != nil {
```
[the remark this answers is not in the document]
harshavardhana: all of these should be addressed in minio-go @ebozduman open an issue there.

PR #2514, opened 2018-08-17
Remove workaround for http response tracer

harshavardhana:


cmd/client-s3-trace_v2.go
```diff
@@ -66,23 +65,7 @@ func (t traceV2) Response(resp *http.Response) (err error) {
...
-			}
-			respTrace = buffer.Bytes()
-			respTrace = append(respTrace, []byte("\r\n")...)
-		} else {
-			respTrace, err = httputil.DumpResponse(resp, false)
-			if err != nil {
-				return err
-			}
-		}
+		respTrace, err = httputil.DumpResponse(resp, false)
```
kannappanr: Why are we not returning on error here? If we don't return this code will get executed,

```if globalInsecure && resp.TLS != nil {
       dumpTLSCertificates(resp.TLS)
    }
```

Is that intended?

harshavardhana: it really doesn't matter..

kannappanr: ok

PR #2520, opened 2018-08-23
Fix cp crash when resuming from session

harshavardhana:
We should make sure to handle errors properly
when reading from the session and also parsing.

Current master ignores all these situations
which leads to crash, this PR handles this to avoid
nil pointer crashes.

### DIFF
--- a/cmd/cp-main.go
+++ b/cmd/cp-main.go
@@ -338,9 +338,19 @@ func doCopySession(session *sessionV8) error {
 					gracefulStop()
 					return
 				}
+
+				if e := urlScanner.Err(); e != nil {
+					// Error while reading. quit immediately
+					gracefulStop()
+					return
+				}
+
 				var cpURLs URLs
 				// Unmarshal copyURLs from each line.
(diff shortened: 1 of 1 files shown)


cmd/cp-main.go
```diff
@@ -338,9 +338,19 @@ func doCopySession(session *sessionV8) error {
...
+					gracefulStop()
+					return
+				}
+
 				var cpURLs URLs
 				// Unmarshal copyURLs from each line.
-				json.Unmarshal([]byte(urlScanner.Text()), &cpURLs)
+				if e := json.Unmarshal([]byte(urlScanner.Text()), &cpURLs); e != nil {
+					errorIf(probe.NewError(e), "Unable to unmarshal %s", urlScanner.Text())
+					continue
```
vadmeste: It is not clear if this is an odd situation or can be easily triggered by a user input, can't we just exit here ?

harshavardhana: We can exit, but my idea to throw an error and continue if possible for other entries which might be valid @vadmeste - so will throw an error and proceed forward.

harshavardhana:
ping @vadmeste

Issue #2366, opened 2018-01-23
Improvement request: make mc able to consume 10Gb/s throughput while uploading large files


sl0n:
## Expected behaviour
I have two 10Gb/s connected servers: one is Minio server, the other one is Minio client. Both servers are equipped with SSD local storage.
I want to upload more than 1GiB in size backup file with "mc".
I expect "mc" to consume all available bandwidth, ie 10Gb/s (given that server is not busy and can handle such a workload)
## Actual behaviour
I tried both "mc copy" and "mc mirror", but couldn't see throughput of more than ~120MiB/s per file.
## Steps to reproduce the behavior
1. Install Minio server and client on physical servers
2. Connect both servers to 10Gb network.
3. Generate more than 1GiB test file, eg with dd.
4. Run "mc copy" or "mc mirror" to test throughput.
## mc version

```
$ ./mc version
Version: 2018-01-18T21:18:56Z
Release-tag: RELEASE.2018-01-18T21-18-56Z
Commit-id: d9663e5d96d5f32725867167de2a0403704ae537

$ ./minio version
Version: 2018-01-02T23:07:00Z
Release-Tag: RELEASE.2018-01-02T23-07-00Z
Commit-ID: a6318dbdaf0a33d80e52bb762e153e4d7ca396f7

```
## System information
iperf throughput test, where minio100-04 is Minio server, minio100-02 - Minio client:

```
@minio100-02 ~]$ iperf -c minio100-04
------------------------------------------------------------
Client connecting to minio100-04, TCP port 5001
TCP window size:  502 KByte (default)
------------------------------------------------------------
[  3] local 10.198.232.23 port 38470 connected with 10.198.221.42 port 5001
[ ID] Interval       Transfer     Bandwidth
[  3]  0.0-10.0 sec  11.0 GBytes  9.41 Gbits/sec

```

harshavardhana:
Thanks @sl0n for opening an issue.

vadmeste:
@sl0n currently mc (using minio-go) spawns three threads to upload all parts in parallel. If there are other files, then copy/mirror should be even more boosted. 

If we are going to do this, then we must add an API in minio-go to control the number of threads responsible of uploading parts.

I'll refer to the team first.

deekoder:
@sl0n  we will consider this for the future, it will be a great nice feature. But we will get to this after we wrap up some of our other open issues first. We usually move the ask to future milestone and close it. We will reactivate it again and bring it back to current milestone when we work on it.

[...]

PR #2521, opened 2018-08-23
Fix parallel upload performance issues

harshavardhana:
Tested by transferring 29GiB data, around 246569 files
over 100Gig network.

- master branch
```
time mc mirror /root myminio/testbucket/
real	2m10s
```

- after changes
```
time mc mirror /root myminio/testbucket/
real	1m2s
```

By default runtime.NumCPU() workers are started by default


harshavardhana:
After testing over 100g network we have some very promising results.

kannappanr:
Some performance numbers uploading 16K files of size 1 KB to AWS
1. Released mc (08/18) `146` seconds
2. This mc - `104` seconds
3. This mc with threads set to `128` - `21` seconds

cmd/mirror-main.go
```diff
@@ -626,13 +622,29 @@ func newMirrorJob(srcURL, dstURL string, isFake, isRemove, isOverwrite, isWatch
...
-		status:         status,
 		statusCh:       make(chan URLs),
-		queueCh:        make(chan func() URLs),
 		wgStatus:       new(sync.WaitGroup),
 		watcher:        NewWatcher(UTCNow()),
 	}
 
+	if wCount > maxParallelWorkers {
+		errorIf(errInvalidArgument().Trace(),
+			"Woker count %d is greater than max workers %d, setting to max %d instead",
```
kannappanr: typo, should be `set to max` instead

kannappanr:
Except for the typo and confirmation from AB, that we can get the `worker-count` from user as input, everything looks good. Once we get the confirmation, we can add the same for `cp` command also.

harshavardhana:
This is approved by @abperiasamy we can proceed to review and merge this PR.

kannappanr:
I think based on our tests, for smaller files, around 32 threads make sense. In this case, if someone is using a processor with lots of CPUs, this will work well. For other cases, we will have to see if this brings in a lot of time improvement.

harshavardhana:
> I think based on our tests, for smaller files, around 32 threads make sense. In this case, if someone is using a processor with lots of CPUs, this will work well. For other cases, we will have to see if this brings in a lot of time improvement.

Worker count is increased 2 at a time with 4 secs interval

vadmeste:
@harshavardhana approved, the start number of workers is responsible for the performance enhancement ?

harshavardhana:
Yes @vadmeste and also higher number.

cmd/parallel-manager.go
```diff
@@ -49,16 +51,15 @@ type ParallelManager struct {
...
 		// to create a new one.
 		return
 	}
 
 	// Update number of threads
-	atomic.AddUint32(&p.workersNum, 1)
+	atomic.AddUint32(&p.workersNum, count)
 
 	// Start a new worker
 	p.wg.Add(1)
```
poornas: shouldn't count workers be started here?

cmd/parallel-manager.go
```diff
@@ -119,7 +120,7 @@ func (p *ParallelManager) monitorProgress() {
 					maxBandwidth = bandwidth
 				}
 
-				p.addWorker()
+				p.addWorker(2)
```
poornas: might be better to define a constant, rather than hard code a number?

nitisht:
@harshavardhana can you pls check the appveyor failure

harshavardhana:
Appveyor failure is related to go1.11 upgrade.

llamafilm:
Was anyone able to achieve 10Gbps, like in the linked issue #2366 ?

29GB in 1m2s is only 3.7Gbps.

I'm seeing about 2.5Gbps transferring a single large file over a 20Gb network.

PR #2523, opened 2018-08-27
Fix url parsing of MC_HOSTS_ env variable

poornas:
When credentials contain special characters such as #, the MC_HOSTS_alias env is not parsing the config correctly. Strip out accesskey and secretkey from url if url parsing fails and try to interpret the config.

Fixes https://github.com/minio/minio-boshrelease/issues/90

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"fmt"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


cmd/config.go
```diff
@@ -205,10 +206,48 @@ func parseEnvURL(envURL string) (*url.URL, string, string, *probe.Error) {
...
+// parse url usually obtained from env.
+func parseEnvURLStr(envURL string) (*url.URL, string, string, *probe.Error) {
+	var envURLStr string
+	u, accessKey, secretKey, err := parseEnvURL(envURL)
+	if err != nil {
+		// url parsing can fail when accessKey/secretKey contains non url encoded values
+		// such as #. Strip accessKey/secretKey from envURL and parse again.
+		re := regexp.MustCompile("^(https?://)(.*?):(.*?)@(.*?)$")
+		res := re.FindAllStringSubmatch(envURL, -1)
+		if len(res[0]) != 5 {
```
nitisht: A comment on why this should be 5 will be helpful here..

cmd/config.go
```diff
@@ -205,10 +206,48 @@ func parseEnvURL(envURL string) (*url.URL, string, string, *probe.Error) {
 	return u, accessKey, secretKey, nil
 }
 
+// parse url usually obtained from env.
+func parseEnvURLStr(envURL string) (*url.URL, string, string, *probe.Error) {
```
nitisht: Can you also add a test case for this function here https://github.com/minio/mc/blob/master/cmd/utils_test.go#L24

cmd/config.go
```diff
@@ -205,10 +206,48 @@ func parseEnvURL(envURL string) (*url.URL, string, string, *probe.Error) {
...
+	var envURLStr string
+	u, accessKey, secretKey, err := parseEnvURL(envURL)
+	if err != nil {
+		// url parsing can fail when accessKey/secretKey contains non url encoded values
+		// such as #. Strip accessKey/secretKey from envURL and parse again.
+		re := regexp.MustCompile("^(https?://)(.*?):(.*?)@(.*?)$")
```
kannappanr: will this work only with `https` and not `http`?

cmd/config.go
poornas: @nitisht , done.

cmd/config.go
kannappanr: ok

nitisht:
#2525 is needed to fix Appveyor failure

kannappanr:
@poornas Can you please rebase. This will solve `Appveyor` failure

PR #2538, opened 2018-09-13
Implement select command

harshavardhana:


harshavardhana:
This PR is ready for review @kannappanr @nitisht

cmd/client-s3.go
```diff
@@ -384,6 +386,61 @@ func (c *s3Client) ListNotificationConfigs(arn string) ([]notificationConfig, *p
...
+				FieldDelimiter:  ",",
+			},
+		}
+		opts.OutputSerialization = minio.SelectObjectOutputSerialization{
+			CSV: &minio.CSVOutputOptions{
+				RecordDelimiter: "\n",
+				FieldDelimiter:  ",",
+			},
+		}
+	} else if strings.Contains(origContentType, "json") {
```
kannappanr: we can maybe skip if the file is not either one of `csv` and `json`

harshavardhana: it is skipped not by this function instead by the caller...

kannappanr: This is the output I see when I run the recursive command with one file which is neither csv not json
```
mc: <ERROR> Unable to run select Invalid data source type. Only CSV and JSON are supported at this time.
2.000000
1178008.000000
```
I think this tells the user that they have a non `csv and json` file. I guess this is good enough.

harshavardhana:
There is a CLI question that needs review from @abperiasamy - blocked until then

harshavardhana:
Unlocked `--expression` is changed to `--query`

harshavardhana:
please review @kannappanr @sinhaashish

PR #2551, opened 2018-09-25
new option --delete-transferred for mc copy

gitnchez:
option to allow remove local files uploaded to remote storage

### DIFF
--- a/cmd/cp-main.go
+++ b/cmd/cp-main.go
@@ -388,14 +388,11 @@ func doCopySession(session *sessionV8) error {
 							sourceAlias := cpURLs.SourceAlias
 							sourceURL := cpURLs.SourceContent.URL
 							sourcePath := filepath.ToSlash(filepath.Join(sourceAlias, sourceURL.Path))
-							fmt.Println("will delete")
 							err := os.Remove(sourcePath)
 							if err != nil {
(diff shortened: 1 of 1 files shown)


kannappanr:
@gitnchez Thanks for this PR. We need to discuss internally and see if we can take this in. I will get back 
 with what we decide.

harshavardhana:
@deekoder did we agree on this ?

harshavardhana:
I think this option is not useful @gitnchez you can do `mc cp` and `mc rm` locally which would serve the same functionality.

gitnchez:
I need it in one case - when i copy multiple files via mc from one folder
to minio server and there are multiple writes to this folder at the same
time. Maybe we need one separate command, something like "mc mv" for such
cases?

пт, 8 мар. 2019 г. в 01:34, Harshavardhana <notifications@github.com>:

harshavardhana:
`mc mv` might be a better idea

kannappanr:
ping @gitnchez Can you please make this change suggested here?

gitnchez:
> ping @gitnchez Can you please make this change suggested here?

sorry, i do not have enough experience for this

harshavardhana:
Closing this PR as stale, please re-open with a completed fix.

PR #2553, opened 2018-09-25
Disallow copying/mirroring a folder into itself

vadmeste:
As unix cp command complains when you copy a directory into itself,
(such as `cp -r ~/Desktop/ ~/Desktop/sub/`), mc will follow the
same behavior with this commit.

The main reason is that copy or mirror command can be unpredictable
with this style of copy.

Example:

```

$ mc ls -r myminio/testbucket/
[2018-09-25 19:28:17 CET]     8B dir/object
[2018-09-25 19:28:40 CET]     8B object

$ mc -q cp -r myminio/testbucket/ myminio/testbucket/dir/
`myminio/testbucket/dir/object` -> `myminio/testbucket/dir/dir/object`
`myminio/testbucket/object` -> `myminio/testbucket/dir/object`

```

We can notice that `myminio/testbucket/dir/object` is both source and target
in the copy operation. The last copy operation that will occur will rule the
result of the total copy operation and `myminio/testbucket/dir/dir/object`
will have a different content in each copy operation.

Due to the unpredictability mentionned above, we will disallow copying or
mirroring a folder into itself.

Fixes #6519

### DIFF
--- a/cmd/mirror-url.go
+++ b/cmd/mirror-url.go
@@ -47,31 +47,35 @@ func checkMirrorSyntax(ctx *cli.Context, encKeyDB map[string][]prefixSSEPair) {
 		errorIf(errInvalidArgument().Trace(URLs...), "`--force` is deprecated please use `--overwrite` instead for the same functionality.")
 	}
 
+	tgtClientURL := newClientURL(tgtURL)
+	if tgtClientURL.Host != "" {
+		if tgtClientURL.Path == string(tgtClientURL.Separator) {
+			fatalIf(errInvalidArgument().Trace(tgtURL),
+				fmt.Sprintf("Target `%s` does not contain bucket name.", tgtURL))
+		}
+	}
+
 	/****** Generic rules *******/
 	if !ctx.Bool("watch") {
-		_, srcContent, err := url2Stat(srcURL, false, encKeyDB)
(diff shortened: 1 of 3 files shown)


cmd/mirror-url.go
```diff
@@ -47,31 +47,39 @@ func checkMirrorSyntax(ctx *cli.Context, encKeyDB map[string][]prefixSSEPair) {
 		errorIf(errInvalidArgument().Trace(URLs...), "`--force` is deprecated please use `--overwrite` instead for the same functionality.")
 	}
 
+	if len(tgtURL) == 0 && tgtURL == "" {
```
kannappanr: aren't these both same?

vadmeste: > aren't these both same?

yes, this code was just moved but I can fix it

vadmeste: Actually I am removing that code, it is not necessary, we will never have tgtURL to be empty since we already check the number of arguments in the beginning of checkMirrorSyntax()

vadmeste: > Actually I am removing that code, it is not necessary, we will never have tgtURL to be empty since we already check the number of arguments in the beginning of checkMirrorSyntax()

Fixed

PR #2554, opened 2018-09-25
Add admin users

harshavardhana:


cmd/admin-users.go
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+package cmd
+
+import "github.com/minio/cli"
+
+var adminUsersCmd = cli.Command{
+	Name:   "users",
+	Usage:  "Manage mulitple users",
```
poornas: typo here on multiple - Manage users would be better

harshavardhana:
This PR is changing a bit

docs/minio-admin-complete-guide.md
```diff
@@ -273,19 +275,69 @@ FLAGS:
   Storage : Used 8.2GiB
 ```
 
+<a name="users"></a>
+### Command `users` - Manage regular users
+`users` command to add new, remove, enable, disable, list users on Minio server.
```
poornas: add new, -> add

cmd/admin-users-add.go
```diff
@@ -0,0 +1,158 @@
...
+	Name:   "add",
+	Usage:  "Add new users",
+	Action: mainAdminUsersAdd,
+	Before: setGlobalsFromContext,
+	Flags:  globalFlags,
+	CustomHelpTemplate: `NAME:
+  {{.HelpName}} - {{.Usage}}
+
+USAGE:
+  {{.HelpName}} TARGET USERNAME PASSWORD POLICYTYPE|POLICYFILE
```
vadmeste: Can we add possible values of policy type ? and a short example of policy file content

vadmeste:
One little problem is when trying to upload an object, mc cp, in a bucket.. it doesn't work because for some reasons, mc want to list all buckets, readwrite policy won't guarantee it that, but users can use custom policy document probably and then fix this later

harshavardhana:
> One little problem is when trying to upload an object, mc cp, in a bucket.. it doesn't work because for some reasons, mc want to list all buckets, readwrite policy won't guarantee it that, but users can use custom policy document probably and then fix this later

This is something of a problem in `minio-go` I think it constructs policies with explicit API names but not for everything.

PR #2555, opened 2018-09-26
Add --api to the configure host part

martbhell:


nitisht:
@martbhell can you please explain why is this needed?

martbhell:
Hello @nitisht for our ceph RGW S3 endpoint it doesn't work without --api:

<pre>
$ ./mc config host add pouta https://object.pouta.csc.fi:443 myaccesskey mysecretkey S3v2
mc: <ERROR> Unable to initialize new config from the provided credentials. The access key ID you provided does not exist in our records.
$ ./mc config host add pouta https://object.pouta.csc.fi:443 myaccesskey mysecretkey --api S3v2
Added `pouta` successfully.
</pre>

PR #2557, opened 2018-09-26
Add admin profiling command

vadmeste:
Implement profiling command to start & download profiling data of
a standalone or all nodes of a cluster.

```
1. Start CPU profiling
   $ mc admin profiling start --type "cpu" myminio/

2. Download latest profiling data under the specified filename
   $ mc admin profiling stop --save "/tmp/profiler-data.zip" myminio/
```

### DIFF
--- a/cmd/admin-profiling-start.go
+++ b/cmd/admin-profiling-start.go
@@ -1,0 +1,111 @@
+/*
+ * Minio Client (C) 2018 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"strings"
+
+	"github.com/minio/cli"
+	"github.com/minio/mc/pkg/console"
+	"github.com/minio/mc/pkg/probe"
+	"github.com/minio/minio/pkg/madmin"
+)
+
+var adminProfilingStartFlags = []cli.Flag{
+	cli.StringFlag{
+		Name:  "type",
+		Usage: "Profiler type, possible values are: `cpu`, `mem`, `block`, `mutex` and `trace`",
+		Value: "mem",
+	},
+}
+
+var adminProfilingStartCmd = cli.Command{
+	Name:            "start",
+	Usage:           "Start recording profiling data",
+	Action:          mainAdminProfilingStart,
+	Before:          setGlobalsFromContext,
+	Flags:           append(adminProfilingStartFlags, globalFlags...),
+	HideHelpCommand: true,
+	CustomHelpTemplate: `NAME:
+  {{.HelpName}} - {{.Usage}}
+
+USAGE:
+  {{.HelpName}} [FLAGS] TARGET
+
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}
+EXAMPLES:
+    1. Start CPU profiling
+       $ {{.HelpName}} --type cpu myminio/
+
+`,
+}
+
+func checkAdminProfilingStartSyntax(ctx *cli.Context) {
+	// Check flags combinations
+	if len(ctx.Args()) != 1 {
+		cli.ShowCommandHelpAndExit(ctx, "start", 1) // last argument is exit code
+	}
+
+	profilerTypes := []madmin.ProfilerType{
+		madmin.ProfilerCPU,
+		madmin.ProfilerMEM,
+		madmin.ProfilerBlock,
+		madmin.ProfilerMutex,
+		madmin.ProfilerTrace,
+	}
+
+	// Check if the provided profiler type is known and supported
+	supportedProfiler := false
+	profilerType := strings.ToLower(ctx.String("type"))
+	for _, profiler := range profilerTypes {
+		if profilerType == string(profiler) {
+			supportedProfiler = true
+			break
+		}
+	}
+	if !supportedProfiler {
+		fatalIf(errDummy(), "Profiler type unrecognized. Possible values are: %v.", profilerTypes)
+	}
+}
+
+// mainAdminProfilingStart - the entry function of profiling command
+func mainAdminProfilingStart(ctx *cli.Context) error {
+	// Check for command syntax
+	checkAdminProfilingStartSyntax(ctx)
+
+	// Get the alias parameter from cli
+	args := ctx.Args()
+	aliasedURL := args.Get(0)
+
+	profilerType := ctx.String("type")
+
+	// Create a new Minio Admin Client
+	client, err := newAdminClient(aliasedURL)
+	if err != nil {
+		fatalIf(err.Trace(aliasedURL), "Cannot initialize admin client.")
+		return nil
+	}
+
+	// Start profiling
+	_, cmdErr := client.StartProfiling(madmin.ProfilerType(profilerType))
+	fatalIf(probe.NewError(cmdErr), "Unable to start profiling.")
+
+	console.Infoln("Profiling data successfully started.")
+	return nil
+}
--- a/cmd/admin-profiling-stop.go
+++ b/cmd/admin-profiling-stop.go
@@ -1,0 +1,106 @@
+/*
+ * Minio Client (C) 2018 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


harshavardhana:
@vadmeste as far as I know @abperiasamy wanted download to be automatic as soon as profiling was stopped on server

vadmeste:
> @vadmeste as far as I know @abperiasamy wanted download to be automatic as soon as profiling was stopped on server

Yes, that's what's happening right now, we don't have separate command download & stop here anymore. However I called it download not stop, probably because it is more clear.

I'll take your suggestion and ask AB for review

vadmeste:
@harshavardhana @kannappanr updated, but let's also get AB review

cmd/admin-profiling-start.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/minio/mc/pkg/console"
+	"github.com/minio/mc/pkg/probe"
+	"github.com/minio/minio/pkg/madmin"
+)
+
+var adminProfilingStartFlags = []cli.Flag{
+	cli.StringFlag{
+		Name:  "type",
+		Usage: "Profiler type, possible values are: `cpu`, `mem`, `block`, `mutex` and `trace`",
+		Value: "cpu",
```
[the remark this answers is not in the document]
vadmeste: > should this be `value` instead of `cpu`?

This means "cpu" is the value by default if the user doesn't specify a profiler type using --type flag in `mc admin profiling start` command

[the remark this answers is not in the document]
harshavardhana: I think `block` should be default @vadmeste or `mem` - we don't use much of `cpu` ever.

cmd/admin-profiling-start.go
```diff
@@ -0,0 +1,105 @@
...
+`,
+}
+
+func checkAdminProfilingStartSyntax(ctx *cli.Context) {
+	// Check flags combinations
+	if len(ctx.Args()) != 1 {
+		cli.ShowCommandHelpAndExit(ctx, "start", 1) // last argument is exit code
+	}
+
+	profilerTypes := []string{"cpu", "mem", "block", "mutex", "trace"}
```
harshavardhana: Don't we already have these types defined in `madmin.Cpu` ?

cmd/admin-profiling-stop.go
```diff
@@ -0,0 +1,133 @@
...
+USAGE:
+  {{.HelpName}} [FLAGS] TARGET
+
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}
+EXAMPLES:
+    2. Download latest profiling data under the specified filename
+       $ {{.HelpName}} --save /tmp/profiling-data.zip myminio/
+
```
harshavardhana: Stray extra line in help output.

cmd/admin-profiling-stop.go
```diff
@@ -0,0 +1,133 @@
...
+`,
+}
+
+func checkAdminProfilingStopSyntax(ctx *cli.Context) {
+	if len(ctx.Args()) != 1 {
+		cli.ShowCommandHelpAndExit(ctx, "stop", 1) // last argument is exit code
+	}
+
+	if len(ctx.String("save")) == 0 {
+		fatalIf(errDummy(), "Kindly specify the save path of profiling data using --save flag.")
```
harshavardhana: why not use current directory as default? and keep the name of the zip file same as the profiler type?

cmd/admin-profiling-stop.go
```diff
@@ -0,0 +1,133 @@
...
+
+	downloadPath := ctx.String("save")
+
+	// Add zip suffix if not found
+	if !strings.HasSuffix(downloadPath, ".zip") {
+		downloadPath += ".zip"
+	}
+
+	// Error out if target download file already exists
+	_, e := os.Stat(downloadPath)
```
harshavardhana: We can have a overwrite option as well... or move it automatically and save it with the timestamp? and write our anyways?

kannappanr:
tested `trace` option, it is failing

```
mc admin profiling start --type trace play
mc: Profiling data successfully started.

mc cp ~/Desktop/minio play/test
mc admin profiling stop --save ~/Desktop/trace.zip play
mc: <ERROR> Unable to download profiling data: Unable to perform the requested operation because profiling is not enabled```

**Comment by vadmeste** [comment]
@kannappanr https://github.com/minio/minio/pull/6599 should fix tracing issue

For mutex, I tested, you just need to do operations on the server for some minutes, (look  `while true; do mc cp file myminio/testbucket; done`)

File: cmd/admin-profiling-start.go (hunk 2 above)
Comment (vadmeste) [reply cmd/admin-profiling-start.go]: > Don't we already have these types defined in `madmin.Cpu` ?

Fixed

File: cmd/admin-profiling-stop.go (hunk 3 above)
Comment (vadmeste) [reply cmd/admin-profiling-stop.go]: > Stray extra line in help output.

Fixed

File: cmd/admin-profiling-stop.go (hunk 4 above)
Comment (vadmeste) [reply cmd/admin-profiling-stop.go]: > why not use current directory as default? and keep the name of the zip file same as the profiler type?

Umm, mc start doesn't keep stateful data in the disk so mc stop will remember what type of profiling this is about, but we can update profiling adming API to return the type of profiling data I guess, not sure if this is worth to do

File: cmd/admin-profiling-stop.go (hunk 5 above)
Comment (vadmeste) [reply cmd/admin-profiling-stop.go]: > We can have a overwrite option as well... or move it automatically and save it with the timestamp? and write our anyways?

My taste prefers the current style, let's review with AB

File: cmd/admin-profiling-start.go (hunk 1 above)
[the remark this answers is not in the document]
Comment (vadmeste) [reply cmd/admin-profiling-start.go]: > I think `block` should be default @vadmeste or `mem` - we don't use much of `cpu` ever.

Done

File: cmd/admin-profiling-stop.go (hunk 4 above)
Comment (vadmeste) [reply cmd/admin-profiling-stop.go]: > No, not that I meant simply say `profiling.zip` @vadmeste if the user is expected to know what profile type it is - then it doesn't matter what we choose. We should do this automatically I mean.

Yes, I see now, it makes sense, fixed now

File: cmd/admin-profiling-stop.go (hunk 5 above)
Comment (vadmeste) [reply cmd/admin-profiling-stop.go]: > We should move automatically and avoid this type of error reporting.. and just tell the user that this is what we did.

Done

**Comment by vadmeste** [comment]
Not blocked anymore

File: cmd/admin-profiling-stop.go [hunk 6]
```diff
@@ -0,0 +1,107 @@
...
+	} else {
+		if !os.IsNotExist(e) {
+			fatal(probe.NewError(e), "Unable to download profiling data:")
+		}
+	}
+
+	e = os.Rename(tmpFile.Name(), downloadPath)
+	fatalIf(probe.NewError(e), "Unable to download profiling data:")
+
+	console.Infof("Profiling data successfully downloaded as `%s`\n", downloadPath)
```
harshavardhana:
```
Profiling data successfully downloaded to %s
```

cmd/admin-profiling-stop.go
```diff
@@ -0,0 +1,107 @@
...
+		e = os.Rename(downloadPath, downloadPath+"."+time.Now().Format("2006-01-02T15:04:05.999999-07:00"))
+		fatalIf(probe.NewError(e), "Unable to create a backup of profiling.zip")
+	} else {
+		if !os.IsNotExist(e) {
+			fatal(probe.NewError(e), "Unable to download profiling data:")
+		}
+	}
+
+	e = os.Rename(tmpFile.Name(), downloadPath)
+	fatalIf(probe.NewError(e), "Unable to download profiling data:")
```
harshavardhana: you can do this in one line

```
	fatalIf(probe.NewErroros.Rename(tmpFile.Name(), downloadPath)(), "Unable to download profiling data:")
```

cmd/admin-profiling-stop.go
```diff
@@ -0,0 +1,107 @@
...
+	tmpFile, e := ioutil.TempFile("", "mc-profiling-")
+	fatalIf(probe.NewError(e), "Unable to download profiling data:")
+
+	// Ask for profiling data, which will come compressed with zip format
+	zippedData, adminErr := client.DownloadProfilingData()
+	fatalIf(probe.NewError(adminErr), "Unable to download profiling data:")
+
+	// Copy zip content to target download file
+	_, e = io.Copy(tmpFile, zippedData)
+	fatalIf(probe.NewError(e), "Unable to download profiling data:")
```
harshavardhana: why this `:` at the end, we don't use this convention.

cmd/admin-profiling-stop.go
vadmeste: > Profiling data successfully downloaded to %s

Fixed

cmd/admin-profiling-stop.go
vadmeste: > fatalIf(probe.NewErroros.Rename(tmpFile.Name(), downloadPath)(), "Unable to download profiling data:")

Fixed

cmd/admin-profiling-stop.go
vadmeste: > why this `:` at the end, we don't use this convention.

Fixed

vadmeste:
ping @kannappanr

PR #2563, opened 2018-10-01
Vendorize "github.com/minio/minio/pkg/mimedb" to use its content-type resolutions

### LINKED ISSUE #2559: mc does not recognize csv in windows correctly
## Expected behavior

Right now, MIME type for a CSV file while using with mc select shows xls. We have to use our own MIME type library to reliably get the correct content type for csv objects across platforms.

## Actual behavior
Using a CSV object from windows does not work correctly with mc select SQL feature since the content type is not showing csv.

Upload CSV object from a windows computer and see if s3 select works on Linux mc

## Proposed solution
Update the pkg/mime-db package in our own codebase. Update the MIME DB info in our own package, re-vendorize this and test if the correct type is detected.
## mc version

Praveenrajmani:
`minio/mimedb` does not take os-shipped mimetypes. So that the content-type resolutions are
standardized across cross plaforms. 

Fixes #2559


cmd/client-url.go
```diff
@@ -228,7 +228,12 @@ func isURLPrefixExists(urlPrefix string, incomplete bool) bool {
 // on failure just return 'application/octet-stream'.
 func guessURLContentType(urlStr string) string {
 	url := newClientURL(urlStr)
-	contentType := mime.TypeByExtension(filepath.Ext(url.Path))
+	var contentType string
+	if objectExt := filepath.Ext(url.Path); objectExt != "" {
```
harshavardhana: can you create a function similar to mimedb.TypeByExtension? it would be better original package has this function. This can help avoid the large modifications on the `mc` end, making sure that `mimedb` is a drop in replacement for `mime` package.

Praveenrajmani: ok . Will send a PR to minio/minio and then re-vendorize this

kannappanr:
Marked blocked per discussion in this PR.

Praveenrajmani:
vendor updated and the PR is ready for review

vendor/github.com/minio/minio/pkg/mimedb/Makefile
```diff
@@ -0,0 +1,17 @@
+# Generate db.go from db.json downloaded nodejs mime-db project.
+# NOTE: Autogenerated db.go needs to be vet proofed. \
+	Manually edit json -> JSON for all variable names
```
harshavardhana: don't need to vendorize this, ignore this from `git reset HEAD^ vendor/github.com/minio/minio/pkg/mimedb/Makefile` and ammend.

Praveenrajmani: done

Issue #2570, opened 2018-10-15
Minio does not: implement Region in config.json / Retry when http response is AuthorizationHeaderMalformed


quentinselle:
Hi, I'm a developper from Scaleway and we release our object storage. We love use minio client but we can't use it with a signature V4.

## Expected behaviour
```
$> ./mc mb sis/test-minio -debug
mc: <DEBUG> PUT / HTTP/1.1
Host: test-minio.s3.nl-ams.scw.cloud
User-Agent: Minio (darwin; amd64) minio-go/v6.0.8 mc/2018-10-15T15:09:26Z
Content-Length: 150
Authorization: AWS4-HMAC-SHA256 Credential=**REDACTED**/20181015/nl-ams/s3/aws4_request, SignedHeaders=content-md5;host;x-amz-content-sha256;x-amz-date, Signature=**REDACTED**
Content-Md5: 8p7tByecW7sUlTxMbWAA/A==
X-Amz-Content-Sha256: b86017d61d4ad22a838e75e918325b623687c3e84d631bc2b6435970bf12473d
X-Amz-Date: 20181015T151557Z
Accept-Encoding: gzip

mc: <DEBUG> HTTP/1.1 200 OK
Content-Length: 0
Content-Type: text/html; charset=UTF-8
Date: Mon, 15 Oct 2018 15:15:57 GMT
Location: http://test-minio.s3.nl-ams.scw.cloud/test-minio
X-Amz-Id-2: txb6b044c543ba4564a9a2f-005bc4af2d
X-Amz-Request-Id: txb6b044c543ba4564a9a2f-005bc4af2d

mc: <DEBUG> Response Time:  159.607086ms
Bucket created successfully `sis/test-minio`.
```
## Actual behaviour
```
$> ./mc ls sis --debug
mc: <DEBUG> GET / HTTP/1.1
Host: s3.nl-ams.scw.cloud
User-Agent: Minio (darwin; amd64) minio-go/5.0.1 mc/2018-03-25T01:22:22Z
Authorization: AWS4-HMAC-SHA256 Credential=**REDACTED**/20181015/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=**REDACTED**
X-Amz-Content-Sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
X-Amz-Date: 20181015T095239Z
Accept-Encoding: gzip

[...]

PR #2571, opened 2018-10-15
Add Region in config.json file (#2570)

quentinselle:
If not set in config.json, default is still `us-east-1`.

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -128,7 +128,7 @@ func newFactory() func(config *Config) (Client, *probe.Error) {
 			options := minio.Options{
 				Creds:        creds,
 				Secure:       useTLS,
-				Region:       "",
+				Region:       config.Region,
 				BucketLookup: config.Lookup,
 			}
 			api, e = minio.NewWithOptions(hostName, &options)
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -96,6 +96,7 @@ type Config struct {
 	AccessKey   string
 	SecretKey   string
 	Signature   string
+	Region      string
 	HostURL     string
 	AppName     string
 	AppVersion  string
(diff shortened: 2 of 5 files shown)


quentinselle:
This PR is related to this issue https://github.com/minio/mc/issues/2570. If you have any questions, please feel free to ask.

vadmeste:
> This PR is related to this issue #2570. If you have any questions, please feel free to ask.

Listing buckets does not depend on region as @kannappanr mentioned, the region depends on the location of a bucket.

You can enable --debug flag in aws to see that it uses 'us-east-1' by default when region doesn't make sense in the current operation, (see StringToSign)

```
$ aws --debug s3 ls s3://
....
2018-10-22 17:07:39,629 - MainThread - botocore.auth - DEBUG - StringToSign:
AWS4-HMAC-SHA256                 
20181022T160739Z                
20181022/us-east-1/s3/aws4_request
....
```

quentinselle:
Indeed, I discussed with @kannappanr on issue #2570 and we concluded that this Pull Request is useless. We found a better way to fix my problem (-> `raise AuthorizationMalformed`). 

That's why I linked you the issue in the hope you'll read it and give a update to it before reply here (no reply on the issue since last week).

I now return a `AuthorizationMalformed` http error with the valid bucket region (`<Region>$REGION</Region>`) and I'm waiting for your reply (minio).

How do you suggest to implement the retry in `minio-go` ? 
Could you please answer in the issue and close this Pull Request.

Thank you.

poornas:
@quentinselle, minio-go already retries with region when response code is InvalidRegion. response message for AuthorizationHeaderMalformed needs to be parsed to get the correct region and set as errResp.region in  https://github.com/minio/minio-go/blob/master/api-error-response.go#L179 and 
https://github.com/minio/minio-go/blob/master/api.go#L630 to use the errResp.Region in case of AuthorizationHeaderMalformed

kannappanr:
Closing this PR as suggested by @quentinselle

PR #2573, opened 2018-10-17
Add 'users policy'

harshavardhana:


cmd/admin-users-add-policy.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Get the alias parameter from cli
+	args := ctx.Args()
+	aliasedURL := args.Get(0)
+
+	// Create a new Minio Admin Client
+	client, err := newAdminClient(aliasedURL)
+	fatalIf(err, "Cannot get a configured admin connection.")
+
+	fatalIf(probe.NewError(client.SetUserPolicy(args.Get(1), args.Get(2))).Trace(args...), "Cannot set user policy for new user")
```
poornas: might be better to reword as ```Cannot set user policy for user``` as the mc admin users add command also allows overwrite

cmd/admin-users-add-policy.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"github.com/fatih/color"
+	"github.com/minio/cli"
+	"github.com/minio/mc/pkg/console"
+	"github.com/minio/mc/pkg/probe"
+)
+
+var adminUsersPolicyCmd = cli.Command{
+	Name:   "policy",
+	Usage:  "Add policy for user",
```
kannappanr: Can this be `Set policy for user`?

cmd/admin-users-add-policy.go
harshavardhana: fixed

cmd/admin-users-add-policy.go
harshavardhana: fixed

PR #2578, opened 2018-10-24
Fix admin docs add 'mc admin info'

harshavardhana:
Fixes https://github.com/minio/minio/issues/6695


docs/minio-admin-complete-guide.md
```diff
@@ -108,7 +108,9 @@ mc config host add minio http://192.168.1.51:9000 BKIKJAA5BMMU2RHO6IBB V7f1CwQqA
 Get Minio server information from https://play.minio.io:9000
```
kannappanr: Here it starts off as if we are going to have an example for play. But the example is for a local server.

harshavardhana: oh crap @kannappanr

PR #2584, opened 2018-10-27
Allow resuming and stopping an active heal operation

harshavardhana:


vadmeste:
@harshavardhana overall, I think it is just easier to do this way:

- `stop` will stop all healing processes, I don't think a user will find it useful to stop a particular healing process
- `heal --restart` will stop all ongoing healing processes and start a new one
- `heal` without any parameter will try to find if there is an ongoing healing, (I think this is important because healing process takes time sometimes and the user will necessarily want to see the healing report again, mc can print that it will show current healing status)

What do you think ?

harshavardhana:
The reason to do stop at a prefix was necessary because of an existing customer. They are running healing at different prefixes from different servers. So the ability to stop them independently is needed. 

I could try to simplify the command further by not asking let's say token. But instead just resume automatically.

cmd/admin-heal-ui.go
```diff
@@ -393,46 +396,70 @@ func (ui *uiData) UpdateDisplay(s *madmin.HealTaskStatus) (err error) {
...
+	if ui.HealOpts.Recursive {
+		flags += "--recursive "
+	}
+	if ui.HealOpts.DryRun {
+		flags += "--dry-run "
+	}
+	if ui.ClientToken != "" {
+		flags += fmt.Sprintf("--client-token %s ", ui.ClientToken)
+	}
+	if ui.Bucket != "" {
```
vadmeste: It looks like you don't have to add ui.Bucket & ui.Prefix since they are already present in aliasedURL

cmd/admin-heal-ui.go
```diff
@@ -393,46 +396,70 @@ func (ui *uiData) UpdateDisplay(s *madmin.HealTaskStatus) (err error) {
...
 	firstIter := true
 	for {
-		_, res, err = ui.Client.Heal(ui.Bucket, ui.Prefix, *ui.HealOpts,
-			ui.ClientToken, ui.ForceStart)
-		if err != nil {
-			return err
-		}
+		select {
+		case <-trapCh:
+			return errors.New(trapMsg)
```
vadmeste: For better UX, I think you need to return errUserInterrupted here and construct the command in the caller so it can show a better error message:

This is the current situation:

```
mc: <ERROR> Unable to display follow heal status. Healing is backgrounded, to resume watching use `mc admin heal --recursive --client-token 0d966db8-78c1-4eb3-9dab-abf1ed79ed19  myminio/testbucket//testbucket`
```

harshavardhana: Client-token is not needed anymore.

cmd/admin-heal-ui.go
harshavardhana: Done

harshavardhana:
PR is updated and should be used with the server PR https://github.com/minio/minio/pull/6718

kannappanr:
Marking this blocked as this should be merged after PR minio/minio#6718

PR #2592, opened 2018-11-06
Add support of bash completion

### LINKED ISSUE #2572: Support bash completion
## Expected behavior
Currently, no way to implement bash completion - implement it by using https://github.com/posener/complete 

## Actual behavior
No bash completion

mc <tab> doesn't provide any output.

## mc version
master

## System information
All operating systems

vadmeste:
Users of bash/zsh will have completion config enabled automatically in the next
shell session start. This commit provides completion for commands
and all subcommands, their flags and completion of filesystems paths,
mc aliases, buckets and prefixes.

Fixes #2572


harshavardhana:
ping @vadmeste how is this coming along?

harshavardhana:
> * `mc -install` to install mc bash completion

I would suggest we don't do this...

vadmeste:
@harshavardhana can we discuss cloning posener/complete and adapting it to our need. The current WIP code looks ugly to me and I failed to make it better.

Besides, we need cloning because posener/complete doesn't work fine unless it is able to list files with two level deep which can take a lot of time in case of S3. To overcome this, we can tweak posener/complete.

I also want this so we can use BashComplete of minio.cli and not throw it away.

vadmeste:
This is not WIP anymore

harshavardhana:
Thanks @vadmeste - looking good, this will be an awesome change.

vadmeste:
> One bug is we are listing fully recursively for buckets and objects,

Not fully recursive @harshavardhana , it is two levels listing. Two levels listing is needed by posener/complete so it can know if it wants to add a space or not after completing the current level.

> we should only list buckets on first `<tab>` and once a bucket is chosen we should go to next with another `<tab>`

I thought about this and this is a good idea, because preventing two levels listing and make it only one can gain a lot of time for users.

I think in general,

harshavardhana:
> I thought about this and this is a good idea, because preventing two levels listing can make users gain a lot of time, but we need to fork posener/complete library for this

Is it possible to get this upstream first and get an opinion before we fork?

vadmeste:
@harshavardhana I rewrote S3 paths completion code after reading posener/complete.PredictFiles which completes filesystem paths, now our S3 paths completion code behaves as the library expects it. It does only one level listing now (the only exception is when only one directory is found so it recursively scans it)

Listing S3 should be quicker now.

vadmeste:
@harshavardhana @poornas PTAL

vadmeste:
> Can you update the documentation in README.md and remove files in autocomplete/* ?

Done

PR #2594, opened 2018-11-08
Update hint message of `user add`'s usage

zacharyzhao:
Since `ACCESSKEY/SECRETKEY` is a common knowledge for S3 user,
so we should make to clear to avoid confusing people.

### DIFF
--- a/cmd/admin-users-add.go
+++ b/cmd/admin-users-add.go
@@ -35,7 +35,7 @@ var adminUsersAddCmd = cli.Command{
   {{.HelpName}} - {{.Usage}}
 
 USAGE:
-  {{.HelpName}} TARGET USERNAME PASSWORD POLICYNAME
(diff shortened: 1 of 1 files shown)


zacharyzhao:
Hi harshavardhana, thanks for your advise, and I've updated this branch, please check it.

kannappanr:
Since this is a UI change, marked this as blocked to have feedback also from @abperiasamy

kannappanr:
@zacharyzhao We spoke about this. Instead of adding a hint, can you please change the name from `UserName` to `AccessKey` and `Password` to `SecretKey`.

zacharyzhao:
@kannappanr I've updated the branch, please check it, thanks a lot.

ebozduman:
My `LGTM` is still valid.

PR #2598, opened 2018-11-13
Fallback detecting filetype from magic numbers

harshavardhana:


cmd/common-methods.go
```diff
@@ -24,9 +24,11 @@ import (
 	"regexp"
 	"strings"
 
+	"golang.org/x/net/http/httpguts"
+	"gopkg.in/h2non/filetype.v1"
```
ebozduman: We could also have used `net/http` package `DetectContentType()` function instead of `filetype` package.

cmd/common-methods.go
```diff
@@ -115,6 +117,27 @@ func getSourceStream(alias string, urlStr string, fetchStat bool, sseKey string)
 				metadata[k] = v
 			}
 		}
+		// If our reader is a seeker try to detect content-type further.
+		if s, ok := reader.(io.ReadSeeker); ok {
+			if ctype := metadata["Content-Type"]; ctype == "application/octet-stream" {
```
ebozduman: Why do we check and set the file type **only** when the content type metadata is set to the default `"application/octet-stream"`?

That is, we can be paranoid and don't trust the set file type even when it is set to something other than `"application/octet-stream"` and always set it to the detected file type regardless of what content type the source file comes with?

cmd/common-methods.go
harshavardhana: DetectContentType gives wrong values on certain files @ebozduman

cmd/common-methods.go
harshavardhana: because "guessURLContentType" always gives `application/octet-stream` for unrecognized files @ebozduman

harshavardhana:
This is needed for file types without extension @vadmeste

harshavardhana:
ping @vadmeste can you review?

PR #2605, opened 2018-11-20
Add in-place update feature

harshavardhana:


cmd/update-main.go
```diff
@@ -17,274 +17,478 @@
...
 
 USAGE:
-  {{.HelpName}}{{if .VisibleFlags}} [FLAGS]{{end}}
+   {{.HelpName}}{{if .VisibleFlags}} [FLAGS]{{end}}
 {{if .VisibleFlags}}
 FLAGS:
   {{range .VisibleFlags}}{{.}}
   {{end}}{{end}}
+EXIT STATUS:
+   0 - You are already running the most recent version.
```
vadmeste: It is probably easier to merge exit status 0 and 1 to 0, if the user wants more information, he can pass --json and parse output.

cmd/update-main.go
```diff
@@ -17,274 +17,478 @@
...
 
-// GetCurrentReleaseTime - returns this process's release time.  If it is official minio version,
-// parsed version is returned else minio binary's mod time is returned.
+// GetCurrentReleaseTime - returns this process's release time.  If it
+// is official mc version, parsed version is returned else mc
+// binary's mod time is returned.
 func GetCurrentReleaseTime() (releaseTime time.Time, err *probe.Error) {
-	return getCurrentReleaseTime(Version, os.Args[0])
+	if releaseTime, err = mcVersionToReleaseTime(Version); err == nil {
+		return releaseTime, err.Trace(Version)
```
vadmeste: It looks like more `return releaseTime, nil`, isn't it ?

cmd/update-main.go
```diff
@@ -17,274 +17,478 @@
...
+	}
+
+	printMsg(updateMessage{
+		Status:  "success",
+		Message: updateMsg,
+	})
+
+	// if the in-place update is disabled then we shouldn't ask the
+	// user to update the binaries.
+	if strings.Contains(updateMsg, mcReleaseURL) {
```
vadmeste: Not sure when in-place update is disabled, and what's the relation with this strings.Contains(), can you elaborate ?

harshavardhana: There is no reason to perform an update if its not a released binary.

cmd/update-main.go
harshavardhana: oh yes

cmd/update-main.go
harshavardhana: No it is not both needs to be differentiated - see the server implementation.

cmd/update-main.go
```diff
@@ -17,274 +17,478 @@
...
 USAGE:
-  {{.HelpName}}{{if .VisibleFlags}} [FLAGS]{{end}}
+   {{.HelpName}}{{if .VisibleFlags}} [FLAGS]{{end}}
 {{if .VisibleFlags}}
 FLAGS:
   {{range .VisibleFlags}}{{.}}
   {{end}}{{end}}
+EXIT STATUS:
+   0 - you are already running the most recent version
+   1 - new update was applied successfully
```
vadmeste: In shell scripting, it is just weird that 1 means success:

e.g. you can't do like this:
   `mc update && mc config host add ... && ...`

Besides, already running the most recent version and new update was applied successfully both means everything is okay, it doesn't seem to me that some people would need to specifically know that mc did actually update itself, even if so, it seems you can do that using --json flag

cmd/update-main.go
```diff
@@ -17,274 +17,478 @@
...
+// binary's mod time is returned.
 func GetCurrentReleaseTime() (releaseTime time.Time, err *probe.Error) {
-	return getCurrentReleaseTime(Version, os.Args[0])
+	if releaseTime, err = mcVersionToReleaseTime(Version); err == nil {
+		return releaseTime, nil
+	}
+
+	// Looks like version is mc non-standard, we use mc
+	// binary's ModTime as release time:
+	return getModTime(os.Args[0])
```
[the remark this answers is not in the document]
harshavardhana: no, we shouldn't do that @vadmeste we decided that we will not use 1970 because accidentally things can get replaced.   The current implementation is fine as it is. will fix the path construction.

harshavardhana:
ping @vadmeste

PR #2613, opened 2018-11-23
Allow empty objects to be uploaded

harshavardhana:


harshavardhana:
@nitisht @kannappanr can you add reviewers?

harshavardhana:
ping @vadmeste

harshavardhana:
We need this PR before we can make an `mc` release.

PR #2617, opened 2018-12-03
Allow copying file onto itself

harshavardhana:
This allows encrypting an unencrypted file
using server side CopyObject API.

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -365,58 +365,19 @@ func readFile(fpath string) (io.ReadCloser, error) {
 	return fileData, nil
 }
 
-// createFile creates an empty file at the provided filepath
-// if one does not exist already.
-func createFile(fpath string) (io.WriteCloser, error) {
-	if e := os.MkdirAll(filepath.Dir(fpath), 0777); e != nil {
-		return nil, e
-	}
-	file, e := os.Create(fpath)
-	if e != nil {
-		return nil, e
-	}
-	return file, nil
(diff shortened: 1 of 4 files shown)


harshavardhana:
> Tested this, copying file onto itself seems to be working in all cases except this.
> If mc cp is attempted on encrypted object without correct keys, the error message is misleading..
> 
> ```
> ➜  mc git:(0a4cc528) ✗ mc cp --insecure mymino/bucket1/issue  myminio/bucket1/issue
> mc: <ERROR> Failed to copy `https://10.1.1.2:9000/bucket1/issue`. Object does not exist
> ```

Yeah this is a local error message @poornas

poornas:
can you remove the InvalidArgument on https://github.com/minio/mc/blob/master/cmd/client-s3.go#L649 ? 
If an encrypted object is copied onto itself without encryption keys, it will give a better error message of "Bad Request" rather than Object does not exist

vadmeste:
@harshavardhana I don't understand, how we can use mc to copy a file into itself while disabling the encryption ? (I understand the idea but it doesn't seem it is already possible in practice using mc)

harshavardhana:
> can you remove the InvalidArgument check on https://github.com/minio/mc/blob/master/cmd/client-s3.go#L649 ?
> If an encrypted object is copied onto itself without encryption keys, it will give a better error message of "Bad Request" rather than Object does not exist

oh yes sure @poornas

PR #2655, opened 2019-01-23
Add mc admin trace command

poornas:
This will replace MINIO_HTTP_TRACE on minio server and instead make it available as a mc admin trace command. Associated minio PR is https://github.com/minio/minio/pull/7128
 
Current PR needs to be vendorized and rebased after PR 7128 is merged.


donatello:
@poornas Due to compile errors here, I am not sure I can easily test the server PR

cmd/admin-trace.go
```diff
@@ -0,0 +1,164 @@
...
+ FLAGS:
+	 {{range .VisibleFlags}}{{.}}
+	 {{end}}
+ EXAMPLES:
+		 1. Show console trace for a Minio server with alias 'play'
+				$ {{.HelpName}} play
+ `,
+}
+
+const timeFormat = "2006-01-02 15:04:05 -0700"
```
harshavardhana: is this http.TimeFormat?

harshavardhana: so why not use that? why have a custom constant..

poornas:
> @poornas you can also add completion for this new command like this:

done @vadmeste, thanks!

cmd/admin-trace.go
```diff
@@ -0,0 +1,224 @@
...
+
+func (t traceMessage) String() string {
+	var nodeNameStr string
+	if t.NodeName != "" {
+		nodeNameStr = fmt.Sprintf("%s ", colorizedNodeName(t.NodeName))
+	}
+	ri := t.ReqInfo
+	rs := t.RespInfo
+	msg := nodeNameStr + console.Colorize("Request", fmt.Sprintf("[REQUEST %s] ", t.FuncName))
+	msg += fmt.Sprintf("[%s] [%s]\n", t.ReqID, ri.Time.Format(timeFormat))
```
donatello: If this function is called often, it is more efficient to avoid string concatenation with `+` and `+=`. `strings.Builder` provides a way that does not involve much copying.

cmd/admin-trace.go
```diff
@@ -0,0 +1,295 @@
...
+	// strip off extra newline added by json encoder
+	bufStr := string(buf.Bytes())
+	return strings.TrimSuffix(bufStr, "\n")
+}
+
+func (t traceMessage) String() string {
+	var nodeNameStr string
+	trc := t.Trace
+	if trc.NodeName != "" {
+		nodeNameStr = fmt.Sprintf("%s ", colorizedNodeName(trc.NodeName))
```
krisis: In the interest of efficient use of memory, we could use [strings.Builder](https://golang.org/pkg/strings/#Builder) type to construct the `String` representation of `traceMessage` value.

poornas: done.

PR #2662, opened 2019-01-29
Add mc rb command to remove empty bucket

### LINKED ISSUE #2635: `mc rm` with --older-than flag removes buckets, too
## Expected behavior

There should be a way to delete all *objects* in a bucket using the `--older-than` flag, but not the *bucket* itself.
I would expect that this happens when running `mc rm -r --older-than 30 --force minio/bucket/`.

## Actual behavior

The bucket was deleted as well. (Just for the record, it *was* older than 30 days.)

1. Have a bucket older than `x` days.
2. Run `mc rm -r --older-than x --force minio/bucket/`.

Alternatively:

```
mc mb minio/bucket/
mc rm -r --older-than 0 --force minio/bucket/
```

## mc version

```
$ mc version
Version: 2018-12-27T00:37:49Z
Release-tag: RELEASE.2018-12-27T00-37-49Z
Commit-id: 4e7eeb0725abcd36b54a37636ff89f29c3238fd8
```

## System information

Fedora 29

poornas:
Fixes #2635 - Change mc rm --recursive behavior to
delete objects in bucket but not the
bucket itself.

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -36,7 +36,7 @@ import (
 
 	"github.com/minio/mc/pkg/httptracer"
 	"github.com/minio/mc/pkg/probe"
-	"github.com/minio/minio-go"
+	minio "github.com/minio/minio-go"
 	"github.com/minio/minio-go/pkg/credentials"
 	"github.com/minio/minio-go/pkg/encrypt"
 	"github.com/minio/minio-go/pkg/policy"
@@ -779,15 +779,13 @@ func (c *s3Client) removeIncompleteObjects(bucket string, objectsCh <-chan strin
 }
 
 // Remove - remove object or bucket(s).
-func (c *s3Client) Remove(isIncomplete bool, contentCh <-chan *clientContent) <-chan *probe.Error {
+func (c *s3Client) Remove(isIncomplete, isRemoveBucket bool, contentCh <-chan *clientContent) <-chan *probe.Error {
 	errorCh := make(chan *probe.Error)
 
 	prevBucket := ""
 	// Maintain objectsCh, statusCh for each bucket
 	var objectsCh chan string
 	var statusCh <-chan minio.RemoveObjectError
-
-	isRemoveBucket := false
 
 	go func() {
 		defer close(errorCh)
@@ -819,7 +817,6 @@ func (c *s3Client) Remove(isIncomplete bool, contentCh <-chan *clientContent) <-
 					}
 				}
 				// Re-init objectsCh for next bucket
-				isRemoveBucket = false
 				objectsCh = make(chan string)
 				if isIncomplete {
 					statusCh = c.removeIncompleteObjects(bucket, objectsCh)
@@ -844,7 +841,6 @@ func (c *s3Client) Remove(isIncomplete bool, contentCh <-chan *clientContent) <-
 				}
 			} else {
 				// end of bucket - close the objectsCh
-				isRemoveBucket = true
 				if objectsCh != nil {
 					close(objectsCh)
 				}
--- a/cmd/mirror-main.go
+++ b/cmd/mirror-main.go
@@ -232,8 +232,8 @@ func (mj *mirrorJob) doRemove(sURLs URLs) URLs {
 	contentCh := make(chan *clientContent, 1)
 	contentCh <- &clientContent{URL: *newClientURL(sURLs.TargetContent.URL.Path)}
 	close(contentCh)
-
-	errorCh := clnt.Remove(false, contentCh)
+	isRemoveBucket := false
+	errorCh := clnt.Remove(false, isRemoveBucket, contentCh)
 	for pErr := range errorCh {
 		if pErr != nil {
 			switch pErr.ToGoError().(type) {
--- a/cmd/rb-main.go
+++ b/cmd/rb-main.go
@@ -1,0 +1,210 @@
+/*
+ * Minio Client (C) 2019 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"encoding/json"
+	"path/filepath"
+
+	"github.com/fatih/color"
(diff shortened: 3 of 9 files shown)


harshavardhana:
We should mention this as a breaking change...

cmd/mirror-main.go
```diff
@@ -232,8 +232,8 @@ func (mj *mirrorJob) doRemove(sURLs URLs) URLs {
 	contentCh := make(chan *clientContent, 1)
 	contentCh <- &clientContent{URL: *newClientURL(sURLs.TargetContent.URL.Path)}
 	close(contentCh)
-
-	errorCh := clnt.Remove(false, contentCh)
+	isRemoveBucket := true
```
Praveenrajmani: I think we should not remove the bucket in the target if the source has zero contents. This should be `false`.
 
image

Fine if it is left intentionally to work like this.

harshavardhana:
The intention here is bucket is never removed from `mc mirror` or `mc cp` operation. Even with `mc rm` we have decided to not remove buckets anymore. 

To remove buckets explicitly only command is `mc rb`

cmd/mirror-main.go
poornas: @Praveenrajmani , changed it so mirror will not remove buckets

docs/minio-client-complete-guide.md
```diff
@@ -314,6 +315,30 @@ mc mb s3/mybucket --region=us-west-1
 Bucket created successfully ‘s3/mybucket’.
 ```
 
+<a name="rb"></a>
+### Command `rb` - Remove a Bucket
+`rb` command removes a bucket and all its contents on an object storage. On a filesystem, it behaves like `rmdir` command.
+
+Note that when a bucket is removed all policies associated with the bucket will also be removed. If you would like to just
```
Praveenrajmani: One comment - missed a `,` here. ` when a bucket is removed, all policies `

vadmeste:
@poornas after introducing rb command, do you expect that rm is still able to remove objects from different buckets at the same time ?

poornas:
> @poornas after introducing rb command, do you expect that rm is still able to remove objects from different buckets at the same time ?

@vadmeste, yes - both the commands below will remove objects from different buckets, except that mc rb will delete bucket in addition. 
```
➜  mc git:(rb) mc rb --force s3/tbucket11 s3/tbucket13
Bucket removed successfully `s3/tbucket11`.
Bucket removed successfully `s3/tbucket13`.
```
and 
```
➜  mc git:(rb) mc rm --r --force s3/tbucket11 s3/tbucket13
```

vadmeste:
@poornas I see, but I also mean this:
     `mc rm alias/`

does it also need to traverse several buckets and remove all objects in them ?

poornas:
> @poornas I see, but I also mean this:
> `mc rm alias/`
> 
> does it also need to traverse several buckets and remove all objects in them ?
@vadmeste , that functionality also remains the same as before.

cmd/rb-main.go
```diff
@@ -0,0 +1,207 @@
...
+		if !isForce {
+			if isNamespaceRemoval(targetURL) {
+				fatalIf(errDummy().Trace(),
+					"This operation results in **site-wide** removal of buckets. If you are really sure, retry this command with ‘--force’ flags.")
+			}
+			if !isEmpty {
+				fatalIf(errDummy().Trace(), "`"+targetURL+"` is not empty. Retry this command with ‘--force’ flag if you want to remove `"+targetURL+"` and all its contents")
+			}
+		}
+		if e := deleteBucket(targetURL); e == nil {
```
vadmeste: Why we don't print errors (`err != nil` case) ?

cmd/rb-main.go
```diff
@@ -0,0 +1,207 @@
...
+	for _, targetURL := range ctx.Args() {
+		// Instantiate client for URL.
+		clnt, err := newClient(targetURL)
+		if err != nil {
+			errorIf(err.Trace(targetURL), "Invalid target `"+targetURL+"`.")
+			cErr = exitStatus(globalErrorExitStatus)
+			continue
+		}
+		isEmpty := true
+		for range clnt.List(true, false, DirNone) {
```
vadmeste: I think you need to pass DirFirst here, we can have some empty directories sometimes

poornas: @vadmeste, it is my understanding that DirFirst option would allow listing directory entries ahead of files. Here, the purpose is to ascertain if the bucket in targeturl is empty , hence listing with DirNone. fwiw, I tested with empty directory and it behaves correctly.
```
➜  minio git:(master) mc ls myminio/newb --r
[2019-02-13 23:01:58 PST]     0B em3/
➜  minio git:(master) mc mb myminio/newb    
Bucket created successfully `myminio/newb`.
➜  minio git:(master) mc ls myminio/newb --r    
[2019-02-13 23:03:47 PST]     0B em3/
➜  minio git:(master) mc rb myminio/newb
mc: <ERROR> `myminio/newb` is not empty. Retry this command with ‘--force’ flag if you want to remove `myminio/newb` and all its contents 
➜  minio git:(master) mc rb myminio/newb --force
Bucket removed successfully `myminio/newb`.
```

cmd/rb-main.go
poornas: @vadmeste,changed this to print error also.

cmd/client-s3.go
```diff
@@ -36,7 +36,7 @@ import (
 
 	"github.com/minio/mc/pkg/httptracer"
 	"github.com/minio/mc/pkg/probe"
-	"github.com/minio/minio-go"
+	minio "github.com/minio/minio-go"
```
kannappanr: @poornas This code has not been used anywhere else in this file. Is this change required?

docs/minio-client-complete-guide.md
```diff
@@ -255,7 +256,7 @@ Skip SSL certificate verification.
 | [**diff** - Diff buckets](#diff) |[**policy** - Set public policy on bucket or prefix](#policy)  |[**session** - Manage saved sessions](#session) |
 | [**config** - Manage config file](#config)  | [**watch** - Watch for events](#watch)  | [**event** - Manage events on your buckets](#event)  |
 | [**update** - Manage software updates](#update)  | [**version** - Show version](#version)  | [**stat** - Stat contents of objects and folders](#stat) |
-| [**head** - Display first 'n' lines of an object](#head) | [**sql** - Run sql queries on objects](#sql) | |
+| [**head** - Display first 'n' lines of an object](#head) | [**sql** - Run sql queries on objects](#sql) | [**rb** - Remove a bucket](#rb) |
```
kannappanr: It is better to place `rb` next to `mb`

cmd/client-s3.go
harshavardhana: This is perhaps automatic with `goimports` through the editor.

docs/minio-client-complete-guide.md
poornas: done @kannappanr

PR #2663, opened 2019-01-29
Edit the delete object message

sinhaashish:
Upon deletion of object the message was `Removing <<Object-Name>>` which has been changed to
`Attempting to remove <<Object-Name>>`

### DIFF
--- a/cmd/rm-main.go
+++ b/cmd/rm-main.go
@@ -129,7 +129,7 @@ type rmMessage struct {
 
 // Colorized message for console printing.
 func (r rmMessage) String() string {
-	return console.Colorize("Remove", fmt.Sprintf("Removing `%s`.", r.Key))
+	return console.Colorize("Remove", fmt.Sprintf("Attempting to remove `%s`.", r.Key))
 }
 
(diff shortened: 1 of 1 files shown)


cmd/rm-main.go
```diff
@@ -129,7 +129,7 @@ type rmMessage struct {
 
 // Colorized message for console printing.
 func (r rmMessage) String() string {
-	return console.Colorize("Remove", fmt.Sprintf("Removing `%s`.", r.Key))
+	return console.Colorize("Remove", fmt.Sprintf("Attempting to remove `%s`.", r.Key))
```
harshavardhana: We should not add this message, we should print the message after successful remove @sinhaashish so the message should be "Removed %s"

kannappanr:
We should not try to delete a bucket if any object deletion fails, as the bucket deletion will definitely fail.

poornas:
@sinhaashish, quiet mode should not print info messages, only failures if any.

kannappanr:
After talking to @abperiasamy  we have decided not to make this change.

Issue #2636, opened 2018-12-31
metadata option in mc


lavvy:
## Expected behavior
./mc  cp myobj play/bucket/myobj  --metadata KeyName1=string,KeyName2=string
## Actual behavior
Error . 
Every other sdk has a way to add custom metadata but mc.

harshavardhana:
mc is not an SDK - you may use an SDK to do this right now.

lavvy:
thanks for the quick response. Yea we are using sdk for it now but mc allows us so much portability between applications . But thanks though cos i thought it was intentionally omitted were it mattered most. 
By the way aws cli has it covered but for unknown reason its not as fast as mc.
Is there any timeline for it ? @harshavardhana

harshavardhana:
> thanks for the quick response. Yea we are using sdk for it now but mc allows us so much portability between applications . But thanks though cos i thought it was intentionally omitted were it mattered most.
> By the way aws cli has it covered but for unknown reason its not as fast as mc.
> Is there any timeline for it ?

There is no easier way to do it @lavvy , we would like to limit `mc's` command line options.  mc is a large scale tool copies million of files. Does this option apply for one file or all files? is complicated to reason.

This can lead to unexpected metadata accumulation on files which were not meant to have this metadata. `aws cli` implements lot more features which we do not implement because it simply doesn't make sense, `mc` is built with a different idea in mind to keep things simple and not be equivalent of what an SDK can do.  Our SDKs are as portable as `mc` written in a portable language, so I see no reason why you would lose portability in your application. Unless you are making platform-specific assumptions.

lavvy:
i have read a lot about your advises in many metadata related issues where you discouraged much use of it. I really could not decipher if its bcos of technical or ethical reasons. But assuming it's any of those or both.
In my opinion it should be better use case if you could include it in mc too cos you know one can easily use the go sdk or python script to achieve it .
##Current behaviour
So at the end the developer will have more scripts to manage .
Every dev trying to implement it will end up with different ethics  which will clutter or fragment more. 
Uncontrolled use of metadata will still find its way into the servers
##Expected behaviour
More homogenity 
More predictable outcomes
Controlled ethics of how metadata is used in the system.

harshavardhana:
> i have read a lot about your advises in many metadata related issues where you discouraged much use of it. I really could not decipher if its bcos of technical or ethical reasons. But assuming it's any of those or both.

[...]

PR #2671, opened 2019-02-02
Add metadata in mc copy

sinhaashish:
mc copy provisioned to pass metadata as string.

Example : mc cp --metadata key1=value1,key=value2 play/mybucket/burningman2011/ ~/latest/

Fixes: [2639](https://github.com/minio/mc/issues/2636)

### DIFF
--- a/cmd/cp-main.go
+++ b/cmd/cp-main.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -57,8 +58,15 @@ var (
 			Name:  "encrypt",
 			Usage: "encrypt/decrypt objects (using server-side encryption with server managed keys)",
 		},
+		cli.StringFlag{
+			Name:  "attr",
+			Usage: "add custom metadata for the object",
+		},
 	}
 )
+
+// ErrInvalidMetadata reflects invalid metadata format
+var ErrInvalidMetadata = errors.New("specified metadata should be of form key1=value1,key2=value2,... and so on")
 
 // Copy command.
 var cpCmd = cli.Command{
@@ -107,7 +115,14 @@ EXAMPLES:
 
    9. Copy a folder with encrypted objects recursively from Amazon S3 to Minio cloud storage.
       $ {{.HelpName}} --recursive --encrypt-key "s3/documents/=32byteslongsecretkeymustbegiven1,myminio/documents/=32byteslongsecretkeymustbegiven2" s3/documents/ myminio/documents/
-`,
+
+  10. Copy a list of objects from local file system to Minio cloud storage with specified metadata.
+			$ {{.HelpName}} --attr key1=value1,key2=value2 Music/*.mp4 play/mybucket/
+			
+	11. Copy a folder recursively from Minio cloud storage to Amazon S3 cloud storage with specified metadata.
+			$ {{.HelpName}} --attr key1=value1,key2=value2 --recursive play/mybucket/burningman2011/ s3/mybucket/
+
+ `,
 }
 
 // copyMessage container for file copy messages
@@ -368,6 +383,19 @@ func doCopySession(session *sessionV8, encKeyDB map[string][]prefixSSEPair) erro
 					cpURLs.TargetContent.Metadata["X-Amz-Storage-Class"] = session.Header.CommandStringFlags["storage-class"]
 				}
 
+				//	metaMap, metaSet := session.Header.UserMetaData
+
+				// Check and handle metadata if passed in command line args
+
+				if len(session.Header.UserMetaData) != 0 {
+					if cpURLs.TargetContent.UserMetadata == nil {
+						cpURLs.TargetContent.UserMetadata = make(map[string]string)
+					}
+					for metaDataKey, metaDataVal := range session.Header.UserMetaData {
+						cpURLs.TargetContent.UserMetadata[metaDataKey] = metaDataVal
+					}
+				}
+
 				// Verify if previously copied, notify progress bar.
 				if isCopied(cpURLs.SourceContent.URL.String()) {
 					queueCh <- func() URLs {
@@ -439,11 +467,32 @@ loop:
 	return retErr
 }
 
+// validate the passed metadataString and populate the map
+func getMetaDataEntry(metadataString string) (map[string]string, *probe.Error) {
+	metaDataMap := make(map[string]string)
+	for _, metaData := range strings.Split(metadataString, ",") {
+		metaDataEntry := strings.Split(metaData, "=")
+		if len(metaDataEntry) == 2 {
+			metaDataMap[metaDataEntry[0]] = metaDataEntry[1]
+		} else {
+			return nil, probe.NewError(ErrInvalidMetadata)
+		}
+	}
+	return metaDataMap, nil
+}
+
 // mainCopy is the entry point for cp command.
 func mainCopy(ctx *cli.Context) error {
 	// Parse encryption keys per command.
 	encKeyDB, err := getEncKeys(ctx)
 	fatalIf(err, "Unable to parse encryption keys.")
+
+	// Parse metadata.
+	userMetaMap := make(map[string]string)
+	if ctx.String("attr") != "" {
+		userMetaMap, err = getMetaDataEntry(ctx.String("attr"))
+		fatalIf(err, "Unable to parse attribute %v", ctx.String("attr"))
+	}
 
 	// check 'copy' cli arguments.
 	checkCopySyntax(ctx, encKeyDB)
@@ -469,6 +518,7 @@ func mainCopy(ctx *cli.Context) error {
 	session.Header.CommandStringFlags["storage-class"] = storageClass
 	session.Header.CommandStringFlags["encrypt-key"] = sseKeys
 	session.Header.CommandStringFlags["encrypt"] = sse
+	session.Header.UserMetaData = userMetaMap
 
 	var e error
 	if session.Header.RootPath, e = os.Getwd(); e != nil {
--- a/cmd/cp-main_test.go
+++ b/cmd/cp-main_test.go
@@ -1,0 +1,61 @@
+/*
+ * Minio Client (C) 2019 Minio, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestParseMetaData(t *testing.T) {
+	metaDataCases := []struct {
+		input  string
+		output map[string]string
+		err    error
+		status bool
+	}{
+		// success scenerio
+		{"key1=value1,key2=value2", map[string]string{"key1": "value1", "key2": "value2"}, nil, true},
+		// using different delimitter, other than ',' between multiple meta data
+		{"key1=value1;key2=value2", nil, ErrInvalidMetadata, false},
+		// using different delimitter, other than '=' between key value
+		{"key1:value1,key2:value2", nil, ErrInvalidMetadata, false},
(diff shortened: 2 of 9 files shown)


oliverkane:
Subscribing to this as this looks useful for a personal project of mine.

functional-tests.sh
```diff
@@ -334,6 +334,20 @@ function test_put_object_with_storage_class_error()
...
 }
 
+## Test mc cp command with metadata set to a json string
+function test_put_object_with_metadata()
+{
+    show "${FUNCNAME[0]}"
+
+    start_time=$(get_time)
+    object_name="mc-test-object-$RANDOM"
+    assert_failure "$start_time" "${FUNCNAME[0]}" mc_cmd cp --attr '{"key1":"value1","key1":"value1"}' "${FILE_1_MB}" "${SERVER_ALIAS}/${BUCKET_NAME}/${object_name}"
```
harshavardhana: This functional test is incomplete, we should also add `mc stat` to validate if the attribute is set.

cmd/cp-main.go
```diff
@@ -438,6 +464,29 @@ loop:
 	return retErr
 }
 
+// parse the passed MetaData and populate the map
+func parseMetaData(metadataString string) (map[string]string, error) {
```
harshavardhana: Can be simplified

```go
func parseMetaData(metadataString string) (map[string]string, error) {
	metaDataMap := make(map[string]string)
	for _, metaData := range strings.Split(metadataString, ",") {
		metaDataEntry := strings.Split(metaData, "=")
		if len(metaDataEntry) == 2 {
			metaDataMap[metaDataEntry[0]] = metaDataEntry[1]
			continue
		} else {
			return nil, errors.New("Unable to parse metadata")
		}
	}
}
```

cmd/cp-main.go
```diff
@@ -359,14 +368,31 @@ func doCopySession(session *sessionV8, encKeyDB map[string][]prefixSSEPair) erro
 				// Save totalSize.
 				cpURLs.TotalSize = session.Header.TotalBytes
 
-				// Check and handle storage class if passed in command line args
-				if _, ok := session.Header.CommandStringFlags["storage-class"]; ok {
+				t, storageclassPresent := session.Header.CommandStringFlags["storage-class"]
+				fmt.Print(" ashishshshshshshshshsshshshshshshsh", t)
```
harshavardhana: Remove debug statements @sinhaashish before submitting a PR as practice.

cmd/cp-main.go
```diff
@@ -359,14 +368,31 @@ func doCopySession(session *sessionV8, encKeyDB map[string][]prefixSSEPair) erro
 				// Save totalSize.
 				cpURLs.TotalSize = session.Header.TotalBytes
 
-				// Check and handle storage class if passed in command line args
-				if _, ok := session.Header.CommandStringFlags["storage-class"]; ok {
+				t, storageclassPresent := session.Header.CommandStringFlags["storage-class"]
```
harshavardhana: s/storageClassPresent/scSet/g is better to be concise..

cmd/cp-main.go
```diff
@@ -359,14 +368,31 @@ func doCopySession(session *sessionV8, encKeyDB map[string][]prefixSSEPair) erro
...
+				fmt.Print(" ashishshshshshshshshsshshshshshshsh", t)
+				fmt.Print("\n \n  ashishshshshshshshshsshshshshshshsh \n \n ", t)
+
+				metaDataString, metadataPresent := session.Header.CommandStringFlags["metadata"]
```
harshavardhana: s/metadataPresent/metaSet/g is better to be concise..

cmd/cp-main.go
```diff
@@ -459,6 +508,7 @@ func mainCopy(ctx *cli.Context) error {
 		sseKeys = key
 	}
 	sse := ctx.String("encrypt")
+	userMetadata := ctx.String("attr")
```
harshavardhana: We should always parse the input here, i.e parseMetadata should be used here.. look at checkCopySyntax or getEncKeys for adding validations..

cmd/cp-main.go
```diff
@@ -57,6 +58,10 @@ var (
 			Name:  "encrypt",
 			Usage: "encrypt/decrypt objects (using server-side encryption with server managed keys)",
 		},
+		cli.StringFlag{
+			Name:  "attr",
```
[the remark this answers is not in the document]
sinhaashish: It was discussed and thus `attr` is used. You can find in the comments of [issue](https://github.com/minio/mc/issues/2636)

cmd/cp-main.go
```diff
@@ -438,12 +454,34 @@ loop:
...
 }
 
+// validate the passed metadataString and populate the map
+func getMetaDataEntry(metadataString string) (map[string]string, *probe.Error) {
+	metaDataMap := make(map[string]string)
+	for _, metaData := range strings.Split(metadataString, ",") {
+		metaDataEntry := strings.Split(metaData, "=")
+		if len(metaDataEntry) == 2 {
+			metaDataMap[metaDataEntry[0]] = metaDataEntry[1]
+			continue
```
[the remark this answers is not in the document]
sinhaashish: removed.

cmd/cp-main.go
```diff
@@ -438,12 +454,34 @@ loop:
...
+		} else {
+			return nil, probe.NewError(errors.New("specified metadata should be of form key1=value1,key2=value2,... and so on"))
```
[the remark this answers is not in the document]
sinhaashish: Added

cmd/cp-main_test.go
```diff
@@ -0,0 +1,64 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	//"errors"
```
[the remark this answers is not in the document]
sinhaashish: Removed

cmd/cp-main.go
```diff
@@ -438,12 +454,34 @@ loop:
...
 func mainCopy(ctx *cli.Context) error {
 	// Parse encryption keys per command.
 	encKeyDB, err := getEncKeys(ctx)
 	fatalIf(err, "Unable to parse encryption keys.")
 
+	// Parse metadata.
+	metaMap := make(map[string]string)
+	if ctx.String("attr") != "" {
+		metaMap, err = getMetaDataEntry(ctx.String("attr"))
+		fatalIf(err, "Unable to parse metadata.")
```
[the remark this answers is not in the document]
sinhaashish: modified

cmd/cp-main.go
```diff
@@ -359,14 +368,21 @@ func doCopySession(session *sessionV8, encKeyDB map[string][]prefixSSEPair) erro
 				// Save totalSize.
 				cpURLs.TotalSize = session.Header.TotalBytes
 
-				// Check and handle storage class if passed in command line args
-				if _, ok := session.Header.CommandStringFlags["storage-class"]; ok {
+				_, scSet := session.Header.CommandStringFlags["storage-class"]
+				metaMap, metaSet := session.Header.MetaData["metadata"]
+
+				// Check and handle metadata or storage class  if passed in command line args
+				if scSet || metaSet {
```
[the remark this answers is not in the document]
sinhaashish: Done.

cmd/cp-main.go
```diff
@@ -107,7 +112,11 @@ EXAMPLES:
 
    9. Copy a folder with encrypted objects recursively from Amazon S3 to Minio cloud storage.
       $ {{.HelpName}} --recursive --encrypt-key "s3/documents/=32byteslongsecretkeymustbegiven1,myminio/documents/=32byteslongsecretkeymustbegiven2" s3/documents/ myminio/documents/
-`,
+
+  10. Copy files and set the metadata  provided as json string.
```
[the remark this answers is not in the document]
sinhaashish: updated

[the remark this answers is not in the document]
sinhaashish: updated

cmd/cp-main.go
```diff
@@ -57,6 +58,10 @@ var (
 			Name:  "encrypt",
 			Usage: "encrypt/decrypt objects (using server-side encryption with server managed keys)",
 		},
+		cli.StringFlag{
+			Name:  "attr",
+			Usage: "add custom metadata string",
```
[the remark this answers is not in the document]
sinhaashish: updated

cmd/cp-main.go
```diff
@@ -57,9 +58,16 @@ var (
...
 		},
+		cli.StringFlag{
+			Name:  "attr",
+			Usage: "add custom metadata for the object",
+		},
 	}
 )
 
+// ErrInvalidMetadata reflects invalid metadata format
+var ErrInvalidMetadata = probe.NewError(errors.New("specified metadata should be of form key1=value1,key2=value2,... and so on"))
```
harshavardhana: This is not a correct style of error when using probe package for example look at errDummy() implementation, you should use that style. 

Using probe in this manner adds the line number of this definition meaning mc --debug will show error is originating here which is not we need to stop at the function return.

harshavardhana: Also the error to point to the exact position to indicate the error, basically use the input Arg and show in which position you see this problem. If that's not possible we can simply show the entire arg and indicate it doesn't follow expected style.

krisis: OK, so can we have, `var ErrInvalidMetadata = errors.New("the appropriate error message")` and call `probe.NewError(ErrInvalidMetadata)` at the respective call sites?

cmd/cp-main.go
```diff
@@ -367,6 +379,20 @@ func doCopySession(session *sessionV8, encKeyDB map[string][]prefixSSEPair) erro
...
 				}
 
+				metaMap, metaSet := session.Header.MetaData["metadata"]
+
+				// Check and handle metadata if passed in command line args
+				if metaSet {
+					if cpURLs.TargetContent.Metadata == nil {
+						cpURLs.TargetContent.Metadata = make(map[string]string)
+					}
+					if metaSet && len(metaMap) != 0 {
```
krisis: we don't need to check `metaSet` again.

cmd/cp-main_test.go
```diff
@@ -0,0 +1,63 @@
...
+	metaDataCases := []struct {
+		input  string
+		output map[string]string
+		err    error
+		status bool
+	}{
+		// success scenerio
+		{"key1=value1,key2=value2", map[string]string{"key1": "value1", "key2": "value2"}, nil, true},
+		// using different delimitter, other than ',' between multiple meta data
+		{"key1=value1;key2=value2", nil, errors.New("specified metadata should be of form key1=value1,key2=value2,... and so on"), false},
```
krisis: we must be able to reuse `ErrInvalidMetadata` in the tests instead of `errors.New(..)`.

cmd/cp-main.go
harshavardhana: yes that is possible @krisis

cmd/cp-main_test.go
sinhaashish: done.

cmd/cp-main.go
sinhaashish: removed `metsSet`

sinhaashish:
@Praveenrajmani  @krisis  PTAL

cmd/cp-main.go
```diff
@@ -107,7 +115,11 @@ EXAMPLES:
 
    9. Copy a folder with encrypted objects recursively from Amazon S3 to Minio cloud storage.
       $ {{.HelpName}} --recursive --encrypt-key "s3/documents/=32byteslongsecretkeymustbegiven1,myminio/documents/=32byteslongsecretkeymustbegiven2" s3/documents/ myminio/documents/
-`,
+
+  10. Copy files and set the metadata provided as a string.
```
krisis: The description says, "Copy files ...", but the example command copies from an object store to another, also recursively. We should provide two examples,
1. for copying a file to an object store
2. for copying an object from an object store to another.

On a related note, what happens when the source object already has some metadata set on it matching the supplied key-value pairs via `attr` flag?

sinhaashish: Added the examples for the above scenario.
>On a related note, what happens when the source object already has some metadata set on it matching the supplied key-value pairs via attr flag?

The value is overridden with the new value passed as the parameter in `mc --attr`

sinhaashish:
Added the code to append the metadata in case of server side copy with `--attr` flag
```  
1.  mc cp  --attr key1=val1 <<local_folder>> myminio/test/one 
2. mc cp --attr key2=val2 myminio/test/one myminio/test/two 
```
Object `two` has both `key1=val1`  and `key2=val2` as its metadata

@krisis @Praveenrajmani @harshavardhana  PTAL

PR #2674, opened 2019-02-06
Vendorize all the latest dependencies with go mod

harshavardhana:


kannappanr:
@harshavardhana PTAL at Travis failures

wlan0:
Please update travis.yaml to use GOMODULES111=on. 

Please ensure that this builds with previous version of go as well.

harshavardhana:
> Please update travis.yaml to use GOMODULES111=on.
> 
> Please ensure that this builds with previous version of go as well.

all of it fixed @wlan0

donatello:
So in go 1.11 my understanding is that `$GOPATH` is no longer needed to manage deps - if this is the case, this PR is not building when GOPATH is not set. It builds in go 1.10 though.

wlan0:
@donatello What is the value of the env var `GO111MODULE`

harshavardhana:
> So in go 1.11 my understanding is that `$GOPATH` is no longer needed to manage deps - if this is the case, this PR is not building when GOPATH is not set. It builds in go 1.10 though.

GOPATH issue is fixed @donatello

Issue #2644, opened 2019-01-14
Update readme file with ppc64le binary


mkumatag:
Currently, I see minio client binary is already published for ppc64le but readme is not updated with the relevant link, this task is for updating the readme file with a proper link.

Reference: https://github.com/minio/minio/issues/7062#issuecomment-452970812

Readme link - https://github.com/minio/mc#gnulinux

Readme has been already fixed for [minio/minio project](https://github.com/minio/minio/issues/7063).

kannappanr:
@mkumatag We will add a link to the ppc binary.

PR #2677, opened 2019-02-06
Update mc quickstart guide with PPC binary link

eco-minio:
Linux specific


eco-minio:
Fixes #2644

harshavardhana:
Generally advisable to send one PR for entire fix, rather than fixing three files in three different PRs.

PR #2681, opened 2019-02-11
mirror: Proper exit when preparing diff returns an error

### LINKED ISSUE #2682: Rare crash during mc mirror
## Expected behavior
No crash

## Actual behavior

```
panic: send on closed channel

goroutine 152 [running]:
github.com/minio/mc/cmd.(*ParallelManager).addWorker.func1(0xc0003ae960)
    /q/.q/sources/gopath/src/github.com/minio/mc/cmd/parallel-manager.go:80 +0x70
created by github.com/minio/mc/cmd.(*ParallelManager).addWorker
    /q/.q/sources/gopath/src/github.com/minio/mc/cmd/parallel-manager.go:69 +0x78
```

This happens when Minio server returns an error during listing,

You can fake the error using the following diff: 

```diff
diff --git a/cmd/fs-v1.go b/cmd/fs-v1.go
index 98e9cc60..5bb47b2f 100644
--- a/cmd/fs-v1.go
+++ b/cmd/fs-v1.go
@@ -1098,9 +1098,17 @@ func (fs *FSObjects) getObjectETag(ctx context.Context, bucket, entry string, lo
 	return extractETag(parseFSMetaMap(fsMetaBuf)), nil
 }
 
+var globalListObjects int
+
 // ListObjects - list all objects at prefix upto maxKeys., optionally delimited by '/'. Maintains the list pool
 // state for future re-entrant list requests.
 func (fs *FSObjects) ListObjects(ctx context.Context, bucket, prefix, marker, delimiter string, maxKeys int) (loi ListObjectsInfo, e error) {
+
+	globalListObjects++
+	if globalListObjects > 8 {
+		return loi, errInvalidArgument
+	}
+
 	if err := checkListObjsArgs(ctx, bucket, prefix, marker, delimiter, fs); err != nil {
 		return loi, err
 	}
```

Then mc mirror from source to target where source & target are two different aliases, the server that needs to have a faked error should be the source

## mc version
master

## System information
All

vadmeste:
A bad synchornization between startMirror() and its caller mirror() happens when preparing the difference of objects between source & target returns an error: mirror() quits when it receives an error from startMirror() without give a chance for that latter to properly stop the parallel mirror manager.

This PR will:
1. Quit mirroring when listing returns an error: this means calculating diff needs to stop and no further action needs to be taken since it can be dangerous especially when --remove flag is passed.
2. Properly stop parallel mirroring manager before quitting.

Fixes #2682

### DIFF
--- a/cmd/mirror-main.go
+++ b/cmd/mirror-main.go
@@ -516,8 +516,9 @@ func (mj *mirrorJob) startMirror(ctx context.Context, cancelMirror context.Cance
 				return
 			}
 			if sURLs.Error != nil {
+				stopParallel()
 				errCh <- sURLs.Error
-				continue
+				return
 			}
 
 			if sURLs.SourceContent != nil {


harshavardhana:
We did add `continue` for a reason can you check why that was ? `git blame` perhaps ?

kannappanr:
> We did add `continue` for a reason can you check why that was ? `git blame` perhaps ?

https://github.com/minio/mc/pull/1784

vadmeste:
@harshavardhana I really dislike continuing anything when difference between source & target returns an error. First, it doesn't make sense for me to continue, I don't think we need this, also it is just dangerous to continue especially when --remove flag is passed.

harshavardhana:
> @harshavardhana I really dislike continuing anything when difference between source & target returns an error. First, it doesn't make sense for me to continue, I don't think we need this, also it is just dangerous to continue especially when --remove flag is passed.

Agreed @vadmeste

cmd/mirror-main.go
```diff
@@ -516,8 +516,9 @@ func (mj *mirrorJob) startMirror(ctx context.Context, cancelMirror context.Cance
 				return
 			}
 			if sURLs.Error != nil {
+				stopParallel()
 				errCh <- sURLs.Error
-				continue
+				return
```
Praveenrajmani: A small comment @vadmeste. I can see that this will return on the first failure itself. For eg, If there is a broken symlink in the source and if the listing reports it,  we should not return right away right, Should'nt we ignore the faulty file and continue mirroring the rest?

vadmeste: @Praveenrajmani listing doesn't return broken symlink anymore, this is fixed recently. Otherwise I think it is okay to just quit, I don't see any use case why we should continue otherwise.

Praveenrajmani: :+1:

PR #2683, opened 2019-02-11
use colorjson for all other mc outputs

wlan0:


harshavardhana:
Shouldn't we be adding tags for actual coloring ?

harshavardhana:
> Shouldn't we be adding tags for actual coloring ?

Also shouldn't we be indenting the output?

poornas:
@wlan0, we might want to check if the output should be indented like in ```mc admin monitor ``` command for consistency across commands

wlan0:
> Shouldn't we be adding tags for actual coloring ?

That's a lot more work. This is a good start.

wlan0:
> Also shouldn't we be indenting the output?

This can be done. We should have a discussion about this.

harshavardhana:
Some unmarshalling failure I saw 

```
~ cp ~/run-dist.sh ~/run-dist-ssl.sh
 0 B / 279 B ┃░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░┃   0.00%panic: unmarshalling is not supported by colorjson

goroutine 58 [running]:
github.com/minio/mc/pkg/colorjson.Unmarshal(...)
	/home/harsha/mygo/src/github.com/minio/mc/pkg/colorjson/decode.go:109
github.com/minio/mc/cmd.doCopySession.func1(0xc0004ce720, 0xc0000a32c0, 0xc0004ce600, 0xc0004ce5a0, 0xc00013c480, 0xc000378ff0, 0xc0002d6780, 0xdb1f40, 0xc0000a3260, 0xdb9240, ...)
	/home/harsha/mygo/src/github.com/minio/mc/cmd/cp-main.go:351 +0x1f1
created by github.com/minio/mc/cmd.doCopySession
	/home/harsha/mygo/src/github.com/minio/mc/cmd/cp-main.go:324 +0x36a
```

Can you check?

wlan0:
@harshavardhana great catch. Let me check.

harshavardhana:
One more small change we can bring is making sure accountingReader prints in the JSON form, you make this change in your PR itself. 

```diff
diff --git a/cmd/accounting-reader.go b/cmd/accounting-reader.go
index 92630598..a316a99f 100644
--- a/cmd/accounting-reader.go
+++ b/cmd/accounting-reader.go
@@ -17,12 +17,14 @@
 package cmd
 
 import (
+       "encoding/json"
        "fmt"
        "sync"
        "sync/atomic"
        "time"
 
        "github.com/cheggaaa/pb"
+       "github.com/minio/mc/pkg/probe"
 )
 
 // accounter keeps tabs of ongoing data transfer information.
@@ -78,9 +80,19 @@ func (a *accounter) writer() {
 
 // accountStat cantainer for current stats captured.
 type accountStat struct {
-       Total       int64
-       Transferred int64
-       Speed       float64
+       Status      string  `json:"status"`
+       Total       int64   `json:"total"`
+       Transferred int64   `json:"transferred"`
+       Speed       float64 `json:"speed"`
+}
+
+func (c accountStat) JSON() string {
+       c.Status = "success"
+       accountMessageBytes, e := json.MarshalIndent(c, "", " ")
+       fatalIf(probe.NewError(e), "Failed to marshal account stat.")
+
+       return string(accountMessageBytes)
+
 }
 
 func (c accountStat) String() string {
diff --git a/cmd/cp-main.go b/cmd/cp-main.go
index 5a5cc89e..b79456cd 100644
--- a/cmd/cp-main.go
+++ b/cmd/cp-main.go
@@ -432,7 +432,7 @@ loop:
                }
        } else {
                if accntReader, ok := pg.(*accounter); ok {
-                       console.Println(console.Colorize("Copy", accntReader.Stat().String()))
+                       printMsg(accntReader.Stat())
                }
        }
 
```

Everything else is working fine now.

harshavardhana:
Looks good to me @wlan0 - lets wait on @poornas @vadmeste

cmd/admin-config-get.go
```diff
@@ -90,8 +90,14 @@ func mainAdminConfigGet(ctx *cli.Context) error {
 	c, e := client.GetConfig()
 	fatalIf(probe.NewError(e), "Cannot get server configuration file.")
 
+	config := map[string]interface{}{}
+	e = json.Unmarshal(c, &config)
```
vadmeste: What's the reason for doing this ? only checking the json ?

harshavardhana: Because server already puts it in JSON and to handle double marshalling..

wlan0:
@vadmeste When the `--json` flag is set, we print json double quotes in the server output which is harder to read and understand

image

With the changes, it uses a data structure instead of trying to escape characters from a string. The output is aesthetically more pleasing 

image

vadmeste:
One little problem with this PR is that it is a breaking change for scripts. Since mc --json used to print one json document per line.

I tested with jq but it turned out it is smart enough to adapt itself with the new json style, but still, we probably have some users breaking mc output into lines (\n) and parse each line.

The solution for this is to detect the type of the standard output, if this is terminal, we can do the pretty printing, if not we can use the old style printing.

What do you think @wlan0  ?

wlan0:
> One little problem with this PR is that it is a breaking change for scripts. Since mc --json used to print one json document per line.

> I tested with jq but it turned out it is smart enough to adapt itself with the new json style, but still, we probably have some users breaking mc output into lines (\n) and parse each line.

> The solution for this is to detect the type of the standard output, if this is terminal, we can do the pretty printing, if not we can use the old style printing.

> What do you think @wlan0 ?

This is news to me. Ideally users shouldn't be parsing the output of a CLI tool and expecting that to be a GA API. Why are we supporting use cases like this and adding unnecessary technical debt?

harshavardhana:
@vadmeste JSON readers are expected to handle this properly document by document -  We don't need to worry about that - if someone reports we can perhaps ask them to fix their scripts such that they should be looking at JSON documents properly not just `\n`.

vadmeste:
> JSON readers are expected to handle this properly document by document

I doubt it.. For example, using Python JSON parser returns an error when parsing the following output:
{
   "key": "val1"
}
{
  "key": "val2"
}
```

```
14:40 $ echo -e '{ "key": "val1" }\n{ "key": "val2" }' | python -c "import sys, json; print json.load(sys.stdin)"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python2.7/json/__init__.py", line 291, in load
    **kw)
  File "/usr/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 367, in decode
    raise ValueError(errmsg("Extra data", s, end, len(s)))
ValueError: Extra data: line 2 column 1 - line 3 column 1 (char 18 - 36)
```

wlan0:
@vadmeste Made the change to print single line json if it is not a terminal

PR #2685, opened 2019-02-12
pb: Do not exceed the specified total transfer bytes

### LINKED ISSUE #2684: Progress bar is not updated properly when uploading a multipart file
## Expected behavior
Progress bar correctly reflects the update status

## Actual behavior
```
15:14 $ mc cp /mnt/p10/vadmeste/backup/512mb myminio/testbucket/
.../backup/512mb:  226.46 MB / 512.00 MB ┃▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓█░░░░░░░░░░░░░░░░░░░░░┃  44.23% 46.78 MB/s 4s
15:14 $

```

1. Upload a file of 512 megabytes

## mc version
master

## System information
All

vadmeste:
mc, via minio-go, can retries some requests for many reasons, such as, network
disconnect. This commit ensures that the progress bar does not show transfered
bytes greater then the total transfered bytes to avoid confusing users.

Fixes #2684

### DIFF
--- a/cmd/progress-bar.go
+++ b/cmd/progress-bar.go
@@ -17,8 +17,6 @@
 package cmd
 
 import (
-	"errors"
-	"io"
 	"runtime"
 	"strings"
 	"time"
@@ -32,10 +30,6 @@ import (
 // progress extender.
 type progressBar struct {
 	*pb.ProgressBar
-	reader       io.Reader
-	readerLength int64
-	bytesRead    int64
-	isResume     bool
 }
 
 // newProgressBar - instantiate a progress bar.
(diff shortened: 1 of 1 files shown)


cmd/progress-bar.go
```diff
@@ -102,18 +96,17 @@ func (p *progressBar) Set64(length int64) *progressBar {
...
-		return p.ProgressBar.Total - offset, nil
-	}
-	return 0, errors.New("invalid argument")
+func (p *progressBar) Read(buf []byte) (n int, err error) {
+	defer func() {
+		// After updating the internal progress bar, make sure that its
+		// current progress doesn't exceed the specified total progress
+		currentProgress := p.ProgressBar.Get()
+		if currentProgress > p.ProgressBar.Total {
+			p.ProgressBar.Set64(p.ProgressBar.Total)
```
ebozduman: I think this logic has a problem: After so many retries, we'll reach the total but we are not going to be completely done yet. That means, end-user will see the same total number for sometime while the copy is still continuing and the displayed number will show the total size until the copy is  really done...

harshavardhana:
Seek would still be needed if you want to Seek back the stream @vadmeste so can you add it back ?

vadmeste:
> Seek would still be needed if you want to Seek back the stream @vadmeste so can you add it back ?

@harshavardhana I think the logic of Seek is wrong here, because parallel uploads (like multipart) have access to the same progress bar. Calling Seek() will make the progress bar confused. A simple case is Seek(0, 0) which makes the progress bar looks like it is starting from the beginning although other uploads could be already successfully completed.

I think a cleaner way to fix this is to have each upload has its own progress reporting and it connects itself to the general progress bar (by the way, this can be a first step for showing multiple line progress)

But this PR is urgent before the release. mc master currently wrongly reports progress of a multipart upload.

harshavardhana:
> @harshavardhana I think the logic of Seek is wrong here, because parallel uploads (like multipart) have access to the same progress bar. Calling Seek() by one upload will make the progress bar confused. A simple case is Seek(0, 0) which makes the progress bar looks like it is starting from the beginning although other uploads could be already successfully completed.

@vadmeste okay understood

cmd/progress-bar.go
vadmeste: > I think this logic has a problem: After so many retries, we'll reach the total but we are not going to be completely done yet. That means, the progress bar will be still moving forward, while the displayed number is stuck at the total size for a while until the copy is really done.

mc will still show an error when the upload fails even after having the progress bar arrived to 100%. 

This actually can happen, for example, during complete multipart upload.

In the future, we can make the progress bar reduces its progress when a part upload fails

harshavardhana:
In essence we can never go back because multiparts have a single progress reader and they are all parallel..

vadmeste:
> In essence we can never go back because multiparts have a single progress reader and they are all parallel..

Yes, but in the future, we can separate upload progress from the UI progress bar, so the UI progress bar counts the progress of all registered uploaders, this looks like doable

PR #2689, opened 2019-02-19
complete: Add support of rb completion

vadmeste:
Add completion for rb command:
- Support filesystem paths
- S3 paths only supports this format: `alias/bucket/`, for that s3Complete{}
has deepLevel parameter to configure maximum path deep level.


vadmeste:
Can I get some reviewers here ?

PR #2692, opened 2019-02-20
Add normal/deep scan flag to admin heal cmd

vadmeste:
Add support of normal/deep scan flag to admin heal cmd:

```
SCAN MODES:
   normal (default): Heal objects which are missing on one or more disks.
   deep            : Heal objects which are missing on one or more disks. Also heal objects with silent data corruption.
```

### DIFF
--- a/cmd/admin-heal.go
+++ b/cmd/admin-heal.go
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"path/filepath"
+	"strings"
 
 	"github.com/fatih/color"
 	"github.com/minio/cli"
@@ -27,7 +28,17 @@ import (
 	"github.com/minio/minio/pkg/madmin"
 )
 
+const (
+	scanNormalMode = "normal"
+	scanDeepMode   = "deep"
+)
+
 var adminHealFlags = []cli.Flag{
+	cli.StringFlag{
+		Name:  "scan",
+		Usage: "select the healing scan mode (normal/deep)",
+		Value: scanNormalMode,
+	},
 	cli.BoolFlag{
 		Name:  "recursive, r",
 		Usage: "heal recursively",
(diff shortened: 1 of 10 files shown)


cmd/admin-heal.go
```diff
@@ -27,7 +28,17 @@ import (
...
+const (
+	scanNormalMode = "normal"
+	scanDeepMode   = "deep"
+)
+
 var adminHealFlags = []cli.Flag{
+	cli.StringFlag{
+		Name:  "scan",
+		Usage: "select the healing scan mode",
+		Value: scanNormalMode,
```
krishnasrinivas: @vadmeste can you provide a way for the user to know how deep scan differs from the normal scan?

vadmeste: > @vadmeste can you provide a way for the user to know how deep scan differs from the normal scan?

I added more text in the HELP of admin heal command

vadmeste:
@krishnasrinivas can we make mc enable "deep" mode by default ? that's more secure for users running heal command. Most of them don't read the help.

harshavardhana:
> @krishnasrinivas can we make mc enable "deep" mode by default ? that's more secure for users running heal command. Most of them don't read the help.

It was agreed that `normal` will be the default, not `deep` - also we will write recommendation documentation regarding scenarios when the users need to run `deep` and how the cluster should be managed in production.

cmd/admin-heal.go
```diff
@@ -66,6 +77,11 @@ USAGE:
 FLAGS:
   {{range .VisibleFlags}}{{.}}
   {{end}}
+
+SCAN MODES:
+   normal: Search for missing and inconsistent data in the disks
+     deep: Cover normal scan and additionally search for data corruption
```
krishnasrinivas: This would be better?:
```
 normal (default): Heal objects which are missing on one or more disks
 deep            : Heal objects which are missing on one or more disks and heal objects with bitrot
```

vadmeste:
PR ready for review, @harshavardhana @krishnasrinivas

Issue #2690, opened 2019-02-19
Error when attempting to transfer from GCP to S3


p-doyle:
I am attempting to transfer a GCP storage bucket to S3 using the following command with the Windows CLI exe:
`mc.exe cp -r gcs/takeout-export-123/ s3/my-bucket/`

It seems to start transferring but then throws this error:

`mc.exe: <ERROR> Failed to copy https://storage.googleapis.com/takeout-export-123/20190215T172254Z/Resource-456/takeout-20190215T172258Z-001.zip. Expires unsupported user defined metadata name`

## Your Environment
Version: 2019-02-13T19:48:27Z
Release-tag: RELEASE.2019-02-13T19-48-27Z
Commit-id: bc1a108e5c8f3c2a9e25ed34ff221985987b3fed
Windows 10 Pro 64-bit

poornas:
Can you run this with --debug flag and post the trace?
`mc.exe cp -r gcs/takeout-export-123/ s3/my-bucket/ --debug`

p-doyle:
So I actually had to move --debug to before -r as when I append it to the command I get:
`mc.exe: <ERROR> Unable to guess the type of copy operation. Invalid arguments provided, please refer mc <command> -h for relevant documentation.`

Anyway here is the trace:

```
ERROR> Failed to copy `https://storage.googleapis.com/takeout-export-123/20190215T172254Z/Resource-456/takeout-20190215T172257Z-001.zip`. Expires unsupported user defined metadata name
 (3) /q/.q/sources/gopath/src/github.com/minio/mc/cmd/cp-main.go:415 cmd.doCopySession(..) Tags: [https://storage.googleapis.com/takeout-export-123/20190215T172254Z/Resource-456/takeout-20190215T172257Z-001.zip]
 (2) /q/.q/sources/gopath/src/github.com/minio/mc/cmd/common-methods.go:252 cmd.uploadSourceToTargetURL(..) Tags: [https://s3.amazonaws.com/my-bucket/20190215T172254Z/Resource-456/takeout-20190215T172257Z-001.zip]
 (1) /q/.q/sources/gopath/src/github.com/minio/mc/cmd/common-methods.go:178 cmd.putTargetStream(..) Tags: [s3, https://s3.amazonaws.com/my-bucket/20190215T172254Z/Resource-456/takeout-20190215T172257Z-001.zip]
[...]

PR #2697, opened 2019-02-26
Stat: display Expires header if present

poornas:
Fixes : #2690


kannappanr:
@poornas Can you please fix Travis failure

poornas:
> @poornas Can you please fix Travis failure
This should go away once minio-go change is vendorized.

poornas:
updated PR after vendorizing minio-go

ebozduman:
There is one more place where `Expiry` needs to be set (line# 1076 in the old code)
```
			if isFetchMeta {
				stat, err := c.getObjectStat(bucket, object, opts)
				if err != nil {
					return nil, err
				}
				objectMetadata.ETag = stat.ETag
				objectMetadata.Metadata = stat.Metadata
				objectMetadata.EncryptionHeaders = stat.EncryptionHeaders
			}
```

poornas:
> There is one more place where `Expiry` needs to be set (line# 1076 in the old code)
fixed @ebozduman

poornas:
> We also need to add `Expiry` setting in our existing tests, like `TestParseStat` in stat_test.go
done

PR #2699, opened 2019-02-28
Avoid ListBuckets to keep Stat() ops on bucket simpler

harshavardhana:
Due to multi-user feature on Minio and `mc` adoption in
many different restrictive environments, most users disable
access to ListBuckets() calls. We can avoid such network
operations as they only add marginal value.


harshavardhana:
https://github.com/golang/lint/issues/436

PR #2703, opened 2019-03-05
rb: Fix removing an empty bucket when passed as argument 

### LINKED ISSUE #2702: rb: Unable to remove an empty bucket
## Expected behavior
`rb` (Remove bucket) should be able to remove an empty bucket

## Actual behavior
`rb` is not able to remove an empty bucket.

```
1. mc mb myminio/testbucket/
2. mc rb myminio/testbucket/
```

## mc version
All

## System information
All

vadmeste:
rb was relying on client.List() to find all elements that needs to be
removed. However it doesn't include 'alias/bucket' in the listing
when 'alias/bucket' itself is passed as an argument in rb command.

This PR will always attempt to delete the passed argument as this is
what the user wants in the first place.

This PR also fixes teardown() in functional-tests.sh to remove
the bucket using rb instead rm since rm does not remove buckets
anymore.

Fixes #2702

### DIFF
--- a/functional-tests.sh
+++ b/functional-tests.sh
@@ -257,7 +257,7 @@ function setup()
 function teardown()
 {
     start_time=$(get_time)
-    assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd rm --force --recursive "${SERVER_ALIAS}/${BUCKET_NAME}"
+    assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd rb --force "${SERVER_ALIAS}/${BUCKET_NAME}"
 }
 
 function test_put_object()
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -792,6 +792,13 @@ func (c *s3Client) Remove(isIncomplete, isRemoveBucket bool, contentCh <-chan *c
 		for content := range contentCh {
 			// Convert content.URL.Path to objectName for objectsCh.
 			bucket, objectName := c.splitPath(content.URL.Path)
+
+			// We don't treat path when bucket is
+			// empty, just skip it when it happens.
+			if bucket == "" {
+				continue
+			}
+
 			// Init objectsCh the first time.
 			if prevBucket == "" {
 				objectsCh = make(chan string)
@@ -858,7 +865,7 @@ func (c *s3Client) Remove(isIncomplete, isRemoveBucket bool, contentCh <-chan *c
 			}
 		}
 		// Remove last bucket if it qualifies.
-		if isRemoveBucket && !isIncomplete {
+		if isRemoveBucket && prevBucket != "" && !isIncomplete {
 			if err := c.api.RemoveBucket(prevBucket); err != nil {
 				errorCh <- probe.NewError(err)
 			}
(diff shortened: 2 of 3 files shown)


functional-tests.sh
```diff
@@ -257,7 +257,7 @@ function setup()
 function teardown()
 {
     start_time=$(get_time)
-    assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd rm --force --recursive "${SERVER_ALIAS}/${BUCKET_NAME}"
+    assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd rb "${SERVER_ALIAS}/${BUCKET_NAME}"
```
poornas: should still pass --force flag in case bucket is not empty

poornas:
@vadmeste, issue happens only when bucket name ends with a "/". Should we allow "/" at the end
of bucketname since [s3 bucketname](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-s3-bucket-naming-requirements.html) convention doesn't permit it?

functional-tests.sh
vadmeste: > should still pass --force flag in case bucket is not empty

yes right, fixed.

vadmeste:
> @vadmeste, issue happens only when bucket name ends with a "/". Should we allow "/" at the end
> of bucketname since [s3 bucketname](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-s3-bucket-naming-requirements.html) convention doesn't permit it?

As per my testing, mc rb doesn't work with and without the trailing slash. By the way, the bucket needs to be empty to reproduce this:

```
$ mc mb myminio/newbucket
$ mc rb myminio/newbucket
```

poornas:
This diff is also needed to validate if non existent bucket is being deleted
```
diff --git a/cmd/rb-main.go b/cmd/rb-main.go
index 2423b587..6120edc9 100644
--- a/cmd/rb-main.go
+++ b/cmd/rb-main.go
@@ -185,21 +185,28 @@ func mainRemoveBucket(ctx *cli.Context) error {
                        cErr = exitStatus(globalErrorExitStatus)
                        continue
                }
+
                isEmpty := true
                for range clnt.List(true, false, DirNone) {
                        isEmpty = false
                        break
                }
+               exists := false
+               _, err = clnt.Stat(false, false, nil)
+               if err == nil {
+                       exists = true
+               }
                // For all recursive operations make sure to check for 'force' flag.
                if !isForce {
```

vadmeste:
@poornas thanks, I incorporated your change with a slight change

poornas:
> > @vadmeste, issue happens only when bucket name ends with a "/". Should we allow "/" at the end
> > of bucketname since [s3 bucketname](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-s3-bucket-naming-requirements.html) convention doesn't permit it?
> 
> As per my testing, mc rb doesn't work with and without the trailing slash. By the way, the bucket needs to be empty to reproduce this:
> 
> ```
> $ mc mb myminio/newbucket
> $ mc rb myminio/newbucket
> ```
@vadmeste, it used to work, regression due to 19e14db50ad4467e7bb10a10a8dddedae7a1e2ee

PR #2705, opened 2019-03-05
mirror: Return non-zero exit code for errors occuring during mirror

vadmeste:
An error, such as, `Bad Request` when copying an object without supplied
keys is not reflected in the exit code.

This commit simplifies a bit the mirror code. It makes mirror before and
after starting to send the status to mj.status.


vadmeste:
Ready for review

vadmeste:
ping @ebozduman, please take a look.

PR #2709, opened 2019-03-08
Add --dangerous flag for mc rb sitewide removal

poornas:


kannappanr:
```
 $ mc rb --dangerous --force myminio1
Bucket removed successfully `myminio1`.
```
Text needs to be changed.

kannappanr:
sorry for the nit pick, but the out put is now

```
$ mc rb --dangerous --force myminio1
Removed `myminio1/test` successfully.
Removed `myminio1/test1` successfully.
Removed `myminio1/test2` successfully.
Removed `myminio1` successfully.
```
It will be better if we do not display the last line, it gives an impression that we are removing the alias itself. Displaying the buckets might be good enough.

PR #2712, opened 2019-03-09
mc admin info changes

wlan0:
- sorts the list of outputs in `mc admin info`
- move status to its own field
- show uptime in a separate field
- shows drives per accurately
- writes version as `<development>` if it is `DEVELOPMENT.GOGET`
- writes error message as `unreachable` if a node is unreachable. It used to show up as `rpc: retry error`
- skips `SQS ARN` if it is empty
- skips `Region` if it is empty

### DIFF
--- a/cmd/admin-info.go
+++ b/cmd/admin-info.go
@@ -19,6 +19,8 @@ package cmd
 import (
 	"fmt"
 	"net/url"
+	"sort"
+	"strings"
 	"time"
 
 	humanize "github.com/dustin/go-humanize"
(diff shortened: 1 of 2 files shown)


cmd/admin-info.go
```diff
@@ -120,49 +122,78 @@ func (u infoMessage) String() (msg string) {
...
 	// ARNs
 	sqsARNs := ""
 	for _, v := range u.ServerInfo.Properties.SQSARN {
 		sqsARNs += fmt.Sprintf("%s ", v)
 	}
-	if sqsARNs == "" {
-		sqsARNs = "<none>"
+	//if sqsARNs == "" {
+	//	sqsARNs = "<none>"
+	//}
```
harshavardhana: Commented out code can be removed..

cmd/admin-info.go
```diff
@@ -120,49 +122,78 @@ func (u infoMessage) String() (msg string) {
...
+		msg += fmt.Sprintf(" SQS ARNs : %s\n", sqsARNs)
 	}
-	msg += fmt.Sprintf(" SQS ARNs : %s\n", sqsARNs)
+
 	// Incoming/outgoing
-	msg += fmt.Sprintf("    Stats : Incoming %s, Outgoing %s\n",
-		humanize.IBytes(u.ServerInfo.ConnStats.TotalInputBytes),
-		humanize.IBytes(u.ServerInfo.ConnStats.TotalOutputBytes))
+	//msg += fmt.Sprintf("    Stats : Incoming %s, Outgoing %s\n",
+	//	humanize.IBytes(u.ServerInfo.ConnStats.TotalInputBytes),
```
harshavardhana: Same as above..

cmd/admin-info.go
```diff
@@ -120,49 +122,78 @@ func (u infoMessage) String() (msg string) {
...
 
 	// Print error if any and exit
 	if u.Err != "" {
-		msg += fmt.Sprintf("%s  %s\n", console.Colorize("InfoFail", dot), u.Addr)
-		msg += fmt.Sprintf("   Uptime : Server is %s\n", console.Colorize("InfoFail", "offline"))
-		msg += fmt.Sprintf("    Error : %s", u.Err)
+		msg += fmt.Sprintf("%s  %s\n", dot, console.Colorize("PrintB", u.Addr))
+		msg += fmt.Sprintf("   Status : %s\n", console.Colorize("InfoFail", "offline"))
+		e := u.Err
+		if strings.Trim(e, " ") == "rpc: retry error" {
```
harshavardhana: How about we do this on server.. ?? Because if server changes the error this will fail..

cmd/admin-info.go
```diff
@@ -120,49 +122,78 @@ func (u infoMessage) String() (msg string) {
...
+	msg += fmt.Sprintf("   Status : %s\n", console.Colorize("Info", "online"))
 
 	// Uptime
-	msg += fmt.Sprintf("   Uptime : %s since %s\n", console.Colorize("Info", "online"),
-		humanize.Time(time.Now().UTC().Add(-u.ServerInfo.Properties.Uptime)))
+	msg += fmt.Sprintf("   Uptime : %s\n", humanize.RelTime(time.Now(), time.Now().Add(-u.ServerInfo.Properties.Uptime), "", ""))
 	// Version
-	msg += fmt.Sprintf("  Version : %s\n", u.ServerInfo.Properties.Version)
+	version := u.ServerInfo.Properties.Version
+	if u.ServerInfo.Properties.Version == "DEVELOPMENT.GOGET" {
```
harshavardhana: This is fine as this is not expected to change

cmd/admin-info.go
```diff
@@ -262,9 +294,37 @@ func mainAdminInfo(ctx *cli.Context) error {
...
+}
+
+func (s *sortInfoWrapper) Len() int {
+	return len(s.infos)
+}
+
+func (s *sortInfoWrapper) Swap(i, j int) {
+	if s.infos != nil {
+		s.infos[i], s.infos[j] = s.infos[j], s.infos[i]
+		return
```
harshavardhana: A naked return is not needed.

cmd/admin-info.go
```diff
@@ -262,9 +294,37 @@ func mainAdminInfo(ctx *cli.Context) error {
...
+	}
+}
+
+func (s *sortInfoWrapper) Less(i, j int) bool {
+	if s.infos != nil {
+		return strings.Compare(s.infos[i].Addr, s.infos[j].Addr) < 0
+	}
```
harshavardhana: It is advised to not use strings.Compare refer godoc just use `s1 < s2`

cmd/admin-info.go
wlan0: Short term fix. I can also add this on the server.

Issue #2710, opened 2019-03-08
mc rm <alias/bucket_name> gives user the wrong success impression


ebozduman:
## Expected behavior
`mc rm <alias/bucket_name>` errors out with a meaningful message, like `rm expects an object name or ---recursive option to clean up the whole bucket content`.

## Actual behavior
`mc rm <alias/bucket_name>` does not error out but the following message in green is shown to the user giving the user a wrong impression as if the removal process is successful. 
`Removing '<alias/bucket_name>'.`
If the bucket and its content are checked with `mc ls <alias/bucket_name>`, the bucket and its content are still there untouched.

## Steps to reproduce the behavior
`mc rm <alias/bucket_name>`

## mc version
$ mc version
Version: 2019-03-08T09:32:30Z
Release-tag: DEVELOPMENT.2019-03-08T09-32-30Z
Commit-id: 2d07e36a593c193de443b761788167a001d56000

sinhaashish:
We can close this issue as [PR](https://github.com/minio/mc/pull/2700) caters this issue.
The new error message `Target <<server_alias/bucketname>> not found`  is much apt than the earlier message which was `Removing '<alias/bucket_name>`.

nitisht:
fixed by https://github.com/minio/mc/pull/2700

PR #2714, opened 2019-03-11
Modify Remove message

sinhaashish:
mc rm <<server_alias/bucketname>> gives an error : rm expects an object name or ---recursive option to clean up the whole bucket content..

[Fixes ](https://github.com/minio/mc/issues/2710)

### DIFF
--- a/cmd/client-errors.go
+++ b/cmd/client-errors.go
@@ -191,3 +191,12 @@ type SameFile struct {
 func (e SameFile) Error() string {
 	return fmt.Sprintf("'%s' and '%s' are the same file", e.Source, e.Destination)
 }
+
+// ObjectNameEmptyInRemove - Object name is empty in remove
+type ObjectNameEmptyInRemove struct {
+	bucket string
+}
+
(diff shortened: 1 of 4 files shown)


cmd/client-errors.go
```diff
@@ -191,3 +191,12 @@ type SameFile struct {
...
 	return fmt.Sprintf("'%s' and '%s' are the same file", e.Source, e.Destination)
 }
+
+// ObjectNameEmptyInRemove - Object name is empty in remove
+type ObjectNameEmptyInRemove struct {
+	bucket string
+}
+
+func (e ObjectNameEmptyInRemove) Error() string {
+	return "rm expects an object name or ---recursive option to clean up the whole bucket content"
```
harshavardhana: We shouldn't add `mc` level commands in the error in client abstraction errors... this cannot be re-purposed for any other commands. 

This error should be only talking about the object name being empty, it is the `mc rm` command which should print the error.

ebozduman:
@sinhaashish,
Please look into the conflict that must be resolved in file, `functional-test.sh`.

PR #2716, opened 2019-03-12
For non-canned policies, show policy status as custom

poornas:
When user has a custom policy that doesn't fall under the canned policies, ```mc policy alias/bucket``` displays `none` which is incorrect. This was reported by a user on slack.

This PR changes this behavior to return `custom` when a noncanned policy is set and also display the policy in --json mode.

### DIFF
--- a/cmd/sql-main.go
+++ b/cmd/sql-main.go
@@ -91,7 +91,7 @@ ENVIRONMENT VARIABLES:
    MC_ENCRYPT_KEY:  list of comma delimited prefix=secret values
 
 SERIALIZATION OPTIONS:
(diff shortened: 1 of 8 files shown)


cmd/sql-main.go
```diff
@@ -91,7 +91,7 @@ ENVIRONMENT VARIABLES:
    MC_ENCRYPT_KEY:  list of comma delimited prefix=secret values
 
 SERIALIZATION OPTIONS:
-  For query serialization options, refer to https://docs.minio.io/docs/minio-client-complete-guide#sql
+   For query serialization options, refer to https://docs.minio.io/docs/minio-client-complete-guide#sql
```
kannappanr: Can the changes in this file be removed?

PR #2717, opened 2019-03-13
tests: Clean per test created buckets

vadmeste:
In addition to the general bucket name created for functinal tests,
some tests inside this latter creates separated buckets but remove
them using rm command which does not support removing buckets anymore.

### DIFF
--- a/functional-tests.sh
+++ b/functional-tests.sh
@@ -233,7 +233,7 @@ function test_make_bucket()
     start_time=$(get_time)
     bucket_name="mc-test-bucket-$RANDOM"
     assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd mb "${SERVER_ALIAS}/${bucket_name}"
-    assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd rm "${SERVER_ALIAS}/${bucket_name}"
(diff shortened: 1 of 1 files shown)


functional-tests.sh
```diff
@@ -763,7 +765,7 @@ function test_mirror_with_sse()
     diff -bB <(ls "$DATA_DIR") <("${MC_CMD[@]}" --json ls "${SERVER_ALIAS}/${bucket_name}/" | jq -r .key) >/dev/null 2>&1
     assert_success "$start_time" "${FUNCNAME[0]}" show_on_failure $? "mirror and list differs"
     # remove recursively with correct encryption key
-    assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd rm --force --recursive --encrypt-key "${cli_flag}" "${SERVER_ALIAS}/${bucket_name}"
+    assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd rb --force "${SERVER_ALIAS}/${bucket_name}"
```
kannappanr: The comment above is stale.

vadmeste: > The comment above is stale.

Right, fixed.

PR #2720, opened 2019-03-20
Fix rm to behave properly deleting prefixes

harshavardhana:


cmd/rm-main.go
```diff
@@ -259,24 +258,23 @@ func removeRecursive(url string, isIncomplete bool, isFake bool, olderThan, newe
...
 
-		// list internally mimics recursive directory listing of object prefixes for s3 similar to FS.
-		// The rmMessage needs to be printed only for actual objects being deleted and not prefixes.
-		if !(content.Time.IsZero() && strings.HasSuffix(urlString, string(filepath.Separator))) {
-			printMsg(rmMessage{
-				Key:  targetAlias + urlString,
-				Size: content.Size,
-			})
-		}
+		printMsg(rmMessage{
```
Praveenrajmani: Should we print the mssage if `content.Time.IsZero()` ?

harshavardhana: why should we not?

Praveenrajmani: :+1

Praveenrajmani: I thought `strings.HasSuffix` is enough to find if the name is a dir or not

PR #2727, opened 2019-03-25
Migrate to go1.12, deprecate vendor directory

harshavardhana:


wlan0:
LGTM - Thanks for checking for xtended FS support - this should prevent any future test/CI debts.

harshavardhana:
Windows build is fixed by fixing xattr package upstream and `go.mod` is updated as well @vadmeste PTAL

cmd/policy-main.go
```diff
@@ -444,13 +444,13 @@ func mainPolicy(ctx *cli.Context) error {
...
-		runPolicyListCmd(ctx)
+		runPolicyListCmd(ctx.Args().Tail())
 	case "links":
 		// policy links alias/bucket/prefix
-		runPolicyLinksCmd(ctx)
+		runPolicyLinksCmd(ctx.Args().Tail(), ctx.Bool("recursive"))
 	default:
 		// policy [download|upload|public|] alias/bucket/prefix
-		runPolicyCmd(ctx)
+		runPolicyCmd(ctx.Args().Tail())
```
vadmeste: You should not Tail() this, the first argument is the access permission (download, upload, public)

harshavardhana:
> I am not sure about mutiple `go: finding` in each make install command

That is upto `go` to decide. I don't know why it does that. But its not an issue @vadmeste

cmd/policy-main.go
harshavardhana: Fixed

PR #2730, opened 2019-03-28
Set correct compressionType for compression select request

harshavardhana:


deekoder:
Fixes a regression caused in the output of mc sql on compressed files.

PR #2700, opened 2019-03-01
rm: Force removing an object when it is encrypted without key specified

vadmeste:
Currently, a single remove HEAD an object, but this latter can return
400 Bad Request when the object is encrypted and the user doesn't specify
any encryption key.

So when HEAD fail, this PR will use GET listing to find the object and
remove it if existent.

We can avoid calling HEAD in the first place and use only listing, but this
has some penality on the server.

PR #2731, opened 2019-03-28
Modify Remove message 

sinhaashish:
mc rm <<server_alias>>/bucketname gives an error : rm expects an object name or ---recursive option to clean up the whole bucket content

[Fixes ](https://github.com/minio/mc/issues/2710)

### DIFF
--- a/cmd/typed-errors.go
+++ b/cmd/typed-errors.go
@@ -99,7 +99,7 @@ var errInvalidTarget = func(URL string) *probe.Error {
 type targetNotFoundErr error
 
 var errTargetNotFound = func(URL string) *probe.Error {
-	msg := "Target `" + URL + "` not found."
+	msg := "Target `" + URL + "` not found, `rm` expects an object name or ---recursive option to clean up the whole bucket content"
 	return probe.NewError(targetNotFoundErr(errors.New(msg))).Untrace()
 }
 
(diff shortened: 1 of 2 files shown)


cmd/typed-errors.go
```diff
@@ -99,7 +99,7 @@ var errInvalidTarget = func(URL string) *probe.Error {
 type targetNotFoundErr error
 
 var errTargetNotFound = func(URL string) *probe.Error {
-	msg := "Target `" + URL + "` not found."
+	msg := "Target `" + URL + "` not found, `rm` expects an object name or ---recursive option to clean up the whole bucket content"
```
harshavardhana: Are you sure that errTargetNotFound is only used by `mc rm`? Because this will make it unusable with other commands..

kannappanr: This is also used by `stat` command, so this is definitely a wrong place. In general, the message here in `errTargetNotFound` is wrong. It should convey that the used URL is either a bucket or a prefix and not Target not found.

harshavardhana: Best way to solve is making it an abstracted error. Look at `mc mb` for the right errors messages.

The error should arrive from client-s3,fs.go abstraction.

sinhaashish:
Closing as [PR](https://github.com/minio/mc/pull/2700) handles the issue.

PR #2733, opened 2019-03-28
Display request header in anonymous mode when debug flag is set

kannappanr:
Currently if authorization header is missing, then request was not displayed when `--debug` flag is passed
This resulted in just the response being displayed, in anonymous mode.

Fix allows printing of headers even if authorization header is not present

### DIFF
--- a/cmd/client-s3-trace_v4.go
+++ b/cmd/client-s3-trace_v4.go
@@ -38,6 +38,14 @@ func newTraceV4() httptracer.HTTPTracer {
 func (t traceV4) Request(req *http.Request) (err error) {
 	origAuth := req.Header.Get("Authorization")
 
+	printTrace := func() error {
+		reqTrace, rerr := httputil.DumpRequestOut(req, false) // Only display header
+		if rerr == nil {
+			console.Debug(string(reqTrace))
+		}
+		return rerr
+	}
+
 	if strings.TrimSpace(origAuth) != "" {
 		// Authorization (S3 v4 signature) Format:
 		// Authorization: AWS4-HMAC-SHA256 Credential=AKIAJNACEGBGMXBHLEZA/20150524/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=bbfaa693c626021bcb5f911cd898a1a30206c1fad6bad1e0eb89e282173bd24c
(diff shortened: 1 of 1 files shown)


cmd/client-s3-trace_v4.go
```diff
@@ -61,6 +61,12 @@ func (t traceV4) Request(req *http.Request) (err error) {
 
 		// Undo
 		req.Header.Set("Authorization", origAuth)
+	} else {
+		var reqTrace []byte
```
Praveenrajmani: Shall we declare `reqTrace` in L40 itself ? after `origAuth := req.Header.Get("Authorization")` ?

sinhaashish: IMO why to declare the variable if its not used. Since `reqTrace` is only used in `else` condition so it is rightly placed.

Praveenrajmani: It is also used in the if clause ( if authorization header is present ). So it can be generically declared

kannappanr: rearranged it a little bit. PTAL

cmd/client-s3-trace_v4.go
```diff
@@ -38,6 +38,14 @@ func newTraceV4() httptracer.HTTPTracer {
 func (t traceV4) Request(req *http.Request) (err error) {
 	origAuth := req.Header.Get("Authorization")
 
+	trace := func() {
+		var reqTrace []byte
+		reqTrace, err = httputil.DumpRequestOut(req, false) // Only display header
```
harshavardhana: Since this is a closure we should avoid `err` to be aliased with the `return` alias such that it is better to call it 

```
trace := func() {
       reqTrace, rerr := httputil.DumpRequestOut(req, false) // Only display header
       if rerr == nil {
              console.Debug(string(reqTrace))
       }
}
```

harshavardhana: Also perhaps call it `printTrace()`

kannappanr: done

cmd/client-s3-trace_v4.go
```diff
@@ -38,6 +38,15 @@ func newTraceV4() httptracer.HTTPTracer {
 func (t traceV4) Request(req *http.Request) (err error) {
 	origAuth := req.Header.Get("Authorization")
 
+	printTrace := func() error {
+		var reqTrace []byte
```
harshavardhana: this extra definition is not needed since you are using `reqTrace, rerr :=`

kannappanr: done

Issue #2640, opened 2019-01-06
Minio Client startup fails with permission error on OpenShift


juv:
## Expected behavior
Minio Client starts without errors on OpenShift 

## Actual behavior
Minio Client starts up and fails immediately with error `mc: <ERROR> Unable to save new mc config. mkdir /.mc: permission denied.`. This is related to https://github.com/helm/charts/issues/4128 -- OpenShift expects containers to not run as `root`. 

## mc version
Unknown (mc fails before `mc version` command can be executed)

## System information
OpenShift: 
```
openshift v3.10.34
kubernetes v1.10.0+b81c8f8
```
Tried with mc docker tags: `latest` (`minio/mc@sha256:7b27ff9a0b9bbc0622fe78b086ddf3a36fe50f8673d906de1f336ed3a5e249d9`) and `edge` (`minio/mc@sha256:091bce3f6f240c731e2c4f203f8a87905677a727cb5d09d04c73be92ab5788b6`)

kannappanr:
@juv Thanks for filing this issue. We will try to resolve this soon.

sinhaashish:
@juv  can you provide the `yaml` file which you are using.

juv:
@sinhaashish I am simply deploying the mc docker image in Openshift. 

```apiVersion: v1
kind: Pod
metadata:
  annotations:
    openshift.io/deployment-config.latest-version: '2'
    openshift.io/deployment-config.name: mc
    openshift.io/deployment.name: mc-2
    openshift.io/generated-by: OpenShiftWebConsole
    openshift.io/scc: restricted
  creationTimestamp: '2019-02-13T11:25:56Z'
  generateName: mc-2-
  labels:
    app: mc
    deployment: mc-2
    deploymentconfig: mc
  name: mc-2-qmgrc
  namespace: my-namespace
  ownerReferences:
    - apiVersion: v1
      blockOwnerDeletion: true
      controller: true
      kind: ReplicationController
      name: mc-2
      uid: 1dffba70-2f82-11e9-9ecc-001a4a160156
  resourceVersion: '44349118'
  selfLink: /api/v1/namespaces/my-namespace/pods/mc-2-qmgrc
  uid: 21bdff70-2f82-11e9-9ecc-001a4a160156
spec:
  containers:
    - image: >-
        minio/mc@sha256:1bda261836dc1d5cb4b9ffde780b18158f5569ebcfe32e3057d0e7f33035acef
      imagePullPolicy: Always
      name: mc
[...]

PR #2734, opened 2019-03-29
Fix for container startup issues on Openshift

juv:
On Openshift, containers will be running under a random uid by default, e.g. `uid=1012330000 gid=0(root) groups=0(root),1012330000`. Dockerfiles should ideally switch the user context e.g. by using `USER 1001`. `1001` will be substituted with a generated uid by Openshift out of a range of uids that is specific to the namespace the deployment is running in. All directories within the container are owned by the `root` user as per Docker default. When writes to certain directories are necessary, containers should use `RUN chgrp -R 0 /some/directory && chmod -R g=u /some/directory` to make the directories accessible from the `root` group (note: not user).

Detailed information for the used fix: https://docs.openshift.com/container-platform/3.10/creating_images/guidelines.html

### DIFF
--- a/Dockerfile.release
+++ b/Dockerfile.release
@@ -7,5 +7,9 @@ RUN \
     apk add --no-cache --virtual .build-deps curl && \
     curl https://dl.minio.io/client/mc/release/linux-amd64/mc > /usr/bin/mc && \
     chmod +x /usr/bin/mc && apk del .build-deps
+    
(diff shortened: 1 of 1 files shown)


juv:
Fixes https://github.com/minio/mc/issues/2640

harshavardhana:
Closing this PR as stale and not needed. Will be fixed in a separate PR.

PR #2748, opened 2019-04-16
Fix: do not check if object listed equals object being removed

### LINKED ISSUE #2746: mc rm does not remove a valid object
## Expected behavior
mc rm succeeds

## Actual behavior
mc rm fails even if the file exists.

```
mc: <ERROR> Failed to remove `myminio/testbucket/hosts`. Target `myminio/testbucket/hosts` not found.
 (0) rm-main.go:181 cmd.removeSingle(..) Tags: [myminio/testbucket/hosts]
 Release-Tag:DEVELOPMENT.2019-04-09T17-42-42Z | Commit:66677d7c21cd | Host:backspace | OS:linux | Arch:amd64 | Lang:go1.12.1 | Mem:2.5 MB/72 MB | Heap:2.5 MB/66 MB
```

```
~ mc ls myminio/testbucket/hosts
[2019-04-16 10:26:49 PDT]  1.0GiB hosts
[2019-04-16 10:27:10 PDT]  1.0GiB hosts.1
```

Just use the latest master
## mc version
master

## System information
Linux

kannappanr:
List within the statURL function considers object name as a prefix
and lists objects that have the same prefix.
currently, it errors out if the object is a subset of another object.

Fixes #2746

### DIFF
--- a/cmd/stat.go
+++ b/cmd/stat.go
@@ -155,7 +155,7 @@ func statURL(targetURL string, isIncomplete, isRecursive bool, encKeyDB map[stri
 		}
 		url := targetAlias + getKey(content)
 
-		if !isRecursive && url != targetURL {
+		if !isRecursive && !strings.HasPrefix(url, targetURL) {
 			return nil, errTargetNotFound(targetURL)
 		}
 


krishnasrinivas:
@harshavardhana 

rm on a single file internally calls statURL() which triggers a list. i.e rm should not trigger a list, do you know why it was done that way?

kannappanr:
S3 returns `200` even if the object is not found
and in `mc` we wanted to show `object not found` error when we issue `mc rm` command
only way to do it is to do stat stat will work if it is not `SSE-C` encrypted object
For an `SSE-C` encrypted object, you will only get `400 Bad request` if you are not passing `encryption keys`
since we also support `newer-than` and `older-than` flags we had to also implement `list` objects

kannappanr:
Here is a sample output from AWS S3, where `ssecsingle` is an object encrypted using `SSE-C`

```
$ aws s3api list-objects --bucket mybucket
{
    "Contents": [
        {
            "Key": "issue1",
            "LastModified": "2019-04-16T21:41:26.000Z",
            "ETag": "\"2fbcdb40c9edd1edc6c16ac2b0a401f8\"",
            "Size": 20,
            "StorageClass": "STANDARD",
        },
        {
            "Key": "issue",
            "LastModified": "2019-03-29T21:18:49.000Z",
            "ETag": "\"2fbcdb40c9edd1edc6c16ac2b0a401f8\"",
            "Size": 20,
            "StorageClass": "STANDARD",
        },
        {
            "Key": "ssecsingle",
            "LastModified": "2019-04-17T01:37:00.000Z",
            "ETag": "\"72ab016a1a04279ecd3287c42e3b83fa\"",
            "Size": 20,
            "StorageClass": "STANDARD",
        }
    ]
}
```
```
$ aws s3api head-object --bucket mybucket --key ssecsingle

An error occurred (400) when calling the HeadObject operation: Bad Request
```
```
$ aws s3api head-object --bucket mybucket --key issue1

{
    "AcceptRanges": "bytes",
    "LastModified": "Tue, 16 Apr 2019 21:41:26 GMT",
    "ContentLength": 20,
    "ETag": "\"2fbcdb40c9edd1edc6c16ac2b0a401f8\"",
    "ContentType": "application/octet-stream",
    "Metadata": {}
}
```
Since we support `newer-than` and `older-than` flags, we have to `list` to get the `LastModified` time on the object, otherwise `stat` on an `SSE-C` encrypted object will return `400 Bad Request`

PR #2760, opened 2019-04-24
Error out in list policy if policy doesn't exist

### LINKED ISSUE #2758: Return code from "mc admin policy list fdsfdsfds" is 0 even when policy doesn't exist
## Expected behavior
provide a return code to confirm a policy doesn't exist.

## Actual behavior
I'm trying to find out if a policy exists or not with the `mc` client tool. 

C:\Users\user>mc admin policy list minio
readonly
readwrite
writeonly

C:\Users\user>echo %errorlevel%
0

C:\Users\user>mc admin policy list minio fdasfdas
fdasfdas

C:\Users\user>echo %errorlevel%
0

The policy fdasfdas does not exist, as you can see by the output of the first `mc admin policy list minio` command. But the exit code of `mc admin policy list minio fdasfdas` is 0 - success.

Examine the return code for a non-existent policy:
mc admin policy list minio fdasfdas

## mc version
Version: 2019-04-10T22:11:28Z
Release-tag: RELEASE.2019-04-10T22-11-28Z
Commit-id: ea469140d50b0240501963666e1de19c6d37f812

sinhaashish:
Error out in case of list policy `mc admin policy list minio fdasfdas` if the policy ` fdasfdas` is not present.

Closes #2758

### DIFF
--- a/cmd/admin-policy-list.go
+++ b/cmd/admin-policy-list.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"fmt"
 	"github.com/fatih/color"
 	"github.com/minio/cli"
 	"github.com/minio/mc/pkg/console"
@@ -76,11 +77,15 @@ func mainAdminPolicyList(ctx *cli.Context) error {
 	fatalIf(probe.NewError(e).Trace(args...), "Cannot list policy")
 
 	if policyName := args.Get(1); policyName != "" {
-		printMsg(userPolicyMessage{
-			op:         "list",
-			Policy:     policyName,
-			PolicyJSON: policies[policyName],
-		})
+		if len(policies[policyName]) != 0 {
+			printMsg(userPolicyMessage{
+				op:         "list",
+				Policy:     policyName,
+				PolicyJSON: policies[policyName],
+			})
+		} else {
+			fatalIf(probe.NewError(fmt.Errorf("%s is not found", policyName)), "Cannot list the policy")
+		}
 	} else {
 		for k := range policies {
 			printMsg(userPolicyMessage{


cmd/admin-policy-list.go
```diff
@@ -76,11 +77,15 @@ func mainAdminPolicyList(ctx *cli.Context) error {
 	fatalIf(probe.NewError(e).Trace(args...), "Cannot list policy")
 
 	if policyName := args.Get(1); policyName != "" {
-		printMsg(userPolicyMessage{
-			op:         "list",
-			Policy:     policyName,
-			PolicyJSON: policies[policyName],
-		})
+		if fmt.Sprintf("%s", policies[policyName]) != "" {
```
ebozduman: `policies` is of type `map[string][]byte`.
So, we can simply check the length of the byte array if zero or not, IMHO:
```
    if len(policies[policyName]) != 0 {
```

sinhaashish: :thumbsup: Modified.

cmd/admin-policy-list.go
```diff
@@ -76,11 +77,15 @@ func mainAdminPolicyList(ctx *cli.Context) error {
...
-			PolicyJSON: policies[policyName],
-		})
+		if len(policies[policyName]) != 0 {
+			printMsg(userPolicyMessage{
+				op:         "list",
+				Policy:     policyName,
+				PolicyJSON: policies[policyName],
+			})
+		} else {
+			fatalIf(probe.NewError(fmt.Errorf("The specified policy: %s not found", policyName)), "Cannot find policy")
```
vadmeste: Neat change: The error message can be better this way:

`fatalIf(probe.NewError(fmt.Errorf("`%s` is not found", policyName)), "Cannot list the policy")`

sinhaashish: Done.

Issue #2771, opened 2019-05-14
Add --ignore-existing to bucket notification event creation


poksiala:
## Expected behavior
It would be great benefit to minio deployment automation if we could ignore `event add` errors that are caused by the rule already existing like we can ignore bucket already existing
```
mc event add myminio/mybucket arn:minio:sqs::1:webhook --event put
mc: <INFO> Identical rule already exists.
```
## Actual behavior
```
mc event add myminio/mybucket arn:minio:sqs::1:webhook --event put
mc: <ERROR> Cannot enable notification on the specified bucket. An object key name filtering rule defined with overlapping prefixes, overlapping suffixes, or overlapping combinations of prefixes and suffixes for the same event types.
```

## Steps to reproduce the behavior
Add same event rule twice to single bucket

## mc version
Version: 2019-05-01T23:27:44Z
Release-tag: RELEASE.2019-05-01T23-27-44Z
Commit-id: b1aa2232ef3babc5f06bacb1b0044022185457ca

## System information
Mac

kannappanr:
@poksiala Will talk about this internally and get back to you.

PR #2773, opened 2019-05-16
Add --ignore-existing/-p flag up addition of event

sinhaashish:
If the event is already present the server throws an error. With the --ignore-existing/ -p flag these error messages are discarded.

Fixes https://github.com/minio/mc/issues/2771

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -230,7 +230,7 @@ func (c *s3Client) GetURL() clientURL {
 }
 
 // Add bucket notification
-func (c *s3Client) AddNotificationConfig(arn string, events []string, prefix, suffix string) *probe.Error {
+func (c *s3Client) AddNotificationConfig(arn string, events []string, prefix, suffix string, ignoreExisting bool) *probe.Error {
 	bucket, _ := c.url2BucketAndObject()
 	// Validate total fields in ARN.
 	fields := strings.Split(arn, ":")
@@ -286,6 +286,9 @@ func (c *s3Client) AddNotificationConfig(arn string, events []string, prefix, su
 
 	// Set the new bucket configuration
 	if err := c.api.SetBucketNotification(bucket, mb); err != nil {
+		if ignoreExisting && strings.Contains(err.Error(), "An object key name filtering rule defined with overlapping prefixes, overlapping suffixes, or overlapping combinations of prefixes and suffixes for the same event types") {
+			return nil
+		}
 		return probe.NewError(err)
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


cmd/client-s3.go
```diff
@@ -251,6 +251,9 @@ func (c *s3Client) AddNotificationConfig(arn string, events []string, prefix, su
 	// Get any enabled notification.
 	mb, e := c.api.GetBucketNotification(bucket)
 	if e != nil {
+		if ignoreExisting {
```
harshavardhana: We should validate the right error here, we shouldn't ignore all errors

cmd/event-add.go
```diff
@@ -65,7 +69,9 @@ EXAMPLES:
 
    2. Enable bucket notification with filters parameters
      $ {{.HelpName}} s3/mybucket arn:aws:sqs:us-west-2:444455556666:your-queue --event put,delete,get --prefix photos/ --suffix .jpg
-
+   
+   3. Ignore duplicate bucket notification with --ignore-existing/-p flag
```
harshavardhana: Use the short form as referenced in the command line @sinhaashish we don't need to talk about `--ignore-existing`

cmd/client-s3.go
sinhaashish: done.

cmd/event-add.go
sinhaashish: done.

sinhaashish:
It should be reviewed once [PR ](https://github.com/minio/minio/pull/7690) is merged .

sinhaashish:
@harshavardhana  @poornas @vadmeste  PTAL

cmd/client-s3.go
```diff
@@ -286,6 +286,9 @@ func (c *s3Client) AddNotificationConfig(arn string, events []string, prefix, su
 
 	// Set the new bucket configuration
 	if err := c.api.SetBucketNotification(bucket, mb); err != nil {
+		if ignoreExisting && err.Error() == "An object key name filtering rule defined with overlapping prefixes, overlapping suffixes, or overlapping combinations of prefixes and suffixes for the same event types." {
```
vadmeste: You might want to remove the dot in the error message and uses strings.Contains() instead

sinhaashish: done. @vadmeste  PTAL

cmd/event-add.go
poornas:
```suggestion
   3. Ignore duplicate bucket notification with -p flag
```

PR #2785, opened 2019-05-31
Accept SECRET_KEY using prompt and standard input.

### LINKED ISSUE #2651: Consider password prompt or env variable when using mc admin user add
## Info
When adding a user using MC, we have to provide the access secret on the CLI. Minio sometimes (depending on setup, usually on distributed) can take upwards to 20 seconds to respond. During this time, the password is shown in ```ps aux``` for any user with access to the machine to see.

## Expected behavior
Ability to securely add users

## Actual behavior
Password can be exposed in running process list if task takes longer than expected

## Possible solution
Environment variable for automation, user password prompt for interactive setup.

1. Set up a Minio, the slower the system the better but doesn't matter for this test case
2. add a user and ps aux | grep mc (eg ```mc admin user add zefie test testpassword readonly & ps aux | grep mc```)
[1] Password is exposed in process list (eg ```root     24913  0.0  0.0 445900 14460 pts/2    Rl   14:00   0:00 mc admin user add zefie test testpassword readonly```

## mc version
- DEVELOPMENT.GOGET ;) ... realistically 0854af6e6d3e9f79e8d9c98db5e0d133a527923b

## System information
* Server type and version: Debian Linux 9 x86_64
* Operating System and version (`uname -a`): Linux s3 4.9.0-8-amd64 #1 SMP Debian 4.9.130-2 (2018-10-27) x86_64 GNU/Linux

sinhaashish:
The `ACCESS_KEY` , `SECRET_KEY` is taken as an input . This is done to avoid the credentials shown
in `ps aux` command.
Adding a new user asks for their credentials.

To test:
1. `mc config host add myminio http://192.168.86.149:9000 `
```
Enter Access Key : BKIKJAA5BMMU2RHO6IBB
Enter Secret Key : V8f1CwQqAcwo80UEIJEjc5gVQUSSx5ohQ9GSrr12
Added `myminio` successfully.
```

2. `mc admin user add  myminio foobar readonly`
``` 
Enter Access Key : testuser 
Enter Secret Key : testuser123
```
Fixes #2651

### DIFF
--- a/cmd/admin-user-add.go
+++ b/cmd/admin-user-add.go
@@ -17,14 +17,18 @@
 package cmd
 
 import (
+	"bufio"
 	"fmt"
+	"os"
 	"strings"
+	"syscall"
 
 	"github.com/fatih/color"
 	"github.com/minio/cli"
 	json "github.com/minio/mc/pkg/colorjson"
 	"github.com/minio/mc/pkg/console"
 	"github.com/minio/mc/pkg/probe"
+	"golang.org/x/crypto/ssh/terminal"
 )
 
 var adminUserAddCmd = cli.Command{
@@ -50,15 +54,19 @@ FLAGS:
   {{end}}
 EXAMPLES:
   1. Add a new user 'foobar' to MinIO server.
-     {{.DisableHistory}}
-     {{.Prompt}} {{.HelpName}} myminio foobar foo12345
-     {{.EnableHistory}}
+     $ set -o history
+     $ {{.HelpName}} myminio foobar foo12345
+     $ set +o history
+  2. Add a new user 'testuser' to MinIO server.
+     $ {{.HelpName}} myminio
+        Enter Access Key : testuser 
+        Enter Secret Key : testuser123
 `,
 }
 
 // checkAdminUserAddSyntax - validate all the passed arguments
 func checkAdminUserAddSyntax(ctx *cli.Context) {
-	if len(ctx.Args()) != 3 {
+	if len(ctx.Args()) > 3 || len(ctx.Args()) < 1 {
 		cli.ShowCommandHelpAndExit(ctx, "add", 1) // last argument is exit code
 	}
 }
@@ -115,6 +123,35 @@ func (u userMessage) JSON() string {
 	return string(jsonMessageBytes)
 }
 
+// fetchUserKeys - returns the access and secret key
+func fetchUserKeys(args cli.Args) (string, string) {
+	accessKey := ""
+	secretKey := ""
+	console.SetColor(cred, color.New(color.FgYellow, color.Italic))
+	reader := bufio.NewReader(os.Stdin)
+
+	switch numberOfArgs := len(args); {
+	case numberOfArgs == 1:
+		fmt.Printf("%s", console.Colorize(cred, "Enter Access Key : "))
+		value, _, _ := reader.ReadLine()
+		accessKey = string(value)
+		fmt.Printf("%s", console.Colorize(cred, "Enter Secret Key : "))
+		bytePassword, _ := terminal.ReadPassword(int(syscall.Stdin))
+		secretKey = string(bytePassword)
+	case numberOfArgs == 2:
+		accessKey = args.Get(1)
+		fmt.Printf("%s", console.Colorize(cred, "Enter Secret Key : "))
+		bytePassword, _ := terminal.ReadPassword(int(syscall.Stdin))
+		secretKey = string(bytePassword)
+	default:
+		accessKey = args.Get(1)
+		secretKey = args.Get(2)
+	}
+	fmt.Printf("\n")
+
+	return accessKey, secretKey
+}
+
 // mainAdminUserAdd is the handle for "mc admin user add" command.
 func mainAdminUserAdd(ctx *cli.Context) error {
 	checkAdminUserAddSyntax(ctx)
@@ -124,17 +161,18 @@ func mainAdminUserAdd(ctx *cli.Context) error {
 	// Get the alias parameter from cli
 	args := ctx.Args()
 	aliasedURL := args.Get(0)
+	accessKey, secretKey := fetchUserKeys(args)
 
 	// Create a new MinIO Admin Client
 	client, err := newAdminClient(aliasedURL)
 	fatalIf(err, "Unable to initialize admin connection.")
 
-	fatalIf(probe.NewError(client.AddUser(args.Get(1), args.Get(2))).Trace(args...), "Cannot add new user")
+	fatalIf(probe.NewError(client.AddUser(accessKey, secretKey)).Trace(args...), "Cannot add new user")
 
 	printMsg(userMessage{
 		op:         "add",
-		AccessKey:  args.Get(1),
-		SecretKey:  args.Get(2),
+		AccessKey:  accessKey,
+		SecretKey:  secretKey,
 		UserStatus: "enabled",
 	})
 
--- a/cmd/config-host-add.go
+++ b/cmd/config-host-add.go
@@ -17,14 +17,21 @@
 package cmd
 
 import (
+	"bufio"
+	"fmt"
 	"math/rand"
+	"os"
+	"syscall"
 	"time"
 
 	"github.com/fatih/color"
 	"github.com/minio/cli"
 	"github.com/minio/mc/pkg/console"
 	"github.com/minio/mc/pkg/probe"
+	"golang.org/x/crypto/ssh/terminal"
 )
+
+const cred = "YellowItalics"
 
 var hostAddFlags = []cli.Flag{
 	cli.StringFlag{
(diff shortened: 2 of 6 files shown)


cmd/admin-user-add.go
```diff
@@ -17,19 +17,28 @@
...
 	"github.com/minio/cli"
 	json "github.com/minio/mc/pkg/colorjson"
 	"github.com/minio/mc/pkg/console"
 	"github.com/minio/mc/pkg/probe"
 )
 
+var userAddFlags = []cli.Flag{
+	cli.BoolFlag{
+		Name:  "stdin",
+		Usage: "allow a stdin remove operation",
```
poornas: should be `add operation` here

cmd/config-host-add.go
```diff
@@ -55,47 +61,57 @@ FLAGS:
...
+	if argsNr < 3 || argsNr > 4 {
 		fatalIf(errInvalidArgument().Trace(ctx.Args().Tail()...),
 			"Incorrect number of arguments for host add command.")
 	}
 
 	alias := args.Get(0)
 	url := args.Get(1)
 	accessKey := args.Get(2)
-	secretKey := args.Get(3)
+	//secretKey := args.Get(3)
```
poornas: this commented line needs to be removed

docs/minio-client-complete-guide.md
```diff
@@ -127,23 +132,43 @@ MinIO server displays URL, access and secret keys.
...
+mc config host add gcs  https://storage.googleapis.com BKIKJAA5BMMU2RHO6IBB  --api S3v2
+```
+```sh
+Enter Secret Key >>>V8f1CwQqAcwo80UEIJEjc5gVQUSSx5ohQ9GSrr12
+```
+
+### Example - MinIO Cloud Storage using standard input
+MinIO server displays URL, access and secret keys.
+
+
```
poornas: extra newline needs to be removed.

cmd/config-host-add.go
```diff
@@ -55,47 +61,57 @@ FLAGS:
   {{range .VisibleFlags}}{{.}}
   {{end}}
 EXAMPLES:
-  1. Add Amazon S3 storage service under "mys3" alias. For security reasons turn off bash history momentarily.
+  1. Add Amazon S3 storage service under "mys3" alias. For security reasons turn off bash history momentarily. Console prompts for the secret key.
      $ set +o history
      $ {{.HelpName}} mys3 https://s3.amazonaws.com \
-                 BKIKJAA5BMMU2RHO6IBB V8f1CwQqAcwo80UEIJEjc5gVQUSSx5ohQ9GSrr12
+				 BKIKJAA5BMMU2RHO6IBB
```
poornas: indentation on line 67 needs to be fixed

cmd/admin-user-add.go
```diff
@@ -108,17 +128,27 @@ func mainAdminUserAdd(ctx *cli.Context) error {
...
 
+	stdin := ctx.Bool("stdin")
+
 	// Get the alias parameter from cli
 	args := ctx.Args()
 	aliasedURL := args.Get(0)
 
+	secretKey := ""
+	if stdin {
+		fmt.Scanln(&secretKey)
```
harshavardhana: We don't need to support stdin reading, it really doesn't make sense from a security point of view. We can simply have prompt and no prompt if secretKey was provided on the command line. 

We should avoid a breaking change here IMO @sinhaashish for example 

Old style is still honored
```
mc admin user add myminio foobar foo12345 writeonly
``` 

Let's say user didn't any inputs, then user should use interactive mode to start the prompt `mc admin user add -i` which starts the prompt
```
mc admin user add -i
Username: foobar
Policy: writeonly
Password: *****
```

This needs to be confirmed with @abperiasamy which I will let you know soon.

sinhaashish: :+1:

aead: Also, if we do that we should read a secret from the command line using [`ReadPassword`](https://godoc.org/golang.org/x/crypto/ssh/terminal#Terminal.ReadPassword).

vadmeste: `cat ~/my_protected_password | mc admin user add --stdin` where ~/my_portected_password is readable only by the user looks like a little more secure way to pass the password

cmd/admin-user-add.go
```diff
@@ -17,19 +17,28 @@
...
 	"github.com/fatih/color"
 	"github.com/minio/cli"
 	json "github.com/minio/mc/pkg/colorjson"
 	"github.com/minio/mc/pkg/console"
 	"github.com/minio/mc/pkg/probe"
 )
 
+var userAddFlags = []cli.Flag{
+	cli.BoolFlag{
+		Name:  "stdin",
```
vadmeste: It looks like --passwd-from-stdin is better name, or --stdin-passwd, ..

cmd/admin-user-add.go
harshavardhana: Yeah but it's not secure from any point of view, it is better to not have such things @vadmeste

harshavardhana:
Okay so @abperiasamy confirmed two things

- preserve the current syntax when all arguments are provided and no change there
- show a prompt if user doesn't type all the input options, let's say alias is available and rest are not available throw a prompt. Where each prompt takes `accessKey`, `secretKey` etc..

So you need to redo this PR to have this behavior.

sinhaashish:
> Okay so @abperiasamy confirmed two things
> 
> * preserve the current syntax when all arguments are provided and no change there
> * show a prompt if user doesn't type all the input options, let's say alias is available and rest are not available throw a prompt. Where each prompt takes `accessKey`, `secretKey` etc..
> 
> So you need to redo this PR to have this behavior.

:+1:

sinhaashish:
@harshavardhana @vadmeste @poornas  modifed the PR. PTAL

docs/minio-admin-complete-guide.md
```diff
@@ -101,6 +101,22 @@ Alias is simply a short name to your MinIO service. MinIO end-point, access and
 mc config host add minio http://192.168.1.51:9000 BKIKJAA5BMMU2RHO6IBB V7f1CwQqAcwo80UEIJEjc5gVQUSSx5ohQ9GSrr12
 ```
 
+Host can also be added by passing access key , secret key using prompt
+
+```sh
+mc config host add <ALIAS> <YOUR-MINIO-ENDPOINT>
+```
+
+This asks for details
```
harshavardhana: Make this default behavior and positional as optional.. meaning move the preference of prompt based input as primary style.

Also the wording can be improved @eco-minio please help here .

cmd/admin-user-add.go
```diff
@@ -111,19 +120,34 @@ func mainAdminUserAdd(ctx *cli.Context) error {
...
 	args := ctx.Args()
 	aliasedURL := args.Get(0)
+	accessKey := ""
+	secretKey := ""
+	policyName := ""
+	if len(args) == 1 {
+		accessKey = prompt.Input("Enter Access Key : ", completer,
+			prompt.OptionPrefixTextColor(prompt.Yellow))
+		secretKey = prompt.Input("Enter Secret Key : ", completer,
+			prompt.OptionPrefixTextColor(prompt.Yellow))
```
aead: The secret key **must** be read differently. PTAL at https://godoc.org/golang.org/x/crypto/ssh/terminal#Terminal.ReadPassword

harshavardhana: And yes thanks @aead we need to use prompt

harshavardhana:
One more thing this PR also needs to do is when `--json`,  `--quiet` or if the tty is not a terminal - it should turn off prompt. This is a script mode or non-interactive.

sinhaashish:
> One more thing this PR also needs to do is when `--json`, `--quiet` or if the tty is not a terminal - it should turn off prompt. This is a script mode or non-interactive.

Done with the other changes . Can you elaborate `if the tty is not a terminal ` .  This PR prompts for access and secret key only if it is not passed.  IMO in script mode access and secret key will be passed so it wont prompt for user input.

docs/minio-client-complete-guide.md
```diff
@@ -131,6 +135,19 @@ MinIO server displays URL, access and secret keys.
...
+mc config host add minio http://192.168.1.51
+```
+
+This prompts for details:
+
+```
+Enter Access Key : BKIKJAA5BMMU2RHO6IBB
+Enter Secret Key : V7f1CwQqAcwo80UEIJEjc5gVQUSSx5ohQ9GSrr12
+Enter API signature. An optional argument. Default value 'S3v4' :
+Enter lookup. An optional argument. Default value 'auto' :
```
vadmeste: I believe both of these prompts are not there

docs/minio-admin-complete-guide.md
```diff
@@ -419,7 +432,22 @@ COMMANDS:
 *Example: Add a new user 'newuser' on MinIO.*
 
 ```
-mc admin user add myminio/ newuser newuser123
+
+*Example: Add a new user 'testuser' on MinIO, with 'writeonly'  using standard input.*
+
+```sh
+$ set -o history
```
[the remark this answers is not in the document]
sinhaashish: removed

cmd/admin-user-add.go
```diff
@@ -50,15 +54,20 @@ FLAGS:
   {{end}}
 EXAMPLES:
   1. Add a new user 'foobar' to MinIO server.
-     {{.DisableHistory}}
-     {{.Prompt}} {{.HelpName}} myminio foobar foo12345
-     {{.EnableHistory}}
+     $ set -o history
+     $ {{.HelpName}} myminio foobar foo12345
+	 $ set +o history
```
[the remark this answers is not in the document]
sinhaashish: corrected

docs/minio-client-complete-guide.md
sinhaashish: Thanks, removed.

cmd/admin-user-add.go
```diff
@@ -115,6 +123,35 @@ func (u userMessage) JSON() string {
...
+	console.SetColor(cred, color.New(color.FgYellow, color.Italic))
+	reader := bufio.NewReader(os.Stdin)
+
+	switch numberOfArgs := len(args); {
+	case numberOfArgs == 1:
+		fmt.Printf("%s", console.Colorize(cred, "Enter Access Key : "))
+		value, _, _ := reader.ReadLine()
+		accessKey = string(value)
+		fmt.Printf("%s", console.Colorize(cred, "Enter Secret Key : "))
+		bytePassword, _ := terminal.ReadPassword(int(syscall.Stdin))
```
aead: Missing proper error handling

cmd/admin-user-add.go
```diff
@@ -115,6 +123,35 @@ func (u userMessage) JSON() string {
...
+		secretKey = string(bytePassword)
+	case numberOfArgs == 2:
+		accessKey = args.Get(1)
+		fmt.Printf("%s", console.Colorize(cred, "Enter Secret Key : "))
+		bytePassword, _ := terminal.ReadPassword(int(syscall.Stdin))
```
aead: Missing proper error handling

cmd/admin-user-add.go
aead: Also according to [Terminal.ReadPassword](https://godoc.org/golang.org/x/crypto/ssh/terminal#Terminal.ReadPassword) we are supposed to pass the prompt. In any case we should use `os.File.Fd()` instead of the `syscall` package.

harshavardhana:
are we going to implement this? @kannappanr @sinhaashish  - if there are no plans can we close them out?

kannappanr:
ping @sinhaashish

h0tw1r3:
please implement this. it is not uncommon for security policies to require (or force) shell logging.

sinhaashish:
Closing this for now . Will send another PR for this

harshavardhana:
@sinhaashish would you take a look at #3083

PR #2793, opened 2019-06-13
doc: add info about --attr flag for cp command

adarrra:


docs/minio-client-complete-guide.md
```diff
@@ -562,6 +563,13 @@ encryptedobject:    14 B / 14 B  ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
 ```
 Notice that two different aliases myminio1 and myminio2 are used for the same endpoint to provide the old secretkey and the newly rotated key.
 
+*Example: Copy file and assign Cache-Control header for it*
+
+```sh
+mc cp --attr Cache-Control=no-cache myscript.js scripts/
```
kannappanr: You are missing the bucket name here. Assuming scripts is the alias name

adarrra: updated

docs/minio-client-complete-guide.md
```diff
@@ -562,6 +563,13 @@ encryptedobject:    14 B / 14 B  ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
 ```
 Notice that two different aliases myminio1 and myminio2 are used for the same endpoint to provide the old secretkey and the newly rotated key.
 
+*Example: Copy file and assign Cache-Control header for it*
+
+```sh
+mc cp --attr Cache-Control=no-cache myscript.js S3/scripts
```
harshavardhana: Can we consistently use alias name as S3 -> play instead?

docs/minio-client-complete-guide.md
```diff
@@ -562,6 +563,13 @@ encryptedobject:    14 B / 14 B  ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
 ```
 Notice that two different aliases myminio1 and myminio2 are used for the same endpoint to provide the old secretkey and the newly rotated key.
 
+*Example: Copy file and assign Cache-Control header for it*
```
harshavardhana: Copy a javascript file to object storage and assign `Cache-Control` header to the uploaded object.

adarrra: updated

docs/minio-client-complete-guide.md
adarrra: updated

CompuRoot:
As of now [documentation](https://docs.min.io/docs/minio-client-complete-guide#cp) that describing `mc cp ` command contained two the same options `--attr`.
```
USAGE:
   mc cp [FLAGS] SOURCE [SOURCE...] TARGET

FLAGS:
  --recursive, -r                    copy recursively
  --older-than value                 copy object(s) older than N days (default: 0)
  --newer-than value                 copy object(s) newer than N days (default: 0)
  --storage-class value, --sc value  set storage class for new object(s) on target
  --attr                             add custom metadata for the object
  --encrypt value                    encrypt/decrypt objects (using server-side encryption with server managed keys)
  --encrypt-key value                encrypt/decrypt objects (using server-side encryption with customer provided keys)
  --attr                             apply metadata to objects (format: KeyName1=string,KeyName2=string)
  --help, -h                         show help
```

I suggest to remove this line as it redundant and confusing:
`      --attr                             add custom metadata for the object`

harshavardhana:
@CompuRoot thanks for the find - feel free to send a PR with the fix.

PR #2802, opened 2019-06-18
add du command support

balamurugana:
This command works like `du` command in unices i.e. summarize disk
usage of given TARGET recursively.

### DIFF
--- a/cmd/du-main.go
+++ b/cmd/du-main.go
@@ -1,0 +1,187 @@
+/*
+ * MinIO Client (C) 2019 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"fmt"
+	"strings"
+
+	neturl "net/url"
+
+	humanize "github.com/dustin/go-humanize"
+	"github.com/fatih/color"
+	"github.com/minio/cli"
+	json "github.com/minio/mc/pkg/colorjson"
+	"github.com/minio/mc/pkg/console"
+	"github.com/minio/mc/pkg/probe"
+)
+
+// du specific flags.
+var (
+	duFlags = []cli.Flag{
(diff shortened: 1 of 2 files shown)


cmd/du-main.go
```diff
@@ -0,0 +1,185 @@
...
+      $ {{.HelpName}} s3/jazz-songs
+
+   2. Summarize disk usage of 'louis' prefix in 'jazz-songs' bucket recursively.
+      $ {{.HelpName}} -H s3/jazz-songs/louis/
+`,
+}
+
+// Structured message depending on the type of console.
+type duMessage struct {
+	Prefix string `json:"prefix"`
```
kannappanr: We seemed to use `Status      string  `json:"status"`` everrywhere, so in the case of json, I think applications will look for `status` to be `success`

balamurugana: I believe you are asking to add an additional field called `"status"` with value `"success"`

harshavardhana: It is an mc json API behavior for all mc commands, check this with other commands..

This is the common guidance for all commands given by @abperiasamy

balamurugana: Got it. Why [rmMessage](https://github.com/minio/mc/blob/master/cmd/rm-main.go#L120) doesn't have `status`? Bug in `mc` somewhere or am I missing something?

harshavardhana: It is a bug..

cmd/du-main.go
```diff
@@ -0,0 +1,187 @@
...
+
+// du specific flags.
+var (
+	duFlags = []cli.Flag{
+		cli.IntFlag{
+			Name:  "max-depth, d",
+			Usage: "print the total for a folder prefix only if it is N or fewer levels below the command line argument",
+		},
+		cli.BoolFlag{
+			Name:  "human-readable, H",
```
kannappanr: Human-readable, can be by default. You can have `-b` to display in bytes. We can also check with AB.

balamurugana: OK. I will do the change once its confirmed.

kannappanr: @balamurugana Can you confirm this change with AB?

kannappanr:
@balamurugana Can you please resolve the conflict? Also, did you talk to AB about whether human readable should be default?

kannappanr:
@balamurugana Can you please resolve the conflicts

kannappanr:
@balamurugana There won't be a flag called `--human-readable, -H `.

The output will be like this
```
$ mc du play/test
7.1 MB	test
```
Also, `--max-depth value, -d` should be `--depth value, -d`

balamurugana:
@kannappanr 

> There won't be a flag called `--human-readable, -H `.
> 
> The output will be like this
> 
> ```
> $ mc du play/test
> 7.1 MB	test
> ```
> 
> Also, `--max-depth value, -d` should be `--depth value, -d`

Done

PR #2803, opened 2019-06-18
[proposal] Enable watch by default

harshavardhana:


deekoder:
Implications are: it won't work on non-minio servers. But mc mirror --watch already didn't work on non-minio servers. We would still do mirror and not --watch on non-minio servers with this proposal.

harshavardhana:
Closing this @abperiasamy as he doesn't want this since its a major breaking change and let it be opt-in.

PR #2804, opened 2019-06-19
`mc tree` command inclusion

Praveenrajmani:
`mc tree` will work similary to the `tree` command utility. Lists all the files,objects, directories in a tree format.


cmd/tree-main.go
```diff
@@ -0,0 +1,211 @@
+/*
+ * MinIO Client (C) 2014-2019 MinIO, Inc.
+ *
```
harshavardhana: This command is new only use 2019

Praveenrajmani: Done

cmd/tree-main.go
```diff
@@ -0,0 +1,218 @@
...
+
+// doTree - list all entities inside a folder in a tree format.
+func doTree(url string, level int, leaf bool) error {
+	targetAlias, targetURL, _ := mustExpandAlias(url)
+	if !strings.HasSuffix(targetURL, "/") {
+		targetURL += "/"
+	}
+
+	clnt, err := newClientFromAlias(targetAlias, targetURL)
+	fatalIf(err.Trace(targetURL), "Unable to initialize target `"+targetURL+"`.")
```
balamurugana: Return this error than `fatalIf()`

cmd/tree-main.go
```diff
@@ -0,0 +1,218 @@
...
+
+	prefixPath := clnt.GetURL().Path
+	separator := string(clnt.GetURL().Separator)
+	if !strings.HasSuffix(prefixPath, separator) {
+		prefixPath = prefixPath[:strings.LastIndex(prefixPath, separator)+1]
```
balamurugana: Use `strings.TrimSuffix()` for readability.

cmd/tree-main.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	bucketNameShowed := false
+	var prev *clientContent
+	show := func(end bool) error {
+		if level == 0 && !bucketNameShowed {
+			bucketNameShowed = true
+			fmt.Printf("%s\n", console.Colorize("Dir", url))
```
balamurugana: You would need to do `treeMessage` type printing. Refer `rm-main.go`

cmd/tree-main.go
```diff
@@ -0,0 +1,218 @@
...
+		prefixPath = strings.TrimPrefix(prefixPath, "."+separator)
+
+		var entry string
+		if strings.HasSuffix(contentURL, "/") {
+			entry = console.Colorize("Dir", strings.TrimPrefix(contentURL, prefixPath))
+		} else {
+			entry = console.Colorize("File", strings.TrimPrefix(contentURL, prefixPath))
+		}
+
+		fmt.Printf("%s%s\n", branchString, entry)
```
balamurugana: same as above

krishnasrinivas:
@Praveenrajmani `--level` argument is also needed by the customer.
`--directory` will also be useful (and easy to implement)

balamurugana:
@ebozduman What is `myminio` here? if it is a directory it works exactly like `tree` command. If it is S3 alias, the underneath S3 client implementation in `mc` returns only bucket names (according to S3 protocol). Tree is an expensive command in listing in S3. We would need to know whether we want to have tree view of entire S3.

cmd/tree-main.go
```diff
@@ -0,0 +1,218 @@
...
+`,
+}
+
+// checkTreeSyntax - validate all the passed arguments
+func checkTreeSyntax(ctx *cli.Context) {
+	args := ctx.Args()
+	if !ctx.Args().Present() {
+		args = []string{"."}
+	}
+	for _, arg := range args {
```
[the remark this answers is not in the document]
Praveenrajmani: Will 

> There is a bug in the implementation.
> The target/alias level is not recursively expanded to list the 2nd and more deeper level dirs and files, but only the 1st level directories and files are listed.
> 
> Example:
> 
> ```
> $ ./mc tree myminio
> myminio
> ├── aaa111
> └── aabb
> $ 
> ```
> 
> versus tree command:
> 
> ```
> $ tree /var/tmp/ersan
> /var/tmp/ersan
> ├── aaa111
> │   └── a
> │       └── b
> │           └── c
> └── aabb
> 5 directories, 0 files
> ```
> 
> One more thing:
> The last status line that shows the total number of found directories and files at the end of the `tree` command output is missing in our implementation, **`5 directories, 0 files`** in our above example.
> 
> Is this by design or is it forgotten to be implemented?
> If we need to have our `mc tree` command behave the same with the regular `tree` command, then we need to implement it.

Fixing this :+1:

Praveenrajmani:
> There is a bug in the implementation.
> The target/alias level is not recursively expanded to list the 2nd and more deeper level dirs and files, but only the 1st level directories and files are listed.
> 
> Example:
> 
> ```
> $ ./mc tree myminio
> myminio
> ├── aaa111
> └── aabb
> $ 
> ```
> 
> versus tree command:
> 
> ```
> $ tree /var/tmp/ersan
> /var/tmp/ersan
> ├── aaa111
> │   └── a
> │       └── b
> │           └── c
> └── aabb
> 5 directories, 0 files
> ```
> 
> One more thing:
> The last status line that shows the total number of found directories and files at the end of the `tree` command output is missing in our implementation, **`5 directories, 0 files`** in our above example.
> 
> Is this by design or is it forgotten to be implemented?
> If we need to have our `mc tree` command behave the same with the regular `tree` command, then we need to implement it.

@ebozduman Will confirm this with @ebozduman . I think it makes sense to print the number of directories and files as well

Praveenrajmani:
Need to discuss some changes with @abperiasamy. It will be WIP untill then

ebozduman:
@balamurugana,
`myminio` is the alias to my minio server or my target.

If you've discussed it before and decided the top level `mc tree <alias>` command is only going to display the buckets, then it is fine.
Maybe @Praveenrajmani can also ask @ab about it if you have not already discussed it before with him.

I was expecting to see something like;
```
$ ./mc tree myminio
myminio
├── aaa111
│   └── a
│       └── b
│           └── c
└── aabb
5 directories, 0 files
```

cmd/tree-main.go
Praveenrajmani: The `err` is of type - `*probe.Error`. It wont be compatible with the top level calls - which listens for `Error` type

cmd/tree-main.go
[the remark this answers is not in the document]
ebozduman: @Praveenrajmani,

Do you mean you are going to work on both of the issues or just one of them?

cmd/ls.go
```diff
@@ -145,6 +145,7 @@ func doList(clnt Client, isRecursive, isIncomplete bool) error {
 		// Trim prefix path from the content path.
 		contentURL = strings.TrimPrefix(contentURL, prefixPath)
 		content.URL.Path = contentURL
+		fmt.Println(content.URL.Path)
```
balamurugana: Is it debug message?

cmd/tree-main.go
```diff
@@ -0,0 +1,227 @@
...
+
+			if err := doTree(url, level+1, end, dirClosed); err != nil {
+				return err
+			}
+		}
+
+		return nil
+	}
+
+	for content := range clnt.List(false, false, DirNone) {
```
balamurugana: There should be a bug in `Client.List()`, For `ListAllMyBuckets` S3 calls, it should return each bucket entry as directory is `Type.IsDir()` should return true, but its not. else S3 level tree works normally.  Please confirm with this @abperiasamy and fix it accordingly.

cmd/tree-main.go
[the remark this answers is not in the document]
Praveenrajmani: @ebozduman AB told the dir/files count is not needed here. I am working on the other fix

docs/minio-client-complete-guide.md
```diff
@@ -285,6 +287,34 @@ mc ls play
...
+  --files, -f                   include files in tree
+  --depth, -d                   set the maximum depth of the tree
+```
+
+_Example: tree all buckets and objects on play/bucket._
+
+```sh
+mc tree play/test-bucket
+play/test-bucket/
+ââ dir_a
```
Praveenrajmani: this is because of my emacs' encoding settings. Will fix this

Praveenrajmani:
As per the discussion with @abperiasamy . By default, we will display only the directories. We will have a flag, say (`--files`) to include the files also. And also one more flag is included - `--depth` which will set the maximum depth of the tree

For eg,
- `mc tree -d 1 /home` will display the first level directories alone 
- `mc tree -f /home` will display all the directories with files
- `mc tree -f -d 2 /home` will display all the second level dirs/files

cmd/tree-main.go
Praveenrajmani: done

cmd/ls.go
Praveenrajmani: yes. removed it now

Praveenrajmani:
Ready for review @ebozduman @balamurugana

cmd/ls.go
```diff
@@ -145,6 +145,7 @@ func doList(clnt Client, isRecursive, isIncomplete bool) error {
 		// Trim prefix path from the content path.
 		contentURL = strings.TrimPrefix(contentURL, prefixPath)
 		content.URL.Path = contentURL
+
```
balamurugana: Remove this whitespace to leave no diff

cmd/tree-main.go
```diff
@@ -0,0 +1,279 @@
...
+// Structured message depending on the type of console.
+type treeMessage struct {
+	Entry        string
+	IsDir        bool
+	BranchString string
+}
+
+// Colorized message for console printing.
+func (t treeMessage) String() string {
+	//fmt.Printf("%s\n", console.Colorize("Dir", url))
```
balamurugana: remove this line

cmd/tree-main.go
```diff
@@ -0,0 +1,279 @@
...
+	if !strings.HasSuffix(targetURL, "/") {
+		targetURL += "/"
+	}
+
+	clnt, err := newClientFromAlias(targetAlias, targetURL)
+	fatalIf(err.Trace(targetURL), "Unable to initialize target `"+targetURL+"`.")
+
+	prefixPath := clnt.GetURL().Path
+	separator := string(clnt.GetURL().Separator)
+	prefixPath = strings.TrimSuffix(prefixPath, prefixPath[strings.LastIndex(prefixPath, separator)+1:])
```
balamurugana: What are you trying to do here? `filepath.Dir()` or `filepath.Base()`

Praveenrajmani: filepath.Dir() is what i was trying. It is doing `filepath.Clean()` additionally. so i preferred this.

cmd/tree-main.go
Praveenrajmani: done

Praveenrajmani:
Can i get some reviews @ebozduman @balamurugana

Praveenrajmani:
PTAL @balamurugana @ebozduman

cmd/tree-main.go
```diff
@@ -0,0 +1,285 @@
...
+
+USAGE:
+  {{.HelpName}} [FLAGS] TARGET [TARGET ...]
+
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}
+EXAMPLES:
+   1. List buckets on Amazon S3 cloud storage in a tree format.
+      $ {{.HelpName}} myS3
```
harshavardhana: We don't have this type of convention in other commands why can't we just say `s3/` ? Also we should provide some MinIO examples as well showing `myminio`

ebozduman: @harshavardhana ,
That was my comment and Praveen changed it per my request. It thought it would be more clear, but if the convention is the other way, then we should use `s3/`.

harshavardhana: Because we already have a default alias for `s3/`  we can just use that.  Also, we have started moving to provide `myminio` in our examples.

Praveenrajmani: changed it to s3/ :+1:

Praveenrajmani:
PTAL @ebozduman @balamurugana

cmd/tree-main.go
```diff
@@ -0,0 +1,285 @@
...
+  {{.HelpName}} - {{.Usage}}
+
+USAGE:
+  {{.HelpName}} [FLAGS] TARGET [TARGET ...]
+
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}
+EXAMPLES:
+   1. List buckets on Amazon S3 cloud storage in a tree format.
```
ebozduman: The default, when `--depth <value>` flag is not provided, is to list `all` the buckets.
So, we should say in the explanation something like,
`    1. List all buckets on Amazon S3 cloud storage in a tree format.`

cmd/tree-main.go
```diff
@@ -0,0 +1,285 @@
...
+      $ {{.HelpName}} s3
+
+   2. List all buckets on Amazon S3 cloud storage in a tree format.
```
ebozduman: `    2. List all buckets in "mybucket" on Amazon S3 cloud storage in a tree format.`

cmd/tree-main.go
```diff
@@ -0,0 +1,285 @@
...
+      $ {{.HelpName}} s3/mybucket/
+
+   3. List all buckets on Amazon S3 cloud storage on Microsoft Windows in a tree format.
```
ebozduman: `   3. List all buckets in "mybucket" on Amazon S3 cloud storage in a tree format.`

cmd/tree-main.go
```diff
@@ -0,0 +1,285 @@
...
+      $ {{.HelpName}} s3\mybucket\
+   
+   4. List all buckets including the objects in a tree format.
```
ebozduman: `4. List all buckets`**`and objects`**`in "mybucket" on Amazon S3 cloud storage in a tree format.`

ebozduman:
I also tested and verified the fix. It is good to go as soon as the last usage comments are addressed.

Praveenrajmani:
Review comments are addressed. PTAL @ebozduman @balamurugana

docs/minio-client-complete-guide.md
```diff
@@ -285,6 +287,34 @@ mc ls play
 [2016-04-08 20:58:18 IST]     0B mybucket/
 ```
 
+<a name="tree"></a>
+### Command `tree` - List Objects in a tree format
+
+`tree` command lists files, objects and directories in a tree format.
```
kannappanr: this probably needs to say `buckets` also

docs/minio-client-complete-guide.md
```diff
@@ -285,6 +287,34 @@ mc ls play
...
+USAGE:
+   mc tree [FLAGS] TARGET [TARGET ...]
+
+FLAGS:
+  --help, -h                    show help
+  --files, -f                   include files in tree
+  --depth, -d                   set the maximum depth of the tree
+```
+
+_Example: tree all buckets and objects on play/bucket._
```
kannappanr: This sentence doesn't make sense

docs/minio-client-complete-guide.md
```diff
@@ -285,6 +287,34 @@ mc ls play
 [2016-04-08 20:58:18 IST]     0B mybucket/
 ```
 
+<a name="tree"></a>
+### Command `tree` - List Objects in a tree format
```
kannappanr: This command can also display `buckets`, so the title should reflect that

cmd/tree-main.go
```diff
@@ -0,0 +1,285 @@
...
+	if t.IsDir {
+		entryType = "Dir"
+	}
+	return fmt.Sprintf("%s%s", t.BranchString, console.Colorize(entryType, t.Entry))
+}
+
+// JSON'ified message for scripting.
+// Does No-op. JSON requests are redirected to `ls -r --json`
+func (t treeMessage) JSON() string {
+	return ""
```
balamurugana: You would need to panic() here instead of returning empty string.

harshavardhana: We should use fatalIf not panic()

Praveenrajmani:
Addressed the comments. PTAL @balamurugana @kannappanr

harshavardhana:
This is blocked for now on my end to discuss with @abperiasamy about why the tree command is not following the `unix` tree style for listing files and directories.  For example, listing files is opt-in which is seems incorrect and may lead to unexpected behavior for users who are familiar with unix `tree` command.

harshavardhana:
Okay verified with @abperiasamy on the behavior here - it is intentional due to the number of objects on object storage can lead to an overload of the output on terminals and we are intentionally differentiating our behavior to keep the output simpler.

But one change that is needed is to avoid having `-1` as default and erroring out for depth with `0` - the following change is needed. 

```diff
diff --git a/cmd/tree-main.go b/cmd/tree-main.go
index 771c1de6..75913b9f 100644
--- a/cmd/tree-main.go
+++ b/cmd/tree-main.go
@@ -66,7 +66,6 @@ var treeFlags = []cli.Flag{
        cli.IntFlag{
                Name:  "depth, d",
                Usage: "sets the depth threshold",
-               Value: -1,
        },
 }
 
@@ -95,15 +94,15 @@ EXAMPLES:
 
    3. List all buckets in "mybucket" on Amazon S3 cloud storage hosted on Microsoft Windows in a tree format.
       $ {{.HelpName}} s3\mybucket\
-      
+
    4. List all buckets and objects in "mybucket" on Amazon S3 cloud storage in a tree format.
       $ {{.HelpName}} -f myminio/mybucket/
-   
+
    5. Set the depth of the tree for listing.
       $ {{.HelpName}} -d 2 myminio/mybucket/
 
-   6. List all the directories irrespective to the depth. -1 is the default value for depth.
-      $ {{.HelpName}} -d -1 myminio/mybucket/
+   6. List all the directories irrespective to the depth.
+      $ {{.HelpName}} myminio/mybucket/
 `,
 }
 
@@ -111,9 +110,12 @@ EXAMPLES:
 func checkTreeSyntax(ctx *cli.Context) {
        args := ctx.Args()
 
-       depth := ctx.Int("depth")
+       var depth = -1
+       if ctx.IsSet("depth") {
+               depth = ctx.Int("depth")
+       }
        if depth != -1 && depth <= 0 {
-               fatalIf(errInvalidArgument().Trace(args...), "depth should have a value greater than 0 or equal to -1")
+               fatalIf(errInvalidArgument().Trace(args...), "depth should have a value greater than 0")
        }
 
        if (args.Present()) && len(args) == 0 {
@@ -263,11 +265,15 @@ func mainTree(ctx *cli.Context) error {
        }
 
        includeFiles := ctx.Bool("files")
+       depth := -1
+       if ctx.IsSet("depth") {
+               depth = ctx.Int("depth")
+       }
 
        var cErr error
        for _, targetURL := range args {
                if !globalJSON {
-                       if e := doTree(targetURL, 1, false, "", ctx.Int("depth"), includeFiles); e != nil {
+                       if e := doTree(targetURL, 1, false, "", depth, includeFiles); e != nil {
                                cErr = e
                        }
                } else {
```

harshavardhana:
Also, cmd/complete.go should be updated to use the `tree` command for autocompletion.

harshavardhana:
```diff
diff --git a/cmd/complete.go b/cmd/complete.go
index 6793763b..0a366306 100644
--- a/cmd/complete.go
+++ b/cmd/complete.go
@@ -156,6 +156,7 @@ var completeCmds = map[string]complete.Predictor{
        "/stat":   complete.PredictOr(s3Completer, fsCompleter),
        "/watch":  complete.PredictOr(s3Completer, fsCompleter),
        "/policy": complete.PredictOr(s3Completer, fsCompleter),
+       "/tree":   complete.PredictOr(s3Complete{deepLevel: 2}, fsCompleter),
 
        "/mb":  aliasCompleter,
        "/sql": s3Completer,
```

Praveenrajmani:
done :+1:  PTAL @harshavardhana @balamurugana @ebozduman @kannappanr

PR #2812, opened 2019-07-02
autocompletion: Add control flag and a prompt

vadmeste:
*) Add --no-autocompletion to disable the automatic install of
the auto completion
*) Add a prompt to ask for permission to install automatic
auto completion when not installed yet

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"bufio"
 	"bytes"
 	"fmt"
 	"os"
@@ -33,6 +34,7 @@ import (
 	"github.com/minio/mc/pkg/probe"
 	"github.com/minio/minio/pkg/words"
 	"github.com/pkg/profile"
+	"golang.org/x/crypto/ssh/terminal"
 
 	completeinstall "github.com/posener/complete/cmd/install"
 )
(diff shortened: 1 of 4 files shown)


cmd/main.go
```diff
@@ -222,6 +220,43 @@ func initMC() {
...
+	if runtime.GOOS == "windows" {
+		return
+	}
+
+	if completeinstall.IsInstalled("mc") {
+		return
+	}
+
+	for {
+		fmt.Printf("Install mc auto-completion in your shell ? (y/n): ")
```
harshavardhana: We should disable the prompt if `mc` is running as a script or --json, --quiet is specified.

harshavardhana: Perhaps no-autocompletion should be true in case of running as scripts...

vadmeste: > We should disable the prompt if `mc` is running as a script or --json, --quiet is specified.

Oups yes, fixed !

> Perhaps no-autocompletion should be true in case of running as scripts...

 Not sure how to detect if this is a script or not

harshavardhana: A script mode is basically non terminal, --json or --quiet mode.

vadmeste: > A script mode is basically non terminal, --json or --quiet mode.

Yes fixed then

PR #2814, opened 2019-07-05
Add env variable for no autocomplete installation

maxnordlund:
This makes it a bit easier for users to avoid being prompted all the time if they don't include the --no-autocomplete flag.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -57,6 +57,7 @@ var globalFlags = []cli.Flag{
 	cli.BoolFlag{
 		Name:  "no-autocompletion",
 		Usage: "disable automatic install of mc auto-completion",
+		EnvVar: "MINIO_NO_AUTOCOMPLETE",
 	},
 }
 
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -57,7 +57,7 @@ var globalFlags = []cli.Flag{
 	cli.BoolFlag{
 		Name:   "no-autocompletion",
 		Usage:  "disable automatic install of mc auto-completion",
-		EnvVar: "MINIO_NO_AUTOCOMPLETE",
+		EnvVar: "MC_NO_AUTOCOMPLETE",
 	},
 }
 
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -55,8 +55,8 @@ var globalFlags = []cli.Flag{
 		Usage: "disable SSL certificate verification",
 	},
 	cli.BoolFlag{
-		Name:  "no-autocompletion",
-		Usage: "disable automatic install of mc auto-completion",
+		Name:   "no-autocompletion",
+		Usage:  "disable automatic install of mc auto-completion",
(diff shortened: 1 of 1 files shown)


harshavardhana:
Can you fix the build issues ?

maxnordlund:
If I understood the error/formatting diff, then it should be fixed now.

vadmeste:
Thanks for the PR @maxnordlund !

Can't the user just use `alias mc="mc --no-autocompletion` and put it in ~/.bashrc or other ? An environment variable does not seem to be needed here

maxnordlund:
To be forced to use an alias is really inconvenient. Also, I use fish, so aliases isn't really a thing, but universal variables are. They act like env variables, except shared between all shells, and persisted between sessions. Perfect place to put stuff like this.

Not to mention the various dotenv commands that reads a `.env` file before running some command. In my experience, that has been used to share env variables between developers, and checked in to a repo (or a sample file).

maxnordlund:
Ping, would be sweet to get this in :)

harshavardhana:
Env should have mc prefix not MinIO so I would suggest change this to MC_NO_AUTOCOMPLETE

maxnordlund:
> Env should have mc prefix not MinIO so I would suggest change this to MC_NO_AUTOCOMPLETE

Done

sinhaashish:
@maxnordlund  please add the details of `MC_NO_AUTOCOMPLETE` in the docs as well.

maxnordlund:
Where, there's no docs for the `--no-autocompletion` as far as I can see?

sinhaashish:
> Where, there's no docs for the `--no-autocompletion` as far as I can see?
> 
> More to the point, #2812 didn't add any docs, so could you please that that and I can modify it accordingly.

You can add the details for `--no-autocompletion`  below the `--insecure` section in 
 1. https://github.com/minio/mc/blob/master/docs/minio-admin-complete-guide.md#option----insecure
 2. https://github.com/minio/mc/blob/master/docs/minio-client-complete-guide.md#option----insecure
 
You need to update this section too. 
  https://github.com/minio/mc/blob/master/README.md#shell-autocompletion

maxnordlund:
I'm just about to go on vacation, so if you please could add something I could then work from?

maxnordlund:
Or even better, just commit directly to this branch and then squash and merge :)

sinhaashish:
> Or even better, just commit directly to this branch and then squash and merge :)

Didn't had the permission to push it to your branch , thus raised another [PR ](https://github.com/maxnordlund/mc/pull/1)  against your branch. Please approve and merge it.

harshavardhana:
PR #2819 supersedes this change, @maxnordlund you may close this PR.

harshavardhana:
Closing this PR as we have moved to no completion installation by default mode, with latest release.  Thanks for the PR nevertheless @maxnordlund

maxnordlund:
Great! I've been on vacation this time, so thanks for fixes this.

PR #2817, opened 2019-07-10
Remove port 9000 for play.min.io

harshavardhana:
Most corporate environments block port 9000,
since we already allow access from port 80 and
port 443 just default to https port instead.


harshavardhana:
> Readme.md, Readme_zh_CN.md and functional-tests.sh also need to change.
> the default config for play in mc config.json would also need migration.

We don't need any migration @poornas - existing users will work as is, on all fresh installations it will use new URL. Fixed remaining parts.

PR #2819, opened 2019-07-11
Fix auto-completion code

### LINKED ISSUE #2816: unsolicited console interaction "Install mc auto-completion in your shell ?"
(I rename execitable to mcli because I use midnight-commander a lot)
REPRO:
curl -sSLo /usr/bin/mcli https://dl.minio.io/client/mc/release/linux-amd64/mc
chmod +x /usr/bin/mcli
EXPECTED:
can use mcli in scripts
CURENT:
it always goes interactive
mcli
mcli: Configuration written to `/root/.mc/config.json`. Please update your access credentials.
mcli: Successfully created `/root/.mc/share`.
mcli: Initialized share uploads `/root/.mc/share/uploads.json` file.
mcli: Initialized share downloads `/root/.mc/share/downloads.json` file.
Install mc auto-completion in your shell ? (y/n):

harshavardhana:
Reverts the code in #2812

Fixes #2816

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -1,5 +1,5 @@
 /*
- * MinIO Client (C) 2014, 2015, 2016, 2017 MinIO, Inc.
+ * MinIO Client (C) 2014-2019 MinIO, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,7 +17,6 @@
 package cmd
 
 import (
-	"bufio"
 	"bytes"
 	"fmt"
 	"os"
@@ -34,7 +33,6 @@ import (
 	"github.com/minio/mc/pkg/probe"
 	"github.com/minio/minio/pkg/words"
 	"github.com/pkg/profile"
-	"golang.org/x/crypto/ssh/terminal"
 
 	completeinstall "github.com/posener/complete/cmd/install"
 )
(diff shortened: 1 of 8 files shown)


sinhaashish:
We should add the details for `autocompletion` below the --insecure section in
https://github.com/minio/mc/blob/master/docs/minio-admin-complete-guide.md#option----insecure
https://github.com/minio/mc/blob/master/docs/minio-client-complete-guide.md#option----insecure

harshavardhana:
It is not a global option to be used always @sinhaashish, that's why it's left out of those docs. This command gets its own mention in our README.md and also as part of a new TIP in our command document.

cmd/main.go
```diff
@@ -221,44 +222,22 @@ func initMC() {
...
-	}
-
+func installAutoCompletion() {
 	if runtime.GOOS == "windows" {
+		console.Infoln("autocompletion feature is not available for this operating system")
 		return
 	}
 
-	if completeinstall.IsInstalled("mc") {
+	if completeinstall.IsInstalled(filepath.Base(os.Args[0])) || completeinstall.IsInstalled("mc") {
```
vadmeste: The second completeinstall.IsInstalled("mc") doesn't seem to be useful here.. completion won't work anyway if the binary name is different to 'mc'

cmd/main.go
```diff
@@ -221,44 +222,22 @@ func initMC() {
...
+		console.Infoln("autocompletion is already enabled in your '$SHELLRC'")
```
vadmeste: Why not simply 'shell' ? I mean did you intend to have a variable substitution here '$SHELLRC'  ?

cmd/main.go
harshavardhana: It will work it's already fixed.

cmd/main.go
harshavardhana: That was AB's recommendation @vadmeste

deekoder:
swapping @nitisht to @poornas  to complete approval and make release today.

PR #2820, opened 2019-07-11
Fix inner link

mausch:


deekoder:
@mausch can you give some context for this change?

kannappanr:
@deekoder This fix seems to have fixed a relative link issue we had in mc's complete guide while navigating through the repo in github website. Needs to be verified in docs site, once this change is deployed on docs site.

PR #2829, opened 2019-07-18
Use long notation for flags in tree example

Praveenrajmani:


Praveenrajmani:
The changes are covered in - #2828

PR #2830, opened 2019-07-19
Fix admin info dot coloring

harshavardhana:


kannappanr:
Will need to show this to @abperiasamy before we can merge this PR

harshavardhana:
I already did @kannappanr it is approved.

PR #2849, opened 2019-08-05
Remove empty parent directories while mirroring fs

### LINKED ISSUE #2826: mirror command does not remove empty directories from local machine
I'm synchronizing a local machine with minio server. Everything what has been deleted from the server should be also removed from the local machine.

I'm executing the following command:
```minio mirror --config-dir /data/config --overwrite --remove s3/prod/ /data/files/```

## Expected behavior
Content of the local machine `/data/files/` directory is exactly the same as `s3` server `prod` bucket:
* all new files should be created on the local machine
* all removed files should be removed from the local machine
* all removed directories should be removed from the local machine

## Actual behavior
Everything except deleting removed directories works properly:
* all new files should be created on the local machine 🆗 
* all removed files should be removed from the local machine 🆗 
* all removed directories should be removed from the local machine 🐛

It causes that after some time I have dozen of empty directories on the local machine.

1. create a directory with deep nested structure (in all leaves directories must be located at least one file)
2. upload to minio server
3. execute mirror remote → local machine (the file structure is not the same)
4. remove some 1-st level directories from minio
5. execute mirror remote → local machine (removed files are deleted, but empty directories stay)

In our case we have the following structure (all directories have the same structure):
```
dir1
    \- artifacts (← directory, stores 3 files)
    |- pipeline-scripts
    |                  \- directory (←directory, stores 3 file)
    |                  |- jobs (← directory, stores 6 files)
    |                  |- utils (← directory, stores 9 files)
    |                  |- .editorconfig
    |                  |- .gitignore
    |                  | ... (additional 4 files)
    |- product (← git repository without .git folder, contain nested directories with 5-6 levels)
    |- resources (← directory, stores 4 files)
    |- pipeline.prooperties
    |- product-fat.zip
```

After removal whole `dir1` from the server, and mirror to local I see that all files have been removed, but directories still exist.

## mc version
```minio version RELEASE.2019-05-23T01-33-27Z```

## System information
```
> uname -a
Linux host-p-15760 4.9.0-8-amd64 #1 SMP Debian 4.9.110-3+deb9u4 (2018-08-21) x86_64 GNU/Linux
```

## Additional comments
I found #2612 about missing empty directories after mirroring. I understand that this behavior is a design decision to not create any empty directories (they store no files, so there is no sense to keep them). It would be really great to make it consistent and remove directories which don't store any files :)

### PR DESCRIPTION
**Pull request by Praveenrajmani**
While mirroring a fs client with a `--remove`. We should also remove the stale/empty parent directores as well. Can reproduce this through steps provided in the issue.

Fixes #2826

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -44,7 +44,8 @@ type fsClient struct {
 }
 
 const (
-	partSuffix = ".part.minio"
+	partSuffix     = ".part.minio"
+	slashSeperator = "/"
 )
 
 var ( // GOOS specific ignore list.
@@ -426,6 +427,50 @@ func (f *fsClient) Get(sse encrypt.ServerSide) (io.ReadCloser, *probe.Error) {
 	return f.get()
 }
 
+// Check if the given error corresponds to ENOTEMPTY for unix
+// and ERROR_DIR_NOT_EMPTY for windows (directory not empty).
+func isSysErrNotEmpty(err error) bool {
+	if err == syscall.ENOTEMPTY {
+		return true
+	}
+	if pathErr, ok := err.(*os.PathError); ok {
+		if runtime.GOOS == "windows" {
+			if errno, _ok := pathErr.Err.(syscall.Errno); _ok && errno == 0x91 {
+				// ERROR_DIR_NOT_EMPTY
+				return true
+			}
+		}
+		if pathErr.Err == syscall.ENOTEMPTY {
+			return true
+		}
+	}
+	return false
+}
+
+// deleteFile deletes a file path if its empty. If it's successfully deleted,
+// it will recursively delete empty parent directories
+// until it finds one with files in it. Returns nil for a non-empty directory.
+func deleteFile(deletePath string) error {
+	// Attempt to remove path.
+	if err := os.Remove((deletePath)); err != nil {
+		if isSysErrNotEmpty(err) {
+			return nil
+		}
+		return err
+	}
+
+	// Trailing slash is removed when found to ensure
+	// slashpath.Dir() to work as intended.
+	parentPath := strings.TrimSuffix(deletePath, slashSeperator)
+	parentPath = path.Dir(parentPath)
+
+	if parentPath != "." {
+		return deleteFile(parentPath)
+	}
+
+	return nil
+}
+
 // Remove - remove entry read from clientContent channel.
 func (f *fsClient) Remove(isIncomplete, isRemoveBucket bool, contentCh <-chan *clientContent) <-chan *probe.Error {
 	errorCh := make(chan *probe.Error)
@@ -440,7 +485,7 @@ func (f *fsClient) Remove(isIncomplete, isRemoveBucket bool, contentCh <-chan *c
 			if isIncomplete {
 				name += partSuffix
 			}
-			if err := os.Remove(name); err != nil {
+			if err := deleteFile(name); err != nil {
 				if os.IsPermission(err) {
 					// Ignore permission error.
 					errorCh <- probe.NewError(PathInsufficientPermission{Path: content.URL.Path})

### CODE REVIEW

**Comment by ebozduman** [comment]
@Praveenrajmani ,
Could you look into the travis failure?

**Comment by Praveenrajmani** [comment]
> I've tested the fix and it doesn't look like the problem is completely fixed.
> Here is the scenario I can reproduce the problem with:
> 
>     * Create a multi-layer directory structure on your source MinIO server,
> 
> 
> ```
> /tmp/ersan
> └─ d1
>    └─ d2
>       ├─ d244
>       ├─ d3
>       │  └─ d4
>       │     └─ d5
>       │        └─ issue
>       └─ issue2
> ```
> 
>     * Mirror it to a local directory, like `/tmp/<user-name>`,
> 
>     * Check it out to make sure it is replica of the source,
> 
>     * Remove `d244` directory and `issue2` object using mc commands,
> 
>     * Mirror the new state to `/tmp/<user-name>`
> 
>     * Check it out at `/tmp/<user-name>`,
>       You'll see `issue2` object is cleaned up correctly, but `d244` directory is still part of the directory structure at  `/tmp/<user-name>`

@ebozduman I cannot reproduce this. Am i missing something?

Screenshot from 2019-08-07 10-50-47

And `mc rm` would not leave empty directories/prefixes behind. It will also delete the parent dir/prefix path if it is empty while removing. I think it is a bug in `mc rm` if it is not removing empty directories.

**Comment by ebozduman** [comment]
Hi Praveen,

My bad. I forgot the explain the crucial step.
There is a corner case/exception, which is not a usual/normal usage, probably not going to happen, but it is still an open door to reproduce the issue.

There is actually a way to create an empty directory; you can use `mc mb <dir-name>` to create a directory with no object in it. So, if you remove this empty directory after mirroring, and then mirror the source  again, the empty directory already existed at the destination local filesystem stays the same and it is not touched.
Here are the complete steps:
- $ ././mc cp /etc/issue play/ersan/d1/d2/d3/d4/d5/file5
- $ ./mc cp /etc/issue play/ersan/d1/d2/file2
- $ ./mc mb play/ersan/d1/d2/d244/
- $ ./mc mirror play/ersan /tmp/ersan
- $ ./mc rm -r --force play/ersan/d1/d2/d244
- $ mc tree -f play/ersan 
```
play/ersan
└─ d1
   └─ d2
      ├─ file2
      └─ d3
         └─ d4
            └─ d5
               └─ file5
```
- $ ./mc mirror play/ersan /tmp/ersan
- ./mc tree -f /tmp/ersan
```
/tmp/ersan
└─ d1
   └─ d2
      ├─ d244
      ├─ d3
      │  └─ d4
      │     └─ d5
      │        └─ file5
      └─ file2
```
One can always challenge this scenario that nobody will go through it, so please feel free to ask @abperiasamy 's, @harshavardhana's and/or other developer's opinion, if you'd like.

Praveenrajmani:
> Hi Praveen,
> 
> My bad. I forgot the explain the crucial step.
> There is a corner case/exception, which is not a usual/normal usage, probably not going to happen, but it is still an open door to reproduce the issue.
> 
> There is actually a way to create an empty directory; you can use `mc mb <dir-name>` to create a directory with no object in it. So, if you remove this empty directory after mirroring, and then mirror the source again, the empty directory already existed at the destination local filesystem stays the same and it is not touched.
> Here are the complete steps:
> 
>     * $ ././mc cp /etc/issue play/ersan/d1/d2/d3/d4/d5/file5
> 
>     * $ ./mc cp /etc/issue play/ersan/d1/d2/file2
> 
>     * $ ./mc mb play/ersan/d1/d2/d244/
> 
>     * $ ./mc mirror play/ersan /tmp/ersan
> 
>     * $ ./mc rm -r --force play/ersan/d1/d2/d244
> 
>     * $ mc tree -f play/ersan
> 
> 
> ```
> play/ersan
> └─ d1
>    └─ d2
>       ├─ file2
>       └─ d3
>          └─ d4
>             └─ d5
>                └─ file5
> ```
> 
>     * $ ./mc mirror play/ersan /tmp/ersan
> 
>     * ./mc tree -f /tmp/ersan
> 
> 
> ```
> /tmp/ersan
> └─ d1
>    └─ d2
>       ├─ d244
>       ├─ d3
>       │  └─ d4
>       │     └─ d5
>       │        └─ file5
>       └─ file2
> ```
> 
> One can always challenge this scenario that nobody will go through it, so please feel free to ask @abperiasamy 's, @harshavardhana's and/or other developer's opinion, if you'd like.

I just tried fixing this.

This problem happens because - We do recursive listing for the delta difference between source and target. We use the `DirNode` recursive list for this, which will not list directories or ghost dirs (0-sized).  This is the reason we miss the `d244` directory (in example) while mirroring

I tried to fix this by listing the dirs in `DirFirst`/`DirLast` mode. The current mirror implementation will not work with this. As the List emits empty prefixes/dirs. We cannot try to copy those empty prefixes to s3/fs. As s3 wont allow empty prefix PUTs. And also order is inconsistent (Order has to be same in both src and tgt while finding the difference) in these modes. 

Any thoughts? @ebozduman @harshavardhana

Praveenrajmani:
> > Hi Praveen,
> > My bad. I forgot the explain the crucial step.
> > There is a corner case/exception, which is not a usual/normal usage, probably not going to happen, but it is still an open door to reproduce the issue.
> > There is actually a way to create an empty directory; you can use `mc mb <dir-name>` to create a directory with no object in it. So, if you remove this empty directory after mirroring, and then mirror the source again, the empty directory already existed at the destination local filesystem stays the same and it is not touched.
> > Here are the complete steps:
> > ```
> > * $ ././mc cp /etc/issue play/ersan/d1/d2/d3/d4/d5/file5
> > 
> > * $ ./mc cp /etc/issue play/ersan/d1/d2/file2
> > 
> > * $ ./mc mb play/ersan/d1/d2/d244/
> > 
> > * $ ./mc mirror play/ersan /tmp/ersan
> > 
> > * $ ./mc rm -r --force play/ersan/d1/d2/d244
> > 
> > * $ mc tree -f play/ersan
> > ```
> > 
> > 
> > ```
> > play/ersan
> > └─ d1
> >    └─ d2
> >       ├─ file2
> >       └─ d3
> >          └─ d4
> >             └─ d5
> >                └─ file5
> > ```
> > 
> > 
> > ```
> > * $ ./mc mirror play/ersan /tmp/ersan
> > 
> > * ./mc tree -f /tmp/ersan
> > ```
> > 
> > 
> > ```
> > /tmp/ersan
> > └─ d1
> >    └─ d2
> >       ├─ d244
> >       ├─ d3
> >       │  └─ d4
> >       │     └─ d5
> >       │        └─ file5
> >       └─ file2
> > ```
> > 
> > 
> > One can always challenge this scenario that nobody will go through it, so please feel free to ask @abperiasamy 's, @harshavardhana's and/or other developer's opinion, if you'd like.
> 
> I just tried fixing this.
> 
> This problem happens because - We do recursive listing for the delta difference between source and target. We use the `DirNone` opt for this, which will not list directories or ghost dirs (0-sized). This is the reason we miss the `d244` directory (in example) while mirroring
> 
> I tried to fix this by listing the dirs in `DirFirst`/`DirLast` mode which will also list the dirs/prefixes. The current mirror implementation will not work with this. As the List emits empty prefixes/dirs. We cannot try to copy those empty prefixes to s3/fs. As s3 wont allow empty prefix PUTs. And also order is inconsistent (Order has to be same in both src and tgt while finding the difference) in these modes.
> 
> Any thoughts? @ebozduman @harshavardhana

@ebozduman Discussed this with Harsha. This requires some change in implementation. And dirs are unexpected entities in mirror. So, we are not mirroring empty dirs for now.

PR #2852, opened 2019-08-09
Add policy set command

donatello:
Companion to https://github.com/minio/minio/pull/8036


kannappanr:
We need to add `mc policy unset` command to remove policies set for a user or a group

kannappanr:
It will also be good to add a `info` command to `mc admin user` so that the current evaluated policy can be displayed

donatello:
> It will also be good to add a info command to mc admin user so that the current evaluated policy can be displayed

Done!

> We need to add mc policy unset command to remove policies set for a user or a group

We can do this subsequently after discussion. Holding on this one.

kannappanr:
@donatello You probably did not checkin code related to `adminUserInfoCmd`.

donatello:
> @donatello You probably did not checkin code related to adminUserInfoCmd.

Oops! Fixed now.

PR #2860, opened 2019-08-14
Fix help message bug in group disable command

donatello:
- Update mc docs for new commands


harshavardhana:
Took this in because we have already made the release of `mc` and it needed doc updates.

PR #2876, opened 2019-08-29
prevent panic if number of endpoints is 0

wlan0:


harshavardhana:
Closing this PR as not needed - since this is a server-side bug // cc @wlan0

PR #2882, opened 2019-09-05
add the `mc admin kms` subcommands

aead:
This commit adds the first set of KMS admin
commands.

For new there is only the
`mc admin kms key status <endpoint> [key-ID]` command
that can be used to fetch status information for a
KMS master key. Listing and rotation commands are
planed.

### How to test
1. `export MINIO_ACCESS_KEY=minio`
2. `export MINIO_SECRET_KEY=minio123`
3. `export MINIO_SSE_MASTER_KEY=my-minio-key:6368616e676520746869732070617373776f726420746f206120736563726574`
4. `minio server /tmp/1`

In a separate terminal you can run:
1. `mc config host add local http://127.1:9000 minio minio123`
2. `mc admin kms key status local`


cmd/admin-kms-key.go
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package cmd
+
+import "github.com/minio/cli"
+
+var adminKMSKeyCmd = cli.Command{
+	Name:   "key",
+	Usage:  "manage KMS keys",
```
poornas: to make it clear that this is only for listing and key-rotation and not for other management operations like key creation/deletion ,the Usage should just list supported key management operations,imo

poornas:
docs need an update

cmd/admin-kms-key.go
aead: We can clarify what KMS operations are supported. However, the list `status, listing, rotation` is not comprehensive. It depends on which future requirements we get in the future. We can cover a lot with these 3 operations but depending on user requirements we may extend that - e.g. disabling a specific key.

But we can extend that list if we add new functionality - we just must not forget that then.

aead:
> docs need an update

In general yes. However, the docs seems to be out-of-sync - e.g.:
```
mc admin

service  restart and stop all MinIO servers
  update   update all MinIO servers
  info     display MinIO server information
  user     manage users
  group    manage groups
  policy   manage policies defined in the MinIO server
  config   manage configuration file
  heal     heal disks, buckets and objects on MinIO server
  profile  generate profile data for debugging purposes
  top      provide top like statistics for MinIO
  trace    show http trace for minio server
  console  show console logs for MinIO server
  kms      perform KMS management operations

vs.

  service      restart or stop all MinIO servers 
  update       updates all MinIO servers
  info         display MinIO server information
  user         manage users
  policy       manage canned policies
  config       manage configuration file
  heal         heal disks, buckets and objects on MinIO server
  top          provide top like statistics for MinIO
```

 Are they still maintained?

poornas:
> Are they still maintained?
 
some of the PR's didn't update docs - #2884 fixed this.

poornas:
could you fix conflicts in go.mod?

poornas:
@aead, there is a bug here - even non-existent keys show a keystatus because we are relying on error to be nil
```
➜  mc git:(c4a1f1f1) mc admin kms key status play my-random-key
Key: my-random-key
         • Encryption ✔
         • Update     ✔
         • Decryption ✔
```

aead:
@poornas 

 ```
mc admin kms key status play my-random-key
Key: my-random-key
         • Encryption ✔
         • Update     ✔
         • Decryption ✔
```

This is actually expected behavior - even though I acknowledge that it looks weird. The master key mimics an actual KMS here. So it actually uses the ID `my-random-key` to derive a separate set of data keys. However, we may change that in the future. In general listing and rotating operations are not useful when a single fixed key is used...

**Comment by poornas** [comment]
@aead, build failures need to be fixed.

**Comment by poornas** [comment]
json mode output doesn't show the en(de)cryption/update status - needs to be fixed
```
 ➜  mc git:(c4a1f1f1) ✗ mc admin kms key status play --json
{
 "key-id": "my-minio-key"
}

```

**Comment by aead** [comment]
@poornas Fixed:
 ```
mc admin kms key status play --json
{"key-id":"my-minio-key","encryption-error":"","update-error":"","decryption-error":""}
```

aead:
@poornas @harshavardhana PTAL

docs/minio-admin-complete-guide.md
```diff
@@ -16,6 +16,7 @@ top         provide top like statistics for MinIO
 trace       show http trace for MinIO server
 console     show console logs for MinIO server
 prometheus  manages prometheus config
```
harshavardhana: Looks like sub-command formatting is off..

aead: Thanks @harshavardhana - fixed

PR #2887, opened 2019-09-06
Fix error message wording in `mc admin` commands

poornas:


cmd/client-admin.go
```diff
@@ -121,7 +121,7 @@ func newAdminClient(aliasedURL string) (*madmin.AdminClient, *probe.Error) {
 	}
 
 	if hostCfg == nil {
-		return nil, probe.NewError(fmt.Errorf("The specified alias: %s not found", urlStrFull))
+		return nil, probe.NewError(fmt.Errorf("No valid configuration found for '%s' host", urlStrFull))
```
harshavardhana:
```suggestion
		return nil, probe.NewError(fmt.Errorf("No valid configuration found for '%s' host alias", urlStrFull))
```

PR #2892, opened 2019-09-14
Add support for attr flag in mirror

harshavardhana:


harshavardhana:
ping @vadmeste

cmd/diff-main.go
```diff
@@ -174,7 +176,7 @@ func doDiffMain(firstURL, secondURL string) error {
 	}
 
 	// Diff first and second urls.
-	for diffMsg := range objectDifference(firstClient, secondClient, firstURL, secondURL) {
+	for diffMsg := range objectDifference(firstClient, secondClient, firstURL, secondURL, false) {
```
vadmeste: Is this an intended false here ? (to avoid breaking the existing deployments ?)

harshavardhana: Yes it's meaningless for `diff` command

PR #2901, opened 2019-09-19
Fix mc functional tests to work with new changed behavior

harshavardhana:
- mc rm doesn't fail for SSE-C encrypted objects because
  we use multi-object delete - this behavior is S3 compatible
- changes needed to ensure that keys are set appropriately
  when doing server side copy.

### DIFF
--- a/cmd/common-methods.go
+++ b/cmd/common-methods.go
@@ -1,5 +1,5 @@
 /*
- * MinIO Client (C) 2015 MinIO, Inc.
+ * MinIO Client (C) 2015-2019 MinIO, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,7 @@ import (
 	"encoding/base64"
 	"errors"
 	"io"
+	"net/http"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -245,9 +246,9 @@ func copySourceToTargetURL(alias string, urlStr string, source string, size int6
 	return nil
 }
 
-// createUserMetadata - returns a map of user defined function
+// getAllMetadata - returns a map of user defined function
 // by combining the usermetadata of object and values passed by attr keyword
-func createUserMetadata(sourceAlias, sourceURLStr string, srcSSE encrypt.ServerSide, urls URLs) (map[string]string, *probe.Error) {
+func getAllMetadata(sourceAlias, sourceURLStr string, srcSSE encrypt.ServerSide, urls URLs) (map[string]string, *probe.Error) {
 	metadata := make(map[string]string)
 	sourceClnt, err := newClientFromAlias(sourceAlias, sourceURLStr)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


functional-tests.sh
```diff
@@ -696,17 +696,13 @@ function test_copy_object_with_sse()
     assert_success "$start_time" "${FUNCNAME[0]}" show_on_failure $? "unable to download object using 'mc cat'"
     assert_success "$start_time" "${FUNCNAME[0]}" check_md5sum "$FILE_1_MB_MD5SUM" "${object_name}.downloaded"
     assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd rm "${object_name}.downloaded"
-    # mc rm on encrypted object without encryption key should fail
-    assert_failure "$start_time" "${FUNCNAME[0]}" mc_cmd rm "${SERVER_ALIAS}/${BUCKET_NAME}/${prefix}/${object_name}"
-    # mc rm on encrypted object with encryption key should pass
-    assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd rm --encrypt-key "${cli_flag}" "${SERVER_ALIAS}/${BUCKET_NAME}/${prefix}/${object_name}"
-    # mc rm on unencrypted destination object should pass
-    assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd rm  "${SERVER_ALIAS}/${BUCKET_NAME}/${object_name}"
```
poornas: cleanup of destination object is still needed.

cmd/common-methods.go
```diff
@@ -298,6 +298,11 @@ func uploadSourceToTargetURL(ctx context.Context, urls URLs, progress io.Reader,
 		}
 
 		sourcePath := filepath.ToSlash(sourceURL.Path)
+		if srcSSE != nil {
+			delete(metadata, "X-Amz-Server-Side-Encryption-Customer-Algorithm")
+			delete(metadata, "X-Amz-Server-Side-Encryption-Customer-Key-Md5")
```
poornas: should also delete `x-amz-server-side-encryption` from metadata for SSE-S3 encrypted objects

kannappanr: `srcSSE != nil` can be removed and all the source metadata with prefix `x-amz-server-side-encryption` should be removed. Because there can also be `x-amz-server-side-encryption-aws-kms-key-id`

functional-tests.sh
harshavardhana: what ??

poornas: ```     assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd rm  "${SERVER_ALIAS}/${BUCKET_NAME}/${object_name}" ```

harshavardhana: the destination object is same as source object  here @poornas

kannappanr: No, they are different Harsha. Please look at https://github.com/minio/mc/pull/2901/files#diff-71b55f1605a99785eb285feff46dc57eL693

cmd/common-methods.go
harshavardhana: done

functional-tests.sh
harshavardhana: Fixed @kannappanr

PR #2906, opened 2019-09-24
Customizes service help for Windows users

### LINKED ISSUE #2776: Service help description wrong for windows users
## Expected behavior
Borrowing from https://github.com/minio/minio/issues/7598 we should make this appropriate behavior just like a server.

## Actual behavior
No platform-specific help

Just type `mc cp -h` on Windows

## mc version
master

## System information
All supported systems

ebozduman:
Fixes #2776

Existing `mc` help examples were not making any distinction between Linux and Windows OS. All examples were in Linux format. 
This change adjusts only the prompt information (`$ ` vs `C:\> `), since the path separator does not matter, that is Windows and MinIO accepts both Unix/Linux style separator `/` and Windows style separator, `\`.
And there were no Environment Variable definitions to adjust in the `mc` help content.

### DIFF
--- a/cmd/admin-config-get.go
+++ b/cmd/admin-config-get.go
@@ -39,7 +39,7 @@ FLAGS:
   {{end}}
 EXAMPLES:
   1. Get server configuration of a MinIO server/cluster.
-     $ {{.HelpName}} play/
+     {{.Prompt}} {{.HelpName}} play/
 `,
 }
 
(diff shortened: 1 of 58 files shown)


ebozduman:
There were so many conflicts, it looked much more faster and easier to create a new PR rather than dealing with the conflicts.
Closing this PR ...

PR #2908, opened 2019-09-24
Customizes service help for Windows/Linux users

### LINKED ISSUE #2776: Service help description wrong for windows users

ebozduman:
Fixes #2776

Existing `mc` help examples were not making any distinction between Linux and Windows OS. All examples were in Linux format. 
This change adjusts only the prompt information (`$ ` vs `C:\> `), since the path separator does not matter, that is Windows and MinIO accepts both Unix/Linux style separator `/` and Windows style separator, `\`.
And there were no Environment Variable definitions to adjust in the `mc` help content.


cmd/admin-group-add.go
```diff
@@ -45,9 +45,7 @@ FLAGS:
   {{end}}
 EXAMPLES:
   1. Add users 'fivecent' and 'tencent' to the group 'allcents':
-     $ set +o history
-     $ {{.HelpName}} myminio allcents fivecent tencent
-     $ set -o history
+     {{.Prompt}} {{.HelpName}} myminio allcents fivecent tencent
```
eco-minio: Is there any windows equivalent to `set +o history` ?  Is it possible to print help based on the OS?  It would be good to keep set -o for linux users at least although it is not a hard requirement.  It looks like you have something that might accomplish this in cmd/admin-user-add.go?

harshavardhana: Its already doing that in another command in cmd/config-host-add.go - it should just use that here as well.

ebozduman: I didn't do it here, because there is no sensitive information to hide here. I think disable/enable history is added here inadvertantly, hence I simply removed them leaving the only adaptive change as `{{.Prompt}}`.

PR #2910, opened 2019-09-26
Refactor mc cp and make session optional 

### LINKED ISSUE #2893: Deprecate mc session command
## Expected behavior
Deprecate `mc session` command and modify `cp` command to handle this scenario.

The new command will look like this
`mc cp -r -s /tmp/bin play/test`

In this case, you will create a session in `/tmp/mc/uniquename`

The file will have `read only` access for the account user who ran `mc`. Other users should not have any access.

On error, the session file is created under `/tmp/mc/uniquename`. Then a command is printed

mc: <ERROR> copy operation safely terminated. To resume operation `mc cp -r --resume /tmp/mc/uniquename`

When you are starting up `mc` if there are existing `session files` in the existing folder, we need to move it to `/tmp/mc`

Resuming the same `mc cp` operation would be via `mc cp --resume /tmp/mc/uniquename`

If attr `-s` is not passed, then session file is not created.

## Actual behavior
mc session resume - resumes previously saved sessions.

This is a breaking change. `cp` is quite slow for large datasets due to session building stage, most users gravitate towards using `mirror` instead without session.

## mc version
All versions

## System information
All OS

balamurugana:
Fixes #2893

### DIFF
--- a/cmd/session-v8.go
+++ b/cmd/session-v8.go
@@ -162,7 +162,7 @@ func loadSessionV8(sid string) (*sessionV8, *probe.Error) {
 }
 
 // newSessionV8 provides a new session.
-func newSessionV8() *sessionV8 {
+func newSessionV8(sessionID string) *sessionV8 {
 	s := &sessionV8{}
 	s.Header = &sessionV8Header{}
 	s.Header.Version = globalSessionConfigVersion
@@ -177,7 +177,7 @@ func newSessionV8() *sessionV8 {
 	s.Header.UserMetaData = make(map[string]string)
 	s.Header.When = UTCNow()
 	s.mutex = new(sync.Mutex)
-	s.SessionID = newRandomID(8)
+	s.SessionID = sessionID
 
 	sessionDataFile, err := getSessionDataFile(s.SessionID)
 	fatalIf(err.Trace(s.SessionID), "Unable to create session data file \""+sessionDataFile+"\".")
@@ -373,6 +373,18 @@ func (s *sessionV8) Delete() *probe.Error {
 func (s sessionV8) CloseAndDie() {
 	s.Close()
 	console.Fatalln("Session safely terminated. To resume session `mc session resume " + s.SessionID + "`")
+}
+
+func (s sessionV8) copyCloseAndDie(sessionFlag bool) {
+	if sessionFlag {
+		s.Close()
+		console.Fatalln("Command terminated safely. Run this command to resume copy again.")
+	} else {
+		s.mutex.Lock()
+		defer s.mutex.Unlock()
+
+		s.DataFP.Close() // ignore error.
+	}
 }
 
 // Create a factory function to simplify checking if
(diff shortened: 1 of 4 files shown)


cmd/session-v8.go
```diff
@@ -375,6 +375,18 @@ func (s sessionV8) CloseAndDie() {
 	console.Fatalln("Session safely terminated. To resume session `mc session resume " + s.SessionID + "`")
 }
 
+func (s sessionV8) copyCloseAndDie(sessionFlag bool) {
+	if sessionFlag {
+		s.Close()
+		console.Fatalln("Command terminated safely. Run this command to resume copy again.")
```
krisis: question: This would result in exit code being non-zero. Is this intentional?

cmd/session-v8.go
```diff
@@ -375,6 +375,18 @@ func (s sessionV8) CloseAndDie() {
...
+	} else {
+		s.mutex.Lock()
```
krisis: question: why do we need to do anything at all when session flag wasn't passed along with `mc cp ...`?

cmd/session-v8.go
balamurugana: `CloseAndDie()` does the same thing. I am keeping the same behaviour.

cmd/session-v8.go
balamurugana: Not sure. I am just keeping session behaviour same as `CloseAndDie()`. When `mc session` command is removed, type `sessionV8` may clean up.

harshavardhana:
Since `mc cp` already has session support can this be retitled as `Refactor `mc cp` and make session optional` ?

cmd/session_test.go
```diff
@@ -35,9 +35,9 @@ func (s *TestSuite) TestSession(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(isSessionDirExists(), Equals, true)
 
-	session := newSessionV8()
+	session := newSessionV8(getHash("cp", []string{"mybucket", "myminio/mybucket"}))
```
krisis: minor nit
```suggestion
	session := newSessionV8(getHash("cp", []string{"object", "myminio/mybucket"}))
```

cmd/session_test.go
```diff
@@ -35,9 +35,9 @@ func (s *TestSuite) TestSession(c *C) {
...
 	c.Assert(session.Header.CommandArgs, IsNil)
-	c.Assert(len(session.SessionID), Equals, 8)
+	c.Assert(len(session.SessionID) >= 8, Equals, true)
```
krisis: How about the following assert? It asserts a more stronger property than `>= 8`. It uses the fact that the sha256 checksum part of the sessionID is 32 bytes. 
```suggestion
	c.Assert(len(session.SessionID) >= sha256.Size, Equals, true)
```

cmd/session_test.go
balamurugana: This is not correct. I am simulating `mc cp -r -c mybucket myminio/mybucket` here.

cmd/session_test.go
balamurugana: This is not correct. Existing session support generates 8 bytes long random string whereas this PR generates guessable session ID using SHA-256 where the length is 32 bytes long. This change makes both to work.

When session support is deprecated and removed, this check will be fine-tuned.

PR #2916, opened 2019-10-02
Doc: Remove text on s3V4 incompatibility with GCS

### LINKED ISSUE #2915: Unable to specify s3v2 when using MC_HOST_ env
In the documentation there's a note about selecting s3v2 when using gcs.
```
NOTE: Google Cloud Storage only supports Legacy Signature Version 2, so you have to pick - S3v2
```
Following on from that the option to specify configuration for the session using `MC_HOST_` env is provided. I cannot see an option here to provide the API without going via `mc config` which is what I'm looking to avoid by using `MC_HOST_`.

## Expected behavior

I should be able to specify the api as S3v2 without using `mc config`, either by cli flag to existing commands, or via an env (making it consistent with `MC_HOST_`).

## Actual behavior

Specifying protocol as part of url (as indicated searching through issues/merge_requests) gives a parsing error:

Specifying with `https://` and using `--debug` appears to be using Authorization: AWS4-HMAC-SHA256, as opposed to just AWS when going via `mc config` and using the `--api` option.

Specifying api as part of protocol gives parsing error:
```
MC_HOST_s3=https+s3v2://ACCESS-KEY-HERE:SECRET-KEY-HERE@storage.googleapis.com mc ls s3/bucketname/
```

Using storage.googleapis.com defaults to s3v4:
```
MC_HOST_s3=https://ACCESS-KEY-HERE:SECRET-KEY-HERE@storage.googleapis.com mc ls s3/bucketname/ --debug
```

## mc version

```
Version: 2019-09-24T01:36:20Z
Release-tag: RELEASE.2019-09-24T01-36-20Z
Commit-id: 643835013047aa27ed35de0ac5a4ab9538f0cd68
```

## System information
```
Release-Tag:RELEASE.2019-09-24T01-36-20Z | Commit:643835013047 | Host:redacted | OS:linux | Arch:amd64 | Lang:go1.13 | Mem:6.8 MB/72 MB | Heap:6.8 MB/66 MB
```

kannappanr:
GCS right now works with S3V4, so removed the documentation
where it says that GCS only supports s3V2.

Fixes #2915

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,8 +149,6 @@ Get your AccessKeyID and SecretAccessKey by following [Google Credentials Guide]
 mc config host add gcs  https://storage.googleapis.com BKIKJAA5BMMU2RHO6IBB V8f1CwQqAcwo80UEIJEjc5gVQUSSx5ohQ9GSrr12
 ```
 
-NOTE: Google Cloud Storage only supports Legacy Signature Version 2, so you have to pick - S3v2
-
 ## Test Your Setup
 `mc` is pre-configured with https://play.min.io, aliased as "play". It is a hosted MinIO server for testing and development purpose.  To test Amazon S3, simply replace "play" with "s3" or the alias you used at the time of setup.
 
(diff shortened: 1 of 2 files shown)


mdsummers:
Please note that a similar change is also required in README.md, as the note re:gcs and s3v2 exists there as well.

kannappanr:
> please update README.md @kannappanr

done

kannappanr:
> Please note that a similar change is also required in README.md, as the note re:gcs and s3v2 exists there as well.

Thanks

PR #2938, opened 2019-10-23
Add options to preserve file system attributes

### LINKED ISSUE #2881: Add options to preserve object create date in mirror and cp
## Expected behavior
There can be an option like `cp -p` in Linux to keep files' create date when moving them, such as `mc mirror -p` or `mc cp -p`.

## Actual behavior
There's no this kind of options in `mc mirror` or `mc cp`.

Sometimes I need to `mc mirror` between servers or buckets. 
But once mirrored from A to B,  B's objects are all newer than A. Then if I mirrored back from B to A, I have to overwrite the entire bucket.

## mc version
Version: 2019-08-29T00:40:57Z
Release-tag: RELEASE.2019-08-29T00-40-57Z
Commit-id: bd7564a2d4524335a594f0f4e10b06c033eb299c

sinhaashish:
Have introduced an option `-a` to preserve the file system attribute while copy. For mirror it will preserve the file attributes and bucket policy if any.

Refer https://github.com/minio/mc/issues/2881 for details.

With `a` option in `cp` and `mirror` we try to preserve the file attribute in the following cases : 

**Case 1** :  Moving file from file system to object storage
 In this case the file system attribute is saved as metadata with the tag `X-Amz-Meta-Mc-Attrs` .
**Case 2** :  Moving file from file system to file system
No change in this case
**Case 3** :  Moving file from object storage to file system.
In this case the value of `X-Amz-Meta-Mc-Attrs` (when uploaded through mc) or `X-Amz-Meta-S3cmd-Attrs` (when uploaded through s3cmd)  is read. Then it is parsed to fetch the mod. uid, atime, mtime etc.. and applied to the file using `os.chmod` , `os.Chtime` , `os.Chown` .

**Case 4** :  Moving file from object storage to  object storage.
Object is copied as is. 

**How to test** : Run 
```
1. mc cp -a ~/Downloads/test.csv  myminio/bucket 
2. mc stat myminio/bucket/test.csv
```
Simmilarly test for mirror using the option `-a`

Closes #2881

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -25,6 +25,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"sort"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -32,6 +33,7 @@ import (
 	"github.com/pkg/xattr"
 	"github.com/rjeczalik/notify"
 
+	"github.com/minio/mc/pkg/disk"
 	"github.com/minio/mc/pkg/hookreader"
 	"github.com/minio/mc/pkg/ioutils"
 	"github.com/minio/mc/pkg/probe"
@@ -338,12 +340,62 @@ func (f *fsClient) put(reader io.Reader, size int64, metadata map[string][]strin
 			err := f.toClientError(e, objectPath)
 			return totalWritten, err.Trace(objectPartPath, objectPath)
 		}
+
+		if len(metadata["mc-attrs"]) != 0 {
+			attr, err := parseAttribute(metadata["mc-attrs"][0])
+			if err != nil {
+				return totalWritten, probe.NewError(err)
+			}
+
+			mode, err := strconv.ParseUint(attr["mode"], 10, 32)
+			if err != nil {
+				return totalWritten, probe.NewError(err)
+			}
+			// Change the mode of file
+			if err := os.Chmod(objectPath, os.FileMode(mode)); err != nil {
+				return totalWritten, probe.NewError(err)
+			}
+
+			uid, err := strconv.Atoi(attr["uid"])
+			if err != nil {
+				return totalWritten, probe.NewError(err)
+			}
+
+			gid, err := strconv.Atoi(attr["gid"])
+			if err != nil {
+				return totalWritten, probe.NewError(err)
+			}
+			// Change the owner
+			if err := os.Chown(objectPath, uid, gid); err != nil {
+				return totalWritten, probe.NewError(err)
+			}
+
+			atime, err := strconv.ParseInt(attr["atime"], 10, 64)
+			if err != nil {
+				return totalWritten, probe.NewError(err)
+			}
+
+			ctime, err := strconv.ParseInt(attr["ctime"], 10, 64)
+			if err != nil {
+				return totalWritten, probe.NewError(err)
+			}
+			// Change the access, modify and change time
+			if err := os.Chtimes(objectPath, time.Unix(atime, 0), time.Unix(ctime, 0)); err != nil {
+				return totalWritten, probe.NewError(err)
+			}
+
+		}
 	}
 	return totalWritten, nil
 }
 
 // Put - create a new file with metadata.
 func (f *fsClient) Put(ctx context.Context, reader io.Reader, size int64, metadata map[string]string, progress io.Reader, sse encrypt.ServerSide) (int64, *probe.Error) {
+	if metadata["mc-attrs"] != "" {
+		meta := make(map[string][]string)
+		meta["mc-attrs"] = append(meta["mc-attrs"], metadata["mc-attrs"])
+		return f.put(reader, size, meta, progress)
+	}
 	return f.put(reader, size, nil, progress)
 }
 
@@ -1012,7 +1064,7 @@ func (f *fsClient) SetAccess(access string, isJSON bool) *probe.Error {
 }
 
 // Stat - get metadata from path.
-func (f *fsClient) Stat(isIncomplete, isFetchMeta bool, sse encrypt.ServerSide) (content *clientContent, err *probe.Error) {
+func (f *fsClient) Stat(isIncomplete, isFetchMeta, isPreserve bool, sse encrypt.ServerSide) (content *clientContent, err *probe.Error) {
 	st, err := f.fsStat(isIncomplete)
 	if err != nil {
 		return nil, err.Trace(f.PathURL.String())
@@ -1037,6 +1089,15 @@ func (f *fsClient) Stat(isIncomplete, isFetchMeta bool, sse encrypt.ServerSide) 
 		}
 		for k, v := range metaData {
 			content.Metadata[k] = v
+		}
+		// Populates meta data with file system attribute only in case of
+		// when preserve flag is passed.
+		if isPreserve {
+			fileAttr, err := disk.GetFileSystemAttrs(path)
+			if err != nil {
+				return content, nil
+			}
+			content.Metadata["mc-attrs"] = fileAttr
 		}
 	}
 
--- a/cmd/cp-main.go
+++ b/cmd/cp-main.go
@@ -65,6 +65,10 @@ var (
 			Name:  "continue, c",
 			Usage: "create or resume copy session",
 		},
+		cli.BoolFlag{
+			Name:  "preserve, a",
+			Usage: "preserve filesystem attributes (mode, ownership, timestamps)",
+		},
 	}
 )
 
@@ -133,8 +137,11 @@ EXAMPLES:
       {{.Prompt}} {{.HelpName}} --storage-class REDUCED_REDUNDANCY myobject.txt play/mybucket
 
   14. Copy a text file to an object storage and create or resume copy session.
-      $ {{.HelpName}} --recursive --continue myobject.txt play/mybucket
- `,
+      {{.Prompt}} {{.HelpName}} --recursive --continue myobject.txt play/mybucket
+	  
+  15. Copy a text file to an object storage and preserve the file system attribute as metadata.
+      {{.Prompt}} {{.HelpName}} -a myobject.txt play/mybucket
+`,
 }
 
 // copyMessage container for file copy messages
@@ -385,6 +392,18 @@ func doCopySession(session *sessionV8, encKeyDB map[string][]prefixSSEPair) erro
 					}
 				}
 
+				// If one needs to store the file system information by passing -a flag
+				if preserve := (session.Header.CommandBoolFlags["preserve"]); preserve {
+					attrValue, pErr := getFileAttrMeta(cpURLs, encKeyDB)
+					if pErr != nil {
+						errorIf(pErr, "Unable to fetch file meta info for %s", urlScanner.Text())
+						continue
+					}
+
+					if attrValue != "" {
+						cpURLs.TargetContent.Metadata["mc-attrs"] = attrValue
+					}
+				}
 				// Verify if previously copied, notify progress bar.
 				if isCopied(cpURLs.SourceContent.URL.String()) {
 					queueCh <- func() URLs {
@@ -519,6 +538,10 @@ func mainCopy(ctx *cli.Context) error {
 	session.Header.CommandStringFlags["encrypt-key"] = sseKeys
 	session.Header.CommandStringFlags["encrypt"] = sse
 	session.Header.CommandBoolFlags["session"] = ctx.Bool("continue")
+
+	if ctx.Bool("preserve") {
+		session.Header.CommandBoolFlags["preserve"] = ctx.Bool("preserve")
+	}
 	session.Header.UserMetaData = userMetaMap
 
 	var e error
--- a/cmd/cp-url-syntax.go
+++ b/cmd/cp-url-syntax.go
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"fmt"
+	"runtime"
 
 	"github.com/minio/cli"
 	"github.com/minio/mc/pkg/console"
@@ -40,7 +41,7 @@ func checkCopySyntax(ctx *cli.Context, encKeyDB map[string][]prefixSSEPair) {
 
 	// Verify if source(s) exists.
 	for _, srcURL := range srcURLs {
-		_, _, err := url2Stat(srcURL, false, encKeyDB)
+		_, _, err := url2Stat(srcURL, false, false, encKeyDB)
 		if err != nil {
 			console.Fatalf("Unable to validate source %s\n", srcURL)
 		}
@@ -72,6 +73,11 @@ func checkCopySyntax(ctx *cli.Context, encKeyDB map[string][]prefixSSEPair) {
 	default:
 		fatalIf(errInvalidArgument().Trace(), "Unable to guess the type of copy operation.")
 	}
+
+	// Preserve functionality not supported for windows
+	if ctx.Bool("preserve") && runtime.GOOS == "windows" {
+		fatalIf(errInvalidArgument().Trace(), "Permissions are not preserved on windows platform.")
+	}
 }
 
 // checkCopySyntaxTypeA verifies if the source and target are valid file arguments.
@@ -81,7 +87,7 @@ func checkCopySyntaxTypeA(srcURLs []string, tgtURL string, keys map[string][]pre
 		fatalIf(errInvalidArgument().Trace(), "Invalid number of source arguments.")
 	}
 	srcURL := srcURLs[0]
-	_, srcContent, err := url2Stat(srcURL, false, keys)
+	_, srcContent, err := url2Stat(srcURL, false, false, keys)
 	fatalIf(err.Trace(srcURL), "Unable to stat source `"+srcURL+"`.")
 
 	if !srcContent.Type.IsRegular() {
@@ -96,7 +102,7 @@ func checkCopySyntaxTypeB(srcURLs []string, tgtURL string, keys map[string][]pre
 		fatalIf(errInvalidArgument().Trace(), "Invalid number of source arguments.")
 	}
 	srcURL := srcURLs[0]
-	_, srcContent, err := url2Stat(srcURL, false, keys)
+	_, srcContent, err := url2Stat(srcURL, false, false, keys)
 	fatalIf(err.Trace(srcURL), "Unable to stat source `"+srcURL+"`.")
 
 	if !srcContent.Type.IsRegular() {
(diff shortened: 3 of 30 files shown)


harshavardhana:
We don't need to introduce new option instead use -a not -p

cmd/cp-main.go
```diff
@@ -133,7 +138,10 @@ EXAMPLES:
       {{.Prompt}} {{.HelpName}} --storage-class REDUCED_REDUNDANCY myobject.txt play/mybucket
 
   14. Copy a text file to an object storage and create or resume copy session.
-      $ {{.HelpName}} --recursive --continue myobject.txt play/mybucket
+      {{.HelpName}} --recursive --continue myobject.txt play/mybucket
```
harshavardhana: Need to use `{{.Prompt}}`

cmd/mirror-main.go
```diff
@@ -282,6 +289,38 @@ func (mj *mirrorJob) doMirror(ctx context.Context, cancelMirror context.CancelFu
...
+				if hostCfgTarget == nil {
+					// if source has file system attribute as metadata
+					if key == "X-Amz-Meta-Mc-Attrs" {
+						sURLs.TargetContent.Metadata["mc-attrs"] = value
+					} else if key == "X-Amz-Meta-S3cmd-Attrs" {
+						sURLs.TargetContent.Metadata["mc-attrs"] = value
+					}
+				}
+			}
+		} else {
```
harshavardhana: This is causing abstraction leak we shouldn't do these things, you should enhance the code such that a filesystem Stat returns appropriate attributes and as well as object storage should also differentiate. 

In this code there are two different paths taken showing the actual differentiation which is exactly we should avoid and have a single call.

You need to redo this PR @sinhaashish

cmd/client-fs.go
```diff
@@ -338,12 +339,55 @@ func (f *fsClient) put(reader io.Reader, size int64, metadata map[string][]strin
...
+			if err != nil {
+				return totalWritten, probe.NewError(err)
+			}
+
+			ctime, err := strconv.ParseInt(attr["ctime"], 10, 64)
+			if err != nil {
+				return totalWritten, probe.NewError(err)
+			}
+			// Change the access, modify and change time
+			os.Chtimes(objectPath, time.Unix(atime, 0), time.Unix(ctime, 0))
```
harshavardhana: errors cannot be ignored for each os calls @sinhaashish you should handle them and return them upwards to the caller.

ebozduman:
@harshavardhana @sinhaashish
In Case 4:
> Case 4 : Moving file from object storage to object storage.
Object is copied as is. 

Don't we also need to make sure we also carry over bucket policies and `X-Amz-Meta-Mc-Attrs` and `X-Amz-Meta-S3cmd-Attrs` if these are already populated, (i.e. if an attr copy, `mc cp -a <a-file> <this-object>`, happened in the past)?

cmd/mirror-main.go
```diff
@@ -677,11 +692,11 @@ func runMirror(srcURL, dstURL string, ctx *cli.Context, encKeyDB map[string][]pr
 	dstClt, err := newClient(dstURL)
 	fatalIf(err, "Unable to initialize `"+srcURL+"`.")
 
-	if ctx.Bool("a") && (srcClt.GetURL().Type != objectStorage || dstClt.GetURL().Type != objectStorage) {
-		fatalIf(errDummy(), "Synchronizing bucket policies is only possible when both source & target point to S3 servers.")
-	}
+	// if ctx.Bool("a") && (srcClt.GetURL().Type != objectStorage || dstClt.GetURL().Type != objectStorage) {
```
harshavardhana: Why is this commented out?

cmd/utils.go
```diff
@@ -318,3 +325,89 @@ func isURLContains(srcURL, tgtURL, sep string) bool {
...
+// Return the file system attribute as string; containing mode,
+// uid, gid, uname, Gname, atime, mtime, ctime and md5
+func getFileSystemAttrs(file string) (string, error) {
+	var uname, gname, atime, mtime, ctime, mode, uid, gid string
+	fi, err := os.Stat(file)
+	if err != nil {
+		return "", err
+	}
+	st := fi.Sys().(*syscall.Stat_t)
+	mode = fmt.Sprint(st.Mode)
```
harshavardhana: Always use `stconv.Itoa()` to avoid fmt.Sprint allocation

cmd/utils.go
```diff
@@ -318,3 +325,89 @@ func isURLContains(srcURL, tgtURL, sep string) bool {
...
+
+	if t.HasChangeTime() {
+		ctime = fmt.Sprint(t.ChangeTime().Unix())
+	}
+
+	md5, err := hashFilemd5(file)
+	if err != nil {
+		return "", err
+	}
+	fileAttr := fmt.Sprintf("atime:%s/ctime:%s/gid:%s/gname:%s/md5:%s/mode:%s/mtime:%s/uid:%s/uname:%s", atime, ctime, gid, gname, md5, mode, mtime, uid, uname)
```
harshavardhana: Avoid multiple allocations in this manner instead construct this last string only once since you can use %d instead of %s for each step, the code will look less verbose and cleaner.

harshavardhana: Or use strings.Builder and keep writing new data like `s.WriteString()`

cmd/utils.go
```diff
@@ -318,3 +325,89 @@ func isURLContains(srcURL, tgtURL, sep string) bool {
...
+
+	return fileAttr, nil
+}
+
+// Return the md5 of the file as string
+func hashFilemd5(filePath string) (string, error) {
```
harshavardhana: This will slow down the process, we don't need to save md5sum of the objects we can totally ignore this part and not implement it.

cmd/utils.go
```diff
@@ -318,3 +325,89 @@ func isURLContains(srcURL, tgtURL, sep string) bool {
...
+}
+
+// Returns a map by parsing the value of X-Amz-Meta-Mc-Attrs/X-Amz-Meta-s3Cmd-Attrs
+func parseAttribute(attrs string) map[string]string {
+	attribute := make(map[string]string)
+	for _, attr := range strings.Split(attrs, "/") {
+		param := strings.Split(attr, ":")
+		attribute[param[0]] = param[1]
+	}
+	return attribute
```
harshavardhana: This code panics with an empty string as input, write tests to ensure that all negative cases are handled. 

https://play.golang.org/p/RWGfz35f-18 - see all the offending inputs

cmd/utils.go
```diff
@@ -318,3 +325,89 @@ func isURLContains(srcURL, tgtURL, sep string) bool {
...
+	}
+	uname = i.Username
+
+	g, err := user.LookupGroupId(strconv.FormatUint(uint64(st.Gid), 10))
+	if err != nil {
+		return "", err
+	}
+	gname = g.Name
+
+	t, err := times.Stat(file)
```
harshavardhana: why do we need this call @sinhaashish we can access the necessary ModTime and AccessTime all with in the same os.Stat() call, I don't think we need to depend on new projects right now.

harshavardhana: https://golang.org/pkg/syscall/#Stat_t already has Atim captured it shall be just `Atim.Sec`

cmd/utils.go
```diff
@@ -318,3 +325,89 @@ func isURLContains(srcURL, tgtURL, sep string) bool {
...
+
+// Return the file system attribute as string; containing mode,
+// uid, gid, uname, Gname, atime, mtime, ctime and md5
+func getFileSystemAttrs(file string) (string, error) {
+	var uname, gname, atime, mtime, ctime, mode, uid, gid string
+	fi, err := os.Stat(file)
+	if err != nil {
+		return "", err
+	}
+	st := fi.Sys().(*syscall.Stat_t)
```
harshavardhana: Also, you need to make sure to write platform-specific code into a separate file like `stat_windows.go`, make sure that you all this code works properly on all other platforms such as FreeBSD etc as well. The best way to validate is first to check if the cross-compile works fine.

cmd/cp-main.go
```diff
@@ -519,6 +539,11 @@ func mainCopy(ctx *cli.Context) error {
 	session.Header.CommandStringFlags["encrypt-key"] = sseKeys
 	session.Header.CommandStringFlags["encrypt"] = sse
 	session.Header.CommandBoolFlags["session"] = ctx.Bool("continue")
+
+	// Preserve functionality not supported for windows
+	if ctx.Bool("preserve") && runtime.GOOS != "windows" {
```
harshavardhana: We shouldn't add code like this, we should error out in checkSyntax() if this occurs and let the user know. Adding code like to avoid saving the flag while on windows platform is buggy coding

functional-tests.sh
```diff
@@ -678,6 +678,34 @@ function test_cat_object_with_sse_error()
...
+    md5_sum="$(get_md5sum "$FILE_1_MB")"
+    if [ $? -ne 0 ]; then
+        echo "unable to get md5sum of $FILE_1_MB"
+        exit 1
+    fi
+    mode="$(stat --format=%a ${FILE_1_MB})"
+    mtime="$(stat --format=%Z ${FILE_1_MB})"
+    userid="$(stat --format=%u ${FILE_1_MB})"
+    uname="$(stat --format=%U ${FILE_1_MB})"
+    FILE_ATTRIBUTE="atime:${atime}/ctime:${ctime}/gid:${gid}/gname:${gname}/md5:${md5_sum}/mode:${mode}/mtime:${mtime}/uid:${userid}/uname:${uname}"
```
harshavardhana: This test is the wrong @sinhaashish using Unix stat command will fail because it is not converting the octal representation properly instead it will show you instead enhance `mc stat --json` output to print the `mode` of the file instead @sinhaashish 

Or enhance the `mc stat` itself to print in the `FILE_ATTRIBUTE` fashion such that comparison becomes easier.

cmd/cp-main.go
```diff
@@ -133,7 +138,10 @@ EXAMPLES:
       {{.Prompt}} {{.HelpName}} --storage-class REDUCED_REDUNDANCY myobject.txt play/mybucket
 
   14. Copy a text file to an object storage and create or resume copy session.
-      $ {{.HelpName}} --recursive --continue myobject.txt play/mybucket
+      {{.Prompt} {{.HelpName}} --recursive --continue myobject.txt play/mybucket
+
+  15. Copy a text file to an object storage and preserve the file system attribute as metadata.
+      {{.Prompt} {{.HelpName}} -p myobject.txt play/mybucket
```
harshavardhana: Its -a now

cmd/client-fs.go
```diff
@@ -338,12 +340,61 @@ func (f *fsClient) put(reader io.Reader, size int64, metadata map[string][]strin
...
+		}
 	}
 	return totalWritten, nil
 }
 
 // Put - create a new file with metadata.
 func (f *fsClient) Put(ctx context.Context, reader io.Reader, size int64, metadata map[string]string, progress io.Reader, sse encrypt.ServerSide) (int64, *probe.Error) {
+	if metadata["mc-attrs"] != "" {
+		meta := make(map[string][]string)
+		val := make([]string, 1)
```
harshavardhana: you don't need to do this it is ideal to have a more idiomatic style slice doesn't need allocation

```
meta["mc-attrs"] = append(meta["mc-attrs"], metadata["mc-attrs"])
```

cmd/cp-url-syntax.go
```diff
@@ -72,6 +73,11 @@ func checkCopySyntax(ctx *cli.Context, encKeyDB map[string][]prefixSSEPair) {
 	default:
 		fatalIf(errInvalidArgument().Trace(), "Unable to guess the type of copy operation.")
 	}
+
+	// Preserve functionality not supported for windows
+	if ctx.Bool("preserve") && runtime.GOOS == "windows" {
+		fatalIf(errInvalidArgument().Trace(), "Preserve functionality not supported for windows.")
```
harshavardhana:
```suggestion
		fatalIf(errInvalidArgument().Trace(), "Permissions are not preserved on windows platform")
```

cmd/mirror-main.go
```diff
@@ -64,8 +64,8 @@ var (
 			Value: "us-east-1",
 		},
 		cli.BoolFlag{
-			Name:  "a",
-			Usage: "preserve bucket policy rules on target bucket(s)",
+			Name:  "preserve, a",
+			Usage: "preserve file system attributes and bucket policy rules on target bucket(s)",
```
harshavardhana:
```suggestion
			Usage: "preserve file(s)/object(s) attributes and bucket policy rules on target bucket(s)",
```

cmd/mirror-main.go
```diff
@@ -157,6 +157,9 @@ EXAMPLES:
 
   13. Update 'Cache-Control' header on existing objects.
       {{.Prompt}} {{.HelpName}} --attr Cache-Control=max-age=90000,min-fresh=9000 myminio/video-files myminio/video-files
+	  
+  14. Mirror a local folder recursively to Amazon S3 cloud storage and preserve the file attribute as metadata.
```
harshavardhana:
```suggestion
  14. Mirror a local folder recursively to Amazon S3 cloud storage and preserve all local file attributes.
```

docs/minio-client-complete-guide.md
```diff
@@ -562,6 +562,7 @@ FLAGS:
   --older-than value                 copy object(s) older than N days (default: 0)
   --newer-than value                 copy object(s) newer than N days (default: 0)
   --storage-class value, --sc value  set storage class for new object(s) on target
+  -a                                 preserve file system attributes and bucket policy rules on target bucket(s)
```
harshavardhana: you need to specify `--preserve` as well?

docs/minio-client-complete-guide.md
```diff
@@ -773,7 +781,7 @@ FLAGS:
   --watch, -w                        watch and synchronize changes
   --remove                           remove extraneous object(s) on target
   --region value                     specify region when creating new bucket(s) on target (default: "us-east-1")
-  -a                                 preserve bucket policy rules on target bucket(s)
+  -a                                 preserve file system attributes and bucket policy rules on target bucket(s)
```
harshavardhana: you need to specify `--preserve` as well?

pkg/disk/stat_windows.go
```diff
@@ -0,0 +1,25 @@
+// +build solaris
+
+/*
+ * MinIO Cloud Storage, (C) 2017 MinIO, Inc.
```
harshavardhana:
```suggestion
 * MinIO Cloud Storage, (C) 2019 MinIO, Inc.
```

pkg/disk/stat_windows.go
```diff
@@ -0,0 +1,25 @@
+// +build solaris
```
harshavardhana: File says windows and `// +build` says solaris - pick one

pkg/probe/wrapper.go
```diff
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-// Package probe implements a simple mechanism to trace and return errors in large programs.
```
harshavardhana: No need to delete this..

harshavardhana:
Looks like `mc cp` crashed 

```
mc cp
panic: template: help:56: unexpected "}" in operand

goroutine 1 [running]:
text/template.Must(...)
	/home/harsha/.gimme/versions/go1.13.linux.amd64/src/text/template/helper.go:23
github.com/minio/cli.printHelpCustom(0xe62f80, 0xc0000ba008, 0xd7505d, 0xd07, 0xd268a0, 0xc0002a0000, 0x0)
	/home/harsha/mygo/pkg/mod/github.com/minio/cli@v1.22.0/help.go:260 +0x514
github.com/minio/cli.ShowCommandHelp(0xc0000c2b60, 0xd2fd8f, 0x2, 0xc0004b9580, 0x40c1b6)
	/home/harsha/mygo/pkg/mod/github.com/minio/cli@v1.22.0/help.go:203 +0x4ae
github.com/minio/cli.ShowCommandHelpAndExit(0xc0000c2b60, 0xd2fd8f, 0x2, 0x1)
	/home/harsha/mygo/pkg/mod/github.com/minio/cli@v1.22.0/help.go:167 +0x3f
github.com/minio/mc/cmd.checkCopySyntax(0xc0000c2b60, 0xc0002abc80)
	/home/harsha/mygo/src/github.com/minio/mc/cmd/cp-url-syntax.go:29 +0x74e
github.com/minio/mc/cmd.mainCopy(0xc0000c2b60, 0x0, 0xc2b60)
	/home/harsha/mygo/src/github.com/minio/mc/cmd/cp-main.go:506 +0xd6
github.com/minio/cli.HandleAction(0xc14360, 0xd857c8, 0xc0000c2b60, 0xc0000bd300, 0x0)
	/home/harsha/mygo/pkg/mod/github.com/minio/cli@v1.22.0/app.go:498 +0xc8
github.com/minio/cli.Command.Run(0xd2fd8f, 0x2, 0x0, 0x0, 0x0, 0x0, 0x0, 0xd38d7c, 0xc, 0x0, ...)
	/home/harsha/mygo/pkg/mod/github.com/minio/cli@v1.22.0/command.go:225 +0x991
github.com/minio/cli.(*App).Run(0xc0001d4340, 0xc0000c0000, 0x2, 0x2, 0x0, 0x0)
	/home/harsha/mygo/pkg/mod/github.com/minio/cli@v1.22.0/app.go:261 +0x6ab
github.com/minio/mc/cmd.Main(0xc0000c0000, 0x2, 0x2)
	/home/harsha/mygo/src/github.com/minio/mc/cmd/main.go:113 +0x29c
main.main()
	/home/harsha/mygo/src/github.com/minio/mc/main.go:34 +0x45
```

Perhaps like missing `{`

cmd/utils.go
sinhaashish: > This code panics with an empty string as input, write tests to ensure that all negative cases are handled.
> 
> https://play.golang.org/p/RWGfz35f-18 - see all the offending inputs

@harshavardhana  Should i add `recover()` in the `parseAttribute` function. How to add the test case for recover. Not clear on this, need your input.

harshavardhana: No, I am asking to write this part of the code such that there is no panic and we recover appropriately - we shouldn't add `recover()`

cmd/client-fs.go
```diff
@@ -338,12 +340,62 @@ func (f *fsClient) put(reader io.Reader, size int64, metadata map[string][]strin
...
+			if err := os.Chmod(objectPath, os.FileMode(mode)); err != nil {
+				return totalWritten, probe.NewError(err)
+			}
+
+			uid, err := strconv.Atoi(attr["uid"])
+			if err != nil {
+				return totalWritten, probe.NewError(err)
+			}
+
+			gid, err := strconv.Atoi(attr["uid"])
```
kannappanr:
```suggestion
			gid, err := strconv.Atoi(attr["gid"])
```

harshavardhana: @sinhaashish ping

sinhaashish: done

PR #2946, opened 2019-10-30
Optionally print disk usage as plain integer

collinwright:
Providing the `--plain-bytes` option to `mc du` will print usage as an integer number of bytes, without humanization.

I believe this would be helpful for using the output directly in scripts or with tools like ElasticSearch, especially when combined with the `--json` flag.

I couldn't find any precedent for unit tests for a command like this --- please point me at them if I missed them :-)

### DIFF
--- a/cmd/du-main.go
+++ b/cmd/du-main.go
@@ -70,13 +70,15 @@ EXAMPLES:
 // Structured message depending on the type of console.
 type duMessage struct {
 	Prefix string `json:"prefix"`
-	Size   string `json:"size"`
+	Size   int64  `json:"size"`
 	Status string `json:"status"`
 }
 
 // Colorized message for console printing.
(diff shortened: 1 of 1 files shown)


kannappanr:
@collinwright thanks for the PR. Let me check internally if printing the size in bytes in the case of `json` is better than adding a new flag and get back to you

harshavardhana:
We should perhaps print plain numbers in `--json` mode without humanized form - let's not introduce --plain-bytes

kannappanr:
@collinwright Instead of adding a new flag, can you please make that change?

collinwright:
@kannappanr sure thing, makes the code way simpler anyways :-)

collinwright:
@kannappanr  changes applied, branch rebased.

PR #2961, opened 2019-11-19
help command is now part of `mc admin config get/set`

harshavardhana:


cmd/admin-config-get.go
```diff
@@ -25,12 +25,19 @@ import (
 	"github.com/minio/minio/pkg/madmin"
 )
 
+var adminConfigGetFlags = []cli.Flag{
+	cli.BoolFlag{
+		Name:  "env",
+		Usage: "list all the env only help",
+	},
```
poornas: `mc admin config get --env myminio` just lists the subsystems - not the env.  The server HelpConfigKVHandler needs to be removed and env functionality brought into the GetConfigKVHandler in https://github.com/minio/minio/pull/8535.

harshavardhana: Its not done yet @poornas

harshavardhana: the APIs will remain different @poornas since the data structure returned are different, what `mc` does is a richer functionality. APIs themselves will remain.

PR #2963, opened 2019-11-20
Retention command to apply retention to objects with a Prefix

kannappanr:


poornas:
we need to unify the PutRetention and PutObjectRetention interface methods to one call - perhaps in another PR.

PR #2964, opened 2019-11-20
mirror object lock configuration to dest bucket

poornas:
if bucket does not exist on the target side.

This PR will allow Object lock configuration on the source bucket to be mirrored to destination bucket if bucket does not exist on the target site.
In addition object retention headers will also be mirrored if retention headers are set on source object

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -72,6 +72,10 @@ const (
 	fileHeaderType           = "fileheader"
 	commentCharType          = "commentchar"
 	typeJSONType             = "type"
+	// AmzObjectLockMode sets object lock mode
+	AmzObjectLockMode = "X-Amz-Object-Lock-Mode"
+	// AmzObjectLockRetainUntilDate sets object lock retain until date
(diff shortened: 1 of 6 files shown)


cmd/client-s3.go
```diff
@@ -80,6 +80,7 @@ var cseHeaders = []string{
 	"X-Amz-Meta-X-Amz-Key",
 	"X-Amz-Meta-X-Amz-Matdesc",
 }
+var timeSentinel = time.Unix(0, 0).UTC()
```
harshavardhana: space before `var`

cmd/client-s3.go
```diff
@@ -799,6 +800,23 @@ func (c *s3Client) Put(ctx context.Context, reader io.Reader, size int64, metada
 	if ok {
 		delete(metadata, "X-Amz-Storage-Class")
 	}
+
+	lockModeStr, ok := metadata["X-Amz-Object-Lock-Mode"]
```
harshavardhana: shall we define some global constants for these names?

poornas: fixed

PR #2965, opened 2019-11-21
Add mirror command to compare metadata and copy

harshavardhana:
mirror now uses the newly introduced
ListObjectsV2WithMetadata API in minio-go

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -68,11 +68,11 @@ func fsNew(path string) (Client, *probe.Error) {
 	}, nil
 }
 
-func isNotSupported(err error) bool {
-	if err == nil {
+func isNotSupported(e error) bool {
+	if e == nil {
 		return false
 	}
-	errno := err.(*xattr.Error)
+	errno := e.(*xattr.Error)
 	if errno == nil {
 		return false
 	}
@@ -83,13 +83,13 @@ func isNotSupported(err error) bool {
 
 // isIgnoredFile returns true if 'filename' is on the exclude list.
 func isIgnoredFile(filename string) bool {
-	matchFile := path.Base(filename)
+	matchFile := filepath.Base(filename)
 
 	// OS specific ignore list.
 	for _, ignoredFile := range ignoreFiles[runtime.GOOS] {
-		matched, err := filepath.Match(ignoredFile, matchFile)
-		if err != nil {
-			panic(err)
+		matched, e := filepath.Match(ignoredFile, matchFile)
+		if e != nil {
+			panic(e)
 		}
 		if matched {
 			return true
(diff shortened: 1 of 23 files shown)


cmd/client-s3.go
```diff
@@ -1012,12 +1012,20 @@ func (c *s3Client) MakeBucket(region string, ignoreExisting, withLock bool) *pro
 		return probe.NewError(BucketNameEmpty{})
 	}
 	if object != "" {
-		if strings.HasSuffix(object, "/") {
+		if !strings.HasSuffix(object, string(c.targetURL.Separator)) {
```
klauspost: I had a hard time understanding this, so I tried to rewrite it so return paths are more clear:
```Go
	if object != "" {
		if !strings.HasSuffix(object, string(c.targetURL.Separator)) {
			object = path.Dir(object)
		}
		if !strings.HasSuffix(object, string(c.targetURL.Separator)) {
			return probe.NewError(BucketNameTopLevel{})
		}
		var retried bool
		for {
			_, e := c.api.PutObject(bucket, object, bytes.NewReader([]byte("")), 0, minio.PutObjectOptions{})
			if e != nil {
				if retried {
					return probe.NewError(e)
				}
				switch minio.ToErrorResponse(e).Code {
				case "NoSuchBucket":
					if withLock {
						e = c.api.MakeBucketWithObjectLock(bucket, region)
					} else {
						e = c.api.MakeBucket(bucket, region)
					}
					if e != nil {
						return probe.NewError(e)
					}
					retried = true
					continue
				}
				return probe.NewError(e)
			}
			return nil
		}
	}
```
Also added a single retry only, so we don't risk an infinite loop if something is broken.

go.mod
```diff
@@ -35,3 +35,5 @@ require (
 )
 
 replace github.com/gorilla/rpc v1.2.0+incompatible => github.com/gorilla/rpc v1.2.0
+
+replace github.com/minio/minio-go/v6 => ../minio-go
```
klauspost: Intended?

kannappanr: Yes, until `minio-go` with the new listing API is released. This is to test the PR

cmd/client-fs.go
```diff
@@ -68,11 +68,11 @@ func fsNew(path string) (Client, *probe.Error) {
 	}, nil
 }
 
-func isNotSupported(err error) bool {
-	if err == nil {
+func isNotSupported(e error) bool {
```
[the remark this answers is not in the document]
harshavardhana: in `mc` we have a convention if Go err is called e and actual probe.NewError() is called err changing this because it can lead to annoying bugs.

cmd/client-fs.go
```diff
@@ -343,46 +343,55 @@ func (f *fsClient) put(reader io.Reader, size int64, metadata map[string][]strin
...
+				return totalWritten, probe.NewError(e)
 			}
+
 			// Change the mode of file
-			if err := os.Chmod(objectPath, os.FileMode(mode)); err != nil {
-				return totalWritten, probe.NewError(err)
+			if e := os.Chmod(objectPath, os.FileMode(mode)); e != nil {
+				if !os.IsPermission(e) {
+					return totalWritten, probe.NewError(e)
+				}
```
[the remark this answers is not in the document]
harshavardhana: That is what I avoided @klauspost  there way too many messages when you are mirroring a large content. I will check with @abperiasamy if we can come up with something better here.

PR #2973, opened 2019-11-25
Refactor duplicated API method(s) for object

poornas:
retention

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -409,13 +409,6 @@ func (f *fsClient) Put(ctx context.Context, reader io.Reader, size int64, metada
 	return f.put(reader, size, nil, progress)
 }
 
-func (f *fsClient) PutRetention(ctx context.Context, metadata map[string]string) *probe.Error {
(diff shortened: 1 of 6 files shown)


kanagarajkm:
when i set retention to `1d`

```

poornas:
> when i set retention to `1d`
> 
> ```
> mc mb local/test2 --with-lock
> mc retention local/test2 governance 1d
> mc cp Dockerfile local/test2
> ```
> 
> ```
> aws s3api --endpoint-url http://localhost:9000 get-object-retention --bucket test2 --key Dockerfile
> {
>     "Retention": {
>         "RetainUntilDate": "0001-01-01T00:00:00Z"
>     }
> }
> ```
@kanagarajkm , the `mc retention local/test2 governance 1d` is applied on object prefixes that already exist in the bucket, not the ones created afterwards. If you intended to set a default object lock configuration for the bucket the correct command is `mc lock local/test2 governance 1d`

kanagarajkm:
@poornas understood, thanks.

PR #2979, opened 2019-12-04
Add finalized command line options

harshavardhana:


cmd/admin-config-set.go
```diff
@@ -30,7 +30,7 @@ import (
 
 var adminConfigSetCmd = cli.Command{
 	Name:   "set",
-	Usage:  "set key to MinIO server/cluster.",
+	Usage:  "interactively set a config key parameters",
```
kannappanr:
```suggestion
	Usage:  "interactively set a config key parameter",
```

harshavardhana:
looks like the github actions are running now PTAL @kannappanr @nitisht

PR #2981, opened 2019-12-05
Avoid re-parsing config history elements

harshavardhana:


Praveenrajmani:
> gofmt seems to be failing in CI, PTAL

I think it is because of `replace github.com/minio/minio => ../minio` which should be removed before merging @nitisht

PR #2982, opened 2019-12-05
New admin info

ebozduman:
Introduces the new `mc admin info <target>` command and removes the sub-commands `server`, `mem`, and `cpu`.


go.mod
```diff
@@ -34,3 +34,5 @@ require (
 )
 
 replace github.com/gorilla/rpc v1.2.0+incompatible => github.com/gorilla/rpc v1.2.0
+
+replace github.com/minio/minio => ../../minio
```
kannappanr:
```suggestion
replace github.com/minio/minio => ../minio
```

ebozduman: Done

cmd/admin-info.go
```diff
@@ -26,17 +39,130 @@ var adminInfoCmd = cli.Command{
...
+	aliasedURL := args.Get(0)
+
+	// Create a new MinIO Admin Client
+	client, err := newAdminClient(aliasedURL)
+	fatalIf(err, "Unable to initialize admin connection.")
+
+	// Fetch info of all servers (cluster or single server)
+	admInfo, _ := client.ServerAdminInfo()
+
+	fmt.Printf("first return:\n%+v\n", admInfo)
```
kannappanr:
```suggestion
```

cmd/admin-info.go
```diff
@@ -26,17 +39,130 @@ var adminInfoCmd = cli.Command{
...
+	// Get the alias parameter from cli
+	args := ctx.Args()
+	aliasedURL := args.Get(0)
+
+	// Create a new MinIO Admin Client
+	client, err := newAdminClient(aliasedURL)
+	fatalIf(err, "Unable to initialize admin connection.")
+
+	// Fetch info of all servers (cluster or single server)
+	admInfo, _ := client.ServerAdminInfo()
```
kannappanr: do not ignore the error here

cmd/admin-info.go
```diff
@@ -26,17 +39,130 @@ var adminInfoCmd = cli.Command{
...
+	msg += fmt.Sprintf("%s online, %s offline\n",
+		english.Plural(totalOnlineDisksCluster, "drive", ""),
+		english.Plural(totalOfflineDisksCluster, "drive", ""))
+
+	return
+}
+
+// JSON jsonified service status message.
+func (u infoMessage) JSON() string {
+	statusJSONBytes, e := json.MarshalIndent(u, "", "    ")
```
kannappanr: based on error, status and error fields need to be added here

cmd/admin-info.go
```diff
@@ -26,17 +39,191 @@ var adminInfoCmd = cli.Command{
 	Action: mainAdminInfo,
 	Before: setGlobalsFromContext,
 	Flags:  globalFlags,
-	Subcommands: []cli.Command{
-		adminInfoServer,
-		adminInfoCPU,
```
sinhaashish: Why are we removing this `adminInfoCPU`. With the present admin info PR https://github.com/minio/minio/pull/8497 , we don't fetch the CPU stats. Moreover the `adminInfoCPU,` uses `ServerCPULoadInfo()` , which is a different API all together. 

Same for `adminInfoMem`

kannappanr: We are going back to the old naming convention. The command will simply be `mc admin info alias`

kannappanr:
> 0

Currently it responds with
```
 $ mc admin info myminio
mc: <ERROR>  Cannot get service status

```

PR #2989, opened 2019-12-08
cp: Only use session when -c is passed

vadmeste:


cmd/progress-bar.go
```diff
@@ -73,9 +73,9 @@ func newProgressBar(total int64) *progressBar {
 	}
 
 	// Start the progress bar.
-	if bar.Total > 0 {
-		bar.Start()
-	}
+	// if bar.Total > 0 {
+	bar.Start()
+	// }
```
balamurugana: May be these comments can be removed.

sinhaashish:
need to add the `-c ` in the examples as well as `minio-client-complete-guide.md` for users to know about this functionality.

cmd/progress-bar.go
kannappanr: done

harshavardhana:
PTAL @sinhaashish @balamurugana

PR #3018, opened 2019-12-19
fix: ignore APINotImplemented errors from filesystem

harshavardhana:


cmd/mirror-main.go
```diff
@@ -544,7 +544,8 @@ func (mj *mirrorJob) watchMirror(ctx context.Context, cancelMirror context.Cance
 		case err := <-mj.watcher.Errors():
 			switch err.ToGoError().(type) {
 			case APINotImplemented:
-				errorIf(err.Trace(), "Unable to Watch on source, ignoring.")
+				errorIf(err.Trace(),
+					"Unable to Watch on source, perhap source doesn't support Watching for events")
```
kannappanr:
```suggestion
					"Unable to Watch on source, perhaps source doesn't support Watching for events")
```

cmd/client-fs.go
```diff
@@ -1015,25 +1018,34 @@ func (f *fsClient) MakeBucket(region string, ignoreExisting, withLock bool) *pro
...
+		API:     "GetObjectLockConfig",
+		APIType: "filesystem",
+	})
 }
 
 // GetAccessRules - unsupported API
 func (f *fsClient) GetAccessRules() (map[string]string, *probe.Error) {
 	return map[string]string{}, probe.NewError(APINotImplemented{
-		API:     "ListBucketPolicies",
+		API:     "GetBucketPolicy",
```
kannappanr: Was the API name wrong here?

harshavardhana: There is no API called ListBucketPolicies?  since its just GetBucketPolicy?

PR #3021, opened 2019-12-23
Mirror: Add STag and ETag on the destination only in Multimaster mode

kannappanr:
Currently, even if you simply do 

```
mc mirror /tmp/test/ play/test/
```
It creates empty `STag` and `ETag` headers on the destination.

Added a check to do that only if `Multimaster` mode is enabled

### DIFF
--- a/cmd/mirror-main.go
+++ b/cmd/mirror-main.go
@@ -303,17 +303,15 @@ func (mj *mirrorJob) doMirror(ctx context.Context, cancelMirror context.CancelFu
 		sURLs.TargetContent.Metadata["X-Amz-Storage-Class"] = mj.storageClass
 	}
 
-	// Set multiMasterETagKey for the target.
-	if sURLs.SourceContent.UserMetadata[multiMasterETagKey] != "" {
-		sURLs.TargetContent.Metadata[multiMasterETagKey] = sURLs.SourceContent.UserMetadata[multiMasterETagKey]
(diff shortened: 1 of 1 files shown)


cmd/mirror-main.go
```diff
@@ -303,17 +303,19 @@ func (mj *mirrorJob) doMirror(ctx context.Context, cancelMirror context.CancelFu
 		sURLs.TargetContent.Metadata["X-Amz-Storage-Class"] = mj.storageClass
 	}
 
-	// Set multiMasterETagKey for the target.
-	if sURLs.SourceContent.UserMetadata[multiMasterETagKey] != "" {
-		sURLs.TargetContent.Metadata[multiMasterETagKey] = sURLs.SourceContent.UserMetadata[multiMasterETagKey]
-	} else {
-		sURLs.TargetContent.Metadata[multiMasterETagKey] = sURLs.SourceContent.ETag
-	}
+	if mj.multiMasterEnable {
```
harshavardhana: perhaps its also `mj.multiMasterEnable || mj.preserve`

kannappanr: done

kannappanr: if the metadata is already set, it gets copied over to the destination by default. Currently, only if `multi-master` is enabled do we generate the `etag` and `stag` metadata on the destination.

PR #3039, opened 2020-01-10
fix: Add snowball support for `mc` with an ENV

harshavardhana:
Add this option to avoid calling APIs which are
not supported or incorrectly implemented in AWS
Snowball.
```
export MC_SNOWBALL_MODE=1
mc mirror snowball/mybucket myminio/mybucket
```


harshavardhana:
Will need to get @abperiasamy approval once.

harshavardhana:
will be doing this differently in `minio-go`

PR #2607, opened 2018-11-20
Ignore broken symlink durring mirror

Praveenrajmani:
The mirror action should fallback on errors and continue mirroring other non-infected objects rather failing on the first catch.

Fixes #2604

PR #3053, opened 2020-01-20
fix: handle symlinks by reading/following as needed

### LINKED ISSUE #3011: mc.exe fails with EvalSymlinks: too many links when using NTFS mounted folders
## Expected behavior

```
mc --config-dir $mc_cfg_dir stat D:\folder_x\folder_y\MNT_FOLDER\folder\subfolder1\subfolder2\image.jpg $mc_cfg_alias/$target_bucket/
Name      : image.jpg
Date      : 2018-01-15 08:35:35 CET 
Size      : 68 KiB 
Type      : file 
Metadata  :
  Content-Type: image/jpeg 
```

Would be the same for mc mirror, mc cp, mc ls etc.

## Actual behavior

```
mc_bin --config-dir $mc_cfg_dir stat D:\folder_x\folder_y\MNT_FOLDER\folder\subfolder1\subfolder2\image.jpg $mc_cfg_alias/$target_bucket/
mc.exe : mc.exe: <ERROR> Unable to stat `D:\folder_x\folder_y\MNT_FOLDER\folder\subfolder1\subfolder2\image.jpg`. EvalSymlinks: too many links.
```

```
mc --config-dir $mc_cfg_dir stat e:\folder\subfolder1\subfolder2\image.jpg $mc_cfg_alias/$target_bucket/
Name      : image.jpg
Date      : 2018-01-15 08:35:35 CET 
Size      : 68 KiB 
Type      : file 
Metadata  :
  Content-Type: image/jpeg 
```

Mount an NTFS formated drive into a NTFS sub-folder (MNT_FOLDER)
Assign a drive letter to the same drive (D:)

## mc version
RELEASE.2019-10-09T22-54-57Z

## System information
Windows Server 2016 Version 1607

harshavardhana:
Avoid using filepath.EvalSymlinks and let the caller
fail appropriately for symlinks.

Fixes #3011

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"io"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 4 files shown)


cmd/client-fs.go
```diff
@@ -765,33 +702,9 @@ func (f *fsClient) listInRoutine(contentCh chan<- *clientContent, isMetadata boo
 			if fi.Mode()&os.ModeSymlink == os.ModeSymlink {
 				fp := filepath.Join(fpath, fi.Name())
 				fi, e = os.Stat(fp)
-				if os.IsPermission(e) {
-					contentCh <- &clientContent{
-						Err: probe.NewError(PathInsufficientPermission{Path: pathURL.Path}),
-					}
-					continue
-				}
-				if os.IsNotExist(e) {
```
Praveenrajmani: We added this 
```
if os.IsNotExist(e) {
	// Lstat makes no attempt to follow the broken link.
	_, e = os.Lstat(fp)
	contentCh <- &clientContent{
		URL:  *newClientURL(fp),
		Size: -1,
		Err:  probe.NewError(e),
	}
	return nil
}
``` 
to ignore any broken symlink during mirroring/listing. else mirror would break in the middle if it sees any broken link while listing

cmd/client-fs.go
```diff
@@ -765,33 +702,9 @@ func (f *fsClient) listInRoutine(contentCh chan<- *clientContent, isMetadata boo
...
-							Err: probe.NewError(TooManyLevelsSymlink{
-								Path: pathURL.Path,
-							}),
-						}
-					} else {
-						contentCh <- &clientContent{
-							Err: probe.NewError(e),
-						}
+					contentCh <- &clientContent{
+						Err: f.toClientError(e, fp),
```
Praveenrajmani: Also, `toClientError` function handles only `PathInsufficientPermission` and `PathNotFound`. We should add cases for `TooManyLevelsSymlink`,`BrokenSymlink` etc.

harshavardhana: toClientError also handles TooManyLevelsSymlink - there is no error for BrokenSymlink @Praveenrajmani

cmd/client-fs.go
harshavardhana: That assumption is no longer true, because the mirror ignores all broken symlinks automatically - we simply have to keep continuously list.

Praveenrajmani: No @harshavardhana, mirroring will fail here in the first catch if it sees any broken link. Will not proceed further mirroring the remaining uninfected files. Is this ok?

Praveenrajmani: For eg,

```
touch 1{01..10}
ln -svf nonexisting 105
mc mb play/test
mc mirror . play/test
```

harshavardhana: This should be fixed,

harshavardhana:
@Praveenrajmani PTAL fixed the mirror issues with symlink

cmd/client-fs.go
```diff
@@ -765,34 +699,8 @@ func (f *fsClient) listInRoutine(contentCh chan<- *clientContent, isMetadata boo
...
-						Err: probe.NewError(PathInsufficientPermission{Path: pathURL.Path}),
-					}
-					continue
-				}
-				if os.IsNotExist(e) {
-					// Lstat makes no attempt to follow the broken link.
-					_, e = os.Lstat(fp)
-					contentCh <- &clientContent{
-						URL:  pathURL,
-						Size: -1,
```
Praveenrajmani: @harshavardhana Can you take a look at this comment here - https://github.com/minio/mc/pull/2607#issuecomment-445264783. The size was set to -1 because of this purpose of file tagging. To identify infected files and color code them. Do you think it is not needed?

cmd/client-fs.go
```diff
@@ -670,36 +632,8 @@ func (f *fsClient) listPrefixes(prefix string, contentCh chan<- *clientContent)
...
-					continue
-				}
-				if e != nil {
-					contentCh <- &clientContent{
-						Err: probe.NewError(e),
-					}
-					continue
-				}
+				// Ignore any errors on symlink
+				continue
```
Praveenrajmani: And also passing `clientContent.Err` will help in logging specific errors in `doList` - https://github.com/minio/mc/blob/master/cmd/ls.go#L117. This change will not log these errrors. is that ok?

harshavardhana: Yes it is okay, we don't need to log errors unless users can do something about it. Right now we happily ignore it anyways so we are wasting a call on os.Lstat() which is not meaningful. 

I say we ignore all broken symlinks because those are useless for us.

cmd/client-fs.go
harshavardhana: We don't actually do any of that, we used it as a way to log some warnings but its not useful in any manner. Every log should have a purpose where the log becomes actionable to the user. 

Our logging is kind of non-actionable we don't need to preserve symlinks because if they can be followed so be it we will copy, if we cannot we just ignore and continue. 

Preserving broken links is not supported so logging that is not useful.

Praveenrajmani: Ok. Agreed :+1:

harshavardhana:
PTAL @vadmeste

PR #3059, opened 2020-01-25
'mc admin heal' created bucket when the bucket doesn't exist

ebozduman:
Fixes the issue where a bucket is created when user runs:
`mc admin heal <alias>/<non-existing-bucket-name>`
I've introduced an extra check to see if bucket exists or not before moving forward with heal process. That check enables mc command to spit out an error message, letting user know `"mc: <ERROR>  Bucket <bucket-name> does not exist."`.

### DIFF
--- a/cmd/admin-heal.go
+++ b/cmd/admin-heal.go
@@ -202,6 +202,18 @@ func mainAdminHeal(ctx *cli.Context) error {
 	splits := splitStr(aliasedURL, "/", 3)
 	bucket, prefix := splits[1], splits[2]
 
+	clnt, err := newClient(aliasedURL)
+	if err != nil {
+		fatalIf(err.Trace(clnt.GetURL().String()), "Unable to create client for URL ", aliasedURL)
+		return nil
+	}
+	for content := range clnt.List(false, false, false, DirNone) {
+		if content.Err != nil {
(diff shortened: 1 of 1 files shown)


kannappanr:
```
$ mc admin heal myminio/bucket
mc: <ERROR> Unable to list folder. Bucket `bucket` does not exist.
```
Instead of `Unable to list folder` it should be something relevant to the operation being performed

cmd/admin-heal.go
```diff
@@ -202,6 +202,18 @@ func mainAdminHeal(ctx *cli.Context) error {
...
 	bucket, prefix := splits[1], splits[2]
 
+	clnt, err := newClient(aliasedURL)
+	if err != nil {
+		fatalIf(err.Trace(clnt.GetURL().String()), "Unable to create client for URL ", aliasedURL)
+		return nil
+	}
+	for content := range clnt.List(false, false, false, DirNone) {
+		if content.Err != nil {
+			fatalIf(content.Err.Trace(clnt.GetURL().String()), "Unable to list folder.")
```
kannappanr:
```suggestion
			fatalIf(content.Err.Trace(clnt.GetURL().String()), "Failed to heal bucket.")
```

ebozduman: ok

Issue #3064, opened 2020-02-01
mc mirror crashes on 32bit ARM system


rkoehn:
## Steps to reproduce the behavior
- Compile mc for arm6 32 bit using:
env GOOS=linux GOARCH=arm GOARM=6 GO111MODULE=on go build github.com/minio/mc

```
mc mirror server/bucket /tmp/localdir
 5.80 MiB / 5.80 MiB ┃ 2.19 MiB/spanic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x11e34]

goroutine 33 [running]:
runtime/internal/atomic.goXadd64(0x1df654c, 0x1, 0x0, 0x629d40, 0x1d97880)
	/usr/local/Cellar/go/1.13.3/libexec/src/runtime/internal/atomic/atomic_arm.go:103 +0x1c
github.com/minio/mc/cmd.(*ProgressStatus).AddCounts(0x1df6540, 0x1, 0x0)
	/Users/erweka/go/pkg/mod/github.com/minio/mc@v0.0.0-20200123103358-b70a79e39bbd/cmd/status.go:182 +0x34
github.com/minio/mc/cmd.(*mirrorJob).startMirror(0x1dc0750, 0x8e3050, 0x1df72c0, 0x1da5678, 0x1cf1fcc)
	/Users/erweka/go/pkg/mod/github.com/minio/mc@v0.0.0-20200123103358-b70a79e39bbd/cmd/mirror-main.go:599 +0x250
github.com/minio/mc/cmd.(*mirrorJob).mirror.func2(0x1e221f0, 0x1dc0750, 0x8e3050, 0x1df72c0, 0x1da5678)
	/Users/erweka/go/pkg/mod/github.com/minio/mc@v0.0.0-20200123103358-b70a79e39bbd/cmd/mirror-main.go:654 +0x84
created by github.com/minio/mc/cmd.(*mirrorJob).mirror
	/Users/erweka/go/pkg/mod/github.com/minio/mc@v0.0.0-20200123103358-b70a79e39bbd/cmd/mirror-main.go:647 +0x80
```

## mc --version
- RELEASE.2020-01-25T03-02-19Z

## System information
Raspberry 2+

## Analysis

This seems to happen because the atomic package has the following limitation on [32-bit platforms](https://golang.org/pkg/sync/atomic/#pkg-note-BUG):

[...]

PR #3065, opened 2020-02-01
Ensure 64bit alignment on 32bit builds for atomics

rkoehn:
Keep elements that are used with atomic.* functions as first element
of struct because it guarantees 64bit alignment on 32 bit machines.
atomic.* functions crash if operand is not aligned at 64bit.
See https://github.com/golang/go/issues/599

This should fix https://github.com/minio/mc/issues/3064.

### DIFF
--- a/cmd/parallel-manager.go
+++ b/cmd/parallel-manager.go
@@ -36,14 +36,17 @@ const (
 
 // ParallelManager - helps manage parallel workers to run tasks
 type ParallelManager struct {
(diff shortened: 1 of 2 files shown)


kannappanr:
please fix formatting as shown here https://github.com/minio/mc/pull/3065/checks?check_run_id=421111175

rkoehn:
Done. Took me some time to understand the format of the error message...

PR #3095, opened 2020-02-26
Add legalhold command to allow setting legal hold

poornas:
on object prefixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ sql       run sql queries on objects
 stat      stat contents of objects
 lock      set and get object lock configuration
 retention set object retention for objects with a given prefix
+legalhold set object legal hold for objects
 diff      list differences in object name, size, and date between buckets
 rm        remove objects
 event     manage object notifications
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -2065,6 +2065,19 @@ func (c *s3Client) PutObjectRetention(mode *minio.RetentionMode, retainUntilDate
 	return nil
 }
 
+// Set object legal hold for a given object.
+func (c *s3Client) PutObjectLegalHold(lhold *minio.LegalHoldStatus) *probe.Error {
+	bucket, object := c.url2BucketAndObject()
+	opts := minio.PutObjectLegalHoldOptions{
+		Status: lhold,
+	}
+	err := c.api.PutObjectLegalHold(bucket, object, opts)
+	if err != nil {
+		return probe.NewError(err)
+	}
+	return nil
+}
+
 // Get object lock configuration of bucket.
 func (c *s3Client) GetObjectLockConfig() (mode *minio.RetentionMode, validity *uint, unit *minio.ValidityUnit, perr *probe.Error) {
 	bucket, _ := c.url2BucketAndObject()
--- a/cmd/legalhold-main.go
+++ b/cmd/legalhold-main.go
@@ -1,0 +1,168 @@
+/*
+ * MinIO Client (C) 2020 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/fatih/color"
+	"github.com/minio/cli"
+	"github.com/minio/mc/pkg/probe"
+	minio "github.com/minio/minio-go/v6"
+	"github.com/minio/minio/pkg/console"
+)
+
+var (
+	lhFlags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "recursive, r",
+			Usage: "apply legal hold recursively",
+		},
+	}
+)
+var legalHoldCmd = cli.Command{
+	Name:   "legalhold",
+	Usage:  "set object legal hold for objects",
+	Action: mainLegalHold,
+	Before: setGlobalsFromContext,
+	Flags:  append(lhFlags, globalFlags...),
+	CustomHelpTemplate: `NAME:
+   {{.HelpName}} - {{.Usage}}
+ 
+ USAGE:
+   {{.HelpName}} [FLAGS] TARGET [ON | OFF]
+ 
+ FLAGS:
+   {{range .VisibleFlags}}{{.}}
+   {{end}}
+ 
+ EXAMPLES:
+	1. Enable object legal hold for objects in a given prefix
+	  $ {{.HelpName}} myminio/mybucket/prefix ON --recursive
+
+	2. Enable legal hold on a specific object
+	  $ {{.HelpName}} myminio/mybucket/prefix/obj.csv ON
+ `,
+}
+
+// Structured message depending on the type of console.
+type legalHoldCmdMessage struct {
+	LegalHold minio.LegalHoldStatus `json:"legalhold"`
+	URLPath   string                `json:"urlpath"`
+	Status    string                `json:"status"`
+	Err       error                 `json:"error,omitempty"`
+}
+
+// Colorized message for console printing.
+func (l legalHoldCmdMessage) String() string {
+	if l.Err != nil {
+		return console.Colorize("LegalHoldMessageFailure", "Cannot set object legal hold status `"+l.URLPath+"`. "+l.Err.Error())
+	}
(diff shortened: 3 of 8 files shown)


go.sum
```diff
@@ -203,9 +203,10 @@ github.com/minio/lsync v1.0.1/go.mod h1:tCFzfo0dlvdGl70IT4IAK/5Wtgb0/BrTmo/jE8pA
 github.com/minio/minio v0.0.0-20200118222113-b849fd7a756d h1:/i9HncNqZaHhRLwaReYT73wpU1XFiKYIfhwCQOrVcRg=
 github.com/minio/minio v0.0.0-20200118222113-b849fd7a756d/go.mod h1:f8MCueGjpibyBY0iVthLZPcrz310Dgwe/Knx+Dx/2+A=
 github.com/minio/minio-go/v6 v6.0.44/go.mod h1:qD0lajrGW49lKZLtXKtCB4X/qkMf0a5tBvN2PaZg7Gg=
+github.com/minio/minio-go/v6 v6.0.45-0.20200117140906-66cf57d21ba4 h1:zjzoKgiRJm7k0J+IXz36u+ltbbU/iRTBlbm5/vZFt1w=
```
harshavardhana: can you run `go mod tidy` ? @poornas

cmd/lock-main.go
```diff
@@ -141,10 +141,6 @@ func mainLock(ctx *cli.Context) error {
 	console.SetColor("Mode", color.New(color.FgCyan, color.Bold))
 	console.SetColor("Validity", color.New(color.FgYellow))
 
-	// Parse encryption keys per command.
-	_, err := getEncKeys(ctx)
-	fatalIf(err, "Unable to parse encryption keys.")
-
```
nitisht: now sure how is this related.

poornas: This is not related - however including in this PR as this code cleanup is so minor that it doesnt warrant a separate PR. Removing this piece of code because encryption keys are needed only if a Stat or Get operation is attempted on an SSE-C encrypted object. `mc lock` is at bucket level and doesn't require encryption keys.

kannappanr: There are no encryption keys involved with that command.

cmd/legalhold-main.go
```diff
@@ -0,0 +1,176 @@
...
+
+	// Quit early if urlStr does not point to an S3 server
+	switch clnt.(type) {
+	case *fsClient:
+		fatal(errDummy().Trace(), "Legal hold not supported for file system.")
+	}
+
+	alias, _, _ := mustExpandAlias(urlStr)
+	if !isRecursive {
+		sse := getSSE(urlStr, encKeyDB[alias])
```
kannappanr: Why do we need `sse` to set `legalhold` ?

cmd/legalhold-main.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// main for retention command.
+func mainLegalHold(ctx *cli.Context) error {
+	console.SetColor("LegalHoldSuccess", color.New(color.FgGreen, color.Bold))
+	console.SetColor("LegalHoldPartialFailure", color.New(color.FgRed, color.Bold))
+	console.SetColor("LegalHoldMessageFailure", color.New(color.FgYellow))
+	args := ctx.Args()
+	// Parse encryption keys per command.
+	encKeyDB, err := getEncKeys(ctx)
```
kannappanr: We shouldn't require encryption keys

cmd/legalhold-main.go
poornas: removed - stat doesn't add much value, so letting list fail if recursive flag is not passed

cmd/legalhold-main.go
poornas: removed

README.md
```diff
@@ -18,6 +18,7 @@ sql       run sql queries on objects
 stat      stat contents of objects
 lock      set and get object lock configuration
 retention set object retention for objects with a given prefix
+legalhold set object legal hold for objects with a given prefix
```
kannappanr: should we remove `with a given prefix` just like we have for `cp`

cmd/legalhold-main.go
```diff
@@ -0,0 +1,166 @@
...
+	  $ {{.HelpName}} myminio/mybucket/prefix/obj.csv ON
+ `,
+}
+
+// Structured message depending on the type of console.
+type legalHoldCmdMessage struct {
+	LegalHold minio.LegalHoldStatus `json:"legalhold"`
+	URLPath   string                `json:"urlpath"`
+	Status    string                `json:"status"`
+	Err       error                 `json:"error"`
```
kannappanr: we should add `omit empty` for error

poornas: done

cmd/client-s3.go
```diff
@@ -2089,6 +2089,19 @@ func (c *s3Client) PutObjectRetention(mode *minio.RetentionMode, retainUntilDate
 	return nil
 }
 
+// Set object legal hold for a given object.
+func (c *s3Client) PutObjectLegalHold(lhold *minio.LegalHoldStatus) *probe.Error {
+	bucket, object := c.url2BucketAndObject()
```
kannappanr: should we check for either bucketname or objectname being empty?

poornas: it is already being done in minio-go, it is not necessary to duplicate this check here

kannappanr:
```
$ mc legalhold play/test/issue ON
Cannot set object legal hold status `/test/issue`.Bucket is missing ObjectLockConfiguration
Errors found while setting legal hold status on objects with prefix `play/test/issue`. 
```
Need space between `.` and `Bucket`. Also, for a regular file, we do not have to print another line.

kannappanr:
```
$ mc legalhold play/test/issue ON
Cannot set object legal hold status `/test/issue`. Bucket is missing ObjectLockConfiguration
Errors found while setting legal hold status on objects with prefix `play/test/issue`. 
```
The second line in the above message should be removed, if it is just a single object. Also, the text should start with `mc: <ERROR>` as is the case with the following example.
```
 $ mc rm play/test/issue1
mc: <ERROR> Failed to remove `play/test/issue1`. Target object is not found 

```

kannappanr:
```
$ mc legalhold play/test1/issue on
Object legal hold successfully set for prefix `play/test1/issue`.

```
This should not say `prefix` in the message

kannappanr:
```
$ mc legalhold play/test1/iss off 
Object legal hold successfully set for prefix `play/test1/iss`.

```
object `iss` doesn't exist, but an object `issue` exists, this seems to do `recursive`, always

kannappanr:
```
 $ mc legalhold play/test1/iss1 off 
Object legal hold successfully set for prefix `play/test1/iss1`.

```
Object `iss1` doesn't exist and there is also no prefix

kannappanr:
```
 $ mc legalhold play/test/iss1 offsasd 
mc: <ERROR> invalid legal hold status '2': invalid argument.
```
What does `2` mean here?

cmd/legalhold-main.go
```diff
@@ -0,0 +1,166 @@
...
+	args := ctx.Args()
+
+	var urlStr string
+	var lhold *minio.LegalHoldStatus
+	switch l := len(args); l {
+	case 2:
+		urlStr = args[0]
+		h := minio.LegalHoldStatus(strings.ToUpper(args[1]))
+		if !h.IsValid() {
+			fatalIf(probe.NewError(errors.New("invalid argument")), "invalid legal hold status '%v'", l)
```
kannappanr:
```suggestion
			fatalIf(probe.NewError(errors.New("invalid argument")), "invalid legal hold status '%v'", h)
```

cmd/legalhold-main.go
```diff
@@ -0,0 +1,166 @@
...
+	// Quit early if urlStr does not point to an S3 server
+	switch clnt.(type) {
+	case *fsClient:
+		fatal(errDummy().Trace(), "Legal hold not supported for file system.")
+	}
+
+	alias, _, _ := mustExpandAlias(urlStr)
+	var cErr error
+	errorsFound := false
+	for content := range clnt.List(isRecursive, false, false, DirNone) {
```
kannappanr: if `isRecursive` is false, we can simply call `PutObjectLegalHold` function here. Also, setLegalHold can be implemented as part of `client` interface, that way we do not have to check if the client type is fs or s3.

cmd/legalhold-main.go
```diff
@@ -0,0 +1,166 @@
...
+
+		newClnt, perr := newClientFromAlias(alias, content.URL.String())
+		if perr != nil {
+			errorIf(content.Err.Trace(clnt.GetURL().String()), "Invalid URL")
+			continue
+		}
+		probeErr := newClnt.PutObjectLegalHold(lhold)
+		if probeErr != nil {
+			errorsFound = true
+			printMsg(legalHoldCmdMessage{
```
kannappanr: should be `ErrorIf`

cmd/legalhold-main.go
poornas: fixed

cmd/legalhold-main.go
poornas: fixed

kannappanr: still see `printMsg`

PR #3097, opened 2020-02-26
mc ilm feature. list, add, remove, export, import.

### LINKED ISSUE #2879: Implement mc ILM cmd
Suggest and implementing `mc ilm` command with expiry feature

BigUstad:
Fixes #2879


cmd/ilm/ilm-util.go
```diff
@@ -0,0 +1,418 @@
...
+	}
+	output := strings.Repeat(toPadWith, leftRptLth) + label + strings.Repeat(toPadWith, rightRptLth)
+	return output
+}
+
+// splitStr splits a string into n parts, empty strings are added
+// if we are not able to reach n elements
+func splitStr(path, sep string, n int) []string {
+	splits := strings.SplitN(path, sep, n)
+	// Add empty strings if we found elements less than nr
```
kannappanr: Comment needs to be cleaned up

gwperry:
Checking to see if there was a timetable associated with this update.  We could really use a Lifecycle feature in the minio client.  Thanks!

cmd/ilm/ilm.go
```diff
@@ -0,0 +1,115 @@
...
+	Rules   []LifecycleRule `xml:"Rule"`
+}
+
+// JSON jsonified lifecycle configuration.
+func (l LifecycleConfiguration) JSON() string {
+	var ilmJSONbytes []byte
+	var err error
+	ilmJSONbytes, err = stdlibjson.MarshalIndent(l, "", "    ")
+	if err != nil {
+		console.Errorln("Failed to get lifecycle configuration in JSON format.")
```
kannappanr, line 111: why is this not calling say `fatalif`?

BigUstad: fatalif would create a circular dependency. Including cmd/ilm/ in cmd/ and vice-versa created problems.

kannappanr: then this should probably return an error

cmd/ilm-add.go
```diff
@@ -0,0 +1,165 @@
...
+	cli.StringFlag{
+		Name:  "transition-days",
+		Usage: "the number of days to transition",
+	},
+	cli.StringFlag{
+		Name:  "storage-class",
+		Usage: "storage class for transition (STANDARD_IA, ONEZONE_IA, GLACIER. Etc)",
+	},
+	cli.BoolFlag{
+		Name:  "disabled",
```
kannappanr, line 95:
```suggestion
		Name:  "disable",
```

cmd/ilm-list.go
```diff
@@ -0,0 +1,285 @@
...
+		// Build & print the line row
+		lineRow := buildILMTableLineRow(rowArr...)
+		row = tbl.buildRow(lineRow...)
+		row = console.Colorize(ilmThemeRow, row)
+		rows += row + "\n"
+	}
+	return rows
+}
+
+func mainILMList(ctx *cli.Context) error {
```
krisis, line 212: I noticed the following while testing this PR locally. Notice that the tags column in the tabular output seems truncated.
```➜  mc git:(ilm-table-display-2) ✗ ./mc ilm list --json play/kp-ilm
{
 "status": "success",
 "target": "play/kp-ilm",
 "ilm": {
  "Rules": [
   {
    "Expiration": {
     "Days": 200
    },
    "ID": "Devices",
    "Filter": {
     "And": {
      "Tags": [
       {
        "Key": "docformat",
        "Value": "docx"
       },
       {
        "Key": "plaintextformat",
        "Value": "txt"
       },
       {
        "Key": "exportFormat",
        "Value": "pdf"
       }
      ]
     }
    },
    "Status": "Enabled"
   }
  ]
 }
}
➜  mc git:(ilm-table-display-2) ✗ ./mc ilm list play/kp-ilm       
     ID     |  Prefix  |  Enabled   | Expiry |  Date/Days   |  Transition  |    Date/Days     |  Storage-Class   |       Tags       
------------|----------|------------|--------|--------------|--------------|------------------|------------------|------------------
 Devices    |          |    ✓       |  ✓     |  200 day(s)  |     ✗        |                  |                  | docformat:docx   
            |          |            |        |              |              |                  |                  | plaintextform..  
            |          |            |        |              |              |                  |                  | exportFormat:..  
------------|----------|------------|--------|--------------|--------------|------------------|------------------|------------------
```

BigUstad: Yes that is by design. We want to restrict tags column to a certain width. The _..._ will indicate to the user that the tags are clipped/trimmed. The way PrettyTable is designed, we need to keep the column widths constant. Shall we change it when we move to console Table?

cmd/ilm-list.go
```diff
@@ -0,0 +1,285 @@
...
+		Name:  "expiry",
+		Usage: "show only expiration fields",
+	},
+	cli.BoolFlag{
+		Name:  "transition",
+		Usage: "show only transition fields",
+	},
+	cli.BoolFlag{
+		Name:  "minimum",
+		Usage: "show minimum fields",
```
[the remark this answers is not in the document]
BigUstad: id, prefix, Enabled status, Expiry/Transition(set or not). Fields id, Enabled/Disabled are mandatory. Prefix is also a very important identifier.
I can list them in Usage..

cmd/ilm-import.go
```diff
@@ -0,0 +1,116 @@
...
+
+// ReadILMConfigJSON read from stdin and set ILM configuration to bucket
+func readILMConfigJSON() (string, *probe.Error) {
+	// User is expected to enter the lifecycleConfiguration instance contents in JSON format
+	var ilmInfo ilm.LifecycleConfiguration
+	var bytes []byte
+	var err error
+	// Consume json from STDIN
+	dec := json.NewDecoder(os.Stdin)
+	if err = dec.Decode(&ilmInfo); err != nil && err != io.EOF {
```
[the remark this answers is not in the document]
BigUstad: `err == io.EOF` implies that the entry (JSON input) has completed. It is encountered both in stdin input if typing or in a file. So, I further check the number of rules to check for an empty configuration.

PR #3099, opened 2020-02-27
add tag command

BigUstad:
add tag command
    mc tag show <object>
    mc tag set <object> --tags <key1:value1> --tags <key2:value2>...

cmd/admin-kms-key-status.go UpdateErr had been removed from (pkg/madmin) KMSKeyStatus in a previous merge of minio.

workaround for undefined compiler error


cmd/admin-kms-key-status.go
```diff
@@ -86,6 +86,6 @@ func (s kmsKeyStatusMsg) String() string {
 }
 
 func (s kmsKeyStatusMsg) JSON() string {
-	const fmtStr = `{"key-id":"%s","encryption-error":"%s","update-error":"%s","decryption-error":"%s"}`
-	return fmt.Sprintf(fmtStr, s.KeyID, s.EncryptionErr, s.UpdateErr, s.DecryptionErr)
+	const fmtStr = `{"key-id":"%s","encryption-error":"%s","decryption-error":"%s"}`
```
kannappanr: Why is this change here in this PR?

kannappanr: Please remove this change

cmd/tag-main.go
```diff
@@ -0,0 +1,66 @@
...
+	tagResultsSuccess string = "Result-Success"
+	tagResultsFailure string = "Result-Failure"
+)
+
+func checkMainTagSyntax(ctx *cli.Context) {
+	cli.ShowCommandHelp(ctx, "")
+}
+
+// getS3Client - To make api calls.
+func getS3ClientForTagOps(urlStr string) s3Client {
```
kannappanr: There should probably be a standard function to get this.

cmd/tag-main.go
```diff
@@ -0,0 +1,66 @@
...
+
+package cmd
+
+import (
+	"github.com/minio/cli"
+)
+
+var tagCmd = cli.Command{
+	Name:   "tag",
+	Usage:  "configure tags for objects",
```
kannappanr: This text needs to reflect that we can also display the tags

cmd/tag-show.go
```diff
@@ -0,0 +1,186 @@
...
+EXAMPLES:
+  1. Show the tags assigned to the object named testobject in the bucket testbucket on alias s3.
+     {{.Prompt}} {{.HelpName}} s3/testbucket/testobject
+
+`,
+}
+
+// tagMessage container for displaying tag
+type tagMessage struct {
+	AliasName string `json:"aliasName"`
```
kannappanr: Why do you need to display `alias`, `host`

cmd/tag-show.go
```diff
@@ -0,0 +1,186 @@
...
+	Host      string `json:"host"`
+	Tags      string `json:"tags"`
```
kannappanr: tags should be a key value array, not a string

cmd/tag-show.go
```diff
@@ -0,0 +1,186 @@
...
+	if globalJSON {
+		var tMsg tagMessage
+		tMsg = parseTagMessage(tagObj.String(), objectURL)
+		console.Println("\n" + tMsg.JSON())
+		return nil
+	}
+
+	switch len(tagObj.TagSet.Tags) {
+	case 0:
+		console.Infoln("No tags set.")
```
kannappanr: This text needs to be improved

cmd/tag-show.go
```diff
@@ -0,0 +1,186 @@
...
+
+	s3c := getS3ClientForTagOps(urlStr)
+	bucketName, objectName := s3c.url2BucketAndObject()
+	tagXML, err := s3c.api.GetObjectTagging(bucketName, objectName)
+	if err != nil {
+		return tagging.Tagging{}, err
+	}
+	var tagObj tagging.Tagging
+	if err = xml.Unmarshal([]byte(tagXML), &tagObj); err != nil {
+		console.Errorln(err.Error() + ", Unable to set Object Tags for display.")
```
kannappanr: Why is there a text about `set tags`

cmd/tag-show.go
```diff
@@ -0,0 +1,186 @@
...
+		return nil
+	}
+
+	switch len(tagObj.TagSet.Tags) {
+	case 0:
+		console.Infoln("No tags set.")
+	case 1:
+		showTagObjectName(objectURL)
+		showTagInfoFieldOne(tagObj.TagSet.Tags[0].Key, tagObj.TagSet.Tags[0].Value)
+	default:
```
kannappanr: why is there a difference between displaying 1 tag and multiple tags

cmd/tag-set.go
```diff
@@ -0,0 +1,113 @@
...
+		cli.ShowCommandHelp(ctx, "set")
+		os.Exit(globalErrorExitStatus)
+	}
+}
+
+func getTagMap(ctx *cli.Context) (map[string]string, error) {
+	ilmTagKVMap := make(map[string]string)
+	tagValues := ctx.StringSlice(strings.ToLower("tags"))
+	for tagIdx, tag := range tagValues {
+		key := splitStr(tag, ":", 2)[0]
```
kannappanr: unable to just set `key` without `value` 
```
mc tag set s3/bucket/issue --tags "key:value" --tags "key1"
mc: <ERROR> error extracting tag argument(#2) key1. Unable to get tags to set.
```

cmd/admin-kms-key-status.go
BigUstad: Compilation will fail without this change. I can remove it once a changeset is merged that has removed instances of UpdateErr in mc. UpdateErr has been removed in minio codebase, which is what is causing the compilation error in mc.

cmd/tag-show.go
BigUstad: I will remove it.

cmd/tag-set.go
BigUstad: Change made, for the sake of uniformity key with empty value is to be entered as --tags "key:"

cmd/admin-kms-key-status.go
kannappanr: Still don't understand the need for this change. If this were the case it should fail even now without any of your other changes, as I do not see this PR using the latest minio package in go.mod

kannappanr: You should also update minio package in go.mod, then this change will make sense

cmd/tag-main.go
```diff
@@ -0,0 +1,51 @@
...
+	tagResultsFailure string = "Result-Failure"
+)
+
+func checkMainTagSyntax(ctx *cli.Context) {
+	cli.ShowCommandHelp(ctx, "")
+}
+
+func mainTag(ctx *cli.Context) error {
+	checkMainTagSyntax(ctx)
+
```
kannappanr:
```suggestion
```

PR #3100, opened 2020-02-28
mc retention: add --recursive and --bypass flags.

poornas:
--bypass flag allows overriding governance if user
has override permissions


cmd/retention-main.go
```diff
@@ -93,6 +109,15 @@ func setRetention(urlStr string, mode *minio.RetentionMode, validity *uint, unit
 	}
 
 	alias, _, _ := mustExpandAlias(urlStr)
+	if !isRecursive {
```
kannappanr: should we skip this check here and instead let the server respond with an object does not exist error? This would eliminate the need to have encryption key

poornas: yes, fixed this

cmd/common-methods.go
```diff
@@ -234,7 +234,7 @@ func putTargetRetention(ctx context.Context, alias string, urlStr string, metada
 			retainUntilDate = t.UTC()
 		}
 	}
-	if err := targetClnt.PutObjectRetention(&lockMode, &retainUntilDate); err != nil {
+	if err := targetClnt.PutObjectRetention(&lockMode, &retainUntilDate, false); err != nil {
```
kannappanr: Why are we forcing `bypassgovernance` to false here?

poornas: this is for cp/mirror functionality and overriding retention dates is not the focus here - so default option of bypassgovernance false is sufficient.

kannappanr: ok. We can fix this if we see any issues here.

Issue #3073, opened 2020-02-08
mc ls no longer work with GCS since 2019-12-17T23-26-28Z


wsxiaoys:
## Expected behavior
mc ls google/... worked.

## Actual behavior
mc: <ERROR> Unable to list folder. A header or query you provided requested a function that is not implemented.
 (1) ls.go:132 cmd.doList(..) Tags: [https://storage.googleapis.com/ps-gcs/]
 (0) client-s3.go:1906 cmd.(*s3Client).listInRoutine(..)
 Commit:b70a79e39bbd | Release-Tag:RELEASE.2020-01-25T03-02-19Z | Host:jupyter-meng | OS:linux | Arch:amd64 | Lang:go1.13.6 | Mem:2.1 MB/72 MB | Heap:2.1 MB/66 MB

## mc --version
mc version RELEASE.2020-01-25T03-02-19Z

kannappanr:
@wsxiaoys Thanks for reporting the issue, we will take a look and get back to you.

kannappanr:
@wsxiaoys It is working as expected with this release. Can you please verify that `mc config host list gcs` prints the accesskey and secret key that you expect. Most probably those creds are not part of the alias. Based on the executable name, mc creates a folder in the home folder and each binary name will have it's own config.json file. You might have to add those creds to that alias again.

lsde:
I can confirm this bug, which appeared in RELEASE.2019-12-17T23-26-28Z.
Since this version, accessing data on GCS is impossible (upload is possible though).
Just try it for yourself:
(both releases have their respective configs with the same access credentials)

**RELEASE.2019-10-09T22-54-57Z**
```
$  ./mc.RELEASE.2019-10-09T22-54-57Z ls --json gcs/test1582795099
{
 "status": "success",
 "type": "file",
 "lastModified": "2020-02-27T10:19:12.098+01:00",
 "size": 9,
 "key": "testfile",
 "etag": "blabla"
}
```

**RELEASE.2019-12-17T23-26-28Z**
```
$ ./mc.RELEASE.2019-12-17T23-26-28Z ls --json gcs/test1582795099
{
 "status": "error",
 "error": {
  "message": "Unable to list folder.",
  "cause": {
   "message": "A header or query you provided requested a function that is not implemented.",
   "error": {
    "Code": "NotImplemented",
    "Message": "A header or query you provided requested a function that is not implemented.",
    "BucketName": "",
    "Key": "",
    "RequestID": "",
[...]

PR #3101, opened 2020-03-03
Use listObjectsV1 for GCS S3 implementation

harshavardhana:


harshavardhana:
possibly a potential fix for #3073

PR #3111, opened 2020-03-11
heal: Show scheduled next healing 

vadmeste:
mc admin heal alias command will show the next healing time.

Also show 'Never completed' when the server returns zero last healing activity

### DIFF
--- a/cmd/admin-heal.go
+++ b/cmd/admin-heal.go
@@ -150,11 +150,24 @@ type backgroundHealStatusMessage struct {
 
 // String colorized to show background heal status message.
 func (s backgroundHealStatusMessage) String() string {
+	dot := console.Colorize("Dot", " ●  ")
+
 	healPrettyMsg := console.Colorize("HealBackgroundTitle", "Background healing status:\n")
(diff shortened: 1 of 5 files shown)


cmd/admin-heal.go
```diff
@@ -153,8 +153,12 @@ func (s backgroundHealStatusMessage) String() string {
 	healPrettyMsg := console.Colorize("HealBackgroundTitle", "Background healing status:\n")
 	healPrettyMsg += fmt.Sprintf("  Total items scanned: %s\n",
 		console.Colorize("HealBackground", s.HealInfo.ScannedItemsCount))
-	healPrettyMsg += fmt.Sprintf("  Last background heal check: %s\n",
-		console.Colorize("HealBackground", timeDurationToHumanizedDuration(time.Since(s.HealInfo.LastHealActivity)).String()+" ago"))
+	lastHealingTime := "Never"
```
harshavardhana: It is better to show when it is anticipated to run too..

vadmeste: > It is better to show when it is anticipated to run too..

Done

vadmeste:
@harshavardhana done

PR #3112, opened 2020-03-11
info: Fix showing offline nodes in the pretty output

vadmeste:
mc admin info was not properly showing the offline status of an offline node.

This commit fixes it.

### DIFF
--- a/cmd/admin-info.go
+++ b/cmd/admin-info.go
@@ -61,50 +61,65 @@ type clusterStruct struct {
 	Info   madmin.InfoMessage `json:"info,omitempty"`
 }
 
-// String provides colorized info messages.
+// String provides colorized info messages depending on the type of a server
+//        FS server                          non-FS server
+// ==============================  ===================================
+// ● <ip>:<port>                   ● <ip>:<port>
+//   Uptime: xxx                     Uptime: xxx
+//   Version: xxx                    Version: xxx
+//   Network: X/Y OK                 Network: X/Y OK
+//
+// U Used, B Buckets, O Objects    Drives: N/N OK
+//
+//                                   U Used, B Buckets, O Objects
+//                                   N drives online, K drives offline
+//
 func (u clusterStruct) String() (msg string) {
+	// Check cluster level "Status" field for error
+	if u.Status == "error" {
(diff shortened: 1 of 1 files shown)


cmd/admin-info.go
```diff
@@ -61,50 +61,65 @@ type clusterStruct struct {
...
 	// Dot represents server status, online (green) or offline (red)
 	dot := "●"
 	// Color palette initialization
 	console.SetColor("Info", color.New(color.FgGreen, color.Bold))
 	console.SetColor("InfoFail", color.New(color.FgRed, color.Bold))
+
 	// MinIO server type default
 	backendType := "Unknown"
+	// Set the type of MinIO server ("FS", "Erasure", "Unknown")
+	v := reflect.ValueOf(u.Info.Backend)
```
harshavardhana: Backend is a proper type not sure why we need a type inference.. @vadmeste

vadmeste: > Backend is a proper type not sure why we need a type inference.. @vadmeste

Recent changes removed backend type field in admin info strucutre and allocated a new backend data with interface type

```
// InfoMessage container to hold server admin related information.
type InfoMessage struct {
        Mode         string             `json:"mode,omitempty"`
        Domain       []string           `json:"domain,omitempty"`
        Region       string             `json:"region,omitempty"`
        SQSARN       []string           `json:"sqsARN,omitempty"`
        DeploymentID string             `json:"deploymentID,omitempty"`
        Buckets      Buckets            `json:"buckets,omitempty"`
        Objects      Objects            `json:"objects,omitempty"`
        Usage        Usage              `json:"usage,omitempty"`
        Services     Services           `json:"services,omitempty"`
        Backend      interface{}        `json:"backend,omitempty"`                              Servers      []ServerProperties `json:"servers,omitempty"`
}
```

harshavardhana: We should keep the types @vadmeste can you fix the server first?

vadmeste:
@harshavardhana if we are going to keep the old API, let's merge this first.. since mc admin info is wrongly showing offline nodes in the output.

harshavardhana:
> @harshavardhana if we are going to keep the old API, let's merge this first.. since mc admin info is wrongly showing offline nodes in the output.

Yes agreed lets move forward @vadmeste

PR #3114, opened 2020-03-12
Onboard Diagnostics

wlan0:
This is an accompanying PR to https://github.com/minio/minio/pull/9024. This should be merged only after the server side is merged

[asciicast](https://asciinema.org/a/309756)


kannappanr:
@wlan0 Unable to compile the PR

```
$ make install
Checking dependencies
Building mc binary to './mc'
go: downloading github.com/mattn/go-isatty v0.0.8
go: downloading github.com/mattn/go-colorable v0.1.4
go: extracting github.com/mattn/go-colorable v0.1.4
go: extracting github.com/mattn/go-isatty v0.0.8
go: finding github.com/mattn/go-isatty v0.0.8
# github.com/minio/mc/cmd
cmd/admin-config-export.go:88:28: not enough arguments in call to client.GetConfig
	have ()
	want (context.Context)
cmd/admin-config-get.go:99:31: not enough arguments in call to client.HelpConfigKV
	have (string, string, bool)
	want (context.Context, string, string, bool)
cmd/admin-config-get.go:112:30: not enough arguments in call to client.GetConfigKV
	have (string)
	want (context.Context, string)
cmd/admin-config-history.go:130:53: not enough arguments in call to client.ClearConfigHistoryKV
	have (string)
	want (context.Context, string)
cmd/admin-config-history.go:137:44: not enough arguments in call to client.ListConfigHistoryKV
	have (int)
	want (context.Context, int)
cmd/admin-config-import.go:99:41: not enough arguments in call to client.SetConfig
	have (*os.File)
	want (context.Context, io.Reader)
cmd/admin-config-reset.go:110:31: not enough arguments in call to client.HelpConfigKV
	have (string, string, bool)
	want (context.Context, string, string, bool)
cmd/admin-config-reset.go:124:43: not enough arguments in call to client.DelConfigKV
	have (string)
	want (context.Context, string)
cmd/admin-config-restore.go:97:54: not enough arguments in call to client.RestoreConfigHistoryKV
	have (string)
	want (context.Context, string)
cmd/admin-config-set.go:108:31: not enough arguments in call to client.HelpConfigKV
	have (string, string, bool)
	want (context.Context, string, string, bool)
cmd/admin-config-set.go:108:31: too many errors
Makefile:75: recipe for target 'build' failed
make: *** [build] Error 2

```

wlan0:
@kannappanr you will only be able to do it after the minio side of it gets merged

kannappanr:
which package should I use `madmin`?

vadmeste:
> @kannappanr you will only be able to do it after the minio side of it gets merged

@wlan0, this should be already done, you only have to rebase.

cmd/client-s3.go
```diff
@@ -20,7 +20,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
-	"encoding/json"
+	json "github.com/minio/mc/pkg/colorjson"
```
klauspost: Not a big fan of these blanket replacements.

1) Output is not JSON and should only be used for printing.
2) Unmarshal has horrible error reporting. Even worse than stdlib.

Could you instead use this as `cjson "github.com/minio/mc/pkg/colorjson"` and only replace where it makes sense?

kannappanr:
PTAL at build failure

cmd/admin-config-import.go
```diff
@@ -17,7 +17,7 @@
 package cmd
 
 import (
-	"encoding/json"
+	json "github.com/minio/mc/pkg/colorjson"
```
klauspost: Move import.

go.mod
```diff
@@ -26,10 +27,11 @@ require (
...
 	golang.org/x/net v0.0.0-20200202094626-16171245cfb2
-	golang.org/x/sys v0.0.0-20200212091648-12a6c2dcc1e4 // indirect
 	golang.org/x/text v0.3.2
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127
 	gopkg.in/h2non/filetype.v1 v1.0.5
 	gopkg.in/ini.v1 v1.52.0 // indirect
 	gopkg.in/yaml.v2 v2.2.4
 )
+
+replace github.com/minio/minio => ../minio
```
klauspost: This can be removed now.

wlan0:
@kannappanr @poornas @klauspost PTAL  -I've addressed your comments

poornas:
@wlan0 , --json mode can just print json on the screen instead of the progress display and saving to .gz file. Can that be changed in this PR?

wlan0:
@poornas I fixed it already - it doesnt print progress and does not save to a file

poornas:
> @poornas I fixed it already - it doesnt print progress and does not save to a file

thanks @wlan0, I see it now.

cmd/admin-obd.go
```diff
@@ -0,0 +1,399 @@
...
+	args := ctx.Args()
+	aliasedURL := args.Get(0)
+
+	// Create a new MinIO Admin Client
+	client, err := newAdminClient(aliasedURL)
+	fatalIf(err, "Unable to initialize admin connection.")
+
+	types := &options
+
+	dot := "●"
```
harshavardhana: I see that we have repeated definitions of this constant can we make it global and use it everywhere? 

```
git grep 'dot :'
cmd/admin-heal.go:      dot := console.Colorize("Dot", " ●  ")
cmd/admin-info.go:      dot := "●"
```

cmd/admin-obd.go
```diff
@@ -0,0 +1,399 @@
...
+
+	// Get the alias parameter from cli
+	args := ctx.Args()
+	aliasedURL := args.Get(0)
+
+	// Create a new MinIO Admin Client
+	client, err := newAdminClient(aliasedURL)
+	fatalIf(err, "Unable to initialize admin connection.")
+
+	types := &options
```
harshavardhana: why do we need access to `options` pointer? and then dereference it later? - can we avoid doing this? if you are trying to copy then have a method for `options.Clone()` 

```
	obdChan := client.ServerOBDInfo(cont, *types)
```

cmd/admin-obd.go
```diff
@@ -0,0 +1,399 @@
...
+	"github.com/minio/minio/pkg/console"
+	"github.com/minio/minio/pkg/madmin"
+)
+
+var adminOBDFlags = []cli.Flag{
+	OBDDataTypeFlag{
+		Name:   "tests",
+		Usage:  "choose OBD tests to run [" + options.String() + "]",
+		Value:  nil,
+		EnvVar: "MINIO_OBD_DATA",
```
harshavardhana: For `mc` we need to prefix MC_ not MINIO_ @wlan0 - is this confirmed with @abperiasamy ?

cmd/admin-obd.go
```diff
@@ -0,0 +1,399 @@
...
+	check := "✔"
+
+	// Color palette initialization
+	console.SetColor("INFO", color.New(color.FgGreen, color.Bold))
+	console.SetColor("WARN", color.New(color.FgRed, color.Bold))
+	console.SetColor("GREEN", color.New(color.FgGreen))
+
+	spinners := []string{"/", "|", "\\", "--", "|"}
+
+	cont, cancel := context.WithCancel(context.Background())
```
harshavardhana: you can use context.WithCancel(globalContext) now

cmd/client-admin.go
```diff
@@ -79,11 +79,11 @@ func newAdminFactory() func(config *Config) (*madmin.AdminClient, *probe.Error)
...
 				Proxy: http.ProxyFromEnvironment,
 				DialContext: (&net.Dialer{
-					Timeout:   30 * time.Second,
-					KeepAlive: 30 * time.Second,
+					Timeout:   60 * time.Second,
+					KeepAlive: 60 * time.Second,
 				}).DialContext,
 				MaxIdleConns:          100,
-				IdleConnTimeout:       90 * time.Second,
+				IdleConnTimeout:       300 * time.Second,
```
harshavardhana: increasing this is not needed @wlan0 is there a good reason to increase this value? I don't think this should adversely affect your implementation because the server will always send data anyways and it won't trigger this.

cmd/client-admin.go
```diff
@@ -79,11 +79,11 @@ func newAdminFactory() func(config *Config) (*madmin.AdminClient, *probe.Error)
 			var transport http.RoundTripper = &http.Transport{
 				Proxy: http.ProxyFromEnvironment,
 				DialContext: (&net.Dialer{
-					Timeout:   30 * time.Second,
-					KeepAlive: 30 * time.Second,
+					Timeout:   60 * time.Second,
```
harshavardhana: we want to keep the dial timeout for 30secs to quickly know if the host is reachable, 60secs is too high IMO @wlan0

cmd/client-admin.go
```diff
@@ -79,11 +79,11 @@ func newAdminFactory() func(config *Config) (*madmin.AdminClient, *probe.Error)
...
+					KeepAlive: 60 * time.Second,
```
harshavardhana: increasing keep alive is also not needed since 30secs means that its `keepAlive` probe timeout which is usually 30 * 9 (net.ipv4.tcp_keepalive_probes)  i.e almost 4.5minutes not sure if we need to keep alive more than that.

cmd/admin-obd.go
```diff
@@ -0,0 +1,399 @@
...
+func checkAdminOBDSyntax(ctx *cli.Context) {
+	if len(ctx.Args()) == 0 || len(ctx.Args()) > 1 {
+		cli.ShowCommandHelpAndExit(ctx, "obd", 1) // last argument is exit code
+	}
+}
+
+//compress and tar obd output
+func tarGZ(c clusterOBDStruct, alias string) error {
+	filename := fmt.Sprintf("%s-obd_%s.json.gz", alias, time.Now().Format("20060102150405"))
+	f, err := os.OpenFile(filename, os.O_CREATE|os.O_RDWR, 0644)
```
harshavardhana: use `0666` instead to honor user umask such that the file is created with default user permissions instead of 0644

cmd/admin-config-export.go
```diff
@@ -89,7 +90,7 @@ func mainAdminConfigExport(ctx *cli.Context) error {
 
 	// Print
 	printMsg(configExportMessage{
-		Value: buf,
+		Value: fmt.Sprintf("%s", buf),
```
harshavardhana: The original reason it's not converted to a string is to ensure that the representation of the `export` command doesn't need to be jsonified and base64 representation keeps the entire file in-tact including its `\n` for further consumption

```
~ echo "cmVnaW9uIG5hbWU9dXMtd2VzdC0xIApzdG9yYWdlX2NsYXNzIHN0YW5kYXJkPSBycnM9RUM6MiAKIyBjYWNoZSBkcml2ZXM9IGV4Y2x1ZGU9IGV4cGlyeT05MCBxdW90YT04MCAKIyBjb21wcmVzc2lvbiBlbmFibGU9b2ZmIGV4dGVuc2lvbnM9LnR4dCwubG9nLC5jc3YsLmpzb24sLnRhciwueG1sLC5iaW4gbWltZV90eXBlcz10ZXh0LyosYXBwbGljYXRpb24vanNvbixhcHBsaWNhdGlvbi94bWwgCiMgZXRjZCBlbmRwb2ludHM9IHBhdGhfcHJlZml4PSBjb3JlZG5zX3BhdGg9L3NreWRucyBjbGllbnRfY2VydD0gY2xpZW50X2NlcnRfa2V5PSAKIyBpZGVudGl0eV9vcGVuaWQgY29uZmlnX3VybD0gY2xpZW50X2lkPSBjbGFpbV9uYW1lPXBvbGljeSBjbGFpbV9wcmVmaXg9IGp3a3NfdXJsPSAKIyBwb2xpY3lfb3BhIHVybD0gYXV0aF90b2tlbj0gCiMga21zX3ZhdWx0IGVuZHBvaW50PSBrZXlfbmFtZT0gYXV0aF90eXBlPWFwcHJvbGUgYXV0aF9hcHByb2xlX2lkPSBhdXRoX2FwcHJvbGVfc2VjcmV0PSBjYXBhdGg9IGtleV92ZXJzaW9uPSBuYW1lc3BhY2U9IAojIGttc19rZXMgZW5kcG9pbnQ9IGtleV9uYW1lPSBjZXJ0X2ZpbGU9IGtleV9maWxlPSBjYXBhdGg9IAojIGxvZ2dlcl93ZWJob29rIGVuYWJsZT1vZmYgZW5kcG9pbnQ9IGF1dGhfdG9rZW49IAojIGF1ZGl0X3dlYmhvb2sgZW5hYmxlPW9mZiBlbmRwb2ludD0gYXV0aF90b2tlbj0gCiMgbm90aWZ5X2FtcXAgZW5hYmxlPW9mZiB1cmw9IGV4Y2hhbmdlPSBleGNoYW5nZV90eXBlPSByb3V0aW5nX2tleT0gbWFuZGF0b3J5PW9mZiBkdXJhYmxlPW9mZiBub193YWl0PW9mZiBpbnRlcm5hbD1vZmYgYXV0b19kZWxldGVkPW9mZiBkZWxpdmVyeV9tb2RlPTAgcXVldWVfbGltaXQ9MCBxdWV1ZV9kaXI9IAojIG5vdGlmeV9rYWZrYSBlbmFibGU9b2ZmIHRvcGljPSBicm9rZXJzPSBzYXNsX3VzZXJuYW1lPSBzYXNsX3Bhc3N3b3JkPSBjbGllbnRfdGxzX2NlcnQ9IGNsaWVudF90bHNfa2V5PSB0bHNfY2xpZW50X2F1dGg9MCBzYXNsPW9mZiB0bHM9b2ZmIHRsc19za2lwX3ZlcmlmeT1vZmYgcXVldWVfbGltaXQ9MCBxdWV1ZV9kaXI9IAojIG5vdGlmeV9tcXR0IGVuYWJsZT1vZmYgYnJva2VyPSB0b3BpYz0gcGFzc3dvcmQ9IHVzZXJuYW1lPSBxb3M9MCBrZWVwX2FsaXZlX2ludGVydmFsPTBzIHJlY29ubmVjdF9pbnRlcnZhbD0wcyBxdWV1ZV9kaXI9IHF1ZXVlX2xpbWl0PTAgCiMgbm90aWZ5X25hdHMgZW5hYmxlPW9mZiBhZGRyZXNzPSBzdWJqZWN0PSB1c2VybmFtZT0gcGFzc3dvcmQ9IHRva2VuPSB0bHM9b2ZmIHRsc19za2lwX3ZlcmlmeT1vZmYgY2VydF9hdXRob3JpdHk9IGNsaWVudF9jZXJ0PSBjbGllbnRfa2V5PSBwaW5nX2ludGVydmFsPTAgc3RyZWFtaW5nPW9mZiBzdHJlYW1pbmdfYXN5bmM9b2ZmIHN0cmVhbWluZ19tYXhfcHViX2Fja3NfaW5fZmxpZ2h0PTAgc3RyZWFtaW5nX2NsdXN0ZXJfaWQ9IHF1ZXVlX2Rpcj0gcXVldWVfbGltaXQ9MCAKIyBub3RpZnlfbnNxIGVuYWJsZT1vZmYgbnNxZF9hZGRyZXNzPSB0b3BpYz0gdGxzPW9mZiB0bHNfc2tpcF92ZXJpZnk9b2ZmIHF1ZXVlX2Rpcj0gcXVldWVfbGltaXQ9MCAKIyBub3RpZnlfbXlzcWwgZW5hYmxlPW9mZiBmb3JtYXQ9bmFtZXNwYWNlIGhvc3Q9IHBvcnQ9IHVzZXJuYW1lPSBwYXNzd29yZD0gZGF0YWJhc2U9IGRzbl9zdHJpbmc9IHRhYmxlPSBxdWV1ZV9kaXI9IHF1ZXVlX2xpbWl0PTAgCiMgbm90aWZ5X3Bvc3RncmVzIGVuYWJsZT1vZmYgZm9ybWF0PW5hbWVzcGFjZSBjb25uZWN0aW9uX3N0cmluZz0gdGFibGU9IGhvc3Q9IHBvcnQ9IHVzZXJuYW1lPSBwYXNzd29yZD0gZGF0YWJhc2U9IHF1ZXVlX2Rpcj0gcXVldWVfbGltaXQ9MCAKIyBub3RpZnlfcG9zdGdyZXM6aG9saSBlbmFibGU9b2ZmIGZvcm1hdD1uYW1lc3BhY2UgY29ubmVjdGlvbl9zdHJpbmc9IHRhYmxlPSBob3N0PSBwb3J0PSB1c2VybmFtZT0gcGFzc3dvcmQ9IGRhdGFiYXNlPSBxdWV1ZV9kaXI9IHF1ZXVlX2xpbWl0PTAgCiMgbm90aWZ5X3Bvc3RncmVzOnRlc3QgZW5hYmxlPW9mZiBmb3JtYXQ9bmFtZXNwYWNlIGNvbm5lY3Rpb25fc3RyaW5nPSB0YWJsZT0gaG9zdD0gcG9ydD0gdXNlcm5hbWU9IHBhc3N3b3JkPSBkYXRhYmFzZT0gcXVldWVfZGlyPSBxdWV1ZV9saW1pdD0wIAojIG5vdGlmeV9lbGFzdGljc2VhcmNoIGVuYWJsZT1vZmYgdXJsPSBmb3JtYXQ9bmFtZXNwYWNlIGluZGV4PSBxdWV1ZV9kaXI9IHF1ZXVlX2xpbWl0PTAgCiMgbm90aWZ5X3JlZGlzIGVuYWJsZT1vZmYgZm9ybWF0PW5hbWVzcGFjZSBhZGRyZXNzPSBrZXk9IHBhc3N3b3JkPSBxdWV1ZV9kaXI9IHF1ZXVlX2xpbWl0PTAg" | base64 --decode
```

Making it string makes it tricky to now consume it properly, requires additionally trickery.  I would say why not avoid this change in this PR?

We can add another version of `madmin.ParseSubSysTarget` which simply returns the correct representation of data structure which can be stringified etc and also when jsonified converts into proper JSON.

kannappanr:
In the json mode, you need to have a status field

```
$ mc ls play/test --json
{
 "status": "success",
 "type": "file",
 "lastModified": "2020-03-31T00:27:39.294-07:00",
 "size": 26,
 "key": "issue",
 "etag": "b954418e6a50d4d4cb8f02776d867550"
}
{
 "status": "success",
 "type": "file",
 "lastModified": "2020-03-31T00:56:46.882-07:00",
 "size": 26,
 "key": "issue1",
 "etag": "b954418e6a50d4d4cb8f02776d867550"
}

```
There is always a status field. In the case of error , the status will be `error` and there will also be an error field.

In the case of error, there will be an error field

kannappanr:
The command seems to hang in the case where the server is old or is a gateway

wlan0:
@kannappanr There is a Status and a Error field. I had set both to omitempty earlier, and wasn't setting either Status or Error. Its fixed now.

Now it errors out in case of gateway mode/old server

cmd/admin-obd.go
wlan0: No, it is not confirmed. I think it should be MC

cmd/admin-obd.go
wlan0: It came in from an earlier version of this impl, where types were provided by the user.

cmd/admin-obd.go
wlan0: Yeah, this should go into a constant

cmd/client-admin.go
wlan0: This isn't needed anymore. It was a debugging session where I introduced this change, and missed removing it later.

cmd/admin-config-export.go
wlan0: yeah, that works

wlan0:
@harshavardhana addressed your comments as well

cmd/admin-config-export.go
```diff
@@ -46,13 +45,13 @@ EXAMPLES:
...
-	Status string `json:"status"`
-	Value  []byte `json:"value"`
+	Status string      `json:"status"`
+	Value  interface{} `json:"value"`
 }
 
 // String colorized service status message.
 func (u configExportMessage) String() string {
-	return string(u.Value)
+	return u.Value.(string)
```
harshavardhana: You need to revert this change as well @wlan0 - this will crash

wlan0: done

PR #3115, opened 2020-03-13
mc cp: Fix prefix handling in TypeC

poornas:
for copy from filesystem to objectstorage.

Avoid prefixing dirname to file on target side
when doing a `mc cp /tmp/dir myminio/bucket --r`
 to make `mc mirror` and `mc cp` behavior consistent
```
with master
➜  mc ls myminio/images     
[2020-03-12 19:13:55 PDT]     26B a.txt
[2020-03-12 19:13:55 PDT]     26B b.txt
➜  mc cp /tmp/data/small myminio/images --r
...mall/b.txt:  130 B / 130 B ┃▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓┃ 8.36 KiB/s 0s➜  data mc ls myminio/images 
[2020-03-12 19:14:29 PDT]      0B small/
```
```
with this fix
➜  mc ls /tmp/data/small
[2020-03-11 17:13:47 PDT]     26B a.txt
[2020-03-11 17:13:54 PDT]     26B b.txt
➜  mc cp /tmp/data/small myminio/images --r
...mall/b.txt:  130 B / 130 B ┃▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓┃ 7.73 KiB/s 0s

```

### DIFF
--- a/cmd/cp-url.go
+++ b/cmd/cp-url.go
@@ -191,6 +191,11 @@ func makeCopyContentTypeC(sourceAlias string, sourceURL clientURL, sourceContent
 	newSourceSuffix := filepath.ToSlash(newSourceURL.Path)
 	if pathSeparatorIndex > 1 {
 		sourcePrefix := filepath.ToSlash(sourceURL.Path[:pathSeparatorIndex])
+		// do not preserve unix cp behavior when copying from filesytem to
+		// objectstore.
+		if sourceAlias == "" && targetAlias != "" {
+			sourcePrefix = sourceURL.Path
+		}
 		newSourceSuffix = strings.TrimPrefix(newSourceSuffix, sourcePrefix)
 	}
 	newTargetURL := urlJoinPath(targetURL, newSourceSuffix)


harshavardhana:
`mc cp` behaves like `unix` cp and `mirror` is like `rsync`  so they are not the same

poornas:
> The prefix is done to behave similar to unix `cp -r`

Ok, @abperiasamy was expecting that mirror and cp would behave identically here.

harshavardhana:
> Ok, @abperiasamy was expecting that mirror and cp would behave identically here.

Then you need to inform him of why this was the case, breaking this now will be unexpected for our users - `cp` intention was to behave like unix `cp` not `mirror`.

poornas:
> > Ok, @abperiasamy was expecting that mirror and cp would behave identically here.
> 
> Then you need to inform him of why this was the case, breaking this now will be unexpected for our users - `cp` intention was to behave like unix `cp` not `mirror`.

@harshavardhana , updated my PR to preserve unix cp style when both src and dst are filesystems - i.e. `mc cp src/dir dst -r` will show dst/dir/filename when dst is a filesystem. If dst is a objectstore, will be copied as dst/filename

cmd/cp-url.go
```diff
@@ -191,6 +191,9 @@ func makeCopyContentTypeC(sourceAlias string, sourceURL clientURL, sourceContent
 	newSourceSuffix := filepath.ToSlash(newSourceURL.Path)
 	if pathSeparatorIndex > 1 {
 		sourcePrefix := filepath.ToSlash(sourceURL.Path[:pathSeparatorIndex])
+		if sourceAlias == "" && targetAlias != "" {
```
harshavardhana: We should add a comment about this behavior difference @poornas

poornas: done

PR #3130, opened 2020-03-26
mc watch windows URL/Absolute path event path 

BigUstad:
Folder is an URI/URL but rename/write event path is absolute path

### DIFF
--- a/cmd/client-fs_windows.go
+++ b/cmd/client-fs_windows.go
@@ -22,7 +22,7 @@ import "github.com/rjeczalik/notify"
 
 var (
 	// EventTypePut contains the notify events that will cause a put (writer)
-	EventTypePut = []notify.Event{notify.Create, notify.Write, notify.Rename}
+	EventTypePut = []notify.Event{notify.Create, notify.Write, notify.Rename, notify.FileNotifyChangeFileName, notify.FileNotifyChangeDirName}
 	// EventTypeDelete contains the notify events that will cause a delete (remove)
 	EventTypeDelete = []notify.Event{notify.Remove}
 	// EventTypeGet contains the notify events that will cause a get (read)
@@ -36,6 +36,9 @@ func IsGetEvent(event notify.Event) bool {
 
 // IsPutEvent checks if the event returned is a put event
 func IsPutEvent(event notify.Event) bool {
+	if event&notify.FileActionRenamedOldName != 0 {
+		return false
+	}
 	for _, ev := range EventTypePut {
 		if event&ev != 0 {
 			return true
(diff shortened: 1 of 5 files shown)


cmd/mirror-main.go
```diff
@@ -405,7 +405,27 @@ func (mj *mirrorJob) watchMirror(ctx context.Context, cancelMirror context.Cance
...
+				shdModifyEventPath := filepath.IsAbs(eventPath) && !filepath.IsAbs(mj.sourceURL) && lenSrcURLSlice > 1
+				if shdModifyEventPath {
+					lastPathPrefix = sourceURLpathList[lenSrcURLSlice-1]
+					lastPathPrefixPos = strings.Index(eventPath, lastPathPrefix)
+				}
+				canModifyEventPath := shdModifyEventPath && lastPathPrefix != "" && lastPathPrefixPos > 0
+				canModifyEventPath = canModifyEventPath && lastPathPrefixPos+len(lastPathPrefix) < len(eventPath)
+				if canModifyEventPath {
+					eventRelPath = eventPath[lastPathPrefixPos+len(lastPathPrefix):]
+					eventRelPath = strings.Replace(eventRelPath, "\\", "/", -1)
```
harshavardhana: filepath.ToSlash() is the helper you want @BigUstad

cmd/client-fs.go
```diff
@@ -183,6 +183,14 @@ func (f *fsClient) Watch(params watchParams) (*watchObject, *probe.Error) {
 				i, e = os.Stat(event.Path())
 				if e != nil {
 					if os.IsNotExist(e) {
+						if runtime.GOOS == "windows" && event.Event()&notify.Rename != 0 {
```
harshavardhana: Does this perhaps need to be done outside? because looking for delete event can be done before doing a stat on the path?

BigUstad: This is for the case of renaming of a file (notify.Rename). The event (notify.Rename) is generated with the older filename and the newer(changed) filename. We need this check to distinguish between the older & newer names. Please advise.

cmd/mirror-main.go
BigUstad: Thanks.

cmd/client-fs.go
harshavardhana: Correct but it's added deep down after os.Stat() why not before?

cmd/client-fs_windows.go
```diff
@@ -37,7 +37,9 @@ func IsGetEvent(event notify.Event) bool {
 // IsPutEvent checks if the event returned is a put event
 func IsPutEvent(event notify.Event) bool {
 	for _, ev := range EventTypePut {
-		if event&ev != 0 {
+		if event&notify.FileActionRenamedOldName != 0 {
```
kannappanr: This has to be outside the `for` loop, why would you check this every time

BigUstad:
There was a mistake in the git push. Closing this PR. Will open another PR.

PR #3132, opened 2020-03-30
Make BuildS3Config public

dvaldivia:
We want to probe credentials from [mcs](https://github.com/minio/m3/tree/master/mcs) and want to leverage `mc` existing code to do so.


dvaldivia:
@cesnietor we can do that on a separate PR to keep this one easy to review

Issue #3148, opened 2020-04-08
mc cat is doing listing before retrieving a file


Baresse:
## Expected behavior

Whatever the number of files in a directory when the command `mc cat  .../file` is used no listing should be done by `mc`. 

## Actual behavior
I’ve a minio cluster serving files stored on an XFS backed filesystem. When I try to retrieve a small CSV file (2 lines) in a directory which contains around 200k files, it takes around 10 seconds to come.

Here is the command I use : 
`mc cat minio/draft/xxxx/yyyy/FR2313202810` 

If I do the same command in a directory with less files inside, the file is returned without delay. 

If I create a shared Download URL using this command : 
 `mc share download -expire 4h minio/draft/xxxx/yyyy/FR2313202810` 

I get an URL and if use the following command : 
  `wget <provided_download_url> -O FR2313202810` 
the file is returned without delay.

Note: We have the same behavior with filename with extension for instance `FR2313202810.csv`

## mc --version
- `mc version RELEASE.2020-04-04T05-28-55Z`

klauspost:
`mc cat local2/somebucket/aprefix/data.csv` results in 

```
s3.ListObjectsV2 localhost:9001/somebucket/?delimiter=%2F&encoding-type=url&fetch-owner=true&list-type=2&prefix=aprefix%2Fdata.csv
s3.GetObject localhost:9001/somebucket/aprefix/data.csv
```

PR #3156, opened 2020-04-13
s3: Use head API before listing in client.Stat()

vadmeste:
Stat() is supposed to return valid information for existing
objects and existing prefixes as well.

The current mechanism issues a listing request and searches
for the object or prefix in the returned list, but this could take
time if there are lot of entries under a parent prefix.

The new commit changes the internal behavior but not the
functional output. It issues a HEAD request first, if there
is no object then assumes it is a prefix and does listing
in order to find it.

Fixes https://github.com/minio/mc/issues/3148

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -1300,11 +1300,43 @@ func (c *S3Client) listObjectWrapper(bucket, object string, isRecursive bool, do
 	return c.api.ListObjectsV2(bucket, object, isRecursive, doneCh)
 }
 
-// Stat - send a 'HEAD' on a bucket or object to fetch its metadata.
-func (c *S3Client) Stat(isIncomplete, isFetchMeta, isPreserve bool, sse encrypt.ServerSide) (*ClientContent, *probe.Error) {
+func (c *S3Client) statIsComplete(bucket, object string) (*ClientContent, *probe.Error) {
+	nonRecursive := false
+	objectMetadata := &ClientContent{}
+	// Prefix to pass to minio-go listing in order to fetch a given object/directory
+	prefix := strings.TrimRight(object, string(c.targetURL.Separator))
+
+	for objectMultipartInfo := range c.api.ListIncompleteUploads(bucket, prefix, nonRecursive, nil) {
+		if objectMultipartInfo.Err != nil {
+			return nil, probe.NewError(objectMultipartInfo.Err)
+		}
+
+		if objectMultipartInfo.Key == object {
+			objectMetadata.URL = *c.targetURL
+			objectMetadata.Time = objectMultipartInfo.Initiated
+			objectMetadata.Size = objectMultipartInfo.Size
+			objectMetadata.Type = os.FileMode(0664)
+			objectMetadata.Metadata = map[string]string{}
+			return objectMetadata, nil
+		}
+
+		if strings.HasSuffix(objectMultipartInfo.Key, string(c.targetURL.Separator)) {
+			objectMetadata.URL = *c.targetURL
+			objectMetadata.Type = os.ModeDir
+			objectMetadata.Metadata = map[string]string{}
+			return objectMetadata, nil
+		}
+	}
+	return nil, probe.NewError(ObjectMissing{})
+}
+
+// Stat - send a 'HEAD' on a bucket or object to fetch its metadata. It also returns
+// a DIR type content if a prefix does exist in the server.
+func (c *S3Client) Stat(isIncomplete, isPreserve bool, sse encrypt.ServerSide) (*ClientContent, *probe.Error) {
 	c.Lock()
 	defer c.Unlock()
 	bucket, object := c.url2BucketAndObject()
+
 	// Bucket name cannot be empty, stat on URL has no meaning.
 	if bucket == "" {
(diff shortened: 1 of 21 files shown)


vadmeste:
> // isFetchMeta is true only in the case of mc stat command which lists any extended attributes
> // present for this object.

Fixed

@ebozduman I also added a separate commit which removes an unused variable

harshavardhana:
The reason we did List before Head was to avoid the Head 404 which will always happen when you want to do something like `mc cp -r` of a directory 

So in such a scenario Listing non recursively was a better option, with the new change now we will send an unnecessary Head call which will always fallback to List.

Check the commit which did by `git blame` to check the history..

cmd/client-s3.go
```diff
@@ -1326,40 +1363,23 @@ func (c *S3Client) Stat(isIncomplete, isFetchMeta, isPreserve bool, sse encrypt.
...
 
-			if objectMultipartInfo.Key == object {
-				objectMetadata.URL = *c.targetURL
-				objectMetadata.Time = objectMultipartInfo.Initiated
-				objectMetadata.Size = objectMultipartInfo.Size
-				objectMetadata.Type = os.FileMode(0664)
-				objectMetadata.Metadata = map[string]string{}
-				return objectMetadata, nil
-			}
+	ctnt, err := c.getObjectStat(bucket, object, opts)
```
Praveenrajmani, line 1369: Shouldn't we do the object stat only if the object name is not suffixed with the slash separator?

cmd/client-s3.go
```diff
@@ -1318,6 +1350,11 @@ func (c *S3Client) Stat(isIncomplete, isFetchMeta, isPreserve bool, sse encrypt.
 		return content, nil
 	}
 
+	// If the request is for incomplete upload stat, handle it here.
+	if isIncomplete {
+		return c.statIsComplete(bucket, object)
+	}
+
 	// The following code tries to calculate if a given prefix/object does really exist
```
Praveenrajmani, line 1358: Can we modify this comment as per the current change

cmd/client-s3.go
```diff
@@ -1368,37 +1388,11 @@ func (c *S3Client) Stat(isIncomplete, isFetchMeta, isPreserve bool, sse encrypt.
 		if strings.HasSuffix(objectStat.Key, string(c.targetURL.Separator)) {
```
Praveenrajmani, line 1388: Why are we omitting this check - `strings.Contains(objectStat.Key, object)` here

cmd/client-s3.go
```diff
@@ -1300,11 +1300,43 @@ func (c *S3Client) listObjectWrapper(bucket, object string, isRecursive bool, do
 	return c.api.ListObjectsV2(bucket, object, isRecursive, doneCh)
 }
 
-// Stat - send a 'HEAD' on a bucket or object to fetch its metadata.
-func (c *S3Client) Stat(isIncomplete, isFetchMeta, isPreserve bool, sse encrypt.ServerSide) (*ClientContent, *probe.Error) {
+func (c *S3Client) statIsComplete(bucket, object string) (*ClientContent, *probe.Error) {
```
Praveenrajmani, line 1303: Can we rename this to `statInComplete`?

cmd/client-s3.go
vadmeste: > Shouldn't we do the object stat only if the object name is not suffixed with the slash separator?

Yes we can optimize it

cmd/client-s3.go
vadmeste: > Can we modify this comment as per the current change

Nothing is supposed to change in the comment, are you referring to something specific ?

cmd/client-s3.go
vadmeste: > Why are we omitting this check - `strings.Contains(objectStat.Key, object)` here? Is it because we already do a stat check before?

Yes right. I wanted to avoid listing at all cost here since listing can take awhile if there are many objects

cmd/client-s3.go
vadmeste: > Can we rename this to `statInComplete`?

Will do

cmd/client-s3.go
```diff
@@ -1263,11 +1263,43 @@ func (c *S3Client) listObjectWrapper(bucket, object string, isRecursive bool, do
 	return c.api.ListObjectsV2(bucket, object, isRecursive, doneCh)
 }
 
-// Stat - send a 'HEAD' on a bucket or object to fetch its metadata.
-func (c *S3Client) Stat(isIncomplete, isFetchMeta, isPreserve bool, sse encrypt.ServerSide) (*ClientContent, *probe.Error) {
+func (c *S3Client) statInComplete(bucket, object string) (*ClientContent, *probe.Error) {
```
ebozduman, line 1266: Actually `statIncomplete` is a misleading name, since stat process is not the incomplete entity here but the file upload is the one.
Can we rename it again? It could be "statIncompleteUploads" or something like that.

vadmeste: > Can we rename it again? It could be `"statIncompleteUploads"` or something like that, like existing `"ListIncompleteUploads"`

Done, I called it `statIncompleteUpload`

vadmeste:
PR rebased.

PR #3159, opened 2020-04-13
Fixes a simple typo

ebozduman:
Fixes just a simple typo


cmd/admin-config-reset.go
```diff
@@ -64,7 +64,7 @@ type configResetMessage struct {
 // String colorized service status message.
 func (u configResetMessage) String() (msg string) {
 	msg += console.Colorize("ResetConfigSuccess",
-		"Reseteting key has been successful.\n")
+		"Reseting key has been successful.\n")
```
klauspost, line 67:
```suggestion
		"Resetting key has been successful.\n")
```

ebozduman: @klauspost 
Good point. Thank you

PR #3161, opened 2020-04-14
sql: Adapt the code to the change in minio-go SQL API

vadmeste:
Can be merged only after merging this https://github.com/minio/minio-go/pull/1268


vadmeste:
Not WIP. Ready for review.

go.mod
```diff
@@ -3,6 +3,8 @@ module github.com/minio/mc
 go 1.13
 
 require (
+	github.com/aliyun/aliyun-oss-go-sdk v0.0.0-20190307165228-86c17b95fcd5 // indirect
```
harshavardhana, line 6: aliyun dependency is deleted from server, we don't need this anymore have you run `go mod tidy?` @vadmeste

vadmeste: > aliyun dependency is deleted from server, we don't need this anymore have you run `go mod tidy?` @vadmeste

Done

PR #3162, opened 2020-04-14
mirror: Use source time in multi master mode

vadmeste:
Still work in progress.

It works but I need to ensure that it conforms exactly to the specification.

### DIFF
--- a/cmd/difference.go
+++ b/cmd/difference.go
@@ -32,13 +32,13 @@ import (
 type differType int
 
 const (
-	differInNone     differType = iota // does not differ
-	differInETag                       // differs in ETag
-	differInSize                       // differs in size
-	differInMetadata                   // differs in metadata
-	differInType                       // differs in type, exfile/directory
-	differInFirst                      // only in source (FIRST)
-	differInSecond                     // only in target (SECOND)
+	differInNone          differType = iota // does not differ
+	differInSize                            // differs in size
+	differInMMSourceMTime                   // differs in multi-master source modtime
+	differInMetadata                        // differs in metadata
+	differInType                            // differs in type, exfile/directory
+	differInFirst                           // only in source (FIRST)
+	differInSecond                          // only in target (SECOND)
 )
 
 func (d differType) String() string {
@@ -49,6 +49,8 @@ func (d differType) String() string {
 		return "size"
 	case differInMetadata:
 		return "metadata"
+	case differInMMSourceMTime:
+		return "mm-source-mtime"
 	case differInType:
 		return "type"
 	case differInFirst:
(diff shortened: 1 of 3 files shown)


vadmeste:
So what this change does:
- Forget everything about old sTag & eTag multi master meta attributes
- Each mirror copy will add a source-mod-time-meta-attribute to the target which is equal to the source object modification time
- In addition to the regular diff: comparison between source & target compares mod time in source object and source-mod-time-meta-attribute in target object
- In watch mode, when an object is received and it has the source-mod-time-meta-attribute, do not mirror it.

krishnasrinivas:
> In addition to the regular diff: comparison between source & target compares mod time in source object and source-mod-time-meta-attribute in target object

the comparison algo should be like this:
```
srcActualModTime := latest(srcMeta["Last-Modified"], srcMeta["X-Amz-Meta-Source-Last-Modified"]
dstActualModTime := latest(dstMeta["Last-Modified"], dstMeta["X-Amz-Meta-Source-Last-Modified"]

if srcActualModTime > dstActualModTime {
  doMirro()
}
```

cmd/difference.go
```diff
@@ -59,34 +61,33 @@ func (d differType) String() string {
 	return "unknown"
 }
 
-const multiMasterETagKey = "X-Amz-Meta-Mm-Etag"
-const multiMasterSTagKey = "X-Amz-Meta-Mm-Stag"
+const multiMasterSourceModTime = "X-Amz-Meta-Mm-Mtime"
```
krishnasrinivas, line 64: `X-Amz-Meta-Mm-Source-Mtime` will be better

vadmeste:
@krishnasrinivas could you take a look ?

cmd/difference.go
```diff
@@ -32,13 +32,13 @@ import (
...
-	differInNone     differType = iota // does not differ
-	differInETag                       // differs in ETag
-	differInSize                       // differs in size
-	differInMetadata                   // differs in metadata
-	differInType                       // differs in type, exfile/directory
-	differInFirst                      // only in source (FIRST)
-	differInSecond                     // only in target (SECOND)
+	differInNone          differType = iota // does not differ
+	differInSize                            // differs in size
+	differInMMSourceMTime                   // differes in multi master source mod time
```
harshavardhana, line 37:
```suggestion
	differInMMSourceMTime                   // differs in multi master source mod time
```

cmd/difference.go
```diff
@@ -59,34 +61,66 @@ func (d differType) String() string {
 	return "unknown"
 }
 
-const multiMasterETagKey = "X-Amz-Meta-Mm-Etag"
+const multiMasterSourceModTimeKey = "X-Amz-Meta-Mm-Source-Mtime"
 const multiMasterSTagKey = "X-Amz-Meta-Mm-Stag"
```
harshavardhana, line 65: this should be removed?

cmd/difference.go
```diff
@@ -59,34 +61,66 @@ func (d differType) String() string {
...
+	if t1.Before(t2) {
+		return t1
+	}
+
+	return t2
+}
+
+// multiMasterModTimeUpdated tries to calculate if the object copy in the target
+// is older than the one in the source by comparing the modtime of the data.
+func multiMasterModTimeUpdated(src, dst *ClientContent, multiMasterCtx string) (ret bool) {
```
harshavardhana, line 83: not using the named return why not just keep it `bool`

vadmeste:
@krishnasrinivas could you take a look ?

PR #3163, opened 2020-04-14
match S3 within URL in full appropriately

### LINKED ISSUE #3140: mc fails with digital ocean ams3 region
## Expected behavior
mc doesn't report an error

## Actual behavior
mc: <ERROR> Unable to stat `done`. Bucket name cannot be shorter than 3 characters.

Configure a bucket in digitalocean spaces ams3 region
```json
                "done": {
                        "url": "https://ams3.digitaloceanspaces.com",
                        "accessKey": "",
                        "secretKey": "",
                        "api": "S3v4",
                        "lookup": "dns"
                },
```

run `mc ls done`

## mc --version
- latest + master

## PROBLEM
I don't know why you are splitting host but the bug is in this line 
https://github.com/minio/mc/blob/master/cmd/client-s3.go#L1404
and another one a few lines lower
https://github.com/minio/mc/blob/master/cmd/client-s3.go#L1426

BigUstad:
Fixes #3140

### DIFF
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"io"
 	"math/rand"
+	"net"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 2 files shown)


cmd/utils.go
```diff
@@ -349,3 +350,25 @@ func parseAttribute(attrs string) (map[string]string, error) {
...
+			cnt, ok := checkMap[fqdn]
+			if !ok {
+				checkMap[fqdn] = 1
+			} else {
+				checkMap[fqdn] = cnt + 1
+			}
+		}
+	}
+	_, ok := checkMap["s3"]
+	return ok
```
klauspost, line 373:
```suggestion
	for _, fqdn := range fqdnParts {
		if fqdn == "s3" {
			return true
		}
	}
	return false
```

BigUstad: @klauspost PTAL

PR #3169, opened 2020-04-16
improve signature probe avoid repeated code

harshavardhana:


cmd/config-host-add.go
```diff
@@ -166,39 +167,44 @@ func probeS3Signature(accessKey, secretKey, url string) (string, *probe.Error) {
...
+		return stype, nil
 	}
 
-	return s3Config.Signature, nil
+	stype, err := probeSignatureType("s3v4")
+	if err != nil {
+		if stype, err = probeSignatureType("s3v2"); err != nil {
+			return "", err.Trace("s3v4", "s3v2")
+		}
+		return stype, nil
```
Praveenrajmani, line 205: Shall we remove this redundant return statement. We can fallback to L207

harshavardhana: No that because err.Trace() is needed to show the return that happened at this place.

PR #3202, opened 2020-05-05
fixes cp not preserving ownership and other attrs

### LINKED ISSUE #3187: mc not preserving ownership
## Expected behavior
mc should preserve ownership

## Actual behavior
mc doesnt

~ ls -l cont/box0/ | grep scripts
drwxr-xr-x 2 sshd daemon 4096 Apr 22 15:10 scripts
~ mc mirror -a --remove --overwrite cont/box0/ dot9/cont/box0/
0 B / ?
~ rm -rf cont/box0
~ mc mirror -a dot9/cont/box0/ cont/box0/
...000/cont/box0/testFile: 149 B / 149 B 7.03 KiB/s 0s
~ ls -l cont/box0/ | grep scripts
drwxr-xr-x 2 root root 4096 Apr 22 15:13 scripts

@klauspost i am using -a flag which is same as --preserve as per ticket #3186 
Also tested with --preserve instead of -a, same result

## mc --version
mc version RELEASE.2020-04-19T19-17-53Z

ebozduman:
Fixes #3187 

When the preserve attributes flag, `-a`, and functionality was introduced, the assumption was that metadata index is going to be `"mc-attrs"`, however, it is `"X-Amz-Meta-Mc-Attrs"`.
The fix is to accept both indexes as valid values.

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -51,6 +51,7 @@ type fsClient struct {
 const (
 	partSuffix     = ".part.minio"
 	slashSeperator = "/"
+	metaDataKey    = "X-Amz-Meta-Mc-Attrs"
 )
 
 var ( // GOOS specific ignore list.
@@ -302,8 +303,8 @@ func (f *fsClient) put(reader io.Reader, size int64, metadata map[string][]strin
 	}
 
 	attr := make(map[string]string)
-	if len(metadata["mc-attrs"]) != 0 {
-		attr, e = parseAttribute(metadata["mc-attrs"][0])
+	if len(metadata[metaDataKey]) != 0 {
+		attr, e = parseAttribute(metadata[metaDataKey][0])
 		if e != nil {
 			return 0, probe.NewError(e)
 		}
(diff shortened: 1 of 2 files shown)


cmd/client-fs.go
```diff
@@ -409,9 +409,13 @@ func (f *fsClient) put(reader io.Reader, size int64, metadata map[string][]strin
...
 func (f *fsClient) Put(ctx context.Context, reader io.Reader, size int64, metadata map[string]string, progress io.Reader, sse encrypt.ServerSide, md5, disableMultipart bool) (int64, *probe.Error) {
-	if metadata["mc-attrs"] != "" {
+	metaDataKey := "mc-attrs"
+	if metadata["X-Amz-Meta-Mc-Attrs"] != "" {
+		metaDataKey = "X-Amz-Meta-Mc-Attrs"
+	}
+	if metadata[metaDataKey] != "" {
 		meta := make(map[string][]string)
-		meta["mc-attrs"] = append(meta["mc-attrs"], metadata["mc-attrs"])
+		meta["mc-attrs"] = append(meta["mc-attrs"], metadata[metaDataKey])
```
harshavardhana, line 418:
```suggestion
		meta[metaDataKey] = append(meta[metaDataKey], metadata[metaDataKey])
```

ebozduman: @harshavardhana,
I did it this way, because I did not want to cause any regression, in case the same assumption is made in some other caller functions. If so, I also need to change those part of the codes in the same way.
Do you still want to see your suggestion to be implemented?

harshavardhana: it looks a bit confusing @ebozduman we need to use a single metadata key.

ebozduman: @harshavardhana, you suggested 2 ways:

1. I've just tested with your suggestion in your first comment.
As I expected, there is at least one or more callers which also expect the metadata key to be `"mc-attrs"`, hence the fix doesn't work. To make it work, there has to be more changes in this PR, if we choose to go this way.

2. Single key sounds better to me also
I can replace `"mc-attrs"` key with `"X-Amz-Meta-Mc-Attrs"` every where. (I'd still use a variable).
But, the only thing I am not %100 sure is that if there is any scenario where the key comes as `"mc-attrs"` and if we miss that code path, this PR is going to cause regression.

3. My original change (confusing, short and fastest solution)

Please let me know which option you want to go.

ebozduman: After discussing with @kannappanr and @poornas, we've decided to go with single key, `"X-Amz-Meta-Mc-Attrs"` everywhere. This way, we also become S3 compliant.

kannappanr:
PTAL build failure

PR #3213, opened 2020-05-07
support session tokens as part of ENV

harshavardhana:


cesnietor:
I tested it using AddbucketNotification and DeleteBucketNotification APIs, it works.

Issue #3215, opened 2020-05-08
Received excess data on input reader on mc cp from bucket to local fs


michagruner:
## Expected behavior
File is copied from a minio S3 bucket to the local filesystem
## Actual behavior
Copy job exits and throws the error message "Received excess data on input reader. Expected only `4504670` bytes, but received `18009689` bytes."
## Steps to reproduce the behavior
`user@server-management:~$ ./mc mb play/xinurg`
Bucket created successfully `play/xinurg`.
`user@server-management:~$ ./mc cp /tmp/000f8e5d889bb88eb7499132e6acbcdc0ff746f9a278d561e5282f12b207b082 play/xinurg/`
...f12b207b082:  4.30 MiB / 4.30 MiB ┃┃ 46.61 KiB/s 1m34s
`user@server-management:~$ ./mc cp play/xinurg/000f8e5d889bb88eb7499132e6acbcdc0ff746f9a278d561e5282f12b207b082 /tmp/`
mc: <ERROR> Failed to copy `https://play.min.io/xinurg/000f8e5d889bb88eb7499132e6acbcdc0ff746f9a278d561e5282f12b207b082`. Received excess data on input reader. Expected only `4504670` bytes, but received `18009689` bytes.
...f12b207b082:  4.30 MiB / 4.30 MiB ┃┃ 341.77 KiB/s 12s
user@server-management:~$ ./mc cp --debug play/xinurg/000f8e5d889bb88eb7499132e6acbcdc0ff746f9a278d561e5282f12b207b082 /tmp/
mc: <DEBUG> GET /xinurg/?location= HTTP/1.1
Host: play.min.io
User-Agent: MinIO (linux; amd64) minio-go/v6.0.54 mc/2020-04-17T08:55:48Z
Authorization: AWS4-HMAC-SHA256 Credential=**REDACTED**/20200508/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=**REDACTED**
X-Amz-Content-Sha256: UNSIGNED-PAYLOAD
X-Amz-Date: 20200508T073331Z
Accept-Encoding: gzip

mc: <DEBUG> HTTP/1.1 200 OK
Content-Length: 128
Accept-Ranges: bytes
Connection: keep-alive
Content-Security-Policy: block-all-mixed-content
Content-Type: application/xml
Date: Fri, 08 May 2020 07:37:02 GMT
Server: nginx/1.10.3 (Ubuntu)
[...]

PR #3216, opened 2020-05-08
fs: Simply PUT and fix issue when the source is not seek-able

vadmeste:
Fixes https://github.com/minio/mc/issues/3215

Functional change in this:
- In any case, the downloaded data will be stored in [file].part.minio first and then renamed to [file]

Bug fix:
The code also does not behave correctly when the source reader is not seekable. reader is wrapped by a reader hook and then tested if it is seekable or not, though the reader hook is always seekable.

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -288,15 +288,24 @@ func (f *fsClient) put(reader io.Reader, size int64, metadata map[string][]strin
 	}
 
 	objectPath := f.PathURL.Path
-	avoidResumeUpload := isStreamFile(objectPath)
+
+	_, seeker := reader.(io.Seeker)
+	avoidResumeUpload := isStreamFile(objectPath) || isStdIO(reader) || !seeker || size < 0
+
 	// Write to a temporary file "object.part.minio" before commit.
 	objectPartPath := objectPath + partSuffix
+
 	if avoidResumeUpload {
-		objectPartPath = objectPath
+		// We cannot resume this operation, then we
+		// should remove any partial download if any.
+		e := os.Remove(objectPartPath)
+		if e != nil && !os.IsNotExist(e) {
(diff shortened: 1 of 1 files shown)


harshavardhana:
I think we should remove the resumable feature it is unnecessary and always made things a bit complicated @vadmeste 

It is worth if we simply it by writing the entire file again rather than seeking.

vadmeste:
> I think we should remove the resumable feature it is unnecessary and always made things a bit complicated @vadmeste
> 
> It is worth it if we simply by writing the entire file again rather than seeking.

Are you sure ? @harshavardhana 

This PR makes it much more simpler than before, and the old code had some issues

harshavardhana:
> > I think we should remove the resumable feature it is unnecessary and always made things a bit complicated @vadmeste
> > 
> > It is worth it if we simply by writing the entire file again rather than seeking.
> 
> Are you sure ? @harshavardhana 
> 
> This PR makes it much more simpler than before, and the old code had some issues

For now it's okay, but users are generally not aware that we are resuming also we should resume with ETag, so that we resume the same  object. If not we end up corrupting the content.

harshavardhana:
> > > I think we should remove the resumable feature it is unnecessary and always made things a bit complicated @vadmeste
> > > It is worth it if we simply by writing the entire file again rather than seeking.
> > 
> > 
> > Are you sure ? @harshavardhana
> > This PR makes it much more simpler than before, and the old code had some issues
> 
> For now it's okay, but users are generally not aware that we are resuming also we should resume with ETag, so that we resume the same object. If not we end up corrupting the content.

Is this PR ready to be out of draft @vadmeste ?

vadmeste:
> Is this PR ready to be out of draft @vadmeste ?

Yes

harshavardhana:
PTAL @krisis

Issue #2694, opened 2019-02-21
mc cp autocomplete does not expand ~/


eco-minio:
## Expected behavior
When using autocomplete for mc commands, ~/ should expand to users home directory
## Actual behavior
Tab autocomplete does not expand ~/
## Steps to reproduce the behavior
`mc cp ~/Down<tab>` to try and expand to /home/user/Downloads
## mc version
```
mc version
Version: 2019-02-20T22:21:50Z
Release-tag: RELEASE.2019-02-20T22-21-50Z
Commit-id: e538599e2ffa4ee8d8040d5cd14475b380da531a
```
## System information
N/A

moonsn:
not work too

mc --autocompletion                                                                                        
mc: autocompletion is already enabled in your '$SHELLRC'

Version: 2019-10-09T22:54:57Z
Release-tag: RELEASE.2019-10-09T22-54-57Z
Commit-id: f93fe1330a3647b1afaff0ed8c188d2897bf391e

PR #3217, opened 2020-05-11
completion: Complete FS path starting with ~/

vadmeste:
posener/complete library does not support the completion of
paths starting with a tilde. This commit implements a workaround 
for that.

Fixes https://github.com/minio/mc/issues/2694

### DIFF
--- a/cmd/auto-complete.go
+++ b/cmd/auto-complete.go
@@ -29,7 +29,31 @@ import (
 // fsComplete knows how to complete file/dir names by the given path
 type fsComplete struct{}
 
-func (fs fsComplete) Predict(a complete.Args) (prediction []string) {
+// predictPathWithTilde completes an FS path which starts with a `~/`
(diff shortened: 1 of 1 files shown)


cmd/auto-complete.go
```diff
@@ -29,7 +29,31 @@ import (
 // fsComplete knows how to complete file/dir names by the given path
 type fsComplete struct{}
 
-func (fs fsComplete) Predict(a complete.Args) (prediction []string) {
+// predictPathWithTilde completes an FS path which starts with a `~/`
+func (fs fsComplete) predictPathWithTilde(a complete.Args) []string {
+	homeDir := os.Getenv("HOME")
```
harshavardhana, line 34: @vadmeste did you look at my suggestion?

vadmeste: > @vadmeste did you look at my suggestion?

done

PR #3218, opened 2020-05-11
set watch params struct as public

cesnietor:
This is needed so that mcs can make use of it.


cesnietor:
> We should perhaps take the opportunity to fix the structure name and rename this to WatchOptions instead of watchParams.

done

cesnietor:
PTAL @harshavardhana, @kannappanr

PR #3221, opened 2020-05-16
mc ilm command 

BigUstad:


cmd/ilm/options.go
```diff
@@ -74,19 +74,22 @@ func getRightAligned(label string, maxLen int) string {
...
+	}
 	n := 0
 	for _, rule := range lfcCfg.Rules {
 		if rule.ID != ilmID {
 			lfcCfg.Rules[n] = rule
 			n++
 		}
 	}
-	lfcCfg.Rules = lfcCfg.Rules[:n]
 	if n == len(lfcCfg.Rules) && len(lfcCfg.Rules) > 0 {
```
harshavardhana, line 87: We don't need `len(lfcCfg.Rules)` check anymore.

cmd/ilm-export.go
```diff
@@ -86,6 +88,10 @@ func mainILMExport(ctx *cli.Context) error {
 
 	ilmCfg, err := client.GetLifecycle()
 	fatalIf(err.Trace(args...), "Unable to get lifecycle configuration")
+	if len(ilmCfg.Rules) == 0 {
```
harshavardhana, line 91: We should allow default config export @BigUstad to allow displaying the data structure, to allow for helping them write the configuration IMO.

BigUstad:
@harshavardhana PTAL

Issue #3224, opened 2020-05-21
"ERROR: Unhandled difference between" when using `./mc diff`


BLuFeNiX:
## Expected behavior
diff works

## Actual behavior
`./mc diff` behaves differently depending on whether the local or remote path is listed first, as well as what level of the directory structure is compared. 

## Steps to reproduce the behavior
```
./mc mirror -a <local> <remote>
./mc diff <local> <remote>
./mc diff <remote> <local>
```
error example:
```
mc: <ERROR> Unhandled difference between `https://s3.wasabisys.com/REDACTED/Bahamas Cruise 2014/2014-12-15 17.02.35.jpg` and `/data/REDACTED/Bahamas Cruise 2014/2014-12-15 17.02.35.jpg`.
```

**Update**
When using `--json`, we can see an undefined diff value:
```
root@b76efd39aa1b:/# ./mc --json diff "/data/REDACTED/" "wasabi2/REDACTED/"
{
 "status": "success",
 "first": "",
 "second": "https://s3.wasabisys.com/REDACTED/",
 "diff": 6
}
```
This is strange, since it seems that the diff value only goes from 0 to 5, according to: https://docs.min.io/docs/minio-client-complete-guide.html

Note: this issue appears to be filename related. I have renamed some of these files and the issue went away for that file only. The issue does not occur if I swap `<local>` and `<remote>`. 

## mc --version
mc version RELEASE.2020-05-16T01-44-37Z

## System information
Ubuntu 20.04 docker image

BLuFeNiX:
Updated original post to better describe issue.

harshavardhana:
Thanks for opening the issue we shall take a look.

PR #3226, opened 2020-05-21
diff: Disable comparing modtimes when multimaster context is not found

vadmeste:
Fixes https://github.com/minio/mc/issues/3224


harshavardhana:
Can you take this out of draft @vadmeste ?

harshavardhana:
This is a good fix which is needed.

vadmeste:
> Can you take this out of draft @vadmeste ?

yes, so actually without this fix, mc diff works like before when time difference is enabled

Issue #2693, opened 2019-02-21
mc mb autocomplete adds additional whitespace


eco-minio:
## Expected behavior
No whitespace should be added after the autocomplete so that you can instantly type the bucket name.
## Actual behavior
When using new autocomplete feature, `mc mb pl<tab complete>` will fill as `mc mb play/ `.  The extra whitespace has to be deleted to add the bucket name.
## Steps to reproduce the behavior
Press tab after autocompleting `mc mb`.
## mc version
```
Version: 2019-02-20T22:21:50Z
Release-tag: RELEASE.2019-02-20T22-21-50Z
Commit-id: e538599e2ffa4ee8d8040d5cd14475b380da531a
```
## System information
N/A

vadmeste:
This will be fixed after we redo the autocomplete without posener/complete.

PR #3227, opened 2020-05-21
[WIP] completion: Escape special characters

vadmeste:
Fixes https://github.com/minio/mc/issues/2693

### DIFF
--- a/cmd/auto-complete.go
+++ b/cmd/auto-complete.go
@@ -1,5 +1,5 @@
 /*
- * MinIO Client (C) 2019 MinIO, Inc.
+ * MinIO Client (C) 2019-2020 MinIO, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"bytes"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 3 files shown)


cmd/auto-complete.go
```diff
@@ -26,6 +26,24 @@ import (
...
+// Predictor function and escapes all found predictions.
+type escaper struct {
+	complete.Predictor
+}
+
+func (e escaper) Predict(args complete.Args) []string {
+	predictions := e.Predictor.Predict(args)
+	for i := range predictions {
+		for _, escape := range escapeCharacters {
+			predictions[i] = strings.Replace(predictions[i], string(escape), "\\"+string(escape), -1)
```
klauspost, line 41: What if parts are already escaped?

So if you have `file\"name<autocomplete>` - will it not double escape the `\"`?

vadmeste: > So if you have `file\"name<autocomplete>` - will it not double escape the `\"`?

yes you are right, this is broken and does not work in all cases.

I think we have two options here:
- We need to send a PR to the original completion library (posener/complete) to accept some special config to make bash auto-escape filenames but the library supports multiple shells (zsh, fish, ..) at the same time and I am not sure if this is doable.

Another option is we do it and make our code more sophisticated.

I am trying the second option for now and see how easy it is

vadmeste:
It turned out this is more complex than I thought.

The problem here is that posener/complete does not understand an escaped space which is supposed to belong to the last written argument, it thinks the user is starting to write a new argument in the command line.

posener/complete doesn't know also how to complete an argument starting with a quote.

Doing this without posener/complete is much easier actually. 

I think we can easily can get rid of posener/complete and supports only bash for the time being. There won't be any breakage in mc side.

harshavardhana:
> It turned out this is more complex than I thought.
> 
> The problem here is that posener/complete does not understand an escaped space which is supposed to belong to the last written argument, it thinks the user is starting to write a new argument in the command line.
> 
> posener/complete doesn't know also how to complete an argument starting with a quote.
> 
> Doing this without posener/complete is much easier actually.
> 
> I think we can easily can get rid of posener/complete and supports only bash for the time being. There won't be any breakage in mc side.

Can you explain further? @vadmeste

vadmeste:
What `posener/complete` fails to do is:
1.  Complete filenames with spaces
```touch "/tmp/fi le"; mc ls "/tmp/fi<TAB>"```
2. Complete filenames starting with a quote:
Try 
```mc ls "/usr/bin/grou<TAB>```
3. Complete a file name with escaped characters
```mc ls /tmp/fi\ l<TAB>```

I investigated them today, all of above represents different issues in the posener/complete. Fixing them is so easy actually: just adding `-o filenames` in the mc autocompletion cmd in ~/.bashrc, but that won't work with posener/complete for other subsequent issues.

Dropping the library seems to be be better, and doesn't seem to need to take more than two days to do that.

harshavardhana:
Is there a plan to fix this ? Can we close this PR?

vadmeste:
> Is there a plan to fix this ? Can we close this PR?

Closing this. This is not a good solution. We need to avoid posener/complete. There is a simple solution for bash only which handles all edge cases that posener/complete is not able to do.

PR #3236, opened 2020-05-26
Avoid annoying error when alias has a leading slash

vadmeste:
```
mc config host list
mc config host add
mc config host remove
mc admin prometheus generate
```

The above mc commands generate an annoying error message complaining
about the leading slash in the alias argument.

Although alias does not support any slash or backslash, we allow the
user to enter it for convenience and the code will just clean it before
using it.

### DIFF
--- a/cmd/admin-prometheus-generate.go
+++ b/cmd/admin-prometheus-generate.go
@@ -138,7 +138,7 @@ func checkAdminPrometheusSyntax(ctx *cli.Context) {
 func generatePrometheusConfig(ctx *cli.Context) error {
 	// Get the alias parameter from cli
 	args := ctx.Args()
-	alias := args.Get(0)
+	alias := cleanAlias(args.Get(0))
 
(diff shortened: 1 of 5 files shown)


harshavardhana:
If there is a beginning `/` we can assume that its not an alias and the intention is to access the local filesystem @vadmeste

vadmeste:
> This would make it impossible to use /data because `data` might be an alias?

The change only relaxes the check on the alias in the following commands:

```
mc config host list
mc config host add
mc config host remove
mc admin prometheus generate
```

where the argument is always an alias and could never be a file system path

harshavardhana:
> > This would make it impossible to use /data because `data` might be an alias?
> 
> The change only relaxes the check on the alias in the following commands:
> 
> ```
> mc config host list
> mc config host add
> mc config host remove
> mc admin prometheus generate
> ```
> 
> where the argument is always an alias and could never be a file system path

Understood @vadmeste

PR #3246, opened 2020-06-07
simplify FS put behavior for partial files

harshavardhana:


kannappanr:
PTAL at build failures

harshavardhana:
> PTAL at build failures

#3247 should be taken first @kannappanr and then I shall rebase this PR.

cmd/client-fs.go
```diff
@@ -316,59 +306,34 @@ func (f *fsClient) put(reader io.Reader, size int64, metadata map[string][]strin
...
 	}
 
-	// Close the file before rename.
-	if e = partFile.Close(); e != nil {
-		return totalWritten, probe.NewError(e)
-	}
+	// Close the file before renaming, we need to do this
+	// specifically for windows users - windows explicitly
+	// disallows renames on Open() fd's by default.
+	tmpFile.Close()
```
vadmeste, line 336: Let's return an error only in this tmpFile.Close() (easier to debug when one user complains)

harshavardhana: @vadmeste why are we bothered about error from Close() in which scenario?

PR #3252, opened 2020-06-10
Filter X-Amz-Tagging-Count from metadata in Put/Copy

poornas:
x-amz-tagging-count is a computed field - it should not
be passed as part of metadata


harshavardhana:
@poornas can we close this PR if nothing that is going to be addressed here?

PR #3264, opened 2020-06-19
fix: support STS for admin commands

harshavardhana:


harshavardhana:
To generate sts

```py
import boto3
import json
from botocore.config import Config

def getTempCredentials():
    stsClient = boto3.client(
        'sts',
        region_name='us-east-1',
        use_ssl=True,
        endpoint_url='https://play.minio.io',
        aws_access_key_id='minio',
        aws_secret_access_key='minio123'
    )
    print(stsClient)
    response = stsClient.assume_role(
        RoleArn='arn:aws:iam::123456789012:user/minio',
        RoleSessionName='test',
        DurationSeconds=3600
    )
    return response

print(getTempCredentials())
```

PR #3266, opened 2020-06-22
Stops looping when api is notImplemented

### LINKED ISSUE #3237: Invalid query parameters when using `mc mirror` with Google Cloud
## Expected behavior

`mc mirror` reports no error when using it to sync files from a Google Cloud bucket to a local directory.

## Actual behavior

The following errors occur when I try to sync from a GCP bucket:

```
mc: <DEBUG> GET /?object-lock= HTTP/1.1
Host: <removed>.storage.googleapis.com
User-Agent: MinIO (linux; amd64) minio-go/v6.0.56 mc/2020-05-16T01:44:37Z
Authorization: AWS4-HMAC-SHA256 Credential=**REDACTED**/20200528/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=**REDACTED**
X-Amz-Content-Sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
X-Amz-Date: 20200528T033928Z
Accept-Encoding: gzip

mc: <DEBUG> HTTP/1.1 400 Bad Request
Content-Length: 177
Alt-Svc: h3-27=":443"; ma=2592000,h3-25=":443"; ma=2592000,h3-T050=":443"; ma=2592000,h3-Q050=":443"; ma=2592000,h3-Q049=":443"; ma=2592000,h3-Q048=":443"; ma=2592000,h3-Q046=":443"; ma=2592000,h3-Q043=":443"; ma=2592000,quic=":443"; ma=2592000; v="46,43"
Cache-Control: private, max-age=0
Content-Type: application/xml; charset=UTF-8
Date: Thu, 28 May 2020 03:39:28 GMT
Expires: Thu, 28 May 2020 03:39:28 GMT
Server: UploadServer
X-Guploader-Uploadid: AAANsUl0cuhxe3jjh0raKQusX8FZ7H8FSVtY6qpQcG0Rw3LUFfXMUbBRA3zcF1dEhJcykJEfyIubqmvmyLvGYWxx5A

<?xml version='1.0' encoding='UTF-8'?><Error><Code>InvalidArgument</Code><Message>Invalid argument.</Message><Details>Invalid query parameter(s): [object-lock]</Details></Error>mc: <DEBUG> Response Time:  111.443263ms

mc: <DEBUG> GET /?policy= HTTP/1.1
Host: <removed>.storage.googleapis.com
User-Agent: MinIO (linux; amd64) minio-go/v6.0.56 mc/2020-05-16T01:44:37Z
Authorization: AWS4-HMAC-SHA256 Credential=**REDACTED**/20200528/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=**REDACTED**
X-Amz-Content-Sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
X-Amz-Date: 20200528T033928Z
Accept-Encoding: gzip

mc: <DEBUG> HTTP/1.1 400 Bad Request
Content-Length: 172
Alt-Svc: h3-27=":443"; ma=2592000,h3-25=":443"; ma=2592000,h3-T050=":443"; ma=2592000,h3-Q050=":443"; ma=2592000,h3-Q049=":443"; ma=2592000,h3-Q048=":443"; ma=2592000,h3-Q046=":443"; ma=2592000,h3-Q043=":443"; ma=2592000,quic=":443"; ma=2592000; v="46,43"
Cache-Control: private, max-age=0
Content-Type: application/xml; charset=UTF-8
Date: Thu, 28 May 2020 03:39:28 GMT
Expires: Thu, 28 May 2020 03:39:28 GMT
Server: UploadServer
X-Guploader-Uploadid: AAANsUkn3jIkhUB_TiGLsD5-jD_hOqbuYIRm3aDDseOzO_qQ2Ldq2TH-RZ-bl9TkYeZ-fzj9UhS-v0RP4htkJ_w4pw

<?xml version='1.0' encoding='UTF-8'?><Error><Code>InvalidArgument</Code><Message>Invalid argument.</Message><Details>Invalid query parameter(s): [policy]</Details></Error>mc: <DEBUG> Response Time:  89.870126ms```
```

Please note that these errors don't block the sync process: files are being downloaded correctly from the bucket.

 1. Create an S3 bucket in Google Cloud using "fine-grained access control" (I did not test it with bucket level access control.)
 2. Mirror that bucket to a local directory like:
```console
mc mirror gcs/<my_bucket>/some/path /my/local/path --json --debug --overwrite
```

## mc --version

```
mc version RELEASE.2020-05-16T01-44-37Z
```

ebozduman:
Fixes #3237
mc mirror fails for GCS, since GCS is not compatible with Amazon S3. So, some apis like Stat, List etc. fails with `notImplemented` error message:
`"A header or query you provided requested a function that is not implemented."`

During `mc mirror`, we hit the same `notImplemented` issue for `Stat` api, and continue to `Compare` to check if there is a difference between source and destination. We loop for 30 seconds meaninglessly even though we hit the `notImplemented` error message. 
So, the fix is to stop trying as soon as we hit `notImplemented` condition.

### DIFF
--- a/cmd/difference.go
+++ b/cmd/difference.go
@@ -341,7 +341,15 @@ func difference(ctx context.Context, sourceClnt, targetClnt Client, sourceURL, t
 			if err != nil {
 				// handle this specifically for filesystem related errors.
 				switch err.ToGoError().(type) {
-				case PathNotFound, PathInsufficientPermission:
+				case PathNotFound, PathInsufficientPermission, APINotImplemented:
+					diffCh <- diffMessage{
+						Error: err,
+					}
+					return
+				}
+
+				stopAtErr := "A header or query you provided requested a function that is not implemented."
+				if strings.Contains(err.String(), stopAtErr) {
 					diffCh <- diffMessage{
 						Error: err,
 					}


ebozduman:
I also piggybacked a change to remove the 2nd definition of ctx and cancelMirror() function. One definition was in mainMirror() function (line# 915), and the second one was in runMirror() function (line# 746) mirror-main.go.

cmd/mirror-main.go
```diff
@@ -802,8 +802,8 @@ func runMirror(srcURL, dstURL string, cli *cli.Context, encKeyDB map[string][]pr
 		os.Exit(globalErrorExitStatus)
 	}()
 
-	ctx, cancelMirror := context.WithCancel(globalContext)
-	defer cancelMirror()
+	// ctx, cancelMirror := context.WithCancel(globalContext)
```
klauspost, line 805: (Remove if not needed, but see other)

cmd/mirror-main.go
```diff
@@ -743,7 +743,7 @@ func getEventPathURLWin(srcURL, eventPath string) string {
 }
 
 // runMirror - mirrors all buckets to another S3 server
-func runMirror(srcURL, dstURL string, cli *cli.Context, encKeyDB map[string][]prefixSSEPair) bool {
+func runMirror(ctx context.Context, cancelMirror context.CancelFunc, srcURL, dstURL string, cli *cli.Context, encKeyDB map[string][]prefixSSEPair) bool {
```
klauspost, line 746: Not really a fan of sending a `context.CancelFunc` What is the reason for that?

The typical assumption around contexts are that they are immutable when passed down, meaning that `runMirror` shouldn't be modifying the sent context. I assume this is what the cancelfunc does.

If you want to signal a cancelation upstream you typically so that by returning an error. Is that feasible?

cmd/mirror-main.go
ebozduman: ok

harshavardhana:
Actually the first fix should happen in minio-go where it shouldn't retry NotImplemented calls @ebozduman and then we can add a top level retry to be avoided in mc.

cmd/mirror-main.go
ebozduman: @klauspost, 

Defining `ctx` and `cancel<mc-command>` in `main<mc-command>` function is the standard way for almost all mc commands.

For `mc mirror` command though, `ctx` and `cancelMirror` were defined in 2 files twice in `mainMirror()` and in `runMirror()` functions. I noticed the duplicate definition and fixed it by removing the 2nd one in `runMirror` function. However, removing 2nd definition caused a compile error and needed to be fixed by passing it through `runMirror().

It is totally an unrelated change and not needed to fix issue #3237. I can remove it if it is desired, since it is an independent problem.

klauspost: It is still a weird pattern, but I guess it works.

PR #3272, opened 2020-06-23
Add replication commands

poornas:
to manage replication targets on MinIO server

This PR adds 
`mc admin bucket replication set|get|remove`
`mc admin bucket remote`
`mc replicate` commands

Please see [minio PR](https://github.com/minio/minio/blob/c1e3495f82a664a46a8835741b990a2b1d945cc0/docs/bucket/replication/README.md) for the flow.


donatello:
@poornas I'm just waiting for this to pass CI before reviewing. Let me know if I should review before that also.

poornas:
> @poornas I'm just waiting for this to pass CI before reviewing. Let me know if I should review before that also.

This command has changed a little, will push another commit later today that will have additional commands to generate replication arn and use that to configure replication on a bucket. The CI will continue to fail because minio madmin pkg and minio-go v7 dependencies are not from released versions.

poornas:
> > @poornas I'm just waiting for this to pass CI before reviewing. Let me know if I should review before that also.
> 
> This command has changed a little, will push another commit later today that will have additional commands to generate replication arn and use that to configure replication on a bucket. The CI will continue to fail because minio madmin pkg and minio-go v7 dependencies are not from released versions.

@donatello, this can be reviewed now - testing this PR will require the minio-go PR and minio API changes referenced below.

poornas:
> to manage replication targets on MinIO server

`mc admin bucket replication`
`mc admin bucket remote`
`mc replicate` 
commands have been added. Please see [minio PR](https://github.com/minio/minio/blob/c1e3495f82a664a46a8835741b990a2b1d945cc0/docs/bucket/replication/README.md) for flow

go.mod
```diff
@@ -2,34 +2,36 @@ module github.com/minio/mc
 
 go 1.13
 
+replace github.com/minio/minio => ../minio
```
vadmeste, line 5: Could you remove replace tag in this file ?

poornas: this is needed until minio PR is merged

cmd/admin-bucket-replication-set.go
```diff
@@ -0,0 +1,202 @@
...
+	// Get the alias parameter from cli
+	args := ctx.Args()
+	aliasedURL := args.Get(0)
+	// Create a new MinIO Admin Client
+	client, cerr := newAdminClient(aliasedURL)
+	fatalIf(cerr, "Unable to initialize admin connection.")
+
+	sourceBucket, replTarget, perr := fetchReplicationTarget(ctx)
+	fatalIf(probe.NewError(perr), "Unable to parse input arguments.")
+	fatalIf(probe.NewError(client.SetBucketReplicationTarget(globalContext, sourceBucket, replTarget)).Trace(args...), "Cannot add new replication target")
```
donatello, line 188: Just a q: why is replication not supported in fs mode? It should be noted in the [server guide](https://github.com/minio/minio/blob/c1e3495f82a664a46a8835741b990a2b1d945cc0/docs/bucket/replication/README.md).

harshavardhana: replication needs versioning enabled on both source and destination - FS mode doesn't support versioning.

donatello: Ah ok, i was able to enable versioning via mc for fs mode - the server should return an error for that.

harshavardhana: yes that should be the case @donatello

harshavardhana:
> Looks like `mc admin bucket replication get` and `mc admin bucket remote` are similar in functionality. Since only one replication target may be configured per bucket (at least according to the implemented mc interface here), maybe the ARN can be returned in the get call and the remote command does not seem necessary (especially since it needs an extra url argument).

I asked poorna to return the ARN in the `mc admin bucket replication set` itself

donatello:
> I asked poorna to return the ARN in the `mc admin bucket replication set` itself 

I'm actually talking about the get call and the remote call.

harshavardhana:
> > I asked poorna to return the ARN in the `mc admin bucket replication set` itself
> 
> I'm actually talking about the get call and the remote call.

What I meant is `mc admin bucket replicate remote` is not needed - I didn't elaborate that.

poornas:
> Looks like `mc admin bucket replication get` and `mc admin bucket remote` are similar in functionality. Since only one replication target may be configured per bucket (at least according to the implemented mc interface here), maybe the ARN can be returned in the get call and the remote command does not seem necessary (especially since it needs an extra url argument).

yes, we can eliminate mc admin bucket remote; will send a PR to minio to return replication Arn in addition.

harshavardhana:
> > Looks like `mc admin bucket replication get` and `mc admin bucket remote` are similar in functionality. Since only one replication target may be configured per bucket (at least according to the implemented mc interface here), maybe the ARN can be returned in the get call and the remote command does not seem necessary (especially since it needs an extra url argument).
> 
> yes, we can eliminate mc admin bucket remote; will send a PR to minio to return replication Arn in addition.

WE don't need any special changes on MinIO - in lieu of time we should just merge what you do with `mc admin bucket remote` right after you call `mc admin bucket replication set`

poornas:
> > > Looks like `mc admin bucket replication get` and `mc admin bucket remote` are similar in functionality. Since only one replication target may be configured per bucket (at least according to the implemented mc interface here), maybe the ARN can be returned in the get call and the remote command does not seem necessary (especially since it needs an extra url argument).
> > 
> > 
> > yes, we can eliminate mc admin bucket remote; will send a PR to minio to return replication Arn in addition.
> 
> WE don't need any special changes on MinIO - in lieu of time we should just merge what you do with `mc admin bucket remote` right after you call `mc admin bucket replication set`

currently replication Arn is not returned in the response.fixing that [here](https://github.com/minio/minio/pull/10114)

cmd/replicate-main.go
```diff
@@ -0,0 +1,182 @@
...
+   {{end}}
+ DESCRIPTION:
+   Manage replication configuration on the bucket.
+ 
+ EXAMPLES:
+   1. Add a replication configuration to the "Devices" bucket
+      {{.Prompt}} {{.HelpName}} myminio/Devices --config "config.json"
+
+   2. List replication configuration on the "Devices" bucket
+      {{.Prompt}} {{.HelpName}} myminio/Devices"
```
vadmeste, line 58: s/"//

cmd/replicate-main.go
```diff
@@ -0,0 +1,182 @@
...
+
+   3. Clear replication configuration on the "Devices" bucket
+      {{.Prompt}} {{.HelpName}} myminio/Devices" --clear
```
vadmeste, line 61: s/"//

cmd/replicate-main.go
```diff
@@ -0,0 +1,182 @@
...
+   {{.HelpName}} [COMMAND FLAGS] TARGET
+ 
+ FLAGS:
+   {{range .VisibleFlags}}{{.}}
+   {{end}}
+ DESCRIPTION:
+   Manage replication configuration on the bucket.
+ 
+ EXAMPLES:
+   1. Add a replication configuration to the "Devices" bucket
```
vadmeste, line 54: Let's have Devices bucket names to be low capital

cmd/admin-bucket-replication-get.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"github.com/fatih/color"
+	"github.com/minio/cli"
+	"github.com/minio/mc/pkg/probe"
+	"github.com/minio/minio/pkg/console"
+)
+
+var adminBucketReplicationGetCmd = cli.Command{
+	Name:   "get",
+	Usage:  "get bucket replication target",
```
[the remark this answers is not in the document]
poornas: this is actually the replication endpoint not the configuration, leaving it as target here.

cmd/admin-bucket-replication-set.go
```diff
@@ -0,0 +1,202 @@
...
+			"Unrecognized bucket path style. Valid options are `[on,off, auto]`.")
+	}
+	u, cerr := url.Parse(replicaURL)
+	if cerr != nil {
+		fatalIf(probe.NewError(cerr), "Malformed replication target URL")
+	}
+	isSSL := u.Scheme == "https"
+	accessKey := u.User.Username()
+	secretKey, _ := u.User.Password()
+	replicaBucket := strings.TrimPrefix(u.Path, slashSeperator)
```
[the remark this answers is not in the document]
poornas: that's a good idea, done.

cmd/replicate-main.go
```diff
@@ -0,0 +1,182 @@
...
+	switch r.Op {
+	case "Add":
+		return console.Colorize("ReplicateMessage", "Replication configuration added successfully to "+r.Source+".")
+	case "Remove":
+		return console.Colorize("ReplicateMessage", "Replication configuration removed successfully from "+r.Source+".")
+	default:
+		if r.ReplicationConfig.Empty() {
+			return console.Colorize("ReplicateNMessage", "No replication configuration found for "+r.Source+".")
+		}
+		msgBytes, e := json.MarshalIndent(r.ReplicationConfig, "", " ")
```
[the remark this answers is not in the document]
poornas: the output of this command is the replication config - with or without the --json flag. Hence the marshalling

Issue #3258, opened 2020-06-16
[resolved] mc fails downloading objects that have incomplete preserved attributes metadata


Josua-SR:
## Expected behavior

When fetching objects with `mc mirror -a` or `mc cp -a` from a minio bucket to the local filesystem, minio should use those parts of *X-Amz-Meta-Mc-Attrs* that are present, and use defaults for those missing.

## Actual behavior

`mc` throws an error and aborts the download of each affected file leaving behind a *.part*:

`mc: <ERROR> Failed to copy `secret`. strconv.ParseUint: parsing "": invalid syntax`

## Steps to reproduce the behavior

It's a little tricky and reveals where I come from intentionally:

Copy a new object to a minio bucket by manually adding *X-Amz-Meta-Mc-Attrs* to the object metadata:
`mc cp --attr "X-Amz-Meta-Mc-Attrs=atime:0/ctime:0/mtime:0" file instance/bucket/`

Then fetch it with `-a`:
`mc cp -a instance/bucket/file copy`

## mc --version
- *mc version RELEASE.2020-05-28T23-43-36Z*

## System information

openSUSE Tumbleweed ...

Josua-SR:
It could be argued that I am abusing internal state here - however sometimes there can be a need to customize the file ownership, mode or change times the easy way, without jumping through local filesystem hoops.

And then one can also argue that sometimes only parts of the usually preserved attributes are relevant and therefore to be added as metadata.

Therefore, I believe the MinIO Client should be partially preserve attributes without failing - by using defaults where metadata is missing.

Josua-SR:
And this is how I would solve it:
https://github.com/Josua-SR/mc/commit/7263606831ba29d7489b8c3183f467257bce0bae

So far only done for uid, gid and mode - timestamps to be revisited ...

Josua-SR:
cool! :)

PR #3282, opened 2020-07-01
Fixes copy with partial metadata failure

ebozduman:
Fixes issue #3258 

When user tries to copy using `-attr` option to specify partial metadata information only, mc fails. since the code assumes all metadata is always going to be provided. This PR provides support for specifying partial metadata.

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -224,7 +224,7 @@ func (f *fsClient) Watch(ctx context.Context, options WatchOptions) (*WatchObjec
 
 func preserveAttributes(fd *os.File, attr map[string]string) *probe.Error {
 	if val, ok := attr["mode"]; ok {
-		mode, e := strconv.ParseUint(val, 10, 32)
+		mode, e := strconv.ParseUint(val, 0, 32)
 		if e != nil {
 			return probe.NewError(e)
 		}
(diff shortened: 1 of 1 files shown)


cmd/client-fs.go
```diff
@@ -223,29 +223,42 @@ func (f *fsClient) Watch(ctx context.Context, options WatchOptions) (*WatchObjec
 }
 
 func preserveAttributes(fd *os.File, attr map[string]string) *probe.Error {
-	mode, e := strconv.ParseUint(attr["mode"], 10, 32)
-	if e != nil {
-		return probe.NewError(e)
-	}
+	if val, ok := attr["mode"]; ok {
+		mode, e := strconv.ParseUint(val, 10, 32)
```
donatello, line 227: Since the mode is usually expressed in base 8 as `0644` or `0777` ParseUint should specify 0 as the base, to allow interpretation based on the prefix:
```suggestion
		mode, e := strconv.ParseUint(val, 0, 32)
```

Otherwise I think there is a bug: `0777` interpreted as base 10 is not the same as `0777` interpreted as base  8.

cmd/client-fs.go
```diff
@@ -223,29 +223,42 @@ func (f *fsClient) Watch(ctx context.Context, options WatchOptions) (*WatchObjec
...
+			return probe.NewError(e)
+		}
+		gidExists = true
 	}
 
 	// Attempt to change the owner.
-	if e := fd.Chown(uid, gid); e != nil {
-		return probe.NewError(e)
+	if gidExists && uidExists {
+		if e := fd.Chown(uid, gid); e != nil {
```
krisis, line 259: for `Chown`, we can pass `-1` for whichever value was *not* specified.
e.g if uid was *not* specified,
```go
if e := fd.Chown(-1, gid); e != nil {
...
```

cmd/client-fs.go
ebozduman: @donatello 
Great catch? This is the old code and already existing bug.
Fixed it.

cmd/client-fs.go
ebozduman: Ok.
Incorporated your comment.

PR #3294, opened 2020-07-14
fix attr argument parsing - support quoted keys and values

dafanasiev:
Minio cli unable to parse `--attr` arg with value like `Content-Disposition='attachment; filename="README.md"'`.
This PR fix that issue.

howto check (example):
```sh
echo 'test-test-test' > README.md
$MINIO_CLI cp --attr "Content-Disposition='attachment; filename="README.md"'" $PWD/README.md minio/0123456789-README.md
```


harshavardhana:
PTAL @vadmeste

PR #3302, opened 2020-07-16
Revert "cli: Load global flags in initBeforeRunningCmd (#3283)"

harshavardhana:


harshavardhana:
This PR also fixes all the broken tests after 712ff333195eeeb11215e092442f9fa55519d275 change

cmd/config-host.go
```diff
@@ -27,6 +27,7 @@ var configHostCmd = cli.Command{
 	Name:   "host",
 	Usage:  "add, remove and list hosts in configuration file",
 	Action: mainConfigHost,
+	Before: initBeforeRunningCmd,
```
vadmeste, line 30: I think no point in adding this, because initBeforeRunningCmd() will be called again in any of the sub-commands

harshavardhana:
> One comment, also what does the PR fix ?

https://subnet.min.io/support/957 this and @vadmeste and the changes in this PR are needed without them `mc` functional tests are not running and have been broken since https://github.com/minio/mc/commit/712ff333195eeeb11215e092442f9fa55519d275

PR #3304, opened 2020-07-16
Implement rewind flag in ls and cp commands

vadmeste:
1. --rewind for ls and cp will back in time to select the object versions at that time 
2. ls with --versions will list all versions and will print delete markers with a strike through
3. cp will just ignore if the selected object version is a delete marker

### DIFF
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -69,6 +69,16 @@ type ListOptions struct {
 	showDir           DirOpt
 }
 
+// CopyOptions holds options for copying operation
+type CopyOptions struct {
+	versionID        string
+	size             int64
+	srcSSE, tgtSSE   encrypt.ServerSide
+	metadata         map[string]string
+	disableMultipart bool
+	isPreserve       bool
+}
+
 // Client - client interface
 type Client interface {
 	// Common operations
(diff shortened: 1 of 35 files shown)


cmd/client.go
```diff
@@ -61,13 +88,14 @@ type Client interface {
 	SetAccess(ctx context.Context, access string, isJSON bool) *probe.Error
 
 	// I/O operations
-	Copy(ctx context.Context, source string, size int64, progress io.Reader, srcSSE, tgtSSE encrypt.ServerSide, metadata map[string]string, disableMultipart, isPreserve bool) *probe.Error
+	Copy(ctx context.Context, source, versionID string, size int64, progress io.Reader, srcSSE, tgtSSE encrypt.ServerSide, metadata map[string]string, disableMultipart, isPreserve bool) *probe.Error
```
klauspost, line 91: Time to have options in a struct?

kannappanr:
Please resolve the conflicts, otherwise LGTM

cmd/client.go
harshavardhana: 🍰

vadmeste: > Time to have options in a struct?

Done

PR #3311, opened 2020-07-20
Move from mc config host to alias command

vadmeste:
```
mc alias set
   alias remove
   alias list
```

are new commands that replace mc config host sub-commands.

mc config host will still be available but hidden for backward
compatiblity. The JSON output is also kept if the user is running mc
config host command.


vadmeste:
PR updated and ready for review

vadmeste:
> PR updated and ready for review

Conflict fixed

cmd/config-deprecated.go
```diff
@@ -0,0 +1,100 @@
+/*
+ * MinIO Client (C) 2014, 2015 MinIO, Inc.
```
kannappanr, line 2:
```suggestion
 * MinIO Client (C) 2020 MinIO, Inc.
```

kannappanr: If you have simply moved the file, we can ignore it

cmd/alias-list.go
```diff
@@ -43,110 +45,122 @@ FLAGS:
...
 
-	// Additional command speific theme customization.
+	// Additional command specific theme customization.
 	console.SetColor("Alias", color.New(color.FgCyan, color.Bold))
 	console.SetColor("URL", color.New(color.FgYellow))
 	console.SetColor("AccessKey", color.New(color.FgCyan))
 	console.SetColor("SecretKey", color.New(color.FgCyan))
 	console.SetColor("API", color.New(color.FgBlue))
-	console.SetColor("Lookup", color.New(color.FgCyan))
+	console.SetColor("Path", color.New(color.FgCyan))
```
krisis, line 75: shouldn't we conditionally set "Path" or "Lookup" based on `deprecated` flag?

cmd/alias-main.go
```diff
@@ -23,29 +23,44 @@ import (
...
 }
 
-// mainConfigHost is the handle for "mc config host" command.
-func mainConfigHost(ctx *cli.Context) error {
+// mainAlias is the handle for "mc alias" command. provides sub-commands which write configuration data in json format to config file.
+func mainAlias(ctx *cli.Context) error {
 	cli.ShowCommandHelp(ctx, ctx.Args().First())
 	return nil
-	// Sub-commands like "remove", "list" have their own main.
+	// Sub-commands like add, list and remove have their own main.
```
krisis, line 59: minor nit:
```suggestion
	// Sub-commands like set, list and remove have their own main.
```

cmd/alias-remove.go
```diff
@@ -61,28 +63,31 @@ func checkConfigHostRemoveSyntax(ctx *cli.Context) {
...
 
-	console.SetColor("HostMessage", color.New(color.FgGreen))
+	console.SetColor("AliasMessage", color.New(color.FgGreen))
 
 	args := ctx.Args()
 	alias := args.Get(0)
-	removeHost(alias) // Remove a host.
+
+	aliasMsg := removeAlias(alias) // Remove an alias
+	aliasMsg.op = "remove"
```
krisis, line 76: is there a reason for not setting `aliasMsg.op` inside `removeAlias` function?

cmd/alias-list.go
vadmeste: > shouldn't we conditionally set "Path" or "Lookup" based on `deprecated` flag?

Correct, but I guess the deprecated code will be removed later. I don't think this is needed.

cmd/alias-remove.go
vadmeste: > is there a reason for not setting `aliasMsg.op` inside `removeAlias` function?

Setting the operation name outside is actually only needed in alias-set.go, because I renamed `add` to `set` and I wanted that `mc config host add` to return 'json.op = add' and `mc alias set` to return 'json.op = set'.

This is not a big deal for me. I can change it if you want.

kannappanr:
@vadmeste PR LGTM. Please resolve the conflict, so it can be merged

vadmeste:
> @vadmeste PR LGTM. Please resolve the conflict, so it can be merged

Done

PR #3313, opened 2020-07-22
Fixes wrong destination path

### LINKED ISSUE #3259: Mirror Command uses absolute path when mirroring to minio root
```
MC_HOST_source=xxx
MC_HOST_dest=yyy
mc mirror source /home/user/mirror
mc mirror /home/user/mirror dest
```
I'm aware that you can do server to server transfers (though haven't tested them).  This is about performing a backup and later a restore.

## Expected behavior
The bucket structure from the source server, which is captured inside of the mirror directory should be copied to the dest server

## Actual behavior
mc mirror throws an error because it tries to create a bucket called /home/user/mirror

## Workaround
Don't use root copies, perform one mirror operation per bucket, though this is not idea.

## mc --version
mc version RELEASE.2020-05-05T06-03-47Z

## System information
Kubernetes 1.18.3 using official minio/mc images.

ebozduman:
Fixes #3259 

Fixes the issue with the creation of destination path where destination was created with full absolute path instead of the base file name.

I tested the fix for regression and could not find anything.
It is a pretty simple fix, but I still would like to remind and suggest the reviewers to pay attention for regression risk.

### DIFF
--- a/cmd/mirror-main.go
+++ b/cmd/mirror-main.go
@@ -748,7 +748,7 @@ func runMirror(ctx context.Context, cancelMirror context.CancelFunc, srcURL, dst
 	// Check if we are only trying to mirror one bucket from source.
 	if dstClt.GetURL().Type == objectStorage &&
 		dstClt.GetURL().Path == string(dstClt.GetURL().Separator) && !mirrorAllBuckets {
-		dstURL = urlJoinPath(dstURL, filepath.Base(srcClt.GetURL().Path))
+		dstURL = urlJoinPath(dstURL, srcClt.GetURL().Path+"/")
 
 		dstClt, err = newClient(dstURL)
 		fatalIf(err, "Unable to initialize `"+dstURL+"`.")
--- a/cmd/mirror-main.go
+++ b/cmd/mirror-main.go
@@ -745,15 +745,6 @@ func runMirror(ctx context.Context, cancelMirror context.CancelFunc, srcURL, dst
 		(srcClt.GetURL().Type == objectStorage &&
 			srcClt.GetURL().Path == string(srcClt.GetURL().Separator))
 
-	// Check if we are only trying to mirror one bucket from source.
-	if dstClt.GetURL().Type == objectStorage &&
-		dstClt.GetURL().Path == string(dstClt.GetURL().Separator) && !mirrorAllBuckets {
-		dstURL = urlJoinPath(dstURL, srcClt.GetURL().Path+"/")
-
-		dstClt, err = newClient(dstURL)
-		fatalIf(err, "Unable to initialize `"+dstURL+"`.")
-	}
-
 	// Create a new mirror job and execute it
 	mj := newMirrorJob(srcURL, dstURL, mirrorOptions{
 		isFake:           cli.Bool("fake"),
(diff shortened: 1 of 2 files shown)


cmd/mirror-main.go
```diff
@@ -748,7 +748,7 @@ func runMirror(ctx context.Context, cancelMirror context.CancelFunc, srcURL, dst
 	// Check if we are only trying to mirror one bucket from source.
 	if dstClt.GetURL().Type == objectStorage &&
 		dstClt.GetURL().Path == string(dstClt.GetURL().Separator) && !mirrorAllBuckets {
-		dstURL = urlJoinPath(dstURL, srcClt.GetURL().Path)
+		dstURL = urlJoinPath(dstURL, filepath.Base(srcClt.GetURL().Path))
```
krisis, line 751: We should copy top-level directories from the source directory into the destination object storage as buckets. But this change mirrors everything to a single bucket.

ebozduman: @krisis,
Fixed.

kannappanr:
ping @krisis @poornas PTAL

cmd/mirror-main.go
```diff
@@ -748,7 +748,7 @@ func runMirror(ctx context.Context, cancelMirror context.CancelFunc, srcURL, dst
 	// Check if we are only trying to mirror one bucket from source.
 	if dstClt.GetURL().Type == objectStorage &&
 		dstClt.GetURL().Path == string(dstClt.GetURL().Separator) && !mirrorAllBuckets {
-		dstURL = urlJoinPath(dstURL, srcClt.GetURL().Path)
+		dstURL = urlJoinPath(dstURL, srcClt.GetURL().Path+"/")
```
harshavardhana, line 751:
```suggestion
		dstURL = urlJoinPath(dstURL, srcClt.GetURL().Path+string(srcClt.GetURL().Separator))
```

poornas:
`mc mirror` on absolute path still seems incorrect - bucket should be created on final directory on path not the entire url. `mc mirror ~/small` or within the directory behaves correctly.
```
mc mirror /home/kp/data/small source
[2020-08-19 14:44:41 PDT]     26B bucket/issue
[2020-08-19 14:44:50 PDT]     26B bucket/pre/issue
[2020-08-19 14:45:03 PDT]     26B bucket2/issue
[2020-08-19 14:45:07 PDT]     26B bucket2/ss
[2020-08-19 14:45:12 PDT]     26B bucket2/ssd/issue
[2020-08-19 14:50:03 PDT]     26B home/kp/data/small/a.txt
[2020-08-19 14:50:03 PDT]     26B home/kp/data/small/b.txt
[2020-08-19 14:50:03 PDT]     26B home/kp/data/small/bucket/issue
[2020-08-19 14:50:03 PDT]     26B home/kp/data/small/bucket/pre/issue

```

ebozduman:
Fixed the issue such that the backup and restore mirrors the same way generating the same directory structure.
@krisis @poornas @harshavardhana PTAL again.

cmd/mirror-main.go
ebozduman: Replaced/

PR #3318, opened 2020-07-24
Change mc stat output format for bucket info

poornas:
`mc stat myminio/bucket` should return bucket metadata instead of generic folder like info returned today.

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -1178,3 +1178,11 @@ func (f *fsClient) DeleteEncryption(ctx context.Context) *probe.Error {
 		APIType: "filesystem",
 	})
 }
+
+// Gets bucket info
+func (f *fsClient) GetBucketInfo(ctx context.Context) (BucketInfo, *probe.Error) {
+	return BucketInfo{}, probe.NewError(APINotImplemented{
+		API:     "GetBucketInfo",
+		APIType: "filesystem",
+	})
+}
--- a/cmd/stat.go
+++ b/cmd/stat.go
@@ -340,7 +340,6 @@ func (v bucketInfoMessage) String() string {
 	keyStr := getKey(&ClientContent{URL: v.Metadata.URL, Type: v.Metadata.Type})
 	key := fmt.Sprintf("%-10s: %s", "Name", keyStr)
 	fmt.Fprintln(&b, console.Colorize("Name", key))
-	fmt.Fprintf(&b, fmt.Sprintf("%-10s: %s \n", "Date", v.Metadata.Date.Local().Format(printDate)))
 	fmt.Fprintf(&b, fmt.Sprintf("%-10s: %-6s \n", "Size", humanize.IBytes(uint64(v.Metadata.Size))))
 	fType := func() string {
 		if v.Metadata.Type.IsDir() {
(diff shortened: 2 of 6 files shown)


cmd/client-fs.go
```diff
@@ -1194,3 +1194,11 @@ func (f *fsClient) DeleteEncryption(ctx context.Context) *probe.Error {
 		APIType: "filesystem",
 	})
 }
+
+// Gets bucket info
+func (f *fsClient) GetInfo(ctx context.Context) (BucketInfo, *probe.Error) {
```
vadmeste, line 1199: GetBucketInfo() as a function name looks better.

cmd/stat.go
```diff
@@ -215,5 +231,162 @@ func statURL(ctx context.Context, targetURL, versionID string, isIncomplete, isR
...
+	// Disable escaping special chars to display XML tags correctly
+	enc.SetEscapeHTML(false)
+
+	fatalIf(probe.NewError(enc.Encode(v)), "Unable to marshal into JSON.")
+	return buf.String()
+
+}
+
+func (v bucketInfoMessage) String() string {
+	var b = &strings.Builder{}
```
vadmeste, line 315: Any reason why strings.Builder is used here ?

cmd/stat.go
```diff
@@ -202,10 +209,19 @@ func statURL(ctx context.Context, targetURL, versionID string, isIncomplete, isR
 			}
 		}
 
-		_, stat, err := url2Stat(ctx, url, versionID, true, encKeyDB, time.Time{})
+		clnt, stat, err := url2Stat(ctx, url, versionID, true, encKeyDB, time.Time{})
 		if err != nil {
 			stat = content
 		}
+		// if stat is on a bucket and non-recursive mode, serve the bucket metadata
+		if clnt != nil && clnt.GetURL().Type == objectStorage && !isRecursive && stat.Type.IsDir() {
```
vadmeste, line 217: I guess you can remove 'clnt.GetURL().Type == objectStorage' and 'stat.Type.IsDir()' here since you are testing for 'err == nil' here

cmd/client-fs.go
harshavardhana: Or generalize it GetStatInfo() which returns StatInfo

cmd/stat.go
harshavardhana: strings.Builder is an efficient way to build strings than using `+` @vadmeste

harshavardhana, line 315:
```
var b strings.Builder
```

Is the idiomatic style @poornas you need to allocate a strings builder. 

>  The zero value is ready to use. 

https://golang.org/pkg/strings/#Builder

docs/minio-client-complete-guide.md
```diff
@@ -1479,11 +1479,25 @@ ENVIRONMENT VARIABLES:
 
 ```
 mc stat play/mybucket
-Name      : mybucket/
-Date      : 2018-02-06 18:06:51 PST
-Size      : 0B
-Type      : folder
-```
+Encryption: Not Set
+
```
harshavardhana, line 1483: may need a better formatting @poornas - I think the previous approach is fine instead these features could be `Metadata` field..

```
~ mc stat play/testbucket
Name      : testbucket/
Date      : 2020-08-06 21:09:46 PDT 
Size      : 0 B
Type      : folder 
Metadata  :
  Versioning: Enabled
  LockConfiguration: Configured
  Replication: Disabled
  Notification: Configured
  Policy: upload
  Location: us-west-1
  Tagging: key1=value1, key2=value2
  ILM: Configured
```

harshavardhana: ^^ is this discussed? @poornas @kannappanr - IMO we should avoid different format

harshavardhana:
@poornas if any of the value is not enabled we don't need to print it

```
~ mc stat play/testbucket
Name      : testbucket/
Date      : 1969-12-31 16:00:00 PST 
Size      : 0 B    
Type      : folder 
Metadata  :
  Encryption: Not Set
  Versioning: Un-versioned
  LockConfiguration: 

  Notification: Unset
  Replication: Enabled
  Policy: none
  Location: us-east-1
  Tagging: 
  ILM: Set
```

We should also make sure to align the output properly here and don't print things that are not configured. 

We should also fix the date if possible in this call or avoid printing it? since 1969 looks weird. 

> Date      : 1969-12-31 16:00:00 PST

vadmeste:
@poornas Replication always shows 'Enabled' for all buckets

poornas:
> @poornas Replication always shows 'Enabled' for all buckets

fixed

harshavardhana:
Can you resolve the conflicts @poornas ?

vadmeste:
@poornas I think `mc stat` does not show info anymore
```
11:50 $ mc stat play/foobar-bucket/
11:50 $ 
```

poornas:
> @poornas I think `mc stat` does not show info anymore
> 
> ```
> 11:50 $ mc stat play/foobar-bucket/
> 11:50 $ 
> ```

this seems to be an issue with master branch also, not introduced by this PR. Will fix it separately.

harshavardhana:
> > @poornas I think `mc stat` does not show info anymore
> > ```
> > 11:50 $ mc stat play/foobar-bucket/
> > 11:50 $ 
> > ```
> 
> this seems to be an issue with master branch also, not introduced by this PR. Will fix it separately.

Seems to work fine with master @poornas 

```
~ mc stat play/testbucket
Name      : testbucket/
Date      : 1969-12-31 16:00:00 PST 
Size      : 0 B    
Type      : folder 

```

harshavardhana:
NOTE: we should remove this

> Date      : 1969-12-31 16:00:00 PST

poornas:
> NOTE: we should remove this
> 
> > Date      : 1969-12-31 16:00:00 PST

this is being done in this PR
```
➜  mc git:(bucketinfo) mc stat play/foobar-bucket
Name      : foobar-bucket/
Size      : 0 B    
Type      : folder 
Metadata  :
  Versioning: Un-versioned
  Location: us-east-1
  Policy: none
```


> Seems to work fine with master @poornas

`mc stat play/foobar-bucket/` with a trailing slash treats / as a directory prefix - hence nothing is displayed as bucket in this case is empty. This is the same behavior with master also and not a bug imo since `mc stat` should work both for prefix stat as well as stat on buckets. See [here](https://github.com/minio/mc/blob/master/cmd/client-s3.go#L2146)

harshavardhana (approved):

vadmeste (approved):
LGTM

PR #3320, opened 2020-07-24
move 'mc ilm' to 'mc bucket ilm'

harshavardhana:
update relevant docs and use the same
usage message consistently

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -339,7 +339,6 @@ var appCmds = []cli.Command{
 	diffCmd,
 	rmCmd,
 	eventCmd,
-	ilmCmd,
 	watchCmd,
 	policyCmd,
 	tagCmd,
--- a/docs/minio-client-complete-guide.md
+++ b/docs/minio-client-complete-guide.md
@@ -18,19 +18,18 @@ stat       show object metadata
 mv         move objects
 tree       list buckets and objects in a tree format
 du         summarize disk usage recursively
-lock       set and get object lock configuration
+lock       manage default bucket object lock configuration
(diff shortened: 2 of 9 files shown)


docs/minio-client-complete-guide.md
```diff
@@ -1212,11 +1211,11 @@ mc event remove play/andoria arn:minio:sqs:us-east-1:1:your-queue
 
 <a name="ilm"></a>
 ### Command `ilm`
-``ilm`` provides a convenient way to setup various bucket lifecycle configurations.
+``ilm`` - A convenient way manage bucket lifecycle configuration.
```
poornas, line 1214:
```suggestion
``ilm`` - A convenient way to  bucket lifecycle configuration.
```

cmd/main.go
```diff
@@ -339,7 +339,6 @@ var appCmds = []cli.Command{
 	diffCmd,
 	rmCmd,
 	eventCmd,
-	ilmCmd,
```
vadmeste, line 342: If you want backward compatibility, it is enough to keep this varaible with hidden variable set to true in cli.Command{}

harshavardhana: No backward compatibility needed, we are breaking willfully @vadmeste

harshavardhana:
> Also, we really need to release minio-go before releasing mc because it fixes some issues in lifecycle

We don't need a release unless it's urgent we can simply go mod update the master @vadmeste

harshavardhana:
> We don't need a release unless it's urgent we can simply go mod update the master @vadmeste

minio-go is release v7.0.2

PR #3321, opened 2020-07-24
Add --rewind flag for cat command

vadmeste:
It is possible to go back in time and download a specific version of 
an objet using mc cat --rewind command.

### DIFF
--- a/cmd/cat-main.go
+++ b/cmd/cat-main.go
@@ -1,5 +1,5 @@
 /*
- * MinIO Client, (C) 2015, 2016, 2017 MinIO, Inc.
+ * MinIO Client, (C) 2015-2020 MinIO, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -34,10 +34,12 @@ import (
 )
 
 var (
-	// This is kept dummy for future purposes
-	// and also to add ioFlags and globalFlags
-	// in CLI registration.
-	catFlags = []cli.Flag{}
+	catFlags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "rewind",
+			Usage: "back in time",
+		},
+	}
 )
 
 // Display contents of a file.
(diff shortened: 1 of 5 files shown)


poornas:
1. Can we add an example of --rewind flag usage on cli help also?
2. should we be issuing better error messages when --rewind flag is used on unversioned object or object created more recently than the duration specified in --rewind flag.
```
➜ mc cat --rewind 1m myminio/test/issue
mc: <ERROR> Unable to read from `myminio/test/issue`. Object does not exist.
``` 
maybe just show the contents if object created before time specified OR issue an error message that object is newer than duration specified. In the case of unversioned object, even if it was created within time window specified in --rewind, the error reported is `Object does not exist`. This will be misleading to user.

vadmeste:
> 1. Can we add an example of --rewind flag usage on cli help also?

Sure

> 2\. should we be issuing better error messages when --rewind flag is used on unversioned object or object created more recently than the duration specified in --rewind flag.

I see --rewind flag like going back to time. I think object not found error makes sense here.

> In the case of unversioned object, even if it was created within time window specified in --rewind, the error reported is `Object does not exist`

This looks like a bug. Will investigate it.

vadmeste:
> In the case of unversioned object, even if it was created within time window specified in --rewind, the error reported is `Object does not exist`

@poornas, I wasn't able to reproduce this, could you check again and tell me how to reproduce ?

poornas:
> > In the case of unversioned object, even if it was created within time window specified in --rewind, the error reported is `Object does not exist`
> 
> @poornas, I wasn't able to reproduce this, could you check again and tell me how to reproduce ?

@vadmeste, I tested again - the object was more recent than the rewind duration specified. It shows contents for unversioned object too.

vadmeste:
I also added a more descriptive error:

```
mc: <ERROR> Unable to read from `s3/vadmeste-testsnap/D`. Object did not exist at `Sat, 16 Mar 2019 20:33:38 CET`.   
```

PR #3322, opened 2020-07-25
rb: Delete object versions as well

vadmeste:
Ready but we need to validate this requirement.

Dangerous for users since it is removing all versions as well.

Maybe we can add a confirmation prompt here, or a flag --with-versions.

### DIFF
--- a/cmd/rb-main.go
+++ b/cmd/rb-main.go
@@ -114,7 +114,7 @@ func checkRbSyntax(ctx context.Context, cliCtx *cli.Context) {
 	}
 }
 
-// Delete a bucket and all its contents, versions will be removed as well.
+// Delete a bucket and all its objects and versions will be removed as well.
 func deleteBucket(ctx context.Context, url string) *probe.Error {
 	targetAlias, targetURL, _ := mustExpandAlias(url)
 	clnt, pErr := newClientFromAlias(targetAlias, targetURL)
--- a/cmd/rb-main.go
+++ b/cmd/rb-main.go
@@ -1,5 +1,5 @@
 /*
- * MinIO Client (C) 2019 MinIO, Inc.
+ * MinIO Client (C) 2019-2020 MinIO, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -34,7 +34,7 @@ var (
 	rbFlags = []cli.Flag{
 		cli.BoolFlag{
 			Name:  "force",
-			Usage: "allow a recursive remove operation",
+			Usage: "allow a recursive remove operation with all objects versions",
 		},
 		cli.BoolFlag{
 			Name:  "dangerous",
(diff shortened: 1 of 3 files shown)


cmd/rb-main.go
```diff
@@ -34,7 +34,7 @@ var (
 	rbFlags = []cli.Flag{
 		cli.BoolFlag{
 			Name:  "force",
-			Usage: "allow a recursive remove operation",
+			Usage: "allow a recursive remove operation with all objects versions",
```
harshavardhana, line 37:
```suggestion
			Usage: "force a recursive remove operation on all object versions",
```

cmd/rb-main.go
```diff
@@ -114,7 +114,7 @@ func checkRbSyntax(ctx context.Context, cliCtx *cli.Context) {
 	}
 }
 
-// deletes a bucket and all its contents
+// Delete a bucket and all its contents, versions will be removed as well.
```
harshavardhana, line 117:
```suggestion
// Delete a bucket and all its objects and versions will be removed as well.
```

docs/minio-client-complete-guide.md
```diff
@@ -399,15 +399,15 @@ Bucket created successfully ‘s3/mybucket’.
 ### Command `rb`
 `rb` command removes a bucket and all its contents on an object storage. On a filesystem, it behaves like `rmdir` command.
 
-Note that when a bucket is removed all policies associated with the bucket will also be removed. If you would like to just
-empty the objects in a bucket use `rm` command
+Note that when a bucket is removed all policies associated with the bucket will also be removed. All objects versions will
```
harshavardhana, line 402:
```suggestion
> NOTE:  When a bucket is removed all bucket configurations associated with the bucket will also be removed. All objects and their versions will be removed as well. If you need to preserve bucket and its configuration - only empty the objects and versions in a bucket use `mc rm` instead.
```

docs/minio-client-complete-guide.md
```diff
@@ -399,15 +399,15 @@ Bucket created successfully ‘s3/mybucket’.
...
+be removed as well. If you would like to just empty the objects in a bucket use `rm` command.
 
 ```
 USAGE:
    mc rb [FLAGS] TARGET [TARGET...]
 
 FLAGS:
-  --force                       allow a recursive remove operation
+  --force                       allow a recursive remove operation with all objects versions
```
harshavardhana, line 410:
```suggestion
  --force                       force a recursive remove operation on all object versions
```

harshavardhana:
> Dangerous for users since it is removing all versions as well.

If they need to preserve the bucket they should be using object-locking @vadmeste

docs/minio-client-complete-guide.md
```diff
@@ -399,15 +399,15 @@ Bucket created successfully ‘s3/mybucket’.
 ### Command `rb`
 `rb` command removes a bucket and all its contents on an object storage. On a filesystem, it behaves like `rmdir` command.
 
-Note that when a bucket is removed all policies associated with the bucket will also be removed. If you would like to just
-empty the objects in a bucket use `rm` command
+> NOTE:  When a bucket is removed all bucket configurations associated with the bucket will also be removed. All objects and their versions will be removed as well. If you need to preserve bucket and its configuration - only empty the objects and versions in a bucket use `mc rm` instead.
+be removed as well. If you would like to just empty the objects in a bucket use `rm` command.
```
harshavardhana, line 403: There are some duplicates here @vadmeste you may have to edit..

vadmeste: > There are some duplicates here @vadmeste you may have to edit..

Fixed

PR #3326, opened 2020-07-28
Fix example for mc admin bucket quota

poornas:


cmd/admin-bucket-quota.go
```diff
@@ -103,7 +103,7 @@ EXAMPLES:
 	  {{.Prompt}} {{.HelpName}} myminio/mybucket --fifo 64kB
```
kannappanr, line 103:
```suggestion
	  {{.Prompt}} {{.HelpName}} myminio/mybucket --fifo 64KB
```

harshavardhana, line 103: isn't this also inconsistent?

poornas: this is the convention to write kilo bytes afaics.

harshavardhana: We should support both styles @poornas also do we really care about kilobytes fifo? Are we trying to show that you can only store 1 file? 

Examples can be a bit more geared towards practical usecases.

PR #3329, opened 2020-07-30
Move version commands to top level

poornas:
moving `mc bucket version` to `mc version` - subcommands come before the target.


harshavardhana:
Can we do this in a single PR @poornas @kannappanr ?

PR #3337, opened 2020-08-03
fix: rb command regression deleting bucket

harshavardhana:


cmd/rb-main.go
```diff
@@ -235,7 +235,10 @@ func mainRemoveBucket(cliCtx *cli.Context) error {
 		}
 
 		listCtx, listCancel := context.WithCancel(ctx)
-		for range clnt.List(listCtx, opts) {
+		for obj := range clnt.List(listCtx, opts) {
+			if obj.Err != nil {
+				continue
```
vadmeste, line 240: Good catch, but there is no point continuing here: if we are going to delete a bucket, we should list all of its contents and remove them. fatalIf() is enough.

vadmeste:
The fix is needed.. but just not sure how to reproduce an issue with this.

harshavardhana:
> The fix is needed.. but just not sure how to reproduce an issue with this.

@vadmeste FS mode

cmd/rb-main.go
harshavardhana: continue is to look for situations if there are objects @vadmeste such that we don't prematurely send DeleteBucket()

vadmeste:
> @vadmeste FS mode

oh okay.. maybe FS mode should support versioning listing as well.

harshavardhana:
> > @vadmeste FS mode
> 
> oh okay.. maybe FS mode should support versioning listing as well.

No it shouldn't @vadmeste

PR #3339, opened 2020-08-04
Add versioning support for rm cmd

vadmeste:
Add versioning support of rm command

--version to specify a specific version id of an object when removing
--versions includes older versions along with objects
--rewind go back in time before removing

### DIFF
--- a/cmd/rm-main.go
+++ b/cmd/rm-main.go
@@ -45,7 +45,7 @@ var (
 		},
 		cli.StringFlag{
 			Name:  "version-id",
-			Usage: "pick a particular version id",
+			Usage: "delete a specific version of an object",
 		},
 		cli.BoolFlag{
 			Name:  "recursive, r",
--- a/cmd/rm-main.go
+++ b/cmd/rm-main.go
@@ -37,7 +37,7 @@ var (
 	rmFlags = []cli.Flag{
 		cli.BoolFlag{
 			Name:  "versions",
-			Usage: "remove objects versions as well",
+			Usage: "remove object(s) and all its versions",
 		},
 		cli.StringFlag{
 			Name:  "rewind",
--- a/cmd/rm-main.go
+++ b/cmd/rm-main.go
@@ -184,11 +184,12 @@ func checkRmSyntax(ctx context.Context, cliCtx *cli.Context, encKeyDB map[string
 	isDangerous := cliCtx.Bool("dangerous")
 	isVersions := cliCtx.Bool("versions")
 	versionID := cliCtx.String("version-id")
+	rewind := cliCtx.String("rewind")
 	isNamespaceRemoval := false
 
-	if isVersions && versionID != "" {
+	if versionID != "" && (isRecursive || isVersions || rewind != "") {
 		fatalIf(errDummy().Trace(),
-			"You cannot have both --versions and --version-id specified.")
+			"You cannot specify --version-id with any of --versions, --rewind and --recursive flags.")
 	}
 
 	for _, url := range cliCtx.Args() {
--- a/cmd/rm-main.go
+++ b/cmd/rm-main.go
@@ -41,7 +41,7 @@ var (
 		},
 		cli.StringFlag{
 			Name:  "rewind",
-			Usage: "back in time before removal",
+			Usage: "revert to older version of object(s) in time",
 		},
 		cli.StringFlag{
 			Name:  "version-id",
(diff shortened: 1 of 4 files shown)


harshavardhana:
> The exact specification is not approved yet.
> 
> Currently:
> --version to specify a specific version id of an object when removing
> --versions includes versions along with objects
> --rewind go back in time before removing

can we not do `mc ls` style for deletions? `mc rm alias/bucket/object:1` ? instead of `--version id`?

vadmeste:
> can we not do `mc ls` style for deletions? `mc rm alias/bucket/object:1` ? instead of `--version id`?

The spec is changed, now we use --version style and `mc ls` will print version ids as well.

The reason is that ':0' notation is not static and can change (:0 means the latest version, :1 means the second latest version, etc..)

cmd/rm-main.go
```diff
@@ -35,6 +35,18 @@ import (
...
+			Name:  "versions",
+			Usage: "remove objects versions as well",
+		},
+		cli.StringFlag{
+			Name:  "rewind",
+			Usage: "back in time before removal",
+		},
+		cli.StringFlag{
+			Name:  "version-id",
+			Usage: "pick a particular version id",
```
harshavardhana, line 48:
```suggestion
			Usage: "delete a specific version of an object",
```

cmd/rm-main.go
```diff
@@ -35,6 +35,18 @@ import (
 // rm specific flags.
 var (
 	rmFlags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "versions",
+			Usage: "remove objects versions as well",
+		},
+		cli.StringFlag{
+			Name:  "rewind",
+			Usage: "back in time before removal",
```
harshavardhana, line 44:
```suggestion
			Usage: "revert to older version of object(s) in time",
```

cmd/rm-main.go
```diff
@@ -35,6 +35,18 @@ import (
 // rm specific flags.
 var (
 	rmFlags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "versions",
+			Usage: "remove objects versions as well",
```
harshavardhana, line 40:
```suggestion
			Usage: "remove object(s) and all its versions",
```

cmd/rm-main.go
vadmeste: > Usage: "revert to older version of object(s) in time",

@harshavardhana the thing is that if --rewind is used with --versions, it will remove all versions prior to the specified rewind argument. So I think this comment can be misleading

vadmeste:
> needs updates to docs

@poornas the documentation is updated.

cmd/rm-main.go
harshavardhana: @vadmeste revert means removal as well - it implies overwrite.

cmd/rm-main.go
```diff
@@ -156,8 +182,15 @@ func checkRmSyntax(ctx context.Context, cliCtx *cli.Context, encKeyDB map[string
 	isRecursive := cliCtx.Bool("recursive")
 	isStdin := cliCtx.Bool("stdin")
 	isDangerous := cliCtx.Bool("dangerous")
+	isVersions := cliCtx.Bool("versions")
+	versionID := cliCtx.String("version-id")
 	isNamespaceRemoval := false
 
+	if isVersions && versionID != "" {
```
kannappanr, line 189: to keep things simple, we should fail even if `rewind` is specified when `--version-id` flag is specified. If `version-id` flag is specified, neither `rewind` nor `versions` flags are allowed.

docs/minio-client-complete-guide.md
```diff
@@ -889,6 +889,9 @@ USAGE:
    mc rm [FLAGS] TARGET [TARGET ...]
 
 FLAGS:
+  --versions                    remove object(s) and all its versions
+  --rewind value                back in time before removal
```
poornas, line 893:
```suggestion
  --rewind value                revert to older version of object(s) in time
```

cmd/rm-main.go
poornas: nor `recursive` flag

vadmeste: > to keep things simple, we should fail even if `rewind` is specified when `--version-id` flag is specified. If `version-id` flag is specified, neither `rewind` nor `versions` flags are allowed.

Done

PR #3342, opened 2020-08-05
cp: Add support of --version-id flag

vadmeste:
mc cp --version-id should pick a specific object version during copying.

It errors out when the version is not available.

### DIFF
--- a/cmd/cp-main.go
+++ b/cmd/cp-main.go
@@ -43,7 +43,7 @@ var (
 		},
 		cli.StringFlag{
 			Name:  "version-id",
-			Usage: "Pick a particular object version to copy",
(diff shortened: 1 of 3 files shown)


vadmeste:
PR updated to use --version-id instead of --version

cmd/cp-main.go
```diff
@@ -41,6 +41,10 @@ var (
 			Name:  "rewind",
 			Usage: "Move back in time",
 		},
+		cli.StringFlag{
+			Name:  "version-id",
+			Usage: "Pick a particular object version to copy",
```
harshavardhana, line 46:
```suggestion
			Usage: "Select an object version to copy",
```

PR #3343, opened 2020-08-06
legalhold: Refactor & add versioning support

vadmeste:
Legalhold command refactor and add support of --version-id, --versions, --rewind flags

### DIFF
--- a/cmd/legalhold-clear.go
+++ b/cmd/legalhold-clear.go
@@ -1,0 +1,105 @@
+/*
+ * MinIO Client (C) 2020 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"time"
+
+	"github.com/fatih/color"
+	"github.com/minio/cli"
+	"github.com/minio/minio-go/v7"
+	"github.com/minio/minio/pkg/console"
+)
+
+var (
+	lhClearFlags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "recursive, r",
+			Usage: "apply legal hold recursively",
+		},
+		cli.StringFlag{
+			Name:  "version-id",
+			Usage: "apply legal hold to a specific object version",
+		},
+		cli.StringFlag{
+			Name:  "rewind",
+			Usage: "Move back in time",
+		},
+		cli.BoolFlag{
+			Name:  "versions",
+			Usage: "Pick earlier versions",
+		},
+	}
+)
+var legalHoldClearCmd = cli.Command{
+	Name:   "clear",
+	Usage:  "clear legal hold for object(s)",
+	Action: mainLegalClearHold,
+	Before: setGlobalsFromContext,
+	Flags:  append(lhClearFlags, globalFlags...),
+	CustomHelpTemplate: `NAME:
+  {{.HelpName}} - {{.Usage}}
+
+USAGE:
+  {{.HelpName}} [FLAGS] TARGET
+
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}
+
+EXAMPLES:
+   1. Disable legal hold on a specific object
+      $ {{.HelpName}} myminio/mybucket/prefix/obj.csv
+
+   2. Disable legal hold on a specific object version
+      $ {{.HelpName}} myminio/mybucket/prefix/obj.csv --version-id "HiMFUTOowG6ylfNi4LKxD3ieHbgfgrvC"
+
+   3. Disable object legal hold recursively for all objects at a prefix
+      $ {{.HelpName}} myminio/mybucket/prefix --recursive
+
+   4. Disable object legal hold recursively for all objects versions older than one year
+      $ {{.HelpName}} myminio/mybucket/prefix --recursive --rewind 365d --versions
+ `,
+}
+
+// main for retention command.
+func mainLegalClearHold(ctx *cli.Context) error {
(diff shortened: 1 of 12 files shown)


cmd/legalhold-clear.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	versionID := ctx.String("version-id")
+	rewind := parseRewindFlag(ctx.String("rewind"))
+	recursive := ctx.Bool("recursive")
+	withVersions := ctx.Bool("versions")
+
+	if versionID != "" && recursive {
+		fatalIf(errInvalidArgument(), "Cannot pass --version-id and --recursive at the same time")
+	}
```
poornas, line 97: also need a validation to error out if `version-id` and `versions` are used together

cmd/legalhold-set.go
```diff
@@ -0,0 +1,178 @@
...
+		if errorsFound {
+			console.Print(console.Colorize("LegalHoldPartialFailure", fmt.Sprintf("Errors found while setting legal hold status on objects with prefix `%s`. \n", urlStr)))
+		} else {
+			console.Print(console.Colorize("LegalHoldSuccess", fmt.Sprintf("Object legal hold successfully set for prefix `%s`.\n", urlStr)))
+		}
+	}
+	return cErr
+}
+
+// main for retention command.
```
poornas, line 152:
```suggestion
// main for legalhold set command.
```

cmd/legalhold-set.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	versionID := ctx.String("version-id")
+	rewind := parseRewindFlag(ctx.String("rewind"))
+	recursive := ctx.Bool("recursive")
+	withVersions := ctx.Bool("versions")
+
+	if versionID != "" && recursive {
+		fatalIf(errInvalidArgument(), "Cannot pass --version-id and --recursive at the same time")
+	}
```
poornas, line 170: might also want to disallow --versions and --version-id at the same time.

poornas:
```
 mc legalhold set dminio5/lock/noprefix  --versions --recursive
Object legal hold successfully set for prefix `dminio5/lock/noprefix`.
```
The message is incorrect here - prefix `noprefix` does not exist

cmd/legalhold-clear.go
poornas: also need a validation  when --versions is used, it requires a --recursive flag - otherwise it quietly does nothing and issues a success message

vadmeste: > also need a validation when --versions is used, it requires a --recursive flag - otherwise it quietly does nothing and issues a success message

Actually --versions does not always require --recursive here. For example if you want to apply legalhold for all versions of a particular object, you don't pass --recursive with it.

vadmeste:
> ```
>  mc legalhold set dminio5/lock/noprefix  --versions --recursive
> Object legal hold successfully set for prefix `dminio5/lock/noprefix`.
> ```
> 
> 
> The message is incorrect here - prefix `noprefix` does not exist

Fixed, now it will show:
```
Not objects/versions found while setting legal hold status with prefix `s3/foo/inexistantprefix`
```

All comments addressed @poornas, PTAL

cmd/legalhold-clear.go
```diff
@@ -0,0 +1,92 @@
...
+
+   2. Disable legal hold on a specific object version
+      $ {{.HelpName}} myminio/mybucket/prefix/obj.csv --version-id "HiMFUTOowG6ylfNi4LKxD3ieHbgfgrvC"
+
+   3. Disable object legal hold recursively for all objects at a prefix
+      $ {{.HelpName}} myminio/mybucket/prefix --recursive
+
+   4. Disable object legal hold recursively for all objects versions older than one year
+      $ {{.HelpName}} myminio/mybucket/prefix --recursive --rewind 365d --versions
+ `,
```
poornas, line 76: extra % being displayed

vadmeste: > extra % being displayed

You mean extra space ? I don't see any %

cmd/legalhold-info.go
```diff
@@ -0,0 +1,203 @@
...
+	if l.VersionID != "" {
+		msg += ", Version id: " + l.VersionID
+	}
+	msg += ", "
+	if l.LegalHold == "" {
+		msg += "No legalhold set"
+	} else {
+		msg += "Legalhold: " + string(l.LegalHold)
+	}
+	return console.Colorize("LegalHoldSuccess", msg)
```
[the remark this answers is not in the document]
vadmeste: > when --versions is used, the legal hold info message displayed seems misleading - as it looks at whether legal hold is set on latest version only. Perhaps we need to display all versions here just like with --recursive?

Good catch, I corrected it and pushed the fix

cmd/legalhold-set.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/minio/cli"
+	"github.com/minio/minio-go/v7"
+	"github.com/minio/minio/pkg/console"
+)
+
+var (
+	lhSetFlags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "recursive, r",
+			Usage: "apply legal hold recursively",
```
harshavardhana, line 34:
```suggestion
			Usage: "apply legal hold recursively to all current versions of object(s)",
```

cmd/legalhold-info.go
```diff
@@ -0,0 +1,203 @@
...
+			Usage: "apply legal hold to a specific object version",
+		},
+		cli.StringFlag{
+			Name:  "rewind",
+			Usage: "Move back in time",
+		},
+		cli.BoolFlag{
+			Name:  "versions",
+			Usage: "Pick earlier versions",
+		},
```
harshavardhana, line 49: Look at the changes done in `set` command to ensure we follow it here.

cmd/legalhold-set.go
```diff
@@ -0,0 +1,194 @@
...
+		},
+		cli.StringFlag{
+			Name:  "version-id",
+			Usage: "apply legal hold to a specific object version",
+		},
+		cli.StringFlag{
+			Name:  "rewind",
+			Usage: "Move back in time",
```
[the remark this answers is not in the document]
vadmeste: > Usage: "apply legal hold for older versions, use with `--versions`",

We can use --rewind without --versions, it means apply on the first version created before the specified time. Something similar to `mc cp --rewind 1d`

kannappanr (approved):
LGTM

poornas (approved):
LGTM

PR #3349, opened 2020-08-11
Add --dest-bucket as flag in mc replicate add/set

poornas:


poornas:
> @poornas CI fails.
> 
> Also, not sure but it looks like the dest bucket does exist in the ARN itself, maybe --dest-bucket is not useful ?

@vadmeste, it is now a separate flag so that it works for AWS also - they use a ARN that passes a replication role name, not the destination bucket as in our case.

vadmeste:
> @vadmeste, it is now a separate flag so that it works for AWS also - they use a ARN that passes a replication role name, not the destination bucket as in our case.

Understood

LGTM but there is some conflicts now.

PR #3350, opened 2020-08-11
stat: Add support of --rewind, --versions and --version-id

vadmeste:
As other commands, this commit adds support of versioning to stat command.

### DIFF
--- a/cmd/stat-main.go
+++ b/cmd/stat-main.go
@@ -29,6 +29,18 @@ import (
 // stat specific flags.
 var (
 	statFlags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "rewind",
+			Usage: "stat on older version(s)",
+		},
+		cli.BoolFlag{
+			Name:  "versions",
+			Usage: "stat all versions",
+		},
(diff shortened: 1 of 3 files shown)


cmd/stat-main.go
```diff
@@ -29,6 +29,18 @@ import (
 // stat specific flags.
 var (
 	statFlags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "rewind",
+			Usage: "go back in time",
+		},
+		cli.BoolFlag{
+			Name:  "versions",
+			Usage: "stat earlier versions",
```
harshavardhana, line 38:
```suggestion
			Usage: "stat all versions",
```

cmd/stat-main.go
```diff
@@ -29,6 +29,18 @@ import (
 // stat specific flags.
 var (
 	statFlags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "rewind",
+			Usage: "go back in time",
```
harshavardhana, line 34:
```suggestion
			Usage: "stat on older versions",
```

PR #3352, opened 2020-08-12
cat: Add --version-id flag

vadmeste:
Add support of --version-id in cat command.


kannappanr:
@vadmeste PTAL build failure

cmd/cat-main.go
```diff
@@ -37,7 +37,11 @@ var (
 	catFlags = []cli.Flag{
 		cli.StringFlag{
 			Name:  "rewind",
-			Usage: "back in time",
+			Usage: "display an object version before the specified rewind time",
```
harshavardhana, line 40:
```suggestion
			Usage: "display an earlier object version",
```

PR #3353, opened 2020-08-12
Fixes mirror not checking modtime

### LINKED ISSUE #3331: Minio fails to mirror file after changes if file's size matches
## Expected behavior

mc mirror should always mirror files with changes

## Actual 

As described in the original issue I'm still able to reproduce this behavior. Files with the same size will *not* be synced even if the content has changed. https://github.com/minio/mc/issues/2187

## mc --version
mc version RELEASE.2020-07-11T05-18-52Z

## System information
Ubuntu 18.08

ebozduman:
Fixes #3331 

Issue #3331(Minio fails to mirror file after changes if file's size matches) is a regression caused by the fix #3226 (diff: Disable comparing modtimes when multimaster context is not found).
The PR fixes the regression without reverting back PR#3226.

### DIFF
--- a/cmd/difference.go
+++ b/cmd/difference.go
@@ -82,7 +82,7 @@ func getSourceModTimeKey(metadata map[string]string) string {
 
 // activeActiveModTimeUpdated tries to calculate if the object copy in the target
 // is older than the one in the source by comparing the modtime of the data.
-func activeActiveModTimeUpdated(src, dst *ClientContent) bool {
+func activeActiveModTimeUpdated(src, dst *ClientContent, isActiveActive bool) bool {
 	if src == nil || dst == nil {
 		return false
 	}
@@ -96,7 +96,7 @@ func activeActiveModTimeUpdated(src, dst *ClientContent) bool {
 
 	srcModTime := getSourceModTimeKey(src.UserMetadata)
 	dstModTime := getSourceModTimeKey(dst.UserMetadata)
-	if srcModTime == "" && dstModTime == "" {
+	if srcModTime == "" && dstModTime == "" && isActiveActive {
 		// No active-active mirror context found, consider src & dst as similar
 		return false
 	}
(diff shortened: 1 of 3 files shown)


cmd/difference.go
```diff
@@ -82,7 +82,7 @@ func getSourceModTimeKey(metadata map[string]string) string {
 
 // activeActiveModTimeUpdated tries to calculate if the object copy in the target
 // is older than the one in the source by comparing the modtime of the data.
-func activeActiveModTimeUpdated(src, dst *ClientContent) bool {
+func activeActiveModTimeUpdated(src, dst *ClientContent, isActiveActive bool) bool {
```
krisis, line 85: Should this function be called at all in an active-active mirror setting?

cmd/difference.go
```diff
@@ -274,7 +274,7 @@ func differenceInternal(ctx context.Context, sourceClnt, targetClnt Client, sour
 					firstContent:  srcCtnt,
 					secondContent: tgtCtnt,
 				}
-			} else if activeActiveModTimeUpdated(srcCtnt, tgtCtnt) {
+			} else if activeActiveModTimeUpdated(srcCtnt, tgtCtnt, isActiveActive) {
```
krisis, line 277: Can we call `activeActiveModTimeUpdated` only if `isActiveActive = true`?

vadmeste:
@ebozduman are you saying that mc mirror should check modtime if active/active is not enabled ?

cmd/difference.go
ebozduman: I guess so. 
This part of the code was added/modified by @harshavardhana.
We can consult with him.

cmd/difference.go
ebozduman: The problem with that is we need modification time check for non-active-active setups too.

ebozduman:
@vadmeste,
> Are you saying that mc mirror should check modtime if active/active is not enabled ?
Yes.

ebozduman:
@harshavardhana,

Please check [PR#3226](https://github.com/minio/mc/pull/3226): "_diff: Disable comparing modtimes when multimaster context is not found_" and your comments there. It looks like what is attempted to be fixed here with this PR, was actually deliberately removed then. So, Issue [#3331](https://github.com/minio/mc/pull/3331) doesn't look like a regression.

Your comments:
> This is a good fix which is needed.
Without this fix, mc diff works like when time difference was enabled.

Please advise if we need to call issue [#3331](https://github.com/minio/mc/pull/3331) "won't fix" & "working as intended".

crrodriguez:
> @harshavardhana,
> 
> Please check [PR#3226](https://github.com/minio/mc/pull/3226): "_diff: Disable comparing modtimes when multimaster context is not found_" and your comments there. It looks like what is attempted to be fixed here with this PR, was actually deliberately removed then. So, Issue [#3331](https://github.com/minio/mc/pull/3331) doesn't look like a regression.
> 
> Your comments:
> 
> > This is a good fix which is needed.
> > Without this fix, mc diff works like when time difference was enabled.
> 
> Please advise if we need to call issue [#3331](https://github.com/minio/mc/pull/3331) "won't fix" & "working as intended".

So, mc mirror cannot be trusted to actually mirror content if files are equal size but different content ? or I am understading it incorrectly.

ebozduman:
@crrodriguez,
We'll clarify it as soon as we make a decision.
Please bear with us.

harshavardhana:
> Please check [PR#3226](https://github.com/minio/mc/pull/3226): "_diff: Disable comparing modtimes when multimaster context is not found_" and your comments there. It looks like what is attempted to be fixed here with this PR, was actually deliberately removed then. So, Issue [#3331](https://github.com/minio/mc/pull/3331) doesn't look like a regression.

The reason for this is we have no way to know which is latest @ebozduman because last-modified is not the correct way to know - that's why active/active is necessary if `mtime` based mirroring is required.

krisis (approved):

ebozduman:
PR#3402 fixed issue#3331. 
This PR is not needed anymore.

PR #3371, opened 2020-08-19
fix: alignment in help for multiple commands

harshavardhana:


cmd/ls.go
```diff
@@ -61,7 +62,7 @@ func (c contentMessage) String() string {
 
 	fileDesc := c.Key
 	if c.VersionID != "" {
-		fileDesc += fmt.Sprintf(" (v%d, %s", c.VersionIndex, c.VersionID)
+		fileDesc += fmt.Sprintf(" (%s, vid=%s", c.VersionOridinal, c.VersionID)
```
poornas, line 65: VersionOridinal -> VersionOrdinal

PR #3380, opened 2020-08-26
fix: messages for version-id and rewind

harshavardhana:


harshavardhana:
NOTE: @vadmeste no command should have a `capitalized` Usage to be consistent for all flags, avoid the rewind flag to just say "move back in time"

PR #3381, opened 2020-08-27
cat: Fix a regression when copying from stdin

vadmeste:
`mc cat >/tmp/output` is not working properly due to mc cat versioning
changes, this commit fixes the issue to make mc cat works as expected.

It also adds a unit test for it.

### DIFF
--- a/cmd/cat-main.go
+++ b/cmd/cat-main.go
@@ -154,9 +154,6 @@ func parseCatSyntax(ctx *cli.Context) (args []string, versionID string, timeRef 
 		fatalIf(errInvalidArgument().Trace(), "You need to pass at least one argument if --version-id is specified")
 	}
 
-	if len(args) == 0 {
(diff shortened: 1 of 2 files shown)


cmd/cat-main.go
```diff
@@ -154,9 +154,6 @@ func parseCatSyntax(ctx *cli.Context) (args []string, versionID string, timeRef
 		fatalIf(errInvalidArgument().Trace(), "You need to pass at least one argument if --version-id is specified")
 	}
 
-	if len(args) == 0 {
```
klauspost, line 157: This is a functional change, right?

Why is it needed again?

vadmeste: > This is a functional change, right?
> 
> Why is it needed again?

This is a regression for an earlier change, this is not needed.

klauspost: ok

PR #3348, opened 2020-08-10
Refactor ilm command

nitisht:

PR #3384, opened 2020-08-27
Fix parsing errors for `mc ilm set` command

poornas:
fix on top of PR https://github.com/minio/mc/pull/3348


nitisht (approved):
LGTM

nitisht:
Thanks for the help with this @poornas . I sent a fix with small updates to example text, other than that LGTM

cmd/ilm-add.go
```diff
@@ -44,43 +44,30 @@ FLAGS:
...
 
 var ilmAddFlags = []cli.Flag{
-	cli.StringFlag{
-		Name:  "id",
-		Usage: "id for the rule, should be a unique value",
-	},
-	cli.StringFlag{
-		Name:  "prefix",
-		Usage: "prefix to apply the lifecycle configuration rule",
-	},
```
BenoitKnecht, line 76: Why get rid of the `prefix` option? It's no longer possible to create lifecycle policies that only match a subset of objects based on their prefix.

cmd/ilm-add.go
```diff
@@ -44,43 +44,30 @@ FLAGS:
...
+  1. Add expiration rule on mybucket.
+     {{.Prompt}} {{.HelpName}} --expiry-days "200" myminio/mybucket
+
+  2. Add expiry and transition date rules on a prefix in mybucket.
+     {{.Prompt}} {{.HelpName}} -expiry-date "2025-09-17" --transition-date "2025-05-01" \
+          --storage-class "GLACIER" myminio/mybucket/doc
+
+  3. Add expiry and transition days rules on a prefix in mybucket.
+     {{.Prompt}} {{.HelpName}} -expiry-days "300" --transition-days "200" \
+          --storage-class "GLACIER" myminio/mybucket/doc
```
BenoitKnecht, line 59: And it makes examples 2 and 3 inaccurate here, the description mentions a prefix, but the policy that gets added doesn't match any particular prefix.

BenoitKnecht: Ooohh, never mind, I just noticed that the prefix is now simply added after the bucket name!

Sorry for the noise.

PR #3391, opened 2020-09-01
fix ls output when latest version is delete marker

harshavardhana:


vadmeste:
@harshavardhana what this does fix exactly ?

harshavardhana:
> @harshavardhana what this does fix exactly ?

It fixes a situation where latest is delete marker i.e `mc ls myminio/bucket/hosts`

vadmeste:
> It fixes a situation where latest is delete marker i.e `mc ls myminio/bucket/hosts`

what does `mc ls myminio/bucket/hosts` show in that case ? because this is a non versioned request, it means it should throw an error

kannappanr:
> > It fixes a situation where latest is delete marker i.e `mc ls myminio/bucket/hosts`
> 
> what does `mc ls myminio/bucket/hosts` show in that case ? because this is a non versioned request, it means it should throw an error

right now it will not throw an error, it will display a blank output. I think that is fine. You will have the same issue with `--versions` for an object that doesn't exist

vadmeste:
> right now it will not throw an error, it will display a blank output. I think that is fine. You will have the same issue with `--versions` for an object that doesn't exist

It does if you test with AWS S3. But I am okay with the blank output either and with this change since this will avoid many S3 calls

vadmeste:
ls already has this behavior, so I think it is okay to not show any error when mc ls doesn't find anything to list.

```
$ mc ls s3/bucket/whe
[2020-04-22 18:20:21 CET]     99B where.csv.gz
```

harshavardhana:
> ls already has this behavior, so I think it is okay to not show any error when mc ls doesn't find anything to list.

Correct that is the reason why we can be consistent just like Listing @vadmeste

vadmeste (approved):
LGTM

PR #3401, opened 2020-09-07
cp: Fix url check not considering version-id in some cases

vadmeste:
cp checks the existance of the source object in Type A and Type B, but
forgets to pass version-id as well. Fix this behavior.

How to reproduce:
`mc cp --version-id anything alias/bucket/object /tmp/` where `object` has a delete marker as current version.


harshavardhana (approved):

klauspost (approved):
LGTM

PR #3402, opened 2020-09-10
re-implement metadata handling to preserve nanosecs as well

harshavardhana:


ebozduman (commented):
I found an issue during `mc mirror -a <local-dir> <minio-bucket>` command in case the destination already has the same object with the same size as source, but older modification time than the source. The expectation is to still mirror/copy the object, but it doesn't.

Here are the reproduction steps:
```
1. mkdir tmp tmp1
2. echo "this is some random content" > tmp/content.txt
3. mc mb myminio/dst
4. mc mirror -a tmp/ myminio/dst/
5. echo "this is SOME rand0m c0ntent" > tmp1/content.txt
6. mc mirror -a tmp1/ myminio/dst/
```

and here is the fix for it:
```
$ git diff cmd/client-s3.go
diff --git a/cmd/client-s3.go b/cmd/client-s3.go
index a4d9ffb51..180375f50 100644
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -1331,7 +1331,7 @@ func (c *S3Client) SetAccess(ctx context.Context, bucketPolicy string, isJSON bo
 }
 
 // listObjectWrapper - select ObjectList mode depending on arguments
-func (c *S3Client) listObjectWrapper(ctx context.Context, bucket, object string, isRecursive bool, timeRef time.Time, withVersions, withDeleteMarkers bool, metadata bool) <-chan minio.ObjectInfo {
+func (c *S3Client) listObjectWrapper(ctx context.Context, bucket, object string, isRecursive bool, timeRef time.Time, withVersions, withDeleteMarkers, metadata bool) <-chan minio.ObjectInfo {
        if !timeRef.IsZero() || withVersions {
                return c.listVersions(ctx, bucket, object, isRecursive, timeRef, withVersions, withDeleteMarkers)
        }
@@ -1342,7 +1342,7 @@ func (c *S3Client) listObjectWrapper(ctx context.Context, bucket, object string,
                return c.api.ListObjects(ctx, bucket, minio.ListObjectsOptions{Prefix: object, Recursive: isRecursive, UseV1: true})
        }
        if metadata {
-               return c.api.ListObjects(ctx, bucket, minio.ListObjectsOptions{Prefix: object, Recursive: isRecursive})
+               return c.api.ListObjects(ctx, bucket, minio.ListObjectsOptions{WithMetadata: true, Prefix: object, Recursive: isRecursive})
        }
        return c.api.ListObjects(ctx, bucket, minio.ListObjectsOptions{Prefix: object, Recursive: isRecursive})
 }
```

harshavardhana:
Yes looks like a regression introduced in versioning changes

harshavardhana:
This is fixed @ebozduman

kannappanr (approved):
LGTM

ebozduman:
@harshavardhana 

I checked again and I still see the issue in the same above scenario.
It looks like we also need to update the `"LastModified"` field on the S3 side for the object, in the `client-s3.go` file. 

Here is the diff for my proposal:
```
$ git diff
diff --git a/cmd/client-s3.go b/cmd/client-s3.go
index fc8b327e4..5a4bbd9ed 100644
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -32,8 +32,10 @@ import (
        "path"
        "path/filepath"
        "regexp"
+       "strconv"
        "strings"
        "sync"
+       "syscall"
        "time"
 
@@ -2207,6 +2209,37 @@ func (c *S3Client) listRecursiveInRoutine(ctx context.Context, contentCh chan *C
                                }
                                return
                        }
+                       if object.UserMetadata["X-Amz-Meta-Mc-Attrs"] != "" {
+                               attrs := object.UserMetadata["X-Amz-Meta-Mc-Attrs"]
+                               attrsSplit := strings.Split(attrs, "/")
+
+                               m := make(map[string]string)
+                               for _, elem := range attrsSplit {
+                                       parts := strings.Split(elem, ":")
+                                       m[parts[0]] = parts[1]
+                               }
+                               var mtime syscall.Timespec
+                               vals := strings.SplitN(m["mtime"], "#", 2)
+                               mtim, e := strconv.ParseInt(vals[0], 10, 64)
+                               if e != nil {
+                                       contentCh <- &ClientContent{
+                                               Err: probe.NewError(e),
+                                       }
+                                       return
+                               }
+                               mtime.Sec = mtim
+                               if len(vals) == 2 {
+                                       mtimnsec, e := strconv.ParseInt(vals[1], 10, 64)
+                                       if e != nil {
+                                               contentCh <- &ClientContent{
+                                                       Err: probe.NewError(e),
+                                               }
+                                               return
+                                       }
+                                       mtime.Nsec = mtimnsec
+                               }
+                               object.LastModified = time.Unix(mtime.Sec, mtime.Nsec)
+                       }
                        contentCh <- c.objectInfo2ClientContent(b, object)
                }
        }
```

I've tested some other scenarios, but I am still not 100 percent sure if this will not cause any regression.

harshavardhana:
@ebozduman resolved this fully.

cmd/difference.go
```diff
@@ -265,7 +267,7 @@ func differenceInternal(ctx context.Context, sourceClnt, targetClnt Client, sour
 				}
 				continue
 			}
-			if (srcType.IsRegular() && tgtType.IsRegular()) && srcSize != tgtSize {
+			if srcSize != tgtSize {
```
ebozduman, line 270: If mtime for the source object is older than the destinaton object's mtime, but their sizes are different, then we'll copy over the newer version of the file just because sizes are different.

```
$ mc stat tmp1/content.txt
Name      : content.txt
Date      : 2020-09-11 19:41:15 PDT 
Size      : 30 B   
Type      : file 
Metadata  :
  Content-Type       : text/plain 
  X-Amz-Meta-Mc-Attrs: atime:1599878236#794570223/gid:1000/gname:ersan/mode:33188/mtime:1599878475#869739459/uid:1000/uname:ersan 

$ mc stat tmp/content.txt
Name      : content.txt
Date      : 2020-09-11 03:06:55 PDT 
Size      : 28 B   
Type      : file 
Metadata  :
  X-Amz-Meta-Mc-Attrs: atime:1599820232#311739372/gid:1000/gname:ersan/mode:33188/mtime:1599818815#52759230/uid:1000/uname:ersan 
  Content-Type       : text/plain 

$ mc mirror -a tmp1/ myminio/dst/
...minio/mc/tmp1/content.txt:  30 B / 30 B ┃▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓┃ 2.36 KiB/s 0s
$ mc mirror -a tmp/ myminio/dst/
.../minio/mc/tmp/content.txt:  28 B / 28 B ┃▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓┃ 3.93 KiB/s 0s
```

I don't think we need to check the size anymore since we've mtime check going forward.
We can safely remove it.

ebozduman (changes requested):

cmd/difference.go
harshavardhana: That is okay it is not the intention @ebozduman of this PR to remove older checks.

ebozduman (approved):
LGTM

PR #3406, opened 2020-09-16
obd: Add log support

vadmeste:
Adding log support with a spinner in this PR.


harshavardhana (changes requested):
@kannappanr should we bring flags to query specific obd results?

harshavardhana (approved):

kannappanr (approved):
LGTM

PR #3407, opened 2020-09-16
rm: Show deleted error msg when delete marker is found

vadmeste:
HEAD objet returns 404 with a delete-marker header when the object is a
delete-marker in a versioning bucket. This PR will show a better error
message in that case.

### DIFF
--- a/cmd/client-errors.go
+++ b/cmd/client-errors.go
@@ -173,7 +173,7 @@ type ObjectIsDeleteMarker struct {
 }
 
 func (e ObjectIsDeleteMarker) Error() string {
-	return "Object has been deleted"
+	return "Object is marked as deleted"
 }
 
 // UnexpectedShortWrite - write wrote less bytes than expected.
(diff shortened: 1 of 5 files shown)


cmd/client-errors.go
```diff
@@ -168,6 +168,14 @@ func (e ObjectMissing) Error() string {
 	return "Object does not exist"
 }
 
+// ObjectIsDeleteMarker - object is a delete marker as latest
+type ObjectIsDeleteMarker struct {
+}
+
+func (e ObjectIsDeleteMarker) Error() string {
+	return "Object has been deleted"
```
harshavardhana, line 176:
```suggestion
	return "Object is marked as deleted"
```

harshavardhana (changes requested):
We should also show perhaps help where we say that `Object is marked as deleted, please use '--vid' to delete a specific version`

vadmeste:
> please use '--vid' to delete a specific version

I think you mean please use `mc undo xxxx` to undelete this object ?

harshavardhana:
> I think you mean please use `mc undo xxxx` to undelete this object ?

yeah @vadmeste

harshavardhana (approved):

kannappanr (approved):
LGTM

PR #3411, opened 2020-09-20
update json flag help to indicate the correct format

harshavardhana:
some users may get confused without reading the docs
might end up assuming that --json is a full JSON document.


poornas (approved):

ebozduman (approved):
LGTM

PR #3412, opened 2020-09-20
pretty print KMS key status

harshavardhana:


cmd/admin-kms-key-status.go
```diff
@@ -63,29 +68,48 @@ func mainAdminKMSKeyStatus(ctx *cli.Context) error {
...
+		EncryptionErr: status.EncryptionErr,
+		DecryptionErr: status.DecryptionErr,
+	})
 	return nil
 }
 
-type kmsKeyStatusMsg madmin.KMSKeyStatus
+type kmsKeyStatusMsg struct {
+	KeyID         string `json:"keyId"`
+	Encryption    bool   `json:"encryption"`
```
aead, line 83: This is redundant state. `Encryption` is always `false` if `EncryptionErr != ""` resp. `true` if `EncryptionErr == ""`. Also a boolean doesn't say much - "What does `Encryption == true` means?"

cmd/admin-kms-key-status.go
```diff
@@ -19,9 +19,11 @@ package cmd
 import (
 	"fmt"
 
+	"github.com/fatih/color"
 	"github.com/minio/cli"
+	json "github.com/minio/mc/pkg/colorjson"
 	"github.com/minio/mc/pkg/probe"
-	"github.com/minio/minio/pkg/madmin"
+	"github.com/minio/minio/pkg/console"
```
aead, line 26: Is it intentional that we don't use the types defined in `madmin`? Now, there is no connection between `mc` and `minio` anymore and if you change one you have to remember to change the other as well. Oherwise you may encounter a parsing error at runtime.

cmd/admin-kms-key-status.go
harshavardhana: It is on purpose to make sure that json output is consistent it is okay to have such redundant state

cmd/admin-kms-key-status.go
harshavardhana: Yes because mc needs it's own API and it's own output, each of mc output should have some standard fields.

Every command follows this excluding KMS key status.

aead: That is not really what I'm saying. I'm asking whether it is okay to decouple the types `mc` (here new structs) expects as result from `minio` (madmin structs). By doing so, a change in `madmin` does not have any effect on `mc`. For example if `madmin` changes the representation of type `Foo` it has to be **remembered* to do the same change in `mc`. The compiler will no longer report an error.

Also, the claim that other commands don't rely on `madmin` types is not true. There are various examples - e.g.:
```
type logMessage struct {
	madmin.LogInfo
}
```

harshavardhana: > Also, the claim that other commands don't rely on `madmin` types is not true. There are various examples - e.g.:
> 
> ```
> type logMessage struct {
> 	madmin.LogInfo
> }
> ```

This should be considered a bug @aead - it is decoupled the `pkg/madmin` struct doesn't serve the purpose of `mc` UI - `-json` is a representation of the UI output.

harshavardhana: > By doing so, a change in `madmin` does not have any effect on `mc`. For example if `madmin` changes the representation of type `Foo` it has to be *_remembered_ to do the same change in `mc`. The compiler will no longer report an error.

if you change the struct you change the UI in `mc` if the user needs to be aware of it, if not you don't need to. The point is there shouldn't be such dependency on `pkg/madmin` structs. 

Some structs might be directly usable if they carry all the right values needed for the `mc` UI - if they don't then we should move on. 

NOTE: I can still make this message but

```
// KMSKeyStatus contains some status information about a KMS master key.
// The MinIO server tries to access the KMS and perform encryption and
// decryption operations. If the MinIO server can access the KMS and
// all master key operations succeed it returns a status containing only
// the master key ID but no error.
type KMSKeyStatus struct {
        KeyID         string `json:"key-id"`
        EncryptionErr string `json:"encryption-error,omitempty"` // An empty error == success
        DecryptionErr string `json:"decryption-error,omitempty"` // An empty error == success
}
```

uses a non JSONic key values such as `key-id` instead it should be `keyId` or atleast `key_id` - this is why I have to end up re-defining it.

harshavardhana: In-fact the madmin.LogInfo has the necessary JSON tags properly set, but it still doesn't follow the right requirement for `mc`

kannappanr (approved):
LGTM

aead (approved):

Issue #3413, opened 2020-09-21
mc mirror --watch implies --remove


rafaelsierra:
## Expected behavior

`mc mirror --watch . s3/my-bucket` should not remove files that were removed

## Actual behavior

`mc mirror --watch` behaves like `mc mirror --watch --remove`

## Steps to reproduce the behavior
```
mc mirror --watch . s3/my-bucket
touch a
touch b
rm a
```

File `a` should not be removed from `s3/my-bucket`

## mc --version
mc version DEVELOPMENT.GOGET (from around 2020-09-18)

## System information
go version go1.15.2 linux/arm64
Running on raspberry pi 3 model b

rafaelsierra:
I suppose this is the intended behavior, if that is the case, than I would suggest a new feature to add `--keep` to `--watch` so that it will never attempt to remove files.

My goal running this `mc mirror --watch` is to create a backup on the bucket from files that have to be removed locally due to lack of disk space (i.e. I have way more space on my bucket than on my raspberry)

harshavardhana:
This is by design @rafaelsierra we have gone few iterations over this and it is worthless for us to keep going back and forth on this. 

Some want it `--remove` with `--watch` and some do not, we cannot satisfy everyone.

rafaelsierra:
What about my suggestion to add `--keep` that would negate this behavior? Have this been discussed as well?

rafaelsierra:
I have been running this PR version for some hours now and it is working as intended (with the `--keep` flag set).

@harshavardhana is this an interesting idea or should I scrap the PR?

tannerjfco:
@harshavardhana if this is by design, it is completely terrible design. The existence of the remove flag implies that if the remove flag isn't set, then things should not be removed. That is also the original behavior of mirror, and introducing a behavior change with no documentation, and no change to the flags design for the command, is a really poor show. Anything that can cause unexpected deletions is a terrible design, and deletions and overwrites should be well safeguarded. The current flag design for the mirror command implies that they are, but that simply isn't true if mirror can remove objects without --remove being set. What is the point of --remove if mirror will remove without the flag set? The point becomes even less clear if a --keep flag as proposed in #3414 is required to provide the opposite effect

rafaelsierra:
I would rather have someone from the project to close this as "won't do" or "working as expected" than letting it close due to staleness.

[...]

PR #3414, opened 2020-09-21
Make it possible to mirror without delete

rafaelsierra:
This PR introduces a proof of concept for the idea to make it possible to use `mc mirror --watch` without deleting target files by using a new flag called `--keep`


rafaelsierra:
Related to #3413

PR #3422, opened 2020-09-28
rm: Remove a delete marker

vadmeste:
A code refactor made rm to invoke HEAD request instead of LIST to stat
object to remove, and this introduced a problem removing a delete-marker
since HEAD of a delete-marker returns 405 Method Not Allowed


kannappanr (approved):
LGTM

donatello (approved):
LGTM

PR #3427, opened 2020-09-29
Makefile should check for possible changes in the generated code

vadmeste:
We should not release mc while the code is not according to the
generated code, like when we forgot to run go generate command.


kannappanr (approved):
LGTM

poornas (approved):

PR #3439, opened 2020-10-06
client-s3: List sends bucket names when bucket not specified

vadmeste:
This commit also has some simplifications in the the listing code.

Fixes https://github.com/minio/minio/issues/10630


harshavardhana (approved):

kannappanr (approved):
LGTM

PR #3443, opened 2020-10-07
Add a --label flag to mc admin bucket remote add command

poornas:
This label will be used in lieu of transition storage class for [ilm transition](https://github.com/minio/minio/pull/10565)

### DIFF
--- a/cmd/admin-bucket-remote-add.go
+++ b/cmd/admin-bucket-remote-add.go
@@ -46,6 +46,10 @@ var adminBucketRemoteAddFlags = []cli.Flag{
 		Name:  "region",
 		Usage: "region of the destination bucket (optional)",
 	},
+	cli.StringFlag{
+		Name:  "label",
+		Usage: "set a label to identify this target (optional)",
+	},
 }
 var adminBucketRemoteAddCmd = cli.Command{
 	Name:   "add",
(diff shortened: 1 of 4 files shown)


kannappanr (approved):
LGTM

cmd/admin-bucket-remote-add.go
```diff
@@ -82,7 +86,7 @@ EXAMPLES:
      {{.DisableHistory}}
      {{.Prompt}} {{.HelpName}} myminio/srcbucket \
                  https://foobar:foo12345@minio2:9000/replicabucket \
-                 --service "replication" --region "us-west-1"
+                 --service "replication" --region "us-west-1" --label "replica1"
```
harshavardhana, line 89:
```suggestion
                 --service "replication" --region "us-west-1" --label "hdd-tier"
```

harshavardhana (changes requested):
I am not sure why is this service called `replication` shouldn't the examples be called `service=transition` ?

poornas:
> I am not sure why is this service called `replication` shouldn't the examples be called `service=transition` ?

The ilm transition PR needs to be in before I can do that - currently only valid service is replication on minio master branch

harshavardhana:
> The ilm transition PR needs to be in before I can do that - currently only valid service is replication on minio master branch

Is label feature also available for replication ? if so how is it used?

poornas:
> > The ilm transition PR needs to be in before I can do that - currently only valid service is replication on minio master branch
> 
> Is label feature also available for replication ? if so how is it used?

no, this is just a tag to identify the remote and is ignored by replication. It is repurposed in ilm transition to also sub as the storage-class name.

vadmeste (approved):
LGTM

PR #3456, opened 2020-10-19
URL_HOST_* requires users to encode weird characters

vadmeste:
Characters such as '#', ':' and other need to be encoded
using url encoding in order to correct support them.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -208,56 +208,37 @@ func mustGetHostConfig(alias string) *aliasConfigV10 {
 	return aliasCfg
 }
 
-var (
-	hostKeys      = regexp.MustCompile("^(https?://)(.*?):(.*?)@(.*?)$")
-	hostKeyTokens = regexp.MustCompile("^(https?://)(.*?):(.*?):(.*?)@(.*?)$")
-)
-
-// parse url usually obtained from env.
-func parseEnvURLStr(envURL string) (*url.URL, string, string, string, *probe.Error) {
-	var accessKey, secretKey, sessionToken string
-	var parsedURL string
-	if hostKeyTokens.MatchString(envURL) {
-		parts := hostKeyTokens.FindStringSubmatch(envURL)
-		if len(parts) != 6 {
-			return nil, "", "", "", errInvalidArgument().Trace(envURL)
-		}
-		accessKey = parts[2]
-		secretKey = parts[3]
-		sessionToken = parts[4]
-		parsedURL = fmt.Sprintf("%s%s", parts[1], parts[5])
-	} else if hostKeys.MatchString(envURL) {
-		parts := hostKeys.FindStringSubmatch(envURL)
-		if len(parts) != 5 {
-			return nil, "", "", "", errInvalidArgument().Trace(envURL)
-		}
-		accessKey = parts[2]
-		secretKey = parts[3]
-		parsedURL = fmt.Sprintf("%s%s", parts[1], parts[4])
-	}
-	var u *url.URL
-	var e error
-	if parsedURL != "" {
-		u, e = url.Parse(parsedURL)
-	} else {
-		u, e = url.Parse(envURL)
-	}
+// Parse MC_HOST url obtained from the environment
+// - It supports the following formats:
+//      http(s)://access-key:secret-key@host:port/
+//    AND
+///     http(s)://access-key:secret-key:session-token@host:port/
+// - Users need to manually escape any odds characters (except for @)
+//   in order for this to work correctly.
+func parseEnvURLStr(envURL string) (endpoint, accessKey, secretKey, sessionToken string, err *probe.Error) {
+	u, e := url.Parse(envURL)
 	if e != nil {
-		return nil, "", "", "", probe.NewError(e)
+		return "", "", "", "", probe.NewError(e)
 	}
+
+	endpoint = u.Scheme + "://" + u.Host
+	accessKey = u.User.Username()
+
+	pwd, _ := u.User.Password()
+	pwdFields := strings.Split(pwd, ":")
+	secretKey = pwdFields[0]
+	if len(pwdFields) > 1 {
+		sessionToken = pwdFields[1]
+	}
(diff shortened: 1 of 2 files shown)


harshavardhana:
> Characters such as '#', ':' and other need to be encoded
> using url encoding in order to correct support them.

Looks like you are introducing a regression @vadmeste - you breaking support for STS credentials by removing the existing code

vadmeste:
> Looks like you are introducing a regression @vadmeste - you breaking support for STS credentials by removing the existing code

yes, there is a functional change here, it is needed if we want to support some odd characters in the access or secret or session key. Those odd characters need to be encoded.

Where did you find the regression ?

harshavardhana:
> yes, there is a functional change here, it is needed if we want to support some odd characters in the access or secret or session key. Those odd characters need to be encoded.

why does it need to be encoded? @vadmeste we already support odd characters by parsing using regex, we don't need url.Parse - url.Parse won't support our STS requirements.

https://github.com/minio/mc/blob/master/docs/minio-client-complete-guide.md#rotating-credentials

cmd/config.go
```diff
@@ -208,56 +208,37 @@ func mustGetHostConfig(alias string) *aliasConfigV10 {
...
 	if e != nil {
-		return nil, "", "", "", probe.NewError(e)
+		return "", "", "", "", probe.NewError(e)
 	}
+
+	endpoint = u.Scheme + "://" + u.Host
+	accessKey = u.User.Username()
+
+	pwd, _ := u.User.Password()
+	pwdFields := strings.Split(pwd, ":")
```
harshavardhana, line 228: You cannot assume that `password` doesn't have `:` @vadmeste - that's why the previous regex was aptly positional.

cmd/config_test.go
```diff
@@ -29,7 +33,7 @@ func TestParseEnvURLStr(t *testing.T) {
 		port         string
 	}{
 		{
-			hostURL:   "https://minio:minio1#23@localhost:9000",
+			hostURL:   "https://minio:minio1%2323@localhost:9000",
```
harshavardhana, line 36: This is not correct users won't pass URL encoded passwords in this manner @vadmeste and they are not expected to.

harshavardhana (changes requested):

cmd/config_test.go
vadmeste: > This is not correct users won't pass URL encoded passwords in this manner @vadmeste and they are not expected to.

Basically, this change is needed if the access or secret keys contains '@' or ':'

I think MinIO or AWS has # or @, this should not be a problem. But one user seems to have picked a key with '@' in it and that made an issue with MC_HOST_ parsing in mc code.

cmd/config.go
vadmeste: > You cannot assume that password doesn't have : @vadmeste - that's why the previous regex was aptly positional.

Not sure I understand, if we cannot assume that the password doesn't have ':', how we will extract the session token from the url in this case ?

cmd/config_test.go
harshavardhana: The current fix is wrong because you are expecting users to provide an URL encoded form, they will never be able to provide that as input. 

I think you can simply ignore that user they avoid creating passwords in that manner - I don't think this PR is needed.

harshavardhana:
If you really want to fix this - the correct fix is this 

```diff
git diff
diff --git a/cmd/config.go b/cmd/config.go
index 602de4b1..ca583070 100644
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -209,8 +209,8 @@ func mustGetHostConfig(alias string) *aliasConfigV10 {
 }
 
 var (
-       hostKeys      = regexp.MustCompile("^(https?://)(.*?):(.*?)@(.*?)$")
-       hostKeyTokens = regexp.MustCompile("^(https?://)(.*?):(.*?):(.*?)@(.*?)$")
+       hostKeys      = regexp.MustCompile("^(https?://)(.*?):(.*)@(.*?)$")
+       hostKeyTokens = regexp.MustCompile("^(https?://)(.*?):(.*?):(.*)@(.*?)$")
 )
 
 // parse url usually obtained from env.
diff --git a/cmd/config_test.go b/cmd/config_test.go
index e774f267..a94017d4 100644
--- a/cmd/config_test.go
+++ b/cmd/config_test.go
@@ -36,9 +36,9 @@ func TestParseEnvURLStr(t *testing.T) {
                        port:      "9000",
                },
                {
-                       hostURL:   "https://minio:minio123@localhost:9000",
+                       hostURL:   "https://minio:minio123@@localhost:9000",
                        accessKey: "minio",
-                       secretKey: "minio123",
+                       secretKey: "minio123@",
                        hostname:  "localhost",
                        port:      "9000",
                },
@@ -57,6 +57,14 @@ func TestParseEnvURLStr(t *testing.T) {
                        hostname:     "localhost",
                        port:         "9000",
                },
+               {
+                       hostURL:      "https://minio:minio@123:token@@localhost:9000",
+                       accessKey:    "minio",
+                       secretKey:    "minio@123",
+                       sessionToken: "token@",
+                       hostname:     "localhost",
+                       port:         "9000",
+               },
                {
                        hostURL:   "https://minio@localhost:9000",
                        accessKey: "minio",
```

vadmeste:
> If you really want to fix this - the correct fix

This still doesn't work if the secret key contains a ':' for example. Anyway, I wanted a simpler code with a little trade-off of not supporting certain characters unless the user provides an encoded form of that character.

harshavardhana:
> This still doesn't work if the secret key contains a ':' for example. Anyway, I wanted a simpler code with a little trade-off of not supporting certain characters unless the user provides an encoded form of that character.

No that is not supposed to work @vadmeste for that they shouldn't be using ENV - encoded form no one will ever use on command and it is not practical to ask this to our users.

The current code satisfies all the other 90% requirements.

vadmeste:
Closing this as already fixed here https://github.com/minio/mc/pull/3464

PR #3457, opened 2020-10-19
Update mc retention, remove mc lock links.

ravindk89:
I think we missed updating `mc retention` after removing `mc lock` in the client complete guide. Made a small patch to fix. Future docs already reflect this information.


kannappanr (approved):
LGTM

vadmeste (approved):
LGTM

PR #3461, opened 2020-10-21
Show better looking help when a wrong flag is passed

vadmeste:
To test, type `mc admin user add -zz`

This change just implements onUsageError() function and adds it to cli.Command to various available commands.

Unfortunately there is no way to support `mc admin user -zz` or change its behavior.

### DIFF
--- a/cmd/mb-main.go
+++ b/cmd/mb-main.go
@@ -18,7 +18,6 @@ package cmd
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/fatih/color"
 	"github.com/minio/cli"
--- a/cmd/mb-main.go
+++ b/cmd/mb-main.go
@@ -154,7 +154,6 @@ func mainMakeBucket(cli *cli.Context) error {
 			continue
 		}
 
-
 		// Successfully created a bucket.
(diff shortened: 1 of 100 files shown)


cmd/mb-main.go
```diff
@@ -153,6 +155,8 @@ func mainMakeBucket(cli *cli.Context) error {
 			continue
 		}
 
+		fmt.Println(clnt.GetURL())
```
poornas, line 158: needs to be removed

poornas (approved):
LGTM. Println needs to be removed from mb command

ebozduman (approved):
LGTM

cmd/mb-main.go
harshavardhana, line 158:
```suggestion
```

harshavardhana (changes requested):

harshavardhana (approved):

cmd/mb-main.go
```diff
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"context"
+	"fmt"
```
harshavardhana, line 21:
```suggestion
```

PR #3462, opened 2020-10-21
ls: Fix listing buckets with --versions

vadmeste:
mc ls --versions alias/ has an unexpected output due to missing
consideration of this use case. This commit fixes the behavior.

To test:
```mc ls --versions play/```


klauspost (approved):
LGTM

kannappanr (approved):

PR #3468, opened 2020-10-22
Simplify queueing to parallel manager

vadmeste:
No function change in this commit, just a bit of simplification before the implement of queuing a task with a barrier


harshavardhana (approved):

kannappanr (approved):
LGTM

PR #3469, opened 2020-10-26
Implement barrier in parallel manager

vadmeste:
Mirroring watch supports parallel manager to run actions of the received
events in parallel, but sometimes the order of the execution of those
actions matter.

For example:
- A new bucket is created
- A new file is uploaded in the new bucket

This PR will bring support of the barrier concept. A task with a barrier
will only executed after all the previous tasks finish, the
subsequent tasks will also be only executed when the barrier task is
finished.

Later, we will bring support of tasks dependency to optimize the
parallelism.

### DIFF
--- a/cmd/parallel-manager.go
+++ b/cmd/parallel-manager.go
@@ -34,6 +34,15 @@ const (
 // Number of workers added per bandwidth monitoring.
 var defaultWorkerFactor = runtime.GOMAXPROCS(0)
 
+// A task is a copy/mirror action that needs to be executed
+type task struct {
+	// The function to execute in this task
+	fn func() URLs
+	// If set to true, ensure no tasks are
+	// executed in parallel to this one.
+	barrier bool
+}
+
 // ParallelManager - helps manage parallel workers to run tasks
 type ParallelManager struct {
 	// Calculate sent bytes.
@@ -43,13 +52,14 @@ type ParallelManager struct {
 	sentBytes int64
 
 	// Synchronize workers
-	wg *sync.WaitGroup
+	wg          *sync.WaitGroup
+	barrierSync sync.RWMutex
 
 	// Current threads number
 	workersNum uint32
 
 	// Channel to receive tasks to run
-	queueCh chan func() URLs
+	queueCh chan task
 
 	// Channel to send back results
 	resultCh chan URLs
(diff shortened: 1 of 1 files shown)


vadmeste:
@krisis @harshavardhana can this get reviewed ?

cmd/parallel-manager.go
```diff
@@ -34,6 +34,11 @@ const (
 	defaultWorkerFactor = 2
 )
 
+type task struct {
+	fn      func() URLs
+	barrier bool
```
klauspost, line 39: Please document exactly what this does.

cmd/parallel-manager.go
```diff
@@ -136,7 +150,21 @@ func (p *ParallelManager) monitorProgress() {
...
 func (p *ParallelManager) queueTask(fn func() URLs) {
-	p.queueCh <- fn
+	p.doQueueTask(task{fn: fn})
+}
+
+func (p *ParallelManager) queueTaskWithBarrier(fn func() URLs) {
+	p.doQueueTask(task{fn: fn, barrier: true})
+}
+
+func (p *ParallelManager) doQueueTask(t task) {
```
klauspost, line 160: So this will block *before* the task is executed and wait until it is done before it returns?

It would seem that this function cannot be called concurrently. Please document.

cmd/parallel-manager.go
```diff
@@ -136,7 +150,21 @@ func (p *ParallelManager) monitorProgress() {
...
+	if t.barrier {
+		p.barrierSync.Wait()
```
klauspost, line 162: >  If a WaitGroup is reused to wait for several independent sets of events, new Add calls must happen after all previous Wait calls have returned.

Can you guarantee that?

It seems to me that a goroutine can reach `p.barrierSync.Add(1)` *after* the Wait has been reached here.

klauspost (changes requested):
The sync mechanism will need to be different.

I think we could 'abuse' a `syncRWMutex`. This would mean that for non-barrier calls you grab a read lock and when you have a barrier call you grab a write lock.

The semantics of `RWMutex` makes it so that whenever a Write lock is requested no more read locks will be handed out. Let me see if I can make a proposal.

klauspost:
@vadmeste change struct memmber to  `barrierSync sync.RWMutex` The worker:
```
	go func() {
		for {
			// Wait for jobs
			t, ok := <-p.queueCh
			if !ok {
				// No more tasks, quit
				p.wg.Done()
				return
			}

			// Execute the task and send the result to channel.
			p.resultCh <- t.fn()

			if t.barrier {
				p.barrierSync.Unlock()
			} else {
				p.barrierSync.RUnlock()
			}
		}
	}()
```

The func:
```
func (p *ParallelManager) doQueueTask(t task) {
	if t.barrier {
		p.barrierSync.Lock()
	} else {
		p.barrierSync.RLock()
	}
	p.queueCh <- t
}
```
Since only one can call at the time can be in `doQueueTask` anyway, there is no need to wait for it to finish, so the `barrierDone` can be removed.

klauspost:
To clarify, there is no need to wait since new tasks cannot be added until the mutex has been unlocked.

vadmeste:
Oh I like the RW mutex idea, will do it.

vadmeste:
Can this get reviewed @krisis @harshavardhana ? I need this for another PR

krisis (approved):

harshavardhana (approved):

PR #3470, opened 2020-10-26
Fix client-s3.Makebucket behavior with object in path

vadmeste:
The S3 client supports creating buckets & directories in MakeBucket call, but currently
it calls path.Dir() for no apparent reason.

Remove path.Dir() call and add a trailing slash to the object if it doesnt have one.


harshavardhana (approved):

kannappanr (approved):
LGTM

PR #3471, opened 2020-10-27
fix `mc stat` with --versions flag to show versionID correctly

poornas:
If --versions flag is specified, instead of HEAD on each version, currently
repeated listing is being done and latest version ID reported
for all the versions instead of actual version ID.

`mc stat source/bucket/obj --versions`

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -1470,6 +1470,10 @@ func (c *S3Client) getObjectStat(ctx context.Context, bucket, object string, opt
 		}
 		return nil, probe.NewError(e)
 	}
+	// HEAD with a version ID will not return version in the response headers
+	if objectMetadata.VersionID == "" {
+		objectMetadata.VersionID = opts.VersionID
+	}
 	return objectMetadata, nil
 }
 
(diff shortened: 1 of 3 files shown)


cmd/client-s3.go
```diff
@@ -1416,6 +1416,8 @@ func (c *S3Client) Stat(ctx context.Context, opts StatOptions) (*ClientContent,
 		// so we can check if there is such prefix which exists
 		ctnt, err := c.getObjectStat(ctx, bucket, object, minio.StatObjectOptions{ServerSideEncryption: opts.sse, VersionID: opts.versionID})
 		if err == nil {
+			// HEAD with a version ID will not return version in the metadata
+			ctnt.VersionID = opts.versionID
```
vadmeste, line 1420: Maybe, instead we make c.getObjectStat() or even more lower level to return ctnt.VersionID if available

vadmeste (commented):
LGTM but with one comment

cmd/client-s3.go
poornas: that's a good idea, done.

harshavardhana (approved):

PR #3488, opened 2020-11-12
Copy bucket policy/locking only with new buckets and with -a

vadmeste:
mc was not requiring -a to trigger mirroring of bucket policy & locking
configurations, this commit enforces this requirement. Besides, the config
mirroring will only happen if mirror command is creating new buckets in
the target server.


harshavardhana (approved):

poornas (approved):
LGTM

PR #3491, opened 2020-11-13
multiple bug fixes for mirror, ls, rm

harshavardhana:
- fixes errant top level directory creation, regression introduced in #3313
- fixes `mc ls -r` regression introduced in #3476
- fixes other bugs such as `mc mirror` would perpetually hang if the
  destination bucket doesn't exist, create and proceed instead.
- removes DirOpt style implementations for List, ListIncompleteUploads
  these are not useful anymore.


kannappanr (approved):
LGTM

PR #3505, opened 2020-11-25
mirror: Watch should not imply Active active necessarly

vadmeste:
Also simplify the code by automatically triggering --overwrite and
--remove when --active--active is passed.


harshavardhana:
This was on purpose @vadmeste since usually watch means active-active - no one wants just watch

vadmeste:
> This was on purpose @vadmeste since usually watch means active-active - no one wants just watch

maybe then we need to hide --active-active flag in that case.

I will send another PR anyway to implement your mc-mirror-origin idea

PR #3508, opened 2020-11-26
mirror: Stop mirroring loop using source IPs

vadmeste:
Currently, mc mirror avoid mirroring an object if its contains MM mtime
header, but this causes an issue for users running two mc mirror
commands to synchronize three sites, A -> B and B -> C.

The solution for this is to record the endpoint address of the source
server in the object metadata of each mirroring operation and stops the
mirroring when the target server already exists in the metadata itself.

-- 
Has some issues that need discussion, we might need to get server's deployment ID instead of relying on server addresses

### DIFF
--- a/cmd/difference.go
+++ b/cmd/difference.go
@@ -63,6 +63,19 @@ func (d differType) String() string {
 	return "unknown"
 }
 
+const activeActiveSourceURLsKey = "X-Amz-Meta-Mm-Source-Urls"
+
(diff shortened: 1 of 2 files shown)


harshavardhana:
We have to sort of make it generic that's why I thought server address is okay, server address do not change transparently on the server side. 

We can assume that the alias user is pointing to is a common load balancer or a DNS for each site.

PR #3514, opened 2020-12-01
Move to Red Hat UBI

nitisht:
Also use multi-arch Dockerfile instead of separate
files for each arch.


harshavardhana (approved):

kannappanr (approved):

maoxuner:
I used to run command such as `tar` from base image alpine during my drone ci steps, but now it ends in failure.

PR #3544, opened 2020-12-19
Update to latest minio version RELEASE.2020-12-18T03-27-42Z

cesnietor:


harshavardhana:
@cesnietor this is not a correct fix PTAL #3545

PR #3547, opened 2020-12-23
Remove transition targets

poornas:
from `mc admin bucket remote` command and references to
transition s.c label in `mc ilm add` command and docs.

This is being deprecated in favor of an improved approach
to transition in the works.


krisis (approved):

kannappanr (approved):
LGTM

PR #3548, opened 2020-12-24
Remove IsInstalled() call before auto-completion installation

Frederick888:
At the moment posener/complete [1] doesn't allow specifying shell for
IsInstalled(), which might lead to confusing error messages, since it
just returns true if it's able to find completion definition in *any*
shell configurations even if it's not what the user currently uses.

Hence we should remove this check and call Install() directly, and user
can get more details in this case.

[1] https://github.com/posener/complete

For example, I'm currently using Zsh, but there are some leftover Fish configuration files in my system (I have uninstalled Fish).
```sh
# before this patch
# confusing error message when auto-completion is installed in Fish but not Zsh
$ mcli --autocompletion
mcli: autocompletion is already enabled in your 'zsh' shell.

# after this patch
$ mcli --autocompletion
mcli: <ERROR> Unable to install auto-completion. 1 error occurred:
        * already installed at /home/frederick/.config/fish/completions/mcli.fish.
# user can now know the details and optionally remove/update the files
$ rm -rf .config/fish
# side-effect: the `mcli --autocompletion` call above actually already installed it in Zsh
$ mcli --autocompletion
mcli: <ERROR> Unable to install auto-completion. 1 error occurred:
        * already installed in /home/frederick/.zshrc.
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -18,9 +18,12 @@ package cmd
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"sort"
 	"strconv"
@@ -289,34 +292,52 @@ func installAutoCompletion() {
 
 	shellName := os.Getenv("SHELL")
 	if shellName == "" {
-		shellName = "$SHELL"
+		ppid := os.Getppid()
+		cmd := exec.Command("ps", "-p", strconv.Itoa(ppid), "-o", "comm=")
+		ppName, err := cmd.Output()
+		if err != nil {
+			fatalIf(probe.NewError(err), "Failed to enable autocompletion. Cannot determine shell type and"+
+				"no SHELL environment variable found")
+		}
+		shellName = strings.TrimSpace(string(ppName))
+		console.Infoln("No 'SHELL' env var. Your shell is auto determined as '" + shellName + "'.")
 	} else {
-		shellName = filepath.Base(shellName)
+		console.Infoln("Your shell is set to '" + shellName + "', by env var 'SHELL'.")
 	}
+	shellName = strings.ToLower(filepath.Base(shellName))
+
 	supportedShells := map[string]bool{
 		"bash": true,
 		"zsh":  true,
 		"fish": true,
 	}
 
+	if !supportedShells[shellName] {
+		fatalIf(probe.NewError(errors.New("")),
+			"'"+shellName+"' is not a supported shell. "+
+				"Supported shells are: bash, zsh, fish")
+	}
+
 	err := completeinstall.Install(filepath.Base(os.Args[0]))
-	if err != nil {
-		if completeinstall.IsInstalled(filepath.Base(os.Args[0])) || completeinstall.IsInstalled("mc") {
-			errStr := err.Error()
-			if supportedShells[shellName] {
-				console.Infoln("autocompletion is already enabled in your '"+shellName+"' shell.\n", errStr[strings.Index(errStr, "\n")+1:])
-			} else {
-				console.Infoln("autocompletion is already enabled\n", errStr[strings.Index(errStr, "\n")+1:])
-			}
-			return
-		}
-		fatalIf(probe.NewError(err), "Unable to install auto-completion.")
-	} else {
-		if supportedShells[shellName] {
-			console.Infoln("enabled autocompletion in your '" + shellName + "' rc file. Please restart your shell.")
+	var printMsg string
+	if err != nil && strings.Contains(err.Error(), "* already installed") {
+		errStr := err.Error()[strings.Index(err.Error(), "\n")+1:]
+		re := regexp.MustCompile(`[::space::]*\*.*` + shellName + `.*`)
+		relatedMsg := re.FindStringSubmatch(errStr)
+		if len(relatedMsg) > 0 {
+			printMsg = "\n" + relatedMsg[0]
 		} else {
-			console.Infoln("enabled autocompletion in bash, zsh and fish where possible.")
+			printMsg = ""
 		}
+	}
+	if printMsg != "" {
+		if completeinstall.IsInstalled(filepath.Base(os.Args[0])) || completeinstall.IsInstalled("mc") {
+			console.Infoln("autocompletion is enabled.", printMsg)
+		} else {
+			fatalIf(probe.NewError(err), "Unable to install auto-completion.")
+		}
+	} else {
+		console.Infoln("enabled autocompletion in your '" + shellName + "' rc file. Please restart your shell.")
 	}
 }
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -294,11 +294,6 @@ func installAutoCompletion() {
 		shellName = filepath.Base(shellName)
 	}
 
-	if completeinstall.IsInstalled(filepath.Base(os.Args[0])) || completeinstall.IsInstalled("mc") {
-		console.Infoln("autocompletion is already enabled in your '" + shellName + "' shell.")
-		return
-	}
-
 	err := completeinstall.Install(filepath.Base(os.Args[0]))
 	if err != nil {
 		fatalIf(probe.NewError(err), "Unable to install auto-completion.")
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -296,6 +296,11 @@ func installAutoCompletion() {
 
 	err := completeinstall.Install(filepath.Base(os.Args[0]))
 	if err != nil {
+		if completeinstall.IsInstalled(filepath.Base(os.Args[0])) || completeinstall.IsInstalled("mc") {
+			errStr := err.Error()
+			console.Infoln("autocompletion is already enabled in your '"+shellName+"' shell.\n", errStr[strings.Index(errStr, "\n")+1:])
+			return
+		}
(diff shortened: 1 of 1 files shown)


ebozduman (changes requested):
The purpose of having it there was to show a softer message rather than an error message. That is whenever a user attempts to install auto completion, there wil be an error message in red:
```
$ mc --autocompletion
mc: <ERROR> Unable to install auto-completion. 1 error occurred:
	* already installed in /home/ersan/.bashrc.
```
So, we should not alarm the user with something s/he does not need to do anything about as there is nothing wrong, no error condition here.

If you would like to see more detailed message about where it has been already installed, you might add that information into the message `autocompletion is already enabled in your '" + shellName + "' shell.`

Frederick888:
@ebozduman I'm not familiar with Go programming but I'm under the impression that posener/complete v1 doesn't have an API to list shells with completions installed?

And tbh I think ideally we should be able to instruct the completion library for which shell we want to install the completion, e.g. `completeinstall.Install(filepath.Base(os.Args[0]), shellName)`... I wonder if there are any alternatives of posener/complete in the Go space?

cmd/main.go
```diff
@@ -294,11 +294,6 @@ func installAutoCompletion() {
 		shellName = filepath.Base(shellName)
 	}
 
-	if completeinstall.IsInstalled(filepath.Base(os.Args[0])) || completeinstall.IsInstalled("mc") {
```
ebozduman, line 297: > @ebozduman I'm not familiar with Go programming but I'm under the impression that posener/complete v1 doesn't have an API to list shells with completions installed?

@Frederick888,

OK. Then I suggest the change to be as follows.
This way, it addresses both concerns, yours and mine :

```
:
:
	shellName := os.Getenv("SHELL")
	if shellName == "" {
		shellName = "$SHELL"
	} else {
		shellName = filepath.Base(shellName)
	}

	err := completeinstall.Install(filepath.Base(os.Args[0]))
	if err != nil {
		if completeinstall.IsInstalled(filepath.Base(os.Args[0])) || completeinstall.IsInstalled("mc") {
			errStr := err.Error()
			console.Infoln("autocompletion is already enabled in your '"+shellName+"' shell.\n", errStr[strings.Index(errStr, "\n")+1:])
			return
		}
		fatalIf(probe.NewError(err), "Unable to install auto-completion.")
	} else {
		console.Infoln("enabled autocompletion in your '" + shellName + "' rc file. Please restart your shell.")
	}
:
:
```

ebozduman (changes requested):

cmd/main.go
Frederick888: Just pushed 2 commits. First one applied your suggestion. Second one was some extra precautions so that it doesn't confuse users if current shell is not supported. I admit it's ugly though... Feel free to do any modifications or let me know your suggestions.

cmd/main.go
```diff
@@ -293,17 +293,30 @@ func installAutoCompletion() {
...
-		return
+	supportedShells := map[string]bool{
+		"bash": true,
+		"zsh":  true,
+		"fish": true,
 	}
 
 	err := completeinstall.Install(filepath.Base(os.Args[0]))
 	if err != nil {
+		if completeinstall.IsInstalled(filepath.Base(os.Args[0])) || completeinstall.IsInstalled("mc") {
```
ebozduman, line 304: @Frederick888 ,

My first comment: What is the point of adding ` || complete.Installed("mc")` ?

Since the auto  completion command is `mc --autocompletion`,  `filepath.Base(os.Args[0])` returns `mc`.

So, adding  ` || complete.Installed("mc")` again is unnecessary. 

Do I miss something?

Frederick888: It's renamed to `mcli` since `mc` is already used by https://midnight-commander.org under Arch (and other distros possibly, e.g. Ubuntu).

ebozduman: OK. That makes sense.

After all of our comments and exchange of ideas, I put together some new changes to also address some other issues, like when shell information cannot be determined we set the `shellName` variable to `$SHELL` and it did not make sense. I thought, we can determine the shell if there is no `SHELL` env var is defined, and if we still fail for any reason, we should stop and fail with an error message.

Hence, I propose the following changes. I believe these changes address all of our comments and concerns.

```
$ git diff
diff --git a/cmd/main.go b/cmd/main.go
index d4ea91f6e..ac31e2fb0 100644
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -18,9 +18,12 @@ package cmd
 
 import (
        "bytes"
+       "errors"
        "fmt"
        "os"
+       "os/exec"
        "path/filepath"
+       "regexp"
        "runtime"
        "sort"
        "strconv"
@@ -289,34 +292,49 @@ func installAutoCompletion() {
 
        shellName := os.Getenv("SHELL")
        if shellName == "" {
-               shellName = "$SHELL"
-       } else {
-               shellName = filepath.Base(shellName)
+               ppid := os.Getppid()
+               cmd := exec.Command("ps", "-p", strconv.Itoa(ppid), "-o", "comm=")
+               ppName, err := cmd.Output()
+               if err != nil {
+                       fatalIf(probe.NewError(err), "Failed to enable autocompletion. Cannot determine shell type and"+
+                               "no SHELL environment variable found")
+               }
+               shellName = strings.TrimSpace(string(ppName))
        }
+       shellName = strings.ToLower(filepath.Base(shellName))
+
        supportedShells := map[string]bool{
                "bash": true,
                "zsh":  true,
                "fish": true,
        }
 
+       if !supportedShells[shellName] {
+               fatalIf(probe.NewError(errors.New("")),
+                       "'"+shellName+"' is not a supported shell. "+
+                               "Supported shells are: bash, zsh, fish")
+       }
+
        err := completeinstall.Install(filepath.Base(os.Args[0]))
-       if err != nil {
-               if completeinstall.IsInstalled(filepath.Base(os.Args[0])) || completeinstall.IsInstalled("mc") {
-                       errStr := err.Error()
-                       if supportedShells[shellName] {
-                               console.Infoln("autocompletion is already enabled in your '"+shellName+"' shell.\n", errStr[strings.Index(errStr, "\n")+1:])
-                       } else {
-                               console.Infoln("autocompletion is already enabled\n", errStr[strings.Index(errStr, "\n")+1:])
-                       }
-                       return
+       var printMsg string
+       if err != nil && strings.Contains(err.Error(), "* already installed") {
+               errStr := err.Error()[strings.Index(err.Error(), "\n")+1:]
+               re := regexp.MustCompile(`[::space::]*\*.*` + shellName + `.*`)
+               relatedMsg := re.FindStringSubmatch(errStr)
+               if len(relatedMsg) > 0 {
+                       printMsg = "\n" + relatedMsg[0]
+               } else {
+                       printMsg = ""
                }
-               fatalIf(probe.NewError(err), "Unable to install auto-completion.")
-       } else {
-               if supportedShells[shellName] {
-                       console.Infoln("enabled autocompletion in your '" + shellName + "' rc file. Please restart your shell.")
+       }
+       if printMsg != "" {
+               if completeinstall.IsInstalled(filepath.Base(os.Args[0])) || completeinstall.IsInstalled("mc") {
+                       console.Infoln("autocompletion is enabled.", printMsg)
                } else {
-                       console.Infoln("enabled autocompletion in bash, zsh and fish where possible.")
+                       fatalIf(probe.NewError(err), "Unable to install auto-completion.")
                }
+       } else {
+               console.Infoln("enabled autocompletion in your '" + shellName + "' rc file. Please restart your shell.")
        }
 }
 ```
Check it out and let's discuss if there are more issues to be addressed.

Comment (Frederick888) [reply cmd/main.go:304]: @ebozduman Thanks for all the effort improving this, and blocking unsupported shells did make it cleaner. Could you push the commit from your side? GH converted all tabs to spaces and although I can apply the patch, it's too easy for me to mess up the indentation...

Comment (ebozduman) [reply cmd/main.go:304]: Sure

Comment (ebozduman) [reply cmd/main.go:304]: @Frederick888,

You need to give me permission to contribute to your branch.
Here are the steps to do it:
1. Go to your mc repository at `https://github.com/Frederick888/mc`,
2. Choose `Settings`,
3. Click on `Manage access`
4. Click on `Invite a collaborator`
5. Enter my username `ebozduman` in the pop-up window
6. Click on `Add ebozduman to this repository`
7. A pending invitation will show up on your window, and I'll get the invitation to push to your branch.

Comment (Frederick888) [reply cmd/main.go]: There is a corner case...

1. Completion already installed in `$HOME/.zshrc`
2. `$HOME/.bashrc` and [etc](https://github.com/posener/complete/blob/9a4745ac49b29530e07dc2581745a218b646b7a3/cmd/install/install.go#L86-L89) don't exist
3. In bash, run `mcli --autocompletion`
4. `mc` says it's installed for bash, but because of https://github.com/posener/complete/blob/v1/cmd/install/install.go#L91-L102, bash is actually skipped

FYI `err.Error()` in this case is:

```
mcli: err: 1 error occurred:
        * already installed in /home/frederick/.zshrc
```

Just thinking out loud here, what about running `completeinstall.Install()` twice and check if the second `err` matches the regexp?

ebozduman: I think we don't need any more changes.
We've already beaten the dead horse more than we were expected.
It's good to go.
LGTM

ebozduman (approved):

harshavardhana (approved):

cmd/main.go
Frederick888: Understandable lol... Thanks for the effort.

PR #3553, opened 2020-12-30
Fix CPU info struct in health v1

anjalshireesh:
Each server has only one CPU info that lists all CPUs and time stats.
However, it was being printed as a list (with only one element) in the
health report. Change the health v1 structures related to CPU info to
remove the unnecessary list part.

old output:

```
"cpus": [
  {
    "cpu": [
      {
        "cpu": 0,
        "vendorId": "GenuineIntel"
        ...
      },
      ...
    ],
    "time": [
      ...
    ]
  }
]
```
new output:
```
"cpus": {
  "cpu": [
    {
      "cpu": 0,
      "vendorId": "GenuineIntel"
      ...
    },
    ...
  ],
  "time": [
    ...
  ]
}
```

### DIFF
--- a/cmd/health_v1.go
+++ b/cmd/health_v1.go
@@ -39,14 +39,14 @@ type HwServersV1 struct {
 // HwServerV1 - server health Info
 type HwServerV1 struct {
 	Addr    string      `json:"addr"`
-	CPUs    []HwCPUV1   `json:"cpus,omitempty"`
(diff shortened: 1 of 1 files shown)


anjalshireesh:
This is a breaking change in the health report structure and must accompany a corresponding schema version change. Closing the PR for now; will revive it at a later point of time.

PR #3555, opened 2020-12-31
admin info: Show yellow dot when the cluster is not fully initialized

vadmeste:
When nodes are up but the cluster is not initialized, it is better to show
a yellow dot instead of a green one in mc admin info command.


harshavardhana (approved):

klauspost (approved):
LGTM

PR #3560, opened 2021-01-04
Fixes recursive removal of empty prefix

### LINKED ISSUE #3486: Invalid result for mc rb
Looks like "Removed ...... successfully" is being printed erroneously in this case, as its not actually being removed.

root@sandbox:~# mc rb dot9/electronics
mc: <ERROR> Failed to remove `dot9/electronics`, continuing.. The bucket you tried to delete is not empty
**Removed `dot9/electronics` successfully.**

root@sandbox:~# mc ls dot9/electronics      
[2020-11-11 20:24:48 UTC]     0B esp32/

It also looks like mc mv does not remove directories or prefixes, is this intended behaviour as well ?
mc version RELEASE.2020-10-03T02-54-56Z
minio    Version: 2020-11-10T21:02:24Z

ebozduman:
Fixes #3486

This fix is extremely prone to regression. I've tested it and I'll still be testing it more to make sure there is no regression caused by it.
Reviewers are also advised to be testing it rigorously.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -292,8 +292,8 @@ func expandAlias(aliasedURL string) (alias string, urlStr string, aliasCfg *alia
 		envConfig, ok = os.LookupEnv(mcEnvHostsDeprecatedPrefix + alias)
 		if ok {
 			errorIf(errInvalidArgument().Trace(mcEnvHostsDeprecatedPrefix+alias),
-				"`MC_HOSTS_<alias>` environment variable is deprecated. Please"+
-					"`MC_HOST_<alias>` instead for the same functionality.")
+				"`MC_HOSTS_<alias>` environment variable is deprecated. Please "+
+					"use `MC_HOST_<alias>` instead for the same functionality.")
 		}
 	}
 
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -291,7 +291,9 @@ func expandAlias(aliasedURL string) (alias string, urlStr string, aliasCfg *alia
 	if envConfig, ok = os.LookupEnv(mcEnvHostPrefix + alias); !ok {
 		envConfig, ok = os.LookupEnv(mcEnvHostsDeprecatedPrefix + alias)
 		if ok {
-			errorIf(errInvalidArgument().Trace(mcEnvHostsDeprecatedPrefix+alias), "`MC_HOSTS_<alias>` environment variable is deprecated. Please use `MC_HOST_<alias>` instead for the same functionality.")
+			errorIf(errInvalidArgument().Trace(mcEnvHostsDeprecatedPrefix+alias),
+				"`MC_HOSTS_<alias>` environment variable is deprecated. Please"+
+					"`MC_HOST_<alias>` instead for the same functionality.")
 		}
 	}
 
--- a/cmd/rm-main.go
+++ b/cmd/rm-main.go
@@ -479,6 +479,9 @@ func mainRm(cliCtx *cli.Context) error {
 	// Support multiple targets.
 	for _, url := range cliCtx.Args() {
 		if isRecursive || withVersions {
+			if isAliasURLDir(ctx, url, encKeyDB, time.Time{}) {
+				url = strings.TrimRight(url, string(os.PathSeparator))
+			}
 			e = listAndRemove(url, rewind, withVersions, isRecursive, isIncomplete, isFake, isBypass, olderThan, newerThan, encKeyDB)
 		} else {
 			e = removeSingle(url, versionID, isIncomplete, isFake, isForce, isBypass, olderThan, newerThan, encKeyDB)
--- a/cmd/rm-main.go
+++ b/cmd/rm-main.go
@@ -202,10 +202,14 @@ func checkRmSyntax(ctx context.Context, cliCtx *cli.Context, encKeyDB map[string
 	for _, url := range cliCtx.Args() {
 		// clean path for aliases like s3/.
 		// Note: UNC path using / works properly in go 1.9.2 even though it breaks the UNC specification.
+		urlOrig := url
 		url = filepath.ToSlash(filepath.Clean(url))
 		// namespace removal applies only for non FS. So filter out if passed url represents a directory
 		dir := isAliasURLDir(ctx, url, encKeyDB, time.Time{})
 		if !dir {
+			if string(urlOrig[len(urlOrig)-1]) == string(os.PathSeparator) {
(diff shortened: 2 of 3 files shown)


cmd/config.go
```diff
@@ -291,7 +291,9 @@ func expandAlias(aliasedURL string) (alias string, urlStr string, aliasCfg *alia
 	if envConfig, ok = os.LookupEnv(mcEnvHostPrefix + alias); !ok {
 		envConfig, ok = os.LookupEnv(mcEnvHostsDeprecatedPrefix + alias)
 		if ok {
-			errorIf(errInvalidArgument().Trace(mcEnvHostsDeprecatedPrefix+alias), "`MC_HOSTS_<alias>` environment variable is deprecated. Please use `MC_HOST_<alias>` instead for the same functionality.")
+			errorIf(errInvalidArgument().Trace(mcEnvHostsDeprecatedPrefix+alias),
+				"`MC_HOSTS_<alias>` environment variable is deprecated. Please"+
```
kannappanr, line 295:
```suggestion
				"`MC_HOSTS_<alias>` environment variable is deprecated. Please use "+
```

cmd/rm-main.go
```diff
@@ -202,10 +202,14 @@ func checkRmSyntax(ctx context.Context, cliCtx *cli.Context, encKeyDB map[string
 	for _, url := range cliCtx.Args() {
 		// clean path for aliases like s3/.
 		// Note: UNC path using / works properly in go 1.9.2 even though it breaks the UNC specification.
+		urlOrig := url
 		url = filepath.ToSlash(filepath.Clean(url))
 		// namespace removal applies only for non FS. So filter out if passed url represents a directory
 		dir := isAliasURLDir(ctx, url, encKeyDB, time.Time{})
 		if !dir {
+			if string(urlOrig[len(urlOrig)-1]) == string(os.PathSeparator) {
```
kannappanr, line 210: what does this check do?

ebozduman: When the url is provided with suffix `/` and with no object in it, `listAndRemove` function checks and decides that this location is empty and `fix/` is `not an object`, so errors out: 
```
mc: <ERROR> Failed to remove `mm1/test1/pre/fix/`. Object does not exist
```
To prevent this, we trim the `/` character if the location is a directory and the last character is a `/`.

vadmeste:
@ebozduman I am not sure what this PR fixes (master already removes an empty prefix if you don't pass a leading slash to it)

ebozduman:
@vadmeste ,

Here is the master current behavior with the mc built from the top of the master branch
```
(master) $ mc tree -f mm1
mm1
└─ test1
   └─ pre
      └─ fix
(master) $ mc rm -r --force mm1/test1/pre/fix/ 
mc: <ERROR> Failed to remove `mm1/test1/pre/fix/`. Object does not exist```
(master) $ mc tree -f mm1
mm1
└─ test1
   └─ pre
      └─ fix
```

This is the behavior after the fix:
```
(rm-empty-prefixes-3486) $ mc tree -f mm1
mm1
└─ test1
   └─ pre
      └─ fix
(rm-empty-prefixes-3486) $ mc rm -r --force mm1/test1/pre/fix/ 
Removing `mm1/test1/pre/fix/`.
(rm-empty-prefixes-3486) $ mc tree -f mm1
mm1
└─ test1
```

vadmeste:
> This is the behavior after the fix:
> 
> ```
> (rm-empty-prefixes-3486) $ mc tree -f mm1
> mm1
> └─ test1
>    └─ pre
>       └─ fix
> (rm-empty-prefixes-3486) $ mc rm -r --force mm1/test1/pre/fix/ 
> Removing `mm1/test1/pre/fix/`.
> (rm-empty-prefixes-3486) $ mc tree -f mm1
> mm1
> └─ test1
> ```

Well I think this is a breaking behavior. Let's talk with the team.

ebozduman:
@vadmeste 
Could you explain what is broken here.
The fix addressed the empty prefix removal issue. Are you saying empty prefixes are not supposed to be deleted?

vadmeste:
> @vadmeste
> Could you explain what is broken here?
> The fix addressed the empty prefix removal issue. Are you saying empty prefixes are not supposed to be deleted?

I think users already understand that `mc rm -r /path/to/dir` will remove `dir` but `mc rm -r /path/to/dir/` will only remove the content of `dir`,

ebozduman:
@vadmeste,

Here is the current behavior versus the fixed behavior.

```
# The existing behavior for an empty prefix, without a leading `/`
$ mc rm play/test/dir0/dir1/dir2
mc: <ERROR> Removal requires --recursive flag. This operation is *IRREVERSIBLE*. Please review carefully
before performing this *DANGEROUS* operation.

# The existing behavior is for an empty prefix with a leading `/`
$ mc rm play/test/dir0/dir1/dir2/
mc: <ERROR> Removal requires --recursive flag. This operation is *IRREVERSIBLE*. Please review carefully
before performing this *DANGEROUS* operation. 
```

So, with or without a leading `/` we recognize that prefix `dir2` is a directory/prefix and we recommend user to use `-r` in both cases. This is a consistent behavior.

Now let's try to recursively remove an empty prefix with and without a leading `/`;

```
# Remove recursively an empty prefix with a leading `/`. Fails with error.
$ mc rm -r --force play/test/dir0/dir1/dir2/
mc: <ERROR> Failed to remove `play/test/dir0/dir1/dir2/`. Object does not exist

# Remove recursively an empty prefix without a leading `/`. Successfully removes the empty prefix recursively.
$ mc rm -r --force play/test/dir0/dir1/dir2
Removing `play/test/dir0/dir1/dir2/`.
$ mc tree -f play/test
$
```

Existing `mc rm -r --force` command can only remove `dir2`, but not `dir2/`, i.e. an empty prefix can only be removed recursively if the prefix name is provided without a leading `/`.

So, this PR fixes the case when an empty prefix is provided with a `/` in a recursive remove command.

```
# This is the fixed behavior
$ mc mb play/test/dir0/dir1/dir2
Bucket created successfully `play/test/dir0/dir1/dir2`.
$ mc tree -f play/test
play/test
└─ dir0
   └─ dir1
      └─ dir2
$ mc rm -r --force play/test/dir0/dir1/dir2/
Removing `play/test/dir0/dir1/dir2/`.
$ mc tree -f play/test
$ 
```

ebozduman:
@vadmeste, 

I see your point and it makes sense.
I guess the message could have been better, like `"No objects found"` instead of `"Object does not exist"`, but not a big deal.

@kannappanr,

Do you also agree? 
We've thought this needs to be fixed and this PR changes exactly this behavior, but now I agree with @vadmeste.

I suggest to close this PR and the issue #3486 as `works as intended`.

PR #3563, opened 2021-01-06
cp: Copy recursive paths having the passed prefix 

vadmeste:
Some users used to copy a subset of files stored under one specific prefix
using `mc cp -r alias/bucket/[starting-characters] /dest`.

However this is not working anymore.

The reason is that before, let's say if we have a prefix containing
an object called '1.txt', client-s3.Stat() passed with a prefix '1'
will return the info of the object '1.txt'. However this behavior
was changed recently and client-s3.Stat() now only returns the
information of the exact object name '1'.

To solve this, we add firstURL2Stat() which returns the stat info of the
first object having the specified prefix.

Fixes https://github.com/minio/mc/issues/3562

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -1315,21 +1315,8 @@ func (c *S3Client) SetAccess(ctx context.Context, bucketPolicy string, isJSON bo
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


harshavardhana (approved):

vadmeste:
@poornas could please take a look ?

poornas (approved):
Tested,LGTM

PR #3565, opened 2021-01-07
Add missing lifecycle config parameters

vadmeste:
Add the following flags to mc ilm command:

  --expired-object-delete-marker
  --noncurrentversion-expiration-days
  --noncurrentversion-transition-days
  --noncurrentversion-transition-storage-class

### DIFF
--- a/cmd/ilm-add.go
+++ b/cmd/ilm-add.go
@@ -90,6 +90,22 @@ var ilmAddFlags = []cli.Flag{
 		Name:  "disable",
 		Usage: "disable the rule",
 	},
+	cli.BoolFlag{
+		Name:  "expired-object-delete-marker",
+		Usage: "remove delete markers with no parallel versions",
+	},
+	cli.IntFlag{
+		Name:  "noncurrentversion-expiration-days",
+		Usage: "the number of days to remove noncurrent versions",
+	},
+	cli.IntFlag{
+		Name:  "noncurrentversion-transition-days",
+		Usage: "the number of days to transition noncurrent versions",
+	},
+	cli.StringFlag{
+		Name:  "noncurrentversion-transition-storage-class",
+		Usage: "the transition storage class for noncurrent versions",
+	},
 }
 
 type ilmAddMessage struct {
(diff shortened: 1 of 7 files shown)


cmd/ilm-add.go
```diff
@@ -90,6 +90,26 @@ var ilmAddFlags = []cli.Flag{
 		Name:  "disable",
 		Usage: "disable the rule",
 	},
+	cli.IntFlag{
+		Name:  "abort-incomplete-uploads-days",
+		Usage: "the number of days to clean up incomplete multipart uploads",
+	},
```
harshavardhana, line 96: This is probably not useful @vadmeste - we can defer on adding this flag.

harshavardhana (changes requested):

cmd/ilm-add.go
vadmeste: > This is probably not useful @vadmeste - we can defer on adding this flag.

Why is that ? incomplete uploads causes costs for users, I think it is good to have.

harshavardhana: We automatically cleanup in MinIO - for AWS we don't care @vadmeste - users can set this usng import/export

harshavardhana: Basically, implement things in `mc` that are meaningful with the MinIO server

go.sum
```diff
@@ -368,6 +368,8 @@ github.com/minio/minio v0.0.0-20201231050455-36fc2f98ed12 h1:yTz1GCGRf70a+2HHNOG
 github.com/minio/minio v0.0.0-20201231050455-36fc2f98ed12/go.mod h1:Y2vPTyVMgM/hdmWY4z1FAh0tAiPVAMLz4eeQ5FNcJmk=
 github.com/minio/minio-go/v7 v7.0.7-0.20201217170524-3baf9ea06f7c h1:NgTbI1w/B+2Jcl+YKTULAAXqkwWqMZbkzmVdWNwzKnA=
 github.com/minio/minio-go/v7 v7.0.7-0.20201217170524-3baf9ea06f7c/go.mod h1:pEZBUa+L2m9oECoIA6IcSK8bv/qggtQVLovjeKK5jYc=
+github.com/minio/minio-go/v7 v7.0.7-0.20210107174959-40d41266f363 h1:dMvWRvc6iSuBjeMVOZ4NrebkoIIeG0RclfxpYCFhMso=
+github.com/minio/minio-go/v7 v7.0.7-0.20210107174959-40d41266f363/go.mod h1:pEZBUa+L2m9oECoIA6IcSK8bv/qggtQVLovjeKK5jYc=
 github.com/minio/selfupdate v0.3.1 h1:BWEFSNnrZVMUWXbXIgLDNDjbejkmpAmZvy/nCz1HlEs=
```
harshavardhana, line 373: You should run `go mod tidy`

vadmeste: > You should run `go mod tidy`

Surprisingly, this doesn't have any effect

cmd/ilm-add.go
vadmeste: > Basically, implement things in `mc` that are meaningful with the MinIO server

Done

harshavardhana (approved):

poornas (approved):
LGTM

PR #3580, opened 2021-01-21
session: Fix resuming cp/mv session

vadmeste:
Session files contained a colored json which leaded a problem with
unparsing it, use jsoniter instead of the colorjson when marshal
unmarshal session data.

To test:
`mc cp --continue /bin/cp/ play/bucket/`

### DIFF
--- a/cmd/cp-main.go
+++ b/cmd/cp-main.go
@@ -28,6 +28,7 @@ import (
 	"strings"
 
 	"github.com/fatih/color"
+	jsoniter "github.com/json-iterator/go"
 	"github.com/minio/cli"
 	json "github.com/minio/mc/pkg/colorjson"
 	"github.com/minio/mc/pkg/probe"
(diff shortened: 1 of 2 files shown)


cmd/cp-main.go
```diff
@@ -19,6 +19,7 @@ package cmd
 import (
 	"bufio"
 	"context"
+	stdjson "encoding/json"
```
harshavardhana, line 22: we can use `jsoniter` here @vadmeste

harshavardhana (changes requested):

cmd/cp-main.go
vadmeste: > we can use `jsoniter` here @vadmeste

good idea, done

kannappanr (approved):

harshavardhana (approved):

PR #3582, opened 2021-01-21
mv: Remove --continue flag as not really needed

vadmeste:
mv moves one object per a time, copy & remove the original copy.
Therefore, it doesn't make sense to have --continue flag since
re-executing the command will naturally continue the operation
(processed files in the source path are already deleted)

### DIFF
--- a/cmd/mv-main.go
+++ b/cmd/mv-main.go
@@ -280,9 +280,8 @@ func mainMove(cliCtx *cli.Context) error {
 	fatalIf(err, "Unable to parse encryption keys.")
 
 	// Parse metadata.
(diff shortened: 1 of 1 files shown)


harshavardhana (changes requested):
The idea of `--continue` was to avoid building sessions by default @vadmeste - sessions shall be only built if `--continue` is provided.

vadmeste:
> The idea of `--continue` was to avoid building sessions by default @vadmeste - sessions shall be only built if `--continue` is provided.

What I am saying is that --continue and session is not needed in mv command.

PR #3583, opened 2021-01-21
parallel: Allocate one channel for transferring big files

vadmeste:
Big files consume a lot of memory and they are also able to consume an
important bandwidth. Allocate one channel for them to be processed
separately from small files.

Working but the idea needs to be approved.

### DIFF
--- a/cmd/cp-main.go
+++ b/cmd/cp-main.go
@@ -533,11 +533,11 @@ func doCopySession(ctx context.Context, cancelCopy context.CancelFunc, cli *cli.
 				if isCopied != nil && isCopied(cpURLs.SourceContent.URL.String()) {
 					parallel.queueTask(func() URLs {
 						return doCopyFake(ctx, cpURLs, pg, isMvCmd)
-					})
+					}, false)
 				} else {
 					parallel.queueTask(func() URLs {
 						return doCopy(ctx, cpURLs, pg, encKeyDB, isMvCmd, preserve)
-					})
+					}, cpURLs.SourceContent.Size > 128*1024*1024)
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


cmd/cp-main.go
```diff
@@ -533,11 +533,11 @@ func doCopySession(ctx context.Context, cancelCopy context.CancelFunc, cli *cli.
 				if isCopied != nil && isCopied(cpURLs.SourceContent.URL.String()) {
 					parallel.queueTask(func() URLs {
 						return doCopyFake(ctx, cpURLs, pg, isMvCmd)
-					})
+					}, false)
 				} else {
 					parallel.queueTask(func() URLs {
 						return doCopy(ctx, cpURLs, pg, encKeyDB, isMvCmd, preserve)
-					})
+					}, cpURLs.SourceContent.Size > 128*1024*1024)
```
klauspost, line 540: Move size to a const.

cmd/parallel-manager.go
```diff
@@ -58,8 +58,8 @@ type ParallelManager struct {
 	// Current threads number
 	workersNum uint32
 
-	// Channel to receive tasks to run
-	queueCh chan task
+	// Channels to receive tasks to run
+	queueChannels []chan task
```
klauspost, line 62: This is always 2, correct?

Either just have 2 vars or `queueChannels [2]chan task`. I prefer the former, makes everything clearer and you don't have to pass around 0 and 1, but maybe a bool.

klauspost (commented):
Drive-by review :)

vadmeste:
This can be closed, there is another PR which monitors memory consumption. Besides part size is changed to 16Mb

PR #3590, opened 2021-02-01
mirror: Replicate versions when enabled

vadmeste:
Spec not approved yet.

Still in testing phase

### DIFF
--- a/cmd/client-errors.go
+++ b/cmd/client-errors.go
@@ -214,3 +214,27 @@ type SameFile struct {
 func (e SameFile) Error() string {
 	return fmt.Sprintf("'%s' and '%s' are the same file", e.Source, e.Destination)
 }
+
+// RetentionAlreadyExists - is thrown when we don't want to overwrite existing retention config
+type RetentionAlreadyExists struct {
+}
+
+func (e RetentionAlreadyExists) Error() string {
+	return "Retention already exists in the target location."
+}
+
+// LegalholdAlreadyExists is thrown when we don't want to overwrite existing legalhold config
+type LegalholdAlreadyExists struct {
+}
+
+func (e LegalholdAlreadyExists) Error() string {
+	return "Legalhold already exists in the target location."
+}
+
+// TagsAlreadyExists is thrown when we don't want to overwrite existing tags
+type TagsAlreadyExists struct {
+}
+
+func (e TagsAlreadyExists) Error() string {
+	return "Tags already exists in the target location."
+}
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -82,6 +82,22 @@ type CopyOptions struct {
 	isPreserve       bool
 }
 
+// PutOptions holds options for object putting operation
+type PutOptions struct {
+	metadata                          map[string]string
+	sse                               encrypt.ServerSide
+	md5, disableMultipart, isPreserve bool
+	versionID, etag                   string
+	modTime                           time.Time
+	isDeleteMarker                    bool
+}
+
+// RemoveOptions holds options for removing object operation
+type RemoveOptions struct {
+	isIncomplete, isRemoveBucket, isBypass bool
+	replicateDeleteMarker                  bool
+}
+
 // Client - client interface
 type Client interface {
 	// Common operations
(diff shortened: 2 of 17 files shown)


cmd/client-errors.go
```diff
@@ -214,3 +214,27 @@ type SameFile struct {
 func (e SameFile) Error() string {
 	return fmt.Sprintf("'%s' and '%s' are the same file", e.Source, e.Destination)
 }
+
+// RetentionAlreadyExists - is thrown when we don't want to overwrite existing retention config
+type RetentionAlreadyExists struct {
```
klauspost, line 219:
```suggestion
type ErrRetentionAlreadyExists struct {
```

cmd/client-errors.go
```diff
@@ -214,3 +214,27 @@ type SameFile struct {
...
+}
+
+func (e RetentionAlreadyExists) Error() string {
+	return "Retention already exists in the target location."
+}
+
+// LegalholdAlreadyExists is thrown when we don't want to overwrite existing legalhold config
+type LegalholdAlreadyExists struct {
```
klauspost, line 227:
```suggestion
type ErrLegalholdAlreadyExists struct {
```

cmd/client-errors.go
```diff
@@ -214,3 +214,27 @@ type SameFile struct {
...
+}
+
+func (e LegalholdAlreadyExists) Error() string {
+	return "Legalhold already exists in the target location."
+}
+
+// TagsAlreadyExists is thrown when we don't want to overwrite existing tags
+type TagsAlreadyExists struct {
```
klauspost, line 235:
```suggestion
type ErrTagsAlreadyExists struct {
```

cmd/client.go
```diff
@@ -82,6 +82,22 @@ type CopyOptions struct {
 	isPreserve       bool
 }
 
+// PutOptions holds options for object putting operation
+type PutOptions struct {
```
klauspost, line 86: Exporting this but none of the fields does not really make it usable.

cmd/client.go
```diff
@@ -82,6 +82,22 @@ type CopyOptions struct {
...
+	metadata                          map[string]string
+	sse                               encrypt.ServerSide
+	md5, disableMultipart, isPreserve bool
+	versionID, etag                   string
+	modTime                           time.Time
+	isDeleteMarker                    bool
+}
+
+// RemoveOptions holds options for removing object operation
+type RemoveOptions struct {
```
klauspost, line 96: Same...

cmd/common-methods.go
```diff
@@ -276,50 +277,82 @@ func getSourceStream(ctx context.Context, alias, urlStr, versionID string, fetch
...
-	if ok {
-		lockMode = minio.RetentionMode(lockModeStr)
-		delete(metadata, AmzObjectLockMode)
+	tagsMaps, err := clnt.GetTags(ctx, versionID)
+	if err != nil {
+		return "", err.Trace(alias, urlStr)
 	}
+	var tags string
+	for k, v := range tagsMaps {
+		tags += fmt.Sprintf("%s=%s&", k, v)
```
klauspost, line 292: No escaping needed? What if either contains an `&` or `=`?

cmd/difference.go
```diff
@@ -160,18 +160,49 @@ func metadataEqual(m1, m2 map[string]string) bool {
...
+	return difference(ctx, sourceClnt, targetClnt, sourceURL, targetURL, false, false, false, true, DirFirst)
 }
 
-func differenceInternal(ctx context.Context, sourceClnt, targetClnt Client, sourceURL, targetURL string, isMetadata bool, isRecursive, returnSimilar bool, dirOpt DirOpt, diffCh chan<- diffMessage) *probe.Error {
+type diffEntry struct {
+	path    string
+	modTime time.Time
+}
+
+func cmpDiffEntry(e1, e2 diffEntry) int {
```
klauspost, line 176: Doc

cmd/client-errors.go
vadmeste: @klauspost this doesn't follow the existing style, maybe we can do that in another PR

cmd/client.go
vadmeste: > Exporting this but none of the fields does not really make it usable.

Still this follows the existing style. I can send other PRs once this merged (this PR is not that easy to review in the first place)

cmd/client-errors.go
klauspost: It follows all other styles. No need to introduce exported types that will be renamed.

cmd/common-methods.go
vadmeste: > No escaping needed? What if either contains an & or =?

Done

cmd/difference.go
vadmeste: > Doc

Done

Issue #3591, opened 2021-02-02
regression: mc panics on recursive copy of non-existant file


matthewdarwin:
## Expected behavior

Follow up to: https://github.com/minio/mc/issues/3562

I installed the latest minio (RELEASE.2021-01-30T00-50-42Z) and when running a recursive file copy I expect it to work properly when the file does not exist (generate a nice error message).

## Actual behavior

Get a panic

```
mc cp --recursive play/this-file-does-not-exist /tmp
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x70 pc=0xd0ae4b]

goroutine 1 [running]:
github.com/minio/mc/cmd.guessCopyURLType(0x1109fe0, 0xc0003da340, 0xc0000300d0, 0x1, 0x2, 0x7fffb7acb872, 0x4, 0x1, 0xc0006d42d0, 0x0, ...)
        github.com/minio/mc/cmd/cp-url.go:71 +0xcb
github.com/minio/mc/cmd.checkCopySyntax(0x1109fe0, 0xc0003da340, 0xc0002f1040, 0xc0006d42d0, 0x0)
        github.com/minio/mc/cmd/cp-url-syntax.go:94 +0x465
github.com/minio/mc/cmd.mainCopy(0xc0002f1040, 0x0, 0x0)
        github.com/minio/mc/cmd/cp-main.go:754 +0x165
github.com/minio/cli.HandleAction(0xe4e920, 0x10106a0, 0xc0002f1040, 0xc0002e9600, 0x0)
        github.com/minio/cli@v1.22.0/app.go:498 +0x82
github.com/minio/cli.Command.Run(0xfa5cca, 0x2, 0x0, 0x0, 0x0, 0x0, 0x0, 0xfb03e7, 0xc, 0x0, ...)
        github.com/minio/cli@v1.22.0/command.go:225 +0x9fb
github.com/minio/cli.(*App).Run(0xc0002f0000, 0xc0000300a0, 0x5, 0x5, 0x0, 0x0)
        github.com/minio/cli@v1.22.0/app.go:261 +0x768
github.com/minio/mc/cmd.Main(0xc0000300a0, 0x5, 0x5)
        github.com/minio/mc/cmd/main.go:123 +0x33a
main.main()
        github.com/minio/mc/main.go:34 +0x45
```

## Steps to reproduce the behavior

```mc cp --recursive play/this-file-does-not-exist /tmp```

[...]

PR #3592, opened 2021-02-02
cp: Fix crash with recursive copy with non existant resource

vadmeste:
Recent change in the code introduced a crash when copying a non
existant bucket or prefix when --recursive flag is passed.

This PR fixes it.

Fixes https://github.com/minio/mc/issues/3591


klauspost (approved):
LGTM

kannappanr (approved):

PR #3596, opened 2021-02-04
ilm: Fix adding ExpiredObjectDeleteMarker alone

vadmeste:
It is not possible to add rule with only ExpiredObjectDeleteMarker
activated. This commit fixes the issue.

How to test:
`mc ilm add --expired-object-delete-marker myminio/testbucket/`


harshavardhana (approved):

ebozduman (approved):
LGTM

Issue #3374, opened 2020-08-22
mc: <ERROR> Unable to validate source `test.*`.


fdefilippo:
## Expected behavior
c:\tmp>dir test*exe
...
Directory of C:\tmp

21/08/2020  14:58        776 test1.bin
21/08/2020  14:58        776 test2.bin

mc cp test* play\test\

copy all matching file

## Actual behavior
mc: <ERROR> Unable to validate source `test*`.

## Steps to reproduce the behavior
mc cp test* play\test\

## mc --version
mc version RELEASE.2020-08-20T00-23-01Z

## System information
windows 10, windows 2012, windows 2016

fdefilippo:
nobody else ?

fdefilippo:
the same executable used in the cygwin environment has no problems::

```
$ ./mc.exe --json cp uno* play/test/
{
 "status": "success",
 "source": "uno.txt",
 "target": "play/test/uno.txt",
 "size": 0,
 "totalCount": 1,
 "totalSize": 0
}
{
 "status": "success",
 "source": "uno1.txt",
 "target": "play/test/uno1.txt",
 "size": 0,
 "totalCount": 2,
 "totalSize": 0
}
{
 "status": "success",
 "source": "uno2.txt",
 "target": "play/test/uno2.txt",
 "size": 0,
 "totalCount": 3,
 "totalSize": 0
}
{
 "status": "success",
 "total": 0,
 "transferred": 0,
 "speed": 0
}

```

ebozduman:
I see this also works in Linux, but not on Windows.
So, it is legit.

PR #3604, opened 2021-02-11
Supports wildcard for 'mc ls' and 'mc cp' commands

ebozduman:
Fixes issue #3374

Support for wildcard `*` support for `mc ls` and `mc cp` commands.

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -1,5 +1,5 @@
 /*
- * MinIO Client (C) 2015-2020 MinIO, Inc.
+ * MinIO Client (C) 2015-2021 MinIO, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this fs except in compliance with the License.
@@ -595,7 +595,7 @@ func (f *fsClient) listPrefixes(prefix string, contentCh chan<- *ClientContent) 
 			continue
 		}
 
-		file := filepath.ToSlash(filepath.Join(dirName, fi.Name()))
+		file := filepath.Join(dirName, fi.Name())
 		if fi.Mode()&os.ModeSymlink == os.ModeSymlink {
 			st, e := os.Stat(file)
 			if e != nil {
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -595,7 +595,7 @@ func (f *fsClient) listPrefixes(prefix string, contentCh chan<- *ClientContent) 
 			continue
 		}
 
-		file := filepath.Join(dirName, fi.Name())
+		file := filepath.ToSlash(filepath.Join(dirName, fi.Name()))
 		if fi.Mode()&os.ModeSymlink == os.ModeSymlink {
 			st, e := os.Stat(file)
 			if e != nil {
--- a/cmd/client-fs_test.go
+++ b/cmd/client-fs_test.go
@@ -1,5 +1,5 @@
 /*
- * MinIO Client (C) 2015 MinIO, Inc.
+ * MinIO Client (C) 2021 MinIO, Inc.
  *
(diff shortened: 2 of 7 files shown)


ebozduman:
Looking into the build test failure...

harshavardhana:
How would you handle bash escapes?

harshavardhana:
> How would you handle bash escapes?

These are real issues here @ebozduman while the intention is nice `*` most users do not know how to bash escape. So it is rather not useful for us to implement and support - as it would create more confusion. 

All other changes and cleanup seem nice.

ebozduman:
@harsha,
> How would you handle bash escapes?

I did not do anything specific for bash escapes, and it'll work as before.

If you mean a situation like, when there is a filename with a `*` in it and user wants us to literally take it by escaping it, like `xxx\*xxx`, then it is fine, since we do take it literally with or without in object storage. 
On the filesystem, `mc ls` and `ls` behaves exactly the same with the fix. Any other cases where we don't act as expected, like literal `*` character appears at the end of a filename, then we can tell customer that this is the only case we don't support, but I think that is pretty rare scenario IMHO.


What bash escape scenario in which supporting wildcard character `*` would cause confusion?
I think, supporting wildcard character in most common scenarios, especially when copying a group of files to object storage, would be very handy.

ravindk89:
@harshavardhana "So it is rather not useful for us to implement and support - as it would create more confusion." <- we can resolve this somewhat by documenting what the behavior is for properly escaping `*` in a filename vs `*` as a wildcard.

@ebozduman just to confirm, the semantics would be something like:

```sh

# Copy all files with .exe

mc cp ~/data/*.exe play/rkbucket/data/

# Copy all files with pattern "*xx.exe"

mc cp ~/data/\**.exe play/rkbucket/data/

# This will have unexpected behavior because file ends with "*"

mc cp ~/data/foo\* play/rkbucket/data/

```

ebozduman:
@ravindk89,
```
> // Copy all files with .exe
> mc cp ~/data/\*.exe play/rkbucket/data/
```

Correct and it works.
```
> // Copy all files with pattern "*xx.exe"
> mc cp ~/data/\**.exe play/rkbucket/data/
```

Correct and it works.
``` 
> // This will have unexpected behavior because file ends with "*"
> mc cp ~/data/foo\* play/rkbucket/data/
```

This example also works, because you've escaped the wildcard correctly in your command.
The question/concern was how we handle escape characters and the problems we may face if a user doesn't know how to escape wildcard characters.

The only problem I see is when a file/object name is defined with a literal `"*"` character at its end. 

$ mc ls foo*  // there are 2 more files in the same location: foo1 and foo2
[2021-03-29 16:26:45 PDT]     0B foo*
[2021-03-29 16:27:51 PDT]     0B foo1
[2021-03-29 16:27:51 PDT]     0B foo2
~[2021-03-29 16:27:51 PDT]     0B foo1~
~[2021-03-29 16:27:51 PDT]     0B foo2~

// The last 2 items in the above listing, duplicate info, is an issue, which we may
// address as part of this effort or as a separate issue.

chrisbecke:
Just would like to comment here that basic wildcard support would be nice, on ALL the mc endpoints that take --recursive, including `mc rm`.

Another thing to consider is wildcard path globbing, where `**` is used by lots of tooling to represent a 0 to many folder depth search.
This assumes that `*/test.txt` would match `foo/test.txt` but not `test.txt` or `/foo/bar/test.txt`. whereas `**/test.txt` would match all three.

ebozduman:
I'll bring this PR to the attention of the team members who are/were involved and/or have some saying about it, to decide its faith
.

ebozduman:
Closing this PR as it would create more confusion.

PR #3606, opened 2021-02-11
rm: Better messages when buckets are empty

vadmeste:
- Do not show removing bucket name in rm command since rm is not
supposed to remove buckets in the first place
- Better error message when bucket is empty (no object or version is
removed)
- Fix the example

### DIFF
--- a/cmd/rm-main.go
+++ b/cmd/rm-main.go
@@ -130,7 +130,7 @@ EXAMPLES:
   07. Remove all objects recursively from Amazon S3 cloud storage.
       {{.Prompt}} {{.HelpName}} --recursive --force --dangerous s3
 
-  08. Remove all buckets and objects older than '90' days recursively from host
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

cmd/rm-main.go
```diff
@@ -368,6 +368,11 @@ func listAndRemove(url string, timeRef time.Time, withVersions, isRecursive, isI
 
 		urlString := content.URL.Path
 
+		// rm command is not supposed to remove buckets, ignore if this is a bucket name
```
kannappanr, line 371: can you give an example where this would happen in the current code?

vadmeste: > can you give an example where this would happen in the current code?

mc mb myminio/testbucket1/
mc mb myminio/testbucket2/
mc rm --force --recursive --dangerous myminio/

kannappanr (approved):

PR #3607, opened 2021-02-12
Ignore parallel net perf data when its empty

anjalshireesh:
When parallel net perf tests are not performed, for example in FS mode,
the server response contains an empty node for the net pef data with
value of `addr` as empty string (`""`).

This data should be ignored when mapping the server response to the
health report format. Otherwise we end up with a health report with an
empty server node under `hardware` section.


kannappanr (approved):
LGTM

harshavardhana (approved):

PR #3611, opened 2021-02-12
mirror: Fix behavior under Windows when source url is S3

vadmeste:
Removed one block of code which seems to be useless and no comment with
it. This was causing issue doing mirroring under Windows.

### DIFF
--- a/cmd/mirror-url.go
+++ b/cmd/mirror-url.go
@@ -83,10 +83,6 @@ func checkMirrorSyntax(ctx context.Context, cliCtx *cli.Context, encKeyDB map[st
 				srcURL = origSrcURL
 			}
 		}
-
-		if !strings.HasSuffix(srcURL, string(srcContent.URL.Separator)) {
-			srcURL += string(srcContent.URL.Separator)
-		}
 	}
 
 	return


harshavardhana (changes requested):
It was added here

```
4e8e67fb0 cmd/mirror-url.go     (Anis Elleuch              2020-12-14 23:10:09 +0100  86) 
4e8e67fb0 cmd/mirror-url.go     (Anis Elleuch              2020-12-14 23:10:09 +0100  87)               if !strings.HasSuffix(srcURL, string(srcContent.URL.Separator)) {
4e8e67fb0 cmd/mirror-url.go     (Anis Elleuch              2020-12-14 23:10:09 +0100  88)                       srcURL += string(srcContent.URL.Separator)
4e8e67fb0 cmd/mirror-url.go     (Anis Elleuch              2020-12-14 23:10:09 +0100  89)               }
```

vadmeste:
@harshavardhana yes correct, and I don't know why I added it. I tested mc mirror while that block is removed in all scenarios that I know and it works fine.

harshavardhana (approved):

harshavardhana:
> @harshavardhana yes correct, and I don't know why I added it. I tested mc mirror while that block is removed in all scenarios that I know and it works fine.

from what it looks like you wanted to add `/` at the end for site-wide mirroring.

PR #3616, opened 2021-02-19
Fixes conflicting Failure&Successful msgs in 'mc rb' output

### LINKED ISSUE #3387: mc rm does not delete bucket if empty, but deletes bucket if it has files
After some of the recent "bug fix" releases, some of the behavior we were expecting changed. In short, if we create a bucket with `mc`, and try to delete it, minio says the bucket does not exist. `mc ls` shows the bucket present. if we add anything to that bucket, and run the same `mc rm`, then the object is somehow found, and deletes appropriately. This was strange behavior, and unexpected. (We have changed our code to use `mc rb` instead of `mc rm`, but we felt it appropriate to file this issue anyway)

## Expected Behavior
I expect `mc rm` to remove my bucket, even if it is empty OR I expect `mc rm` to never remove my bucket at all

## Current Behavior
`mc rm` will only remove the bucket if there are files inside, and errors "it cannot find the object" otherwise

## Possible Solution
Unsure. Prior to Aug 17th (whatever version of minio was released then) we did not encounter this issue.

## Steps to Reproduce (for bugs)
```
○ →  mc mb my-bucket
Bucket created successfully `my-bucket`.

○ → mc rm my-bucket --recursive --force
mc: <ERROR> Failed to remove `my-bucket`. Object does not exist

○ → mc ls
[2020-06-24 13:44:08 MDT]    319B main.go
[2020-08-28 11:01:07 MDT]     64B my-bucket/

○ → touch my-bucket/file.txt

○ → mc rm my-bucket --recursive --force
Removing `my-bucket/file.txt`.
```

## Context
We were testing error cases in our codebase using minio.
We wanted to show the user a helpful error message if the bucket was empty (meaning that it did not have any matching products in it for us to use). In the test, we run
`mc mb my-bucket`
[run-test]
`mc rm my-bucket --force --recursive` 
We remove the bucket after every test to ensure a clean test environment for the next test, so contamination does not occur.

## Regression
 Yes. This was not an issue before Aug 17
 Not sure which commit caused the regression

## Your Environment

* Version used (`minio --version`): minio version RELEASE.2020-08-27T05-16-20Z
* Server setup and configuration: N/A
* Operating System and version (`uname -a`): OSX 10.15.6 AND ubuntu bionic

ebozduman:
Fixes #3387

The fix only addresses the conflicting `Removed '<object-path>' successfully.` and `Failed to remove '<object-path>', continuing ...` messages for object storage, when bucket is not empty in the command:
`mc rb <alias>/<bucket>/[<prefix>/][<object>] --force` output 

The team decided not to address the file system side of the issue, as there is no concept of a `"bucket"` for file systems.

### DIFF
--- a/cmd/rb-main.go
+++ b/cmd/rb-main.go
@@ -18,7 +18,6 @@ package cmd
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"path/filepath"
 	"strings"
@@ -168,13 +167,9 @@ func deleteBucket(ctx context.Context, url string) *probe.Error {
 		}
 	}()
 
-	foundErr := false
+	// Give up on the first error.
 	for perr := range errorCh {
-		errorIf(perr.Trace(url), "Failed to remove `"+url+"`, continuing..")
-		foundErr = true
-	}
-	if foundErr {
-		return probe.NewError(errors.New("Bucket not removed"))
+		return perr
 	}
 	return nil
 }
--- a/cmd/rb-main.go
+++ b/cmd/rb-main.go
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"path/filepath"
 	"strings"
@@ -167,8 +168,13 @@ func deleteBucket(ctx context.Context, url string) *probe.Error {
 		}
 	}()
 
+	foundErr := false
 	for perr := range errorCh {
 		errorIf(perr.Trace(url), "Failed to remove `"+url+"`, continuing..")
+		foundErr = true
+	}
+	if foundErr {
+		return probe.NewError(errors.New("Bucket not removed"))
 	}
 	return nil
 }
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -1070,7 +1070,9 @@ func (c *S3Client) Remove(ctx context.Context, isIncomplete, isRemoveBucket, isB
 		defer close(errorCh)
 		if isRemoveBucket {
 			if _, object := c.url2BucketAndObject(); object != "" {
-				errorCh <- probe.NewError(errors.New("cannot delete prefixes with `mc rb` command - Use `mc rm` instead"))
+				errorCh <- probe.NewError(errors.New(
+					"use `mc rm` command to delete prefixes, or point your" +
+						"bucket directly, `mc rb <alias>/<bucket-name>/`"))
 				return
 			}
 		}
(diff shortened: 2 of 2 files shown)


cmd/rb-main.go
```diff
@@ -167,8 +168,13 @@ func deleteBucket(ctx context.Context, url string) *probe.Error {
 		}
 	}()
 
+	foundErr := false
 	for perr := range errorCh {
 		errorIf(perr.Trace(url), "Failed to remove `"+url+"`, continuing..")
+		foundErr = true
+	}
+	if foundErr {
+		return probe.NewError(errors.New("Bucket not removed"))
```
vadmeste, line 177: Can we return the error returned by errorCh instead of a generic error ?

vadmeste (changes requested):
One comment

cmd/rb-main.go
ebozduman: Hi @vadmeste ,

The errors are displayed in the above for loop as they are hit one by one. So, there could be a bunch of error messages on display. 

When we are out of the loop, we may choose to say nothing, show a generic message, or maybe use one of them or the last one but that would repeat an already displayed error message. 
I don' think it would look good to show the last or any one of the displayed error message one more time.

So, I chose to show a generic message, however, we can choose to stop without showing any error message as an alternative.
Please let me know if you think we should go with no error message.

vadmeste: @ebozduman, I think we need to quit when the first error is encountered. If we fail to remove something in a bucket, we know in advance we won't be able to remove the bucket, I am not sure why we have the continuing process here.. are we retrying something ? if we do I think we don't have to, minio-go will retry for us.

Basically, I think the best is to have this:
`mc: <Error> Failed to remove mm1/locked: unable to remove xxxx: object is WORM protected and cannot be..`

ebozduman: @vadmeste ,

We are not retrying anything. I think this is by design.
I believe the original requirement was to have a single complete pass and show the user all of the errors in one shot, instead of running the same command several times and collecting information about each and every error one by one in single runs.

We can definitely stop at the first error, but that would change the api's behavior. The current implementation cleans up all the prefixes and objects if there is no problem removing them. Stopping at the first error would stop the full cleanup of the non-problematic objects/files.
@kannappanr?

ebozduman: @vadmeste,

Actually, the error messages are returned by `deleteBucket` function, so I replaced the code and simplified it such that it returns on the first error as you've requested.

I also changed 2 error messages to better explain the situation. 

When user tries to forcefully remove a bucket, `mc rb --force <alias>/<bucket>`, with one or more `WORM protected` objects, the old error message was not giving the name of the offending object, hence it was too generic. 

Old error message example:
```
mc: <ERROR> Failed to remove `myminio/locked/`. Object is WORM protected
and cannot be overwritten.
```

Now, the name and the version id of the object are part of the error message.
New error message example:
```
mc: <ERROR> Failed to remove `myminio/locked/`. Object, 'prefix/obj1
(Version ID=b1d0b084-9b70-46fe-ae63-a2efdcd3866c)' is WORM protected
and cannot be overwritten.
```

vadmeste (approved):
LGTM

kannappanr (approved):
LGTM

PR #3618, opened 2021-02-20
admin: Implement a better restart command UX

vadmeste:
mc admin restart service clearly indicates to the user that restarting
process is started and also shows progress using:
 `.` in red to indicate the cluster is still offline
 `!` in yellow to indicate the cluster is up but still initializing

e.g.
```
$ mc admin service restart docker                                                                                                                                                 
Restart command successfully sent to `docker`. Type Ctrl-C or wait to see the status of the restart.                                                                                                   
........!!!!
Restarted `docker` successfully.

```

### DIFF
--- a/cmd/admin-service-restart.go
+++ b/cmd/admin-service-restart.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"context"
 	"time"
 
 	"github.com/fatih/color"
@@ -24,6 +25,7 @@ import (
 	json "github.com/minio/mc/pkg/colorjson"
 	"github.com/minio/mc/pkg/probe"
 	"github.com/minio/minio/pkg/console"
+	"github.com/minio/minio/pkg/madmin"
 )
 
 var adminServiceRestartCmd = cli.Command{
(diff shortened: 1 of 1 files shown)


cmd/admin-service-restart.go
```diff
@@ -119,20 +123,36 @@ func mainAdminServiceRestart(ctx *cli.Context) error {
...
 
-	return nil
+	for {
+		select {
+		case <-globalContext.Done():
+			return globalContext.Err()
+		case <-time.NewTimer(3 * time.Second).C:
+			ctx, cancel := context.WithTimeout(globalContext, 1*time.Second)
+			// Fetch the service status of the specified MinIO server
+			info, e := client.ServerInfo(ctx)
```
harshavardhana, line 143:
```suggestion
			info, e := client.ServerInfo(ctx)
```

I always wondered if we can avoid this heavy call instead make it a simpler HTTP request to know if the server is up? like reaching out to the `/cluster` endpoint?  

Or probably introduce another separate API that does a much simpler `objLayer` up or not check?

harshavardhana (changes requested):

cmd/admin-service-restart.go
vadmeste: > I always wondered if we can avoid this heavy call instead make it a simpler HTTP request to know if the server is up? like reaching out to the `/cluster` endpoint?
> 
> Or probably introduce another separate API that does a much simpler `objLayer` up or not check?

I would say, when can make ServerInfo light when the cluster is not initialized yet. So in total, there is only one heavy call when the cluster is initialized, and that would be the last call in this command before prompting 'successful restart.' and quitting.

harshavardhana (approved):

PR #3628, opened 2021-03-01
Update minio dependency to latest master branch

anjalshireesh:
`minio` was recently modified to use the `v3` api of `gopsutil`
Since `mc` is also dependent on the related structures, the
corresponding imports in `health_v1.go` need to be updated
to ensure succesful compilation with the `minio` master branch.


harshavardhana (approved):

vadmeste (approved):
LGTM

PR #3629, opened 2021-03-01
parallel-manager: Reduce parallelism when there is not enough memory

vadmeste:
Some users with low memory and good connection uploading some big files
can see OOM crash, since mc will increase upload parallelism over
time when the bandwidth and the server is able to carry the load.

This PR will check the free OS memory before starting mirror/cp and will
, it will stop parallelism until the memory gets some room.

### DIFF
--- a/cmd/parallel-manager.go
+++ b/cmd/parallel-manager.go
@@ -23,6 +23,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/minio/minio-go/v7"
 	mem "github.com/shirou/gopsutil/v3/mem"
 )
 
@@ -171,7 +172,7 @@ func (p *ParallelManager) queueTaskWithBarrier(fn func() URLs, uploadSize int64)
 	p.doQueueTask(task{fn: fn, barrier: true, uploadSize: uploadSize})
 }
 
-func (p *ParallelManager) enoughMemForUpload(uploadSize uint64) bool {
+func (p *ParallelManager) enoughMemForUpload(uploadSize int64) bool {
 	if uploadSize < 0 {
 		panic("unexpected size")
 	}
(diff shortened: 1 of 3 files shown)


harshavardhana (approved):

harshavardhana:
Please fix the conflicts @vadmeste

vadmeste:
> Please fix the conflicts @vadmeste

Done

cmd/parallel-manager.go
```diff
@@ -153,18 +160,48 @@ func (p *ParallelManager) monitorProgress() {
...
+
+	estimateNeededMemoryForUpload := func(size uint64) uint64 {
+		// TODO: default part size should be exported in minio-go
+		const partSize = 16 * 1024 * 1024
+		parallel := size / partSize
+		if parallel >= 3 {
+			return 3 * partSize
+		}
+		return uint64(size)
+	}
```
harshavardhana, line 191: You can use `minio.OptimalPartInfo(objectSize, partSize)`

harshavardhana (changes requested):

harshavardhana (approved):

cmd/parallel-manager.go
```diff
@@ -153,18 +161,49 @@ func (p *ParallelManager) monitorProgress() {
...
+			panic(e)
+		}
+		if partsCount >= 4 {
+			return 4 * uint64(partSize)
+		}
+		return uint64(size)
+	}
+
+	smem := runtime.MemStats{}
+	runtime.ReadMemStats(&smem)
```
klauspost, line 196:
```suggestion
	runtime.GC()
	runtime.ReadMemStats(&smem)
```

cmd/parallel-manager.go
```diff
@@ -182,12 +221,20 @@ func (p *ParallelManager) stopAndWait() {
 
 // newParallelManager starts new workers waiting for executing tasks
 func newParallelManager(resultCh chan URLs) *ParallelManager {
+
+	var maxMem float64
+	memStats, err := mem.VirtualMemory()
+	if err == nil {
+		maxMem = float64(memStats.Available) * 0.8 // use upto 80% of available memory.
```
klauspost, line 228: This will account for the default `GOGC:100`.
```suggestion
		maxMem = float64(memStats.Available) / 2 // use upto 50% of available memory.
```

harshavardhana (approved):

klauspost (approved):
Sure, let's try it.

mrsrvman:
This PR does not solve the problem when running in docker with memory limits set

mrsrvman:
Look https://github.com/minio/mc/issues/3781

PR #3641, opened 2021-03-09
Add option to upload health report to Subnet

anjalshireesh:
Add an optional text flag `license` to the `mc admin subnet health`
command. If this flag is passed, upload the generated health report to
Subnet passing the license key as auth.

For use in development mode (where the report needs to be uploaded to
Subnet running on localhost), introduce another optional (and hidden)
boolean flag called `dev`. If it is passed, use `http://localhost:9000`
as the base url for Subnet instead of `https://subnet.min.io`

This feature can subsequently be used to automate the health report
upload by either using a cron job or introducing another `frequency`
flag to the command.

### DIFF
--- a/cmd/admin-subnet-health.go
+++ b/cmd/admin-subnet-health.go
@@ -17,11 +17,16 @@
 package cmd
 
 import (
+	"bytes"
 	"context"
 	gojson "encoding/json"
 	"errors"
 	"flag"
 	"fmt"
+	"io"
+	"io/ioutil"
+	"mime/multipart"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
@@ -49,6 +54,15 @@ var adminHealthFlags = []cli.Flag{
 		Value:  3600 * time.Second,
 		EnvVar: "MC_HEALTH_DEADLINE,MC_OBD_DEADLINE",
 	},
+	cli.StringFlag{
+		Name:  "license",
+		Usage: "Subnet license key",
+	},
+	cli.BoolFlag{
+		Name:   "dev",
+		Usage:  "Development mode",
+		Hidden: true,
+	},
 }
 
 var adminSubnetHealthCmd = cli.Command{
@@ -73,7 +87,7 @@ EXAMPLES:
 `,
 }
 
-// checkAdminInfoSyntax - validate arguments passed by a user
+// checkAdminHealthSyntax - validate arguments passed by a user
 func checkAdminHealthSyntax(ctx *cli.Context) {
 	if len(ctx.Args()) == 0 || len(ctx.Args()) > 1 {
 		cli.ShowCommandHelpAndExit(ctx, "health", 1) // last argument is exit code
@@ -81,8 +95,7 @@ func checkAdminHealthSyntax(ctx *cli.Context) {
 }
 
 //compress and tar health report output
-func tarGZ(c HealthReportInfo, alias string) error {
-	filename := fmt.Sprintf("%s-health_%s.json.gz", filepath.Clean(alias), time.Now().Format("20060102150405"))
+func tarGZ(c HealthReportInfo, filename string) error {
 	f, err := os.OpenFile(filename, os.O_CREATE|os.O_RDWR, 0666)
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


cmd/admin-subnet-health.go
```diff
@@ -161,11 +174,85 @@ func mainAdminHealth(ctx *cli.Context) error {
...
+	license := ctx.String("license")
+	if len(license) > 0 {
+		error = uploadHealthReport(filename, license, ctx.Bool("dev"))
+		if error != nil {
+			console.Println(warnText("Unable to upload health report to Subnet: "), error.Error())
+		}
+	}
+	return nil
+}
+
```
harshavardhana, line 197:
```suggestion
	e := tarGZ(clusterHealthInfo, filename)
	fatalIf(probe.NewError(e), "Unable to create health report file")

	license := ctx.String("license")
	if len(license) > 0 {
		e = uploadHealthReport(filename, license, ctx.Bool("dev"))
		fatalIf(probe.NewError(e), "Unable to upload health report to Subnet portal")
	}
	return nil
}

```

cmd/admin-subnet-health.go
```diff
@@ -161,11 +174,85 @@ func mainAdminHealth(ctx *cli.Context) error {
...
+func uploadHealthReport(filename string, license string, dev bool) error {
+	url := subnetUploadURL(filename, license, dev)
+	req, err := subnetUploadReq(url, filename)
```
harshavardhana, line 200:
```suggestion
	req, e := subnetUploadReq(url, filename)
```

cmd/admin-subnet-health.go
```diff
@@ -161,11 +174,85 @@ func mainAdminHealth(ctx *cli.Context) error {
...
+	if err != nil {
+		return err
```
harshavardhana, line 202:
```suggestion
		return e
```

cmd/admin-subnet-health.go
```diff
@@ -161,11 +174,85 @@ func mainAdminHealth(ctx *cli.Context) error {
...
+
+	client := &http.Client{}
+	resp, herr := client.Do(req)
+	if herr != nil {
+		return herr
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusOK {
+		console.Infoln("Health data uploaded to Subnet")
```
harshavardhana, line 213:
```suggestion
		console.Infoln("MinIO Health data was successfully uploaded to Subnet")
```

It would nice to print the `subnet` URL to view the health report  - this should be returned as part of the response to the successful health report as "Location" header in the response IMO.

cmd/admin-subnet-health.go
```diff
@@ -161,11 +174,85 @@ func mainAdminHealth(ctx *cli.Context) error {
...
+}
+
+func uploadHealthReport(filename string, license string, dev bool) error {
+	url := subnetUploadURL(filename, license, dev)
+	req, err := subnetUploadReq(url, filename)
+	if err != nil {
+		return err
+	}
+
+	client := &http.Client{}
```
harshavardhana, line 205: Use a proper HTTP transport such as default transport used in `mc` to avoid perpetual hangs if subnet is not reachable.

harshavardhana (changes requested):

cmd/admin-subnet-health.go
harshavardhana: In `mc` all Go error is `e` and all `err` is `*probe.Error`

cmd/admin-subnet-health.go
anjalshireesh: Done

cmd/admin-subnet-health.go
anjalshireesh: Done

cmd/admin-subnet-health.go
anjalshireesh: Done

cmd/admin-subnet-health.go
anjalshireesh: Actually have updated this PR only for this.

harshavardhana (approved):

klauspost (approved):
LGTM

cmd/admin-subnet-health.go
```diff
@@ -161,11 +177,86 @@ func mainAdminHealth(ctx *cli.Context) error {
...
+		return e
+	}
+
+	if resp.StatusCode == http.StatusOK {
+		msg := "MinIO Health data was successfully uploaded to Subnet."
+		clusterURL, _ := url.PathUnescape(gjson.Get(string(respBody), "cluster_url").String())
+		if len(clusterURL) > 0 {
+			msg += fmt.Sprintf(" Can be viewed at: %s", clusterURL)
+		}
+		console.Infoln(msg)
```
kannappanr, line 222: if `json` flag is passed, a different output should be printed. It has to be consistent with how other commands are doing it now.

```
$ mc admin user add myminio test1 test1234
Added user `test1` successfully.
$ mc admin user add myminio test1 test1234 --json
{
 "status": "success",
 "accessKey": "test1",
 "secretKey": "test1234",
 "userStatus": "enabled"
}
```

anjalshireesh: @kannappanr the `mc admin subnet health` command currently prints the health report json in case of success (when `--json` is passed). I'm not sure if it's a good idea to change that..

anjalshireesh: Also, the new `--license` option is not to be used with `--json` as it relies on the generated `tar.gz` report file.
I'll add validation to ensure that both these options are not passed together.

kannappanr (approved):
LGTM

PR #3642, opened 2021-03-09
trace: Make evaluation filters more user friendly

vadmeste:
Currently trace filters are like OR operations, but I think in debugging
we want AND instead.

Currently, --status-code 503 --path 'bucket/path'  will show 503 requests or
requests with 'bucket/path', this PR will require having both conditions
in the requests in order to be shown in the trace.


klauspost:
I guess it depends on what you are debugging, but generally AND is more useful than OR.

harshavardhana (approved):

klauspost (approved):
LGTM

PR #3644, opened 2021-03-12
cp: Add --tags to tag uploaded objects

vadmeste:
Applying tagging to objects copied in an S3 object using --tags flag.


klauspost (approved):
Maybe add more info to the docs so the usage is more clear.

harshavardhana (approved):

PR #3646, opened 2021-03-15
mirror: Add total_s3uploaded_bytes metric

vadmeste:
The new metric will track the total bytes of copied objects in the
target during mirroring.

### DIFF
--- a/cmd/mirror-main.go
+++ b/cmd/mirror-main.go
@@ -204,6 +204,10 @@ var (
 		Name: "mc_mirror_total_s3ops",
 		Help: "The total number of failed mirror events",
 	})
+	s3mirrorTotalUploadedBytes = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "mc_mirror_total_s3uploaded_bytes",
+		Help: "The total number of bytes uploaded by mirror",
+	})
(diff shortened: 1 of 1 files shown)


cmd/mirror-main.go
```diff
@@ -473,6 +477,7 @@ func (mj *mirrorJob) monitorMirrorStatus() (errDuringMirror bool) {
 		}
 
 		if sURLs.SourceContent != nil {
+			s3mirrorTotalUploadedBytes.Add(float64(sURLs.SourceContent.Size))
```
harshavardhana, line 480: we should add it after its uploaded? @vadmeste

harshavardhana (changes requested):

cmd/mirror-main.go
vadmeste: > we should add it after its uploaded? @vadmeste

yes it is already doing that. This is monitorMirrorStatus() function.

krisis (approved):

vadmeste:
@harshavardhana PTAL

kannappanr (approved):
LGTM

harshavardhana (approved):

PR #3654, opened 2021-03-19
reduce default part size upto 16MiB

harshavardhana:


harshavardhana:
part size moved from 128MiB to 16MiB as default to avoid memory build up in `mc` - let `mc` calculate the necessary multipart size for bigger files on its own. 

@vadmeste has another PR to reduce memory usage by limiting concurrency along with this PR and @vadmeste should sufficiently cover large memory usage in `mc`.

klauspost (approved):
LGTM

vadmeste (approved):
LGTM

PR #3655, opened 2021-03-19
Keep coloring even when Stdout is a pipe

### LINKED ISSUE #3632: Feature Request: mc --color to force colorized output
## Expected behavior
It should be possible to force 'mc' using color output, even if output is a pipe. In addition to the `--no-color` flag, there should be a `--color` flag, too.

## Actual behavior
'mc' switches to no-color-output automatically when output is a pipe --- and there is no flag to influence this behavior.

`mc  admin trace  -a s3/ | less -R`  displays no color, although `less -R` is capable to do so.

## mc --version
- (paste output of `mc --version`)
mc version RELEASE.2021-02-10T07-32-57Z
## System information
FreeBSD 12.2  bash command line

## Workaround
Use `screen` and log screen output

ebozduman:
Fixes #3632

With this fix, together with its minio part minio/minio#11834, `mc` commands coloring schema will be maintained even when Stdout is a pipe.

### DIFF
--- a/cmd/globals.go
+++ b/cmd/globals.go
@@ -73,6 +73,8 @@ func setGlobals(quiet, debug, json, noColor, insecure bool) {
 	// Disable colorified messages if requested.
 	if globalNoColor || globalQuiet {
 		console.SetColorOff()
+	} else {
+		console.SetColorOn()
 	}
 }
 
--- a/go.mod
+++ b/go.mod
@@ -34,3 +34,4 @@ require (
 	gopkg.in/h2non/filetype.v1 v1.0.5
 	gopkg.in/yaml.v2 v2.3.0
 )
+replace github.com/minio/minio => ../minio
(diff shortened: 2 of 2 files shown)


donatello (approved):
Works as intended - coloring is present when piping as well, unless the `--no-color` option is given. However, I feel the existing behavior is more useful. I prefer an option like `--color=always|never` for explicitly turning it on or off.

ebozduman:
If desired, we can implement @donatello's suggestion; a new flag, like `color=always` and `color=never`.
Then, we can deprecate `--no-color` flag in time.

harshavardhana:
> If desired, we can implement @donatello's suggestion; a new flag, like `color=always` and `color=never`.
> Then, we can deprecate `--no-color` flag in time.

We are not planning to do this as the server side PR is closed

PR #3677, opened 2021-04-02
Check all possible locks on an object before 'mc mv'

### LINKED ISSUE #3674: `mc mv` fails with 'locked' error even when no locking is configured yet.
## Expected behavior
`mc mv` successfully moves an object from source to destination, if the source bucket is created with `--with-lock`, and none of the locking features; legalhold, retention, ilm, is configured on that object yet.

## Actual behavior
 `mc: <ERROR> Object lock configuration is enabled on the specified bucket in alias mm1/locked/obj1.`
It looks like we only check if the bucket is created with `--with-lock` flag or not, instead of checking if a lock  mechanism has been configured on the object

1. `mc mb <alias>/regular-bucket/`
2. `mc mb --with-lock <alias>/locked/`
3. `mc cp /etc/issue <alias>/locked/obj1`
4. `mc mv <alias>/locked/obj1 <alias>/regular-bucket/`
5. `mc: <ERROR> Object lock configuration is enabled on the specified bucket in alias <alias>/locked/obj1.`

## mc --version
mc version RELEASE.2021-03-23T05-46-11Z

ebozduman:
Fixes #3674 

The fix introduces checks if any one of the 3 possible lock configurations exists for the object. If one found, then `mc mv` is rejected with a clear error message.

### DIFF
--- a/cmd/mv-main.go
+++ b/cmd/mv-main.go
@@ -20,11 +20,14 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"sync"
 
 	"github.com/fatih/color"
 	"github.com/minio/cli"
 	"github.com/minio/mc/pkg/probe"
+	"github.com/minio/minio-go/v7"
+	"github.com/minio/minio-go/v7/pkg/lifecycle"
 	"github.com/minio/minio/pkg/console"
 )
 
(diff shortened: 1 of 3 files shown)


cmd/mv-main.go
```diff
@@ -243,17 +246,32 @@ func mainMove(cliCtx *cli.Context) error {
...
 		if _, ok := client.(*S3Client); ok {
-			enabled, err := isBucketLockEnabled(ctx, urlStr)
-			if err != nil {
-				fatalIf(err.Trace(), "Unable to get bucket lock configuration of `%s`", urlStr)
+			var lifeCycleConfig *lifecycle.Configuration
+			var legalHoldStatus minio.LegalHoldStatus
+			var retentionMode minio.RetentionMode
+			var err *probe.Error
+			ignoreErr := "The specified object does not have a ObjectLock configuration"
+			if lifeCycleConfig, err = client.GetLifecycle(ctx); err != nil && strings.Contains(err.Cause.Error(), ignoreErr) {
```
vadmeste, line 254: There is no such thing called ILM lock, I don't think we need to call this function.

vadmeste (commented):
I think we don't have to call isBucketLockEnabled() in the first place. It is enough to remove it and we can quit when we encounter the first WORM error.

cmd/mv-main.go
ebozduman: @vadmeste 
Done. Removed it.

ebozduman:
@vadmeste,

After we discussed the issue, I refactored the fix/code, such that we make the inexpensive call, `Stat` to get the metadata, which has the lock configuration information.

PTAL again.

kannappanr:
`mv` should not work on a source bucket which has lock not enabled and probably unversioned. We should atleast check if the bucket is object lock enabled or not. If it is, you should not allow mv to proceed

ebozduman:
Closing per Kannappan's recommendation as the current behavior is what the requirement asks for.

Issue #3621, opened 2021-02-21
mc mirror remove parent directory when all files inside directories are deleted


GreatBahram:
## Expected behavior

1. I'm using `mc mirror` to keep a directory on my computer sync with a bucket on a MinIO bucket. When I create a directory inside my bucket, let's call it `mydir`,  put some files inside mydir directory, and then remove all of those files, `mc mirror` will, in the end, remove mydir as well. I guess I have to say, I tested this behavior on the MinIO Web interface, and the result was the same, So I'm not sure if this is the right place to follow up on this problem. 
### Steps to reproduce the behavior
```shell
mc mirror --watch mydir myminio/mydir & 
# create a new empty directory 'first'
mkdir mydir/first/
# I create some files in mydir/first directory
for i in {1..10}; do touch mydir/first/$i; done

# Now, interestingly, if I remove all files inside mydir/first directory, `mc mirror will remove the first directory as well!!
rm -rf mydir/first/*
```
### Actual behavior
I expected that mydir exists on both MinIO and on my local directory.

2. And, there is another problem that is related to this one and that is if I want to mirror things from minio to a local directory and if I remove all files in that bucket. Then run the `mc mirror` command with the `--remove` option; it will remove mydir on my computer instead of removing all files and directories inside that directory. And then it'll create it again.

## Steps to reproduce the behavior
```shell
mc mirror --remove myminio/mybucket mydir &

# Then, I remove all files and directories on  minio/mybucket/*

# Now, if I run mc mirror with --remove option, it will remove the mydir directory on my computer, and it'll create it again!
mc mirror --remove myminio/mydir mydir &

# You could see this behavior by running watch -n 0.5 ls mydir on another tab/terminal
```
### Actual behavior
This behavior can be problematic for an end-user who doesn't understand the philosophy behind it, for some moment the directory he/she was in it, it's not available. This problem is different from the first problem, in the first problem, both MinIO and mc are acting the same way, while here, I think, mc mirror will remove the target directory out of its volition, maybe it's for efficiency or other reason.

arizvisa:
This might be a dupe of #3379.

GreatBahram:
Yes, it is related to that one. But the previous one didn't get fixed as they thought it was a local filesystem issue. So at this point, we can strongly say it happens at least on two major platforms.

arizvisa:
This is not stale as it seems to be being worked on in PR #3678.

ebozduman:
[...]

PR #3678, opened 2021-04-05
Stops removing empty filesystem  paths

ebozduman:
Fix for #3621
`mc rm`, `mc mv`, or `mc mirror` commands stop removing the empty path after a file is successfully removed.
This fix does not change the behavior of object storage objects and prefixes.
So, it only works if the resource is a filesystem entity.

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -440,9 +440,9 @@ func isSysErrNotEmpty(err error) bool {
 	return false
 }
 
-// deleteFile deletes a file path if its empty. If it's successfully deleted,
-// it will recursively delete empty parent directories
-// until it finds one with files in it. Returns nil for a non-empty directory.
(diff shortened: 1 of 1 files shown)


cmd/client-fs.go
```diff
@@ -457,10 +457,6 @@ func deleteFile(deletePath string) error {
 	parentPath := strings.TrimSuffix(deletePath, slashSeperator)
 	parentPath = path.Dir(parentPath)
 
-	if parentPath != "." {
```
klauspost, line 460: Yeah, everything about `parentPath` can be removed.

ebozduman: Done.

arizvisa:
(not stale)

klauspost (approved):

ebozduman:
I guess there is no reason to keep this PR anymore.
Closing as "wontfix".

PR #3702, opened 2021-04-28
fix: correctly set content-disposition header for share download

### LINKED ISSUE #3697: mc share download doesn't contain content-disposition header
## Expected behavior

A link generated via `mc share download` should have the `content-disposition` header set to the original filename.
This causes some browsers, like Chrome, to rename the file according to the `content-type` when downloading.

## Actual behavior

The `content-disposition` header is not set.
The file is downloaded with the correct filename by Chrome.

1. Create a zip file and rename it's ending to something esle. e.g. `.zip2`
2. Upload the file to minio
3. Generate a share download link via `mc share download <path to file>`
4. Open the generated link in Chrome
5. The file is downloaded with the `.zip` ending, instead of `.zip2`

## mc --version
mc version RELEASE.2021-03-23T05-46-11Z

## System information
Linux machine 5.4.108-1-MANJARO #1 SMP PREEMPT Wed Mar 24 18:24:35 UTC 2021 x86_64 GNU/Linux

## Investigation

I'm not sure if this is an issue in mc or the minio server.

It looks like on the server side the header was added https://github.com/minio/minio/pull/11712

I tested it against minio RELEASE.2021-04-06T23-11-00Z, which is far as I can tell includes https://github.com/minio/minio/pull/11712. But the behavior is still as described above.

Also using the web interface the `content-disposition` header is correctly set. So I assume it is a client issue.

bachp:
fixes #3697

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -2099,6 +2099,10 @@ func (c *S3Client) ShareDownload(ctx context.Context, versionID string, expires 
 	if versionID != "" {
 		reqParams.Set("versionId", versionID)
 	}
+
+	// Make sure files are downloaded with their original name
+	reqParams.Set("Response-Content-Disposition", fmt.Sprintf("attachment; filename=\"%s\"", path.Base(object)))
+
 	presignedURL, e := c.api.PresignedGetObject(ctx, bucket, object, expires, reqParams)
 	if e != nil {
 		return "", probe.NewError(e)


bachp:
@harshavardhana @poornas Is there something missing to get this looked at?

bachp:
@harshavardhana I rebase the PR on latest master

klauspost:
Not the worlds biggest fan of adding it on all URLs. While it may work fine in your case it will change the behavior for everybody.

For example upload an image and share it. Before this it would display in a browser, now it will always be downloaded.

bachp:
@klauspost I didn't consider that use case. Do you have a suggestion how to solve original issue with the wrongly renamed file?

One I could think off is to add an additional paramter to `mc share download`?

harshavardhana:
> @klauspost I didn't consider that use case. Do you have a suggestion how to solve original issue with the wrongly renamed file?

@bachp you can add `mc share download --preserve-filename`  would be better.

bachp:
@harshavardhana After reading some more about what Content-Disposition actually does I think `--force-download` or something similar might be a better option.

Also I'm not sure anymore that this is the right solution to my problem. 

As I now found out the resulting file would also be properly downloaded with the correct name if I set it's `Content-Type` to `application/octet-stream`. Another way to is to add `Content-Disposition` during upload to the file. So maybe this are better solutions.

The only thing I haven't figured out yet is how to change the metadata of an already uploaded file via `mc` or how to change the `Content-Type` of some files when doing a recursive `mc mv` or `mc cp`?

harshavardhana:
> The only thing I haven't figured out yet is how to change the metadata of an already uploaded file via `mc` or how to change the `Content-Type` of some files when doing a recursive `mc mv` or `mc cp`?

That is just `mc cp --attr`  you can look at examples in `mc cp --help`

bachp:
@harshavardhana Is there also a way to change the `Content-Type` of an already uploaded file, without re uploading it?

harshavardhana:
> @harshavardhana Is there also a way to change the `Content-Type` of an already uploaded file, without re uploading it?

@bachp as I had mentioned before `mc cp --help` 

Just like this `m cp --attr "Content-Type=archive/zip" --recursive myminio/mybucket myminio/mybucket` for all objects

vadmeste:
I think --add-request-header is better use case here, some users could need something else other than content-disposition here

bachp:
@vadmeste I think having a dedicated `--preserve-filename` mught still be useful as with the generic option you would need to pass the filename yourself

bachp:
@harshavardhana Thanks for the hint. I figured it out what I was looking for is:

```
mc find myminio/mybucket --name "*.zip2" --exec 'mc cp --attr Content-Type=application/octet-stream {} {}'
```

Which changes the content type of all my zip2 files already uploaded.

PR #3722, opened 2021-05-14
Add flag for replica sync to `mc replicate` command

poornas:
Signed-off-by: Poorna Krishnamoorthy <poorna@minio.io>

Needs https://github.com/minio/minio-go/pull/1491

### DIFF
--- a/cmd/replicate-add.go
+++ b/cmd/replicate-add.go
@@ -62,7 +62,7 @@ var replicateAddFlags = []cli.Flag{
 	},
 	cli.StringFlag{
 		Name:  "replicate",
-		Usage: "comma separated list to enable replication of delete markers, and/or deletion of versioned objects.Valid options are \"delete-marker\", \"delete\" and \"\"",
+		Usage: "comma separated list to enable replication of delete markers, deletion of versioned objects and replica metadata sync(in the case of active-active replication).Valid options are \"delete-marker\", \"delete\" ,\"replica-metadata-sync\" and \"\"",
 	},
 }
 
@@ -162,6 +162,7 @@ func mainReplicateAdd(cliCtx *cli.Context) error {
 	}
 	dmReplicateStatus := disableStatus
 	deleteReplicationStatus := disableStatus
+	replicaSync := enableStatus
 	if cliCtx.IsSet("replicate") {
 		replSlice := strings.Split(cliCtx.String("replicate"), ",")
 		for _, opt := range replSlice {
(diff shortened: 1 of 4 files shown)


cmd/replicate-add.go
```diff
@@ -64,6 +64,10 @@ var replicateAddFlags = []cli.Flag{
 		Name:  "replicate",
 		Usage: "comma separated list to enable replication of delete markers, and/or deletion of versioned objects.Valid options are \"delete-marker\", \"delete\" and \"\"",
 	},
+	cli.BoolFlag{
+		Name:  "disable-replica-sync",
```
ravindk89, line 68: Are we set on this naming? It's a little unintuitive to have a flg `--disable-replica-sync=false` because of the double negative.

Would it be difficult to just do `replica-sync` and have it follow the 'enabled/disabled(default)' structure from minio-go? That would also keep the flag in line w/ replicate-edit.go

cmd/replicate-edit.go
```diff
@@ -60,6 +60,10 @@ var replicateEditFlags = []cli.Flag{
 		Name:  "replicate",
 		Usage: "comma separated list to enable replication of delete markers, and/or deletion of versioned objects.Valid options are \"delete-marker\", \"delete\" and \"\"",
```
ravindk89, line 61: Q: Is there a reason to not include replica-sync here, since its a modifier on replication behavior? Or do we want to keep this only for delete-related behavior (in which case, calling it `--replicate` is a bit odd)

cmd/replicate-add.go
poornas: This flag is set only when replica sync is explicitly disabled by user - if we were to have --replica-sync like in the edit command, there is the gray area of when it is not set. Since this applies to just 2 way replication, we can't have a edit to enforce this flag. Bool flags do not have a value - `replicate add --disable-replica-sync ` command disables this setting

cmd/replicate-edit.go
poornas: that's actually a good idea - we could do that

cmd/replicate-add.go
ravindk89: Ah I misunderstood from the example. If we do keep this, we might state something like:

"Usage: Disables syncing of replica metadata modifications in active-active replication configurations. Omit this flag to enable syncing replica metadata modifications.

cmd/replicate-edit.go
ravindk89: To address the concern above w.r.t. specifying this in active-passive - if I set `--replicate "delete,delete-marker,replica-sync" in active-passive, nothing should actually happen right? Because on Site A, no object would ever have replica status. Setting it right away in active-passive should have no "visible" effect, nor would enabling it after initially starting replication.

If a user then did enable active-active, replica sync would work as expected - though they would need to set `replica-sync` on both ends. The documentation can include explicit warnings against setting the field on only one side of active-active, as you wont see true synchronization on both sides.

cmd/replicate-edit.go
```diff
@@ -97,6 +97,8 @@ EXAMPLES:
   6. Disable delete marker and versioned delete replication on a replication configuration rule with ID "kxYD.491" on a target myminio/bucket.
      {{.Prompt}} {{.HelpName}} myminio/mybucket --id "kxYD.491" --replicate ""
 
+  7. Enable replica metadata sync ,delete marker and versioned delete replication on a replication configuration rule with ID "kxYD.491" on a target myminio/bucket.
```
ravindk89, line 100:
```suggestion
  7. Enable replica metadata sync ,delete marker and versioned delete replication on a replication configuration rule with ID "kxYD.491" on a target myminio/bucket. Replica metadata sync has no effect in active-passive configurations.
```

ravindk89 (approved):
LGTM docs-wise. If I'm thinking wrongly about the effect of `replica-sync` on active-passive configs, let me know.

cmd/replicate-edit.go
poornas: in an active passive setup this setting would have no meaning - since no object would be in Replica status. The default is always replica-sync enabled. So in an active-active scenario, if they edit a rule and don't set the replica-metadata-sync while changing delete or delete-marker replication setting, it would essentially disable it. The docs need to have a caveat on that.

kannappanr (approved):
LGTM

PR #3725, opened 2021-05-18
fix: enhance support for special access/secretkeys in remote targets

harshavardhana:
it latest release and master `mc admin bucket remote add` cannot
parse accesskeys/secretkeys with special characters such as `@`,
this is due to an incorrect implementation.

use the technique used in parsing MC_HOST_alias here to allow
for more complex use-cases.


poornas (approved):
LGTM

kannappanr (approved):
LGTM

PR #3730, opened 2021-05-28
heal: Refactor info output

vadmeste:
mc admin heal <alias> will show simplifed output which aggregates MRF &
new disks healing.

e.g.:

```
$ mc admin heal myminio/
Objects Healed: 7/27 (25.9%), 31 MB/110 MB (0%)
Heal rate: 3 obj/s, 10 MB/s
Estimated Completion: 11 seconds
```


kannappanr (approved):
LGTM

PR #3732, opened 2021-06-01
Fix build on 32-bit systems (OpenBSD/solaris)

danieljakots:
Hi,

This is currently broken with `pkg/disk/stat_other.go:40:91: cannot use st.Atim.Nsec (type int32) as type int64 in argument to strconv.FormatInt` (line number is off because it's an old release).

Does it make sense to  you? I can't test it unfortunately because I don't have access to a 32-bit system.

`go build` works on my amd64 system so no regression on that side I guess.

Cheers,
Daniel

### DIFF
--- a/pkg/disk/stat_other.go
+++ b/pkg/disk/stat_other.go
@@ -38,7 +38,7 @@ func GetFileSystemAttrs(file string) (string, error) {
 
 	var fileAttr strings.Builder
 	fileAttr.WriteString("atime:")
-	fileAttr.WriteString(strconv.FormatInt(st.Atim.Sec, 10) + "#" + strconv.FormatInt(st.Atim.Nsec, 10))
+	fileAttr.WriteString(strconv.FormatInt(int64(st.Atim.Sec), 10) + "#" + strconv.FormatInt(int64(st.Atim.Nsec), 10))
 	fileAttr.WriteString("/gid:")
 	fileAttr.WriteString(strconv.Itoa(int(st.Gid)))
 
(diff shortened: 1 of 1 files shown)


harshavardhana:
> This is currently broken with `pkg/disk/stat_other.go:40:91: cannot use st.Atim.Nsec (type int32) as type int64 in argument to strconv.FormatInt` (line number is off because it's an old release).

We do not support 32bit systems @danieljakots - please report if there is an actual issue.

danieljakots:
Can you please precise what you mean by "We do not support 32bit systems"? Is it the same as "we don't support officially OpenBSD but we have code to make it build/work" or is it "we don't want code for 32bit systems"?
(no strong feeling about it either way, it's just to know what I should do ;))

harshavardhana:
> Can you please precise what you mean by "We do not support 32bit systems"? Is it the same as "we don't support officially OpenBSD but we have code to make it build/work" or is it "we don't want code for 32bit systems"?
> (no strong feeling about it either way, it's just to know what I should do ;))

We do not support 32bit systems, OpenBSD is a stretch. If you wish to get this PR in make sure to update

```
buildscripts/cross-compile.sh
```

```
function _init() {
    ## All binaries are static make sure to disable CGO.
    export CGO_ENABLED=0

    ## List of architectures and OS to test coss compilation.
    SUPPORTED_OSARCH="linux/ppc64le linux/mips64 linux/arm64 linux/s390x darwin/arm64 darwin/amd64 freebsd/amd64 windows/amd64 linux/arm linux/386 netbsd/amd64 linux/mips"
}
```

danieljakots:
Thanks for the clarification!

I asked someone to test and the build is still broken (because of github.com/tklauser/go-sysconf / golang.org/x/sys/unix this time). I'm going to look into it.
Does it make sense to add openbsd/i386 to cross-compile.sh yet? Or do you prefer to wait until mc builds successfully on openbsd/i386?

I think we should add openbsd/amd64 to the file either way.

harshavardhana:
> 
> I think we should add openbsd/amd64 to the file either way.

👍🏽  - this alone is acceptable for now.

danieljakots:
Thanks, I created two PRs: one for minio/minio and one for minio/mc.

harshavardhana (approved):

PR #3737, opened 2021-06-03
du: Use recursive listing when no detailed report is asked

vadmeste:
mc du will always count the size of all objects under a specified
path, du -r will show the disk usage in a more detailed fashion, but if
not specified, let's do a recursive LIST call instead of calling du()
multiple times.

Bonus: check if the passed argument is a directory and error out
otherwise.

### DIFF
--- a/cmd/du-main.go
+++ b/cmd/du-main.go
@@ -18,6 +18,7 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
 	"net/url"
 	"strings"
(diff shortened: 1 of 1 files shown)


cmd/du-main.go
```diff
@@ -122,10 +123,17 @@ func du(urlStr string, timeRef time.Time, withVersions bool, depth int, encKeyDB
 		return 0, exitStatus(globalErrorExitStatus) // End of journey.
 	}
 
-	contentCh := clnt.List(globalContext, ListOptions{
+	recursive := false
+	if depth == 1 {
+		// No disk usage details below this level,
+		// just do a recursive listing
+		recursive = true
+	}
```
klauspost, line 131:
```suggestion
	// No disk usage details below this level,
	// just do a recursive listing
	recursive := depth == 1
```

klauspost (approved):
I guess this covers 95% of use cases anyway.

harshavardhana (approved):

PR #3743, opened 2021-06-11
fix: remove legacy path for prometheus

harshavardhana:
additionally add `--public` flag for unauthenticated
prometheus config.

### DIFF
--- a/cmd/admin-prometheus-generate.go
+++ b/cmd/admin-prometheus-generate.go
@@ -34,9 +34,15 @@ import (
 
 const (
 	defaultJobName     = "minio-job"
-	legacyMetricsPath  = "/minio/prometheus/metrics"
 	defaultMetricsPath = "/minio/v2/metrics/cluster"
 )
+
+var prometheusFlags = []cli.Flag{
+	cli.BoolFlag{
+		Name:  "public",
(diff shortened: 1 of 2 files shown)


poornas (approved):

cmd/admin-prometheus-generate.go
```diff
@@ -166,38 +159,26 @@ func generatePrometheusConfig(ctx *cli.Context) error {
...
+			return e
+		}
 
-	info, e := client.ServerInfo(globalContext)
-	if e != nil {
-		fatalIf(probe.NewError(e), "Failed to get server info.")
+		// Setting the values
+		defaultConfig.ScrapeConfigs[0].BearerToken = token
 	}
-	if info.Servers[0].Version < "2021-01-30T00-20-58Z" {
```
vadmeste, line 191: 2021-01-30T00-20-58Z is considered very recent for systems in production :) I think it is still so soon to remove legacy code.

harshavardhana: yeah but this is not useful @vadmeste we can simply document this properly and whoever is 01-30 can use older `mc`. We don't have to support older servers for long periods.

vadmeste (approved):
LGTM

PR #3766, opened 2021-08-05
Add multisite replication support in `mc replicate` cmd

poornas:
Also adding a flag to `mc replicate reset` to allow granular
replication reset by target

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -1187,8 +1187,8 @@ func (f *fsClient) GetReplicationMetrics(ctx context.Context) (replication.Metri
 
 // ResetReplication - kicks off replication again on previously replicated objects if existing object
 // replication is enabled in the replication config, not implemented
-func (f *fsClient) ResetReplication(ctx context.Context, before time.Duration) (string, *probe.Error) {
-	return "", probe.NewError(APINotImplemented{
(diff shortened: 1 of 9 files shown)


cmd/replicate-edit.go
```diff
@@ -185,12 +181,11 @@ func mainReplicateEdit(cliCtx *cli.Context) error {
 
 	opts := replication.Options{
 		TagString:    cliCtx.String("tags"),
-		RoleArn:      cliCtx.String("arn"),
 		StorageClass: cliCtx.String("storage-class"),
 		RuleStatus:   state,
 		ID:           cliCtx.String("id"),
 		Op:           replication.SetOption,
-		DestBucket:   cliCtx.String("remote-bucket"),
+		DestBucket:   cliCtx.String("arn"),
```
kannappanr, line 188: we cannot reuse the flag with a different meaning. It will mean different in existing scripts

poornas: why? the --arn flag already has the destination bucket name in its naming convention. Since we are dispensing with role Arn  and using it in the Destination bucket section of the replication configuration, this is precisely what we want to do here. Even if this was being used in existing scripts, this is the conversion we will make internally for backward compatibility.

poornas: ok, added back arn flag

kannappanr (approved):
LGTM

harshavardhana (approved):

PR #3777, opened 2021-08-19
Support autotuning of the workload for speedtest

krishnasrinivas:


harshavardhana (changes requested):
@krishnasrinivas I thought we are going to do auto-tuning on the server side, not on the client?

krishnasrinivas:
@harshavardhana I think we did not discuss on the implementation details. I did it this way as it's a simpler approach to keep API and server side functionality simple and drive the tuning from client side.

dvaldivia (approved):
Tested. LGTM.
```
root@my-shell2:/# ./mc admin speedtest myminio
Running speedtest (autotune: true).........
MinIO 2021-08-17T20:53:08Z, 4 nodes, 28 drives
PUT: 1.4 GiB/s, 1012 objs/s
GET: 3.0 GiB/s, 3839 objs/s
```

harshavardhana (approved):

PR #3780, opened 2021-08-20
rb: --force flag forces bucket removal after BucketNotEmpty err

vadmeste:
Some users struggle to remove a bucket when they don't use recommanded
setups, such as NFS disks.

This command will force bucket removal in server side, using
x-minio-force-delete during DELETE bucket call.

### DIFF
--- a/cmd/client-fs.go
+++ b/cmd/client-fs.go
@@ -894,6 +894,15 @@ func (f *fsClient) MakeBucket(ctx context.Context, region string, ignoreExisting
 	return nil
 }
 
+// RemoveBucket - remove a bucket
+func (f *fsClient) RemoveBucket(ctx context.Context, forceRemove bool) *probe.Error {
+	e := os.Remove(f.PathURL.Path)
+	if e != nil {
+		return probe.NewError(e)
+	}
+	return nil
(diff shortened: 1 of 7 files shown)


cmd/client-fs.go
```diff
@@ -894,6 +894,15 @@ func (f *fsClient) MakeBucket(ctx context.Context, region string, ignoreExisting
 	return nil
 }
 
+// RemoveBucket - remove a bucket
+func (f *fsClient) RemoveBucket(ctx context.Context, forceRemove bool) *probe.Error {
+	e := os.Remove(f.PathURL.Path)
```
klauspost, line 899:
```suggestion
	e := os.RemoveAll(f.PathURL.Path)
```

klauspost (commented):
LGTM. One suggestion.

cmd/client-fs.go
vadmeste: I made another change, PTAL.

harshavardhana (approved):

klauspost (approved):
LGTM

harshavardhana (approved):

PR #3783, opened 2021-08-23
validation for namespace removal

### LINKED ISSUE #3776: mc rb --force is enough for site-wide removal of buckets
## Expected behavior

As I understand the code in rb-main.go a site-wide removal (e.g. `mc rb <alias>`) should require both the `--force` and `--dangerous` options.

## Actual behavior

It is possible to simply use `mc rb --force <alias>` to remove all buckets the aliased user has access to.

1. Run minio as a local S3 server: `./minio server ./data`
2. Setup a `mc alias` for the local S3 server: `./mc alias set local http://localhost:9000 minioadmin minioadmin`
3. Create some buckets and upload objects, e.g.
    ```bash
    for i in {0..5}; do ./mc mb "local/new$i"; ./mc cp mc "local/new$i"; done
    ```
4. Execute `./mc rb --force local`.

## mc --version
- mc version RELEASE.2021-07-27T06-46-19Z

## System information

- Using an openSUSE-Leap-15.2 WSL2 instance on Windows.

sinhaashish:
## Expected behavior
 rm-main.go a site-wide removal (e.g. `mc rm <alias>`) should require both the `--force` and `--dangerous` options.

## Actual behavior

It is possible to simply use `mc rm --force <alias>` to remove all buckets the aliased user has access to.

1. Run minio as a local S3 server: `./minio server ./data`
2. Setup a `mc alias` for the local S3 server: `./mc alias set local http://localhost:9000 minioadmin minioadmin`
3. Create some buckets and upload objects, e.g.
    ```bash
    for i in {0..5}; do ./mc mb "local/new$i"; ./mc cp mc "local/new$i"; done
    ```
4. Execute `./mc rm --force local`.

## System information

- Using an openSUSE-Leap-15.2 WSL2 instance on Windows.
Closes #3776

### DIFF
--- a/cmd/rm-main.go
+++ b/cmd/rm-main.go
@@ -210,7 +210,7 @@ func checkRmSyntax(ctx context.Context, cliCtx *cli.Context, encKeyDB map[string
 		url = filepath.ToSlash(filepath.Clean(url))
 		// namespace removal applies only for non FS. So filter out if passed url represents a directory
 		dir := isAliasURLDir(ctx, url, encKeyDB, time.Time{})
-		if !dir {
+		if dir {
 			_, path := url2Alias(url)
 			isNamespaceRemoval = (path == "")
 			break
@@ -231,17 +231,15 @@ func checkRmSyntax(ctx context.Context, cliCtx *cli.Context, encKeyDB map[string
 
 	// For all recursive or versions bulk deletion operations make sure to check for 'force' flag.
 	if (isVersions || isRecursive || isStdin) && !isForce {
-		if isNamespaceRemoval {
-			fatalIf(errDummy().Trace(),
-				"This operation results in site-wide removal of objects. If you are really sure, retry this command with ‘--dangerous’ and ‘--force’ flags.")
-		}
 		fatalIf(errDummy().Trace(),
 			"Removal requires --force flag. This operation is *IRREVERSIBLE*. Please review carefully before performing this *DANGEROUS* operation.")
 	}
-	if (isRecursive || isStdin) && isNamespaceRemoval && !isDangerous {
+
+	if isNamespaceRemoval && !(isDangerous && isForce) {
 		fatalIf(errDummy().Trace(),
 			"This operation results in site-wide removal of objects. If you are really sure, retry this command with ‘--dangerous’ and ‘--force’ flags.")
 	}
+
 }
 
 // Remove a single object or a single version in a versioned bucket


vadmeste (approved):
LGTM

cmd/rm-main.go
```diff
@@ -210,7 +210,7 @@ func checkRmSyntax(ctx context.Context, cliCtx *cli.Context, encKeyDB map[string
 		url = filepath.ToSlash(filepath.Clean(url))
 		// namespace removal applies only for non FS. So filter out if passed url represents a directory
 		dir := isAliasURLDir(ctx, url, encKeyDB, time.Time{})
-		if !dir {
+		if dir {
```
vadmeste, line 213: @sinhaashish, we have a function call isNamespaceRemoval(), we can simplify the code here, can you do that ?

vadmeste (changes requested):
One comment

cmd/rm-main.go
vadmeste, line 213: Can you use isNamespaceRemoval() here ?

cmd/rm-main.go
```diff
@@ -242,6 +242,11 @@ func checkRmSyntax(ctx context.Context, cliCtx *cli.Context, encKeyDB map[string
 		fatalIf(errDummy().Trace(),
 			"This operation results in site-wide removal of objects. If you are really sure, retry this command with ‘--dangerous’ and ‘--force’ flags.")
 	}
+	if isNamespaceRemoval && (!isDangerous || !isForce) {
```
klauspost, line 245:
```suggestion
	if isNamespaceRemoval && !(isDangerous && isForce) {
```
This is more readable to me (if is namespace removal and not is dangerous and force)

klauspost: This overlaps with the checks above.  It seems we are just throwing more cases into the mix.

Can't we simplify these?

sinhaashish: @klauspost @vadmeste  PTAL

cmd/rm-main.go
sinhaashish: In the code we check if its a directory ,  then find the path to decide if its a name space removal or not. So `isNamespaceRemoval` cant be used there.

vadmeste: > In the code we check if its a directory , then find the path to decide if its a name space removal or not. So `isNamespaceRemoval` cant be used there.

I didn't understand. What I am saying is it L210->L216 seems to be checking if this is a namespace removal or not. Which is what isNamespaceRemoval() is doing as well. So my suggestion is to remove L210->L216 line and replace it with isNamespaceRemoval()

klauspost (approved):
LGTM. Can't help you decide the thing you are discussing with Anis.

vadmeste (approved):
LGTM

PR #3787, opened 2021-08-26
Fix the issue remove object in mirror process event if set flag is no…

nguyentankhtn:
In order to fix the bug as remove objects in the mirror cmd event if set the remove argument as false
Change the logic at **if mirrorURL.TargetContent != nil && (mj.opts.isRemove || mj.opts.activeActive)** in the function watchMirrorEvents
become 
			**if mirrorURL.TargetContent != nil && mj.opts.isRemove && mj.opts.activeActive**


luk020 (approved):

lscgc (approved):

klauspost:
@nguyentankhtn Please `go fmt` the code.

Could you explain what you are fixing here, with a reproducer so we can test it?

ravindk89:
@klauspost this *might* be due to some misunderstandings around what the intended default behavior of `mc mirror --watch` is

https://github.com/minio/docs/issues/377

The `--remove` command is causing some confusion here - since our CLI output for `--watch` does not imply any delete functionality.

@nguyentankhtn @luk020 @lscgc 

Based on internal discussion, the *expectation* is that `--watch` synchronizes PUT and DELETE operations between source and destination. `--remove` *specifically removes extra objects that only exist on the destination*. `--overwrite` allows `mc mirror` to stomp a file on the destination if the source does not match . ANd `--active-active` is `--watch --remove --overwrite` in one flag.

This is, I believe, our final determination at this time on what `mc mirror` should do. Given the scenario of two empty buckets, `mc mirror --watch` by default would result in the destination bucket "mirroring" the source bucket. Excluding DELETE operations would result in the destination bucket being a *historical record*, not a mirror. That is a distinct and different use case.

Let me know if this makes sense - it might change whether this PR is fixing an actual bug or modifying existing intended behavior.

nguyentankhtn:
> @klauspost this _might_ be due to some misunderstandings around what the intended default behavior of `mc mirror --watch` is
> 
> [minio/docs#377](https://github.com/minio/docs/issues/377)
> 
> The `--remove` command is causing some confusion here - since our CLI output for `--watch` does not imply any delete functionality. So our docs are contributing, and I have an active issue to fix that.
> 
> @nguyentankhtn @luk020 @lscgc
> 
> Based on internal discussion, the _expectation_ is that `--watch` synchronizes PUT and DELETE operations between source and destination. `--remove` _specifically removes extra objects that only exist on the destination_. `--overwrite` allows `mc mirror` to stomp a file on the destination if the source does not match . ANd `--active-active` is `--watch --remove --overwrite` in one flag.
> 
> This is, I believe, our final determination at this time on what `mc mirror` should do. Given the scenario of two empty buckets, `mc mirror --watch` by default would result in the destination bucket "mirroring" the source bucket. Excluding DELETE operations would result in the destination bucket being a _historical record_, not a mirror. That is a distinct and different use case. `--remove` would handle situations where the destination bucket may not be empty, and `--overwrite` handles situations where the destination bucket does not have versioning (thus requiring an overwrite on object name collision)
> 
> Let me know if this makes sense - it might change whether this PR is fixing an actual bug or modifying existing intended behavior.

@ravindk89, How can I mirror objects from site A to site B if we wanna the delete action in site A does not to apply to site B?
I can see in old versions, it will work if we do not set the flag --remove. So that is confusing to us.

luk020:
> @klauspost this _might_ be due to some misunderstandings around what the intended default behavior of `mc mirror --watch` is
> 
> [minio/docs#377](https://github.com/minio/docs/issues/377)
> 
> The `--remove` command is causing some confusion here - since our CLI output for `--watch` does not imply any delete functionality. So our docs are contributing, and I have an active issue to fix that.
> 
> @nguyentankhtn @luk020 @lscgc
> 
> Based on internal discussion, the _expectation_ is that `--watch` synchronizes PUT and DELETE operations between source and destination. `--remove` _specifically removes extra objects that only exist on the destination_. `--overwrite` allows `mc mirror` to stomp a file on the destination if the source does not match . ANd `--active-active` is `--watch --remove --overwrite` in one flag.
> 
> This is, I believe, our final determination at this time on what `mc mirror` should do. Given the scenario of two empty buckets, `mc mirror --watch` by default would result in the destination bucket "mirroring" the source bucket. Excluding DELETE operations would result in the destination bucket being a _historical record_, not a mirror. That is a distinct and different use case. `--remove` would handle situations where the destination bucket may not be empty, and `--overwrite` handles situations where the destination bucket does not have versioning (thus requiring an overwrite on object name collision)
> 
> Let me know if this makes sense - it might change whether this PR is fixing an actual bug or modifying existing intended behavior.

This was a very good, high quality reply. Much appreciated.

I have two follow up questions:
- Does that mean that `mirror --watch` has the power to remove objects but `mirror` without the flag doesn't?
- Regardless of the answer to the above, is there an argument against adding some sort of a `--no-deletion-mirroring` feature?

ravindk89:
> I can see in old versions, it will work if we do not set the flag --remove.

Yeah, I appreciate that being confusing. However, looking at the code, this behavior appears to be at least [over a year old](https://github.com/minio/mc/blob/df8bb795bac7fffb7b7fc564d27c6543765d2a21/cmd/mirror-main.go#L481-L502) - meaning that since at least that time, `mc watch --mirror` has mirrored PUT and DELETE operations.

> Does that mean that mirror --watch has the power to remove objects but mirror without the flag doesn't?

I'm not certain - from the [code](https://github.com/minio/mc/blob/master/cmd/mirror-main.go) we call `doRemove()` in both the [`watchMirrorEvents`](https://github.com/minio/mc/blob/master/cmd/mirror-main.go#L595-L614) and [`startMirror`](https://github.com/minio/mc/blob/master/cmd/mirror-main.go#L720-L724). But `startMirror` appears to depend on the `--remove` option to have that behavior.

I need to test this to be sure, but this would be consistent with the expectation here - `mc mirror --watch` = we *watch* for new PUT/DELETE operations and synchronize those to the target. If we aren't watching for anything, then `mc mirror` is effectively doing what `rclone` or `rsync` would do - take the source at the time the command runs and synchronize that to the destination. Any new PUT or DELETE ops may not be captured. The code has some logic for when an object on the source does not exist (which is, do nothing). 

I need to spend some time testing, after which I will handle the open issue on my end. 

> How can I mirror objects from site A to site B if we wanna the delete action in site A does not to apply to site B?

The answer today would be to use Server-SIde Bucket Replication. You can explicitly configure server-side replication to *not* replicate delete operations or delete markers, creating a complete historical record on the destination. Our docs talk a bit about that [here](https://docs.min.io/minio/baremetal/replication/enable-server-side-one-way-bucket-replication.html#considerations)

> [Can we add some new flags to support X behaviors/logic?]

There are some creative suggestions here - @luks20 your idea of an `--event [PUT,DELETE]` flag for example - but they require us to assign engineering resources to fully test and confirm that nothing breaks with existing production use cases that are depending on the current `mc mirror --watch` functionality. Each new flag adds to the matrix of interactivity between flags, and generally increases our support burden in the future.

That's not something we can do without taking resources from other priority initiatives. I also cannot promise that we can revisit this at some point. 

As an aside and a personal note, I was reading the docs for `rclone` and saw this on their front page:

> Users call rclone "The Swiss army knife of cloud storage" ...

This is a personal opinion than an official company opinion, but as the (only) tech writer I hope we *dont* aim to be a swiss army knife for S3. Having APIs that provide compact solutions to a well-defined set of problems is easier to document and maintain over time. Others in our org might disagree, I suppose, so I'm just speaking for myself. Looking at `mc mirror` I can already see signs of us trying to solve many problems in one tool. I *think* at this time, we're trying to avoid doing any more of that (at least for `mc mirror`). 

Hopefully this discussion if nothing else provides some insight into what we (or I, at least) am hoping to achieve with our docs + product.

luk020:
> > How can I mirror objects from site A to site B if we wanna the delete action in site A does not to apply to site B?
> 
> The answer today would be to use Server-SIde Bucket Replication. You can explicitly configure server-side replication to _not_ replicate delete operations or delete markers, creating a complete historical record on the destination. Our docs talk a bit about that [here](https://docs.min.io/minio/baremetal/replication/enable-server-side-one-way-bucket-replication.html#considerations)
>
> As an aside and a personal note, I was reading the docs for `rclone` and saw this on their front page:
> 
> > Users call rclone "The Swiss army knife of cloud storage" ...
> 
> This is a personal opinion than an official company opinion, but as the (only) tech writer I hope we _dont_ aim to be a swiss army knife for S3. Having APIs that provide compact solutions to a well-defined set of problems is easier to document and maintain over time. Others in our org might disagree, I suppose, so I'm just speaking for myself. Looking at `mc mirror` I can already see signs of us trying to solve many problems in one tool. I _think_ at this time, we're trying to avoid doing any more of that (at least for `mc mirror`).
> 
> Hopefully this discussion if nothing else provides some insight into what we (or I, at least) am hoping to achieve with our docs + product.

Thanks once again for the thoughtful reply. I quoted above the relevant bits for my answer.

What we are trying to do is exactly what @nguyentankhtn mentioned: upload and delete files to our MinIO buckets freely while AWS S3 serves as a cloud backup/log/dump of all our files.

While I agree with your vision for the project, I feel like it comes very close to achieving that very useful functionality without getting there. `replicate` does it, but only between MinIO servers. `mirror` does it too, but it's a manual/scheduled process. `mirror --watch` was supposed to be the perfect solution, but we've just discussed it at length now as to why it doesn't work.

That leaves us looking for other solutions, from using MinIO event notifications + Webhook + API Gateway + Lambda, which brings its own set of pros and cons, to looking for another product altogether.

kannappanr:
Basically what you are looking for is a archive type of solution where any new object created gets copied over to AWS S3 and gets deleted whenever you wish, need not be right away

ravindk89:
>  I feel like it comes very close to achieving that very useful functionality without getting there.

The constant struggle of engineering :) Without scope constraints, any given feature can solve quite a few problems. But those scope constraints allow us to push the product forward without spending too much time on one thing. There's a balance to strike. 

If the result of this particular feature landing on the out-of-scope side of that balance is MinIO not working for you, that would be unfortunate - but you have to do what is best for your requirements.

harshavardhana:
--watch doesn't automatically imply `--remove` it needs to be specified 

>                         if mirrorURL.TargetContent != nil && (mj.opts.isRemove || mj.opts.activeActive) {

This is behaving as intended. 

This PR is changing the behavior that is not intended for us, closing this PR.

PR #3788, opened 2021-09-03
fix: subnet health spinners not working properly

anjalshireesh:
The "admin info" step was moved from being the first step to the last
step in minio at some point, but the sequence of spinners in mc was not
updated with this change. Because of this, the first spinner itself used
to get stuck for the whole duration of report generation.

Fixed by correcting the sequence of spinners to make them in sync with
the processing in minio.

Signed-off-by: Shireesh Anjal <shireesh@minio.io>


harshavardhana (approved):

vadmeste (approved):
LGTM

PR #3790, opened 2021-09-07
Add ilm restore command

vadmeste:
Add support of ilm restore command.

To test:
```
mc mirror --storage-class GLACIER /usr/bin/ s3/testbucket/
mc ilm restore -r s3/testbucket/
```

### DIFF
--- a/cmd/ilm-restore.go
+++ b/cmd/ilm-restore.go
@@ -1,0 +1,323 @@
+// Copyright (c) 2015-2021 MinIO, Inc.
+//
+// This file is part of MinIO Object Storage stack
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package cmd
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/minio/cli"
+	json "github.com/minio/colorjson"
+	"github.com/minio/mc/pkg/probe"
+)
+
+// ilm restore specific flags.
+var (
+	ilmRestoreFlags = []cli.Flag{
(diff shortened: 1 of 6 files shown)


cmd/ilm-restore.go
```diff
@@ -0,0 +1,323 @@
...
+		case err, ok := <-restoreReqStatus:
+			if !ok {
+				done = true
+				break
+			}
+			errorIf(err.Trace(), "Unable to send restore request.")
+			if err == nil {
+				sent++
+			}
+		case <-ticker.C:
```
krisis, line 245: why do we need the ticker here?

cmd/ilm-restore.go
```diff
@@ -0,0 +1,323 @@
...
+		case err, ok := <-restoreFinishedStatus:
+			if !ok {
+				done = true
+				break
+			}
+			errorIf(err.Trace(), "Unable to check for restore status")
+			if err == nil {
+				finished++
+			}
+		case <-ticker.C:
```
krisis, line 268: why do we need the ticker here?

cmd/ilm-restore.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+// Wait until an object which receives restore request is completely restored in the fast tier
+func waitRestoreObject(ctx context.Context, targetAlias, targetURL, versionID string) *probe.Error {
+	clnt, err := newClientFromAlias(targetAlias, targetURL)
+	if err != nil {
+		return err
+	}
+
+	for {
```
krisis, line 158: shouldn't we wait between successive Stat requests to fetch restore status?

poornas:
@vadmeste , restore API also allows specifying a select query and saving that output in a different bucket. Is that going to come later?

cmd/ilm-restore.go
vadmeste: > why do we need the ticker here?

Just to print status each one second.. the status shows progress with dots so the user feels something is in progress

cmd/ilm-restore.go
vadmeste: > shouldn't we wait between successive Stat requests to fetch restore status?

right, fixed

vadmeste:
> restore API also allows specifying a select query

Probably it will come later, not sure.

> saving that output in a different bucket.

Not fully sure why is this useful. Let's discuss this if you know how it can be useful.

krisis (approved):

harshavardhana (approved):

harshavardhana:
> > restore API also allows specifying a select query
> 
> Probably it will come later, not sure.
> 
> > saving that output in a different bucket.
> 
> Not fully sure why is this useful. Let's discuss this if you know how it can be useful.

If we add it, it can help test that part of the code.

PR #3793, opened 2021-09-08
Fix data race in mc admin subnet health

anjalshireesh:
The code that explicitly stops all spinners was not doing it in the same
sequence as the data returned by the server, resulting in a race
condition.

Fixed by correcting the sequence.


anjalshireesh:
This bug is fixed by another already under-review PR #3788. So closing this one.

Issue #3627, opened 2021-02-27
mcli should respect XDG basedirs for storing its configuration


ddevault:
**Is your feature request related to a problem? Please describe.**

mcli should not store its configuration in the `~/.*` namespace on Unix systems.

**Describe the solution you'd like**

mcli should respect the [XDG basedirs spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html), using `$XDG_CONFIG_HOME/mcli` if set, and falling back to `~/.config/mcli` if not. It can read `~/.mcli` as a last resort to avoid breaking the config for users who have already set it up, but it should not create new files there by default anymore.

klauspost:
This will require moving the config to not break existing installs on update.

vadmeste:
> This will require moving the config to not break existing installs on update.

Maybe we can do as a fallback better. I mean the last check will be ~/.mc/

ddevault:
Don't run bots like that, jeesh

renich:
phew... I'm also waiting for this one.

ddevault:
Don't run bots like that, jeesh

vadmeste:
I sent a PR for this. But I am not sure if we are going to merge it.

> Don't run bots like that, jeesh

We receive a lot of issues and we cannot process all of them. It makes sense that a bot helps us here :)

ddevault:
Those bots are an anti-feature which communicates a lack of investment from maintainers in their community. You should triage issues and close them when you know they're not a good idea, and otherwise leave them open. Having a lot of open issues is not a mark of shame.

harshavardhana:
> Those bots are an anti-feature which communicates a lack of investment from maintainers in their community. You should triage issues and close them when you know they're not a good idea, and otherwise leave them open. Having a lot of open issues is not a mark of shame.

How we manage the issues is our internal requirement, bots will run as per our requirement.

PR #3794, opened 2021-09-08
Follow XDG config dir spec in new mc installations

vadmeste:
config.json & certs directories will be in ~/.config/mc/ and share
downloads.json and uploads.json will be in ~/.local/share/mc/share/

This is draft because I am not sure if we want this.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -48,7 +48,7 @@ func getMcConfigDir() (string, *probe.Error) {
 	if e != nil {
 		return "", probe.NewError(e)
 	}
-	if runtime.GOOS == "windows" {
+	if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
 		return filepath.Join(homeDir, mcDirName()), nil
 	}
 	// Check if ~/.mc already exists, deprecated Sep 2021
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -48,13 +48,24 @@ func getMcConfigDir() (string, *probe.Error) {
 	if e != nil {
 		return "", probe.NewError(e)
 	}
-	configDir := filepath.Join(homeDir, defaultMCConfigDir())
-	return configDir, nil
+	if runtime.GOOS == "windows" {
+		return filepath.Join(homeDir, mcDirName()), nil
+	}
+	// Check if ~/.mc already exists, deprecated Sep 2021
+	oldMcConfigDir := filepath.Join(homeDir, "."+mcDirName())
+	if st, err := os.Stat(oldMcConfigDir); err == nil && st.IsDir() {
+		return oldMcConfigDir, nil
+	}
+	// Construct mc config dir path according to https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html
+	xdgPath := os.Getenv("XDG_CONFIG_HOME")
+	if xdgPath == "" {
+		xdgPath = filepath.Join(homeDir, ".config")
+	}
+	return filepath.Join(xdgPath, mcDirName()), nil
 }
 
-// Return default default mc config directory.
-// Generally you want to use getMcConfigDir which returns custom overrides.
-func defaultMCConfigDir() string {
+// Return mc directory name for config/data
+func mcDirName() string {
 	if runtime.GOOS == "windows" {
 		// For windows the path is slightly different
 		cmd := filepath.Base(os.Args[0])
@@ -63,7 +74,7 @@ func defaultMCConfigDir() string {
 		}
 		return fmt.Sprintf("%s\\", cmd)
 	}
-	return fmt.Sprintf(".%s/", filepath.Base(os.Args[0]))
+	return fmt.Sprintf("%s/", filepath.Base(os.Args[0]))
 }
 
 // mustGetMcConfigDir - construct MinIO Client config folder or fail
--- a/cmd/share.go
+++ b/cmd/share.go
@@ -115,7 +115,7 @@ func getShareDir() (string, *probe.Error) {
 	if e != nil {
 		return "", probe.NewError(e)
 	}
-	if runtime.GOOS == "windows" {
+	if runtime.GOOS == "windows" || runtime.GOOS == "darwin" {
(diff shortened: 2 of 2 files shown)


ddevault:
>This is draft because I am not sure if we want this.

What's the cause for hesitation?

cmd/config.go
```diff
@@ -48,13 +48,24 @@ func getMcConfigDir() (string, *probe.Error) {
 	if e != nil {
 		return "", probe.NewError(e)
 	}
-	configDir := filepath.Join(homeDir, defaultMCConfigDir())
-	return configDir, nil
+	if runtime.GOOS == "windows" {
+		return filepath.Join(homeDir, mcDirName()), nil
+	}
+	// Check if ~/.mc already exists, deprecated Sep 2021
+	oldMcConfigDir := filepath.Join(homeDir, mcDirName())
```
klauspost, line 55: `mcDirName` no longer returns the same.

cmd/config.go
```diff
@@ -48,13 +48,24 @@ func getMcConfigDir() (string, *probe.Error) {
...
+	if st, err := os.Stat(oldMcConfigDir); err == nil && st.IsDir() {
+		return oldMcConfigDir, nil
+	}
+	// Construct mc config dir path according to https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html
+	xdgPath := os.Getenv("XDG_CONFIG_HOME")
+	if xdgPath == "" {
+		xdgPath = filepath.Join(homeDir, ".config")
```
klauspost, line 62: Just return the previous path `~/.mc`. No need to have even more places to look.

cmd/share.go
```diff
@@ -105,13 +107,30 @@ func shareSetColor() {
...
+	// Check if ~/.mc already exists, deprecated Sep 2021
+	oldMcConfigDir := filepath.Join(homeDir, "."+mcDirName())
+	if st, err := os.Stat(oldMcConfigDir); err == nil && st.IsDir() {
+		return filepath.Join(oldMcConfigDir, globalSharedURLsDataDir), nil
+	}
+
+	// Construct mc config dir path according to https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html
+	xdgPath := os.Getenv("XDG_DATA_HOME")
+	if xdgPath == "" {
+		xdgPath = filepath.Join(homeDir, ".local/share")
```
klauspost, line 131: Use previous location. No need for a third.

cmd/config.go
vadmeste: > Just return the previous path `~/.mc`. No need to have even more places to look.

I am following the specification here, if XDG_CONFIG_HOME is not defined, then you need to use ~/.config/mc/

ddevault: Using ~/.config/mc is the correct behavior in this case. The only time ~/.mc should be used is if the config file cannot be found at a better location but exists in ~/.mc, i.e. for installs which pre-date this change.

klauspost: I don't really care about the specification, I just don't want the config to potentially be in 3 different places. 2 is confusing enough.

ddevault: No one else really cares about the spec, either, but to do the right thing you need to set the fallback to ~/.config/mc for users who don't have the environment variable set. Most users do not have the environment variable set and the problem will not be resolved to their satisfaction without this logic.

harshavardhana:
> config.json & certs directories will be in ~/.config/mc/ and share
> downloads.json and uploads.json will be in ~/.local/share/mc/share/

@vadmeste let's not do this as the default unless XDG is available and set - this is not true for macOS, BSD systems. 

XDG is a wider standard than just this you have to follow the following basics to keep properly support for XDG

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

- There is a single base directory relative to which user-specific data files should be written. This directory is defined by the environment variable $XDG_DATA_HOME.

- There is a single base directory relative to which user-specific configuration files should be written. This directory is defined by the environment variable $XDG_CONFIG_HOME.

- There is a single base directory relative to which user-specific state data should be written. This directory is defined by the environment variable $XDG_STATE_HOME.

- There is a single base directory relative to which user-specific executable files may be written.

- There is a set of preference ordered base directories relative to which data files should be searched. This set of directories is defined by the environment variable $XDG_DATA_DIRS.

- There is a set of preference ordered base directories relative to which configuration files should be searched. This set of directories is defined by the environment variable $XDG_CONFIG_DIRS.

- There is a single base directory relative to which user-specific non-essential (cached) data should be written. This directory is defined by the environment variable $XDG_CACHE_HOME.

- There is a single base directory relative to which user-specific runtime files and other file objects should be placed. This directory is defined by the environment variable $XDG_RUNTIME_DIR.

Our code needs to honor these basics, look for each of these ENVs and apply them appropriately. 

Also NOTE: using `.local/mc` and `.config/mc` will conflict with `midnight-commander` so you need to be careful about choosing the same config path.

vadmeste:
> Also NOTE: using `.local/mc` and `.config/mc` will conflict with `midnight-commander` so you need to be careful about choosing the same config path.

Yes, that's why I am not too keen in supporting XDG here. This looks like only important if we want to add mc to GNU/Linux distributions.

ddevault:
The binary also conflicts with midnight-commander, which predates minio. An executable name collision is much worse than a config directory collision, why should it apply to this change but not the other?

harshavardhana:
> The binary also conflicts with midnight-commander, which predates minio. An executable name collision is much worse than a config directory collision, why should it apply to this change but not the other?

https://github.com/minio/mc/blob/master/CONFLICT.md

harshavardhana:
> > Also NOTE: using `.local/mc` and `.config/mc` will conflict with `midnight-commander` so you need to be careful about choosing the same config path.
> 
> Yes, that's why I am not too keen in supporting XDG here. This looks like only important if we want to add mc to GNU/Linux distributions.

Let's not do this for now then - this is not an important requirement for now. We can add it only if it solves some real needs.

ddevault:
What nonsense.

klauspost:
Closed as per #3627

PR #3796, opened 2021-09-14
rb: Fix side wide removal of an alias

vadmeste:
rb --force --dangerous should remove all buckets under an alias. This
PR fixes a recent regression.


klauspost:
Can we add a regression test?

harshavardhana (approved):

.github/workflows/go.yml
```diff
@@ -35,6 +35,10 @@ jobs:
         env:
           GO111MODULE: on
         run: |
+          wget https://dl.min.io/server/minio/release/linux-amd64/minio && chmod +x minio
+          mkdir -p ~/.minio/certs/ && openssl req -x509 -newkey rsa:1024 -keyout ~/.minio/certs/private.key -out ~/.minio/certs/public.crt -days 7 -nodes -subj '/CN=localhost'
+          ./minio server /tmp/test-xl/{1...4}/ & sleep 10
+          export SERVER_ENDPOINT=localhost:9000 ACCESS_KEY=minioadmin SECRET_KEY=minioadmin ENABLE_HTTPS=1
```
harshavardhana, line 41: Shouldn't this export be in multiple lines?

.github/workflows/go.yml
```diff
@@ -35,6 +35,10 @@ jobs:
         env:
           GO111MODULE: on
         run: |
+          wget https://dl.min.io/server/minio/release/linux-amd64/minio && chmod +x minio
+          mkdir -p ~/.minio/certs/ && openssl req -x509 -newkey rsa:1024 -keyout ~/.minio/certs/private.key -out ~/.minio/certs/public.crt -days 7 -nodes -subj '/CN=localhost'
```
harshavardhana, line 39: You can get the certs from minio-go repo if necessary for localhost keys @vadmeste Or commit the certs into mc repo.

functional-tests.sh
```diff
@@ -456,7 +488,7 @@ function test_presigned_get_object()
     out=$("${MC_CMD[@]}" --json share download "${SERVER_ALIAS}/${BUCKET_NAME}/${object_name}")
     assert_success "$start_time" "${FUNCNAME[0]}" show_on_failure $? "unable to get presigned get object url"
     download_url=$(echo "$out" | jq -r .share)
-    curl -sSg --output "${object_name}.downloaded" -X GET "$download_url"
+    curl -ksSg --output "${object_name}.downloaded" -X GET "$download_url"
```
harshavardhana, line 491: Let's just trust the system?

.github/workflows/go.yml
harshavardhana: Or better use mkcert in batch mode to generate and trust the certs

functional-tests.sh
harshavardhana: You can avoid curl -k once certs trusted with mkcert

harshavardhana (changes requested):

harshavardhana (approved):

klauspost (approved):
LGTM 👍🏼

PR #3804, opened 2021-09-21
Order mc admin info output by endpoint

vadmeste:
To make easy for users to navigate through the list of servers.


klauspost (approved):
LGTM

harshavardhana (approved):

PR #3805, opened 2021-09-22
Disallow updating azure tier account-name

krisis:


harshavardhana (changes requested):
what was the side affect of previous feature?

krisis:
> what was the side affect of previous feature?

This is best explained using an example. Imagine a setup where a user configures an azure type tier and configures bucket ILM policies to transition objects to this tier. Over time we would have objects transitioned to this tier which is backed by a storage account, let's call this `storageacct1`. If this user decides to update the storage account to `storageacct2` then objects that were transitioned earlier would be looked up in `storageacct2` but they remain in `storageacct1`. This problem doesn't exist with other tier types, namely s3 and gcs. In each of these the credentials correspond to an IAM user and doesn't change the part of namespace this tier is referring to.

harshavardhana (approved):

kannappanr (approved):

PR #3807, opened 2021-09-24
verifiers: Fix gofmt warning

vadmeste:
gofmt is generating new warning messages now. Fix them.


harshavardhana:
You should add both.

klauspost:
I replaced it with #3811 which also updates more stuff.

PR #3810, opened 2021-09-27
Add new admin-cluster subcommand

donatello:
This change adds support for enabling cluster replication


cmd/admin-cluster-replicate-add.go
```diff
@@ -0,0 +1,111 @@
...
+	fatalIf(err, "Unable to initialize admin connection.")
+
+	pc := make([]madmin.PeerCluster, 0, len(ctx.Args()))
+	for _, clusterName := range ctx.Args() {
+		admClient, err := newAdminClient(clusterName)
+		fatalIf(err, "unable to initialize admin connection")
+
+		ak, sk := admClient.GetAccessAndSecretKey()
+		info, e := admClient.ServerInfo(globalContext)
+		if e != nil {
```
vadmeste, line 94: You can do fatalIf(probe.NewError(e)...) without checking if e is nil or not

vadmeste (commented):
Few comments.

Also, unit tests will fail because you need to add the new commands in auto-complete.go

cmd/admin-cluster-replicate-add.go
donatello: 👍🏾

donatello:
> Also, unit tests will fail because you need to add the new commands in auto-complete.go

Fixed.

PR #3814, opened 2021-09-29
admin info: Show 'OK' in yellow when offline networks/disks

vadmeste:
In mc admin info output, it is better to show 'OK' status in yellow and
not in green when there is a failed disk or a failed network link so it
is easier for users to detect something wrong.

### DIFF
--- a/cmd/admin-info.go
+++ b/cmd/admin-info.go
@@ -175,16 +175,20 @@ func (u clusterStruct) String() (msg string) {
 		msg += fmt.Sprintf("   Version: %s\n", version)
 
 		// Network info, only available for non-FS types
-		var connectionAlive int
-		totalNodes := strconv.Itoa(len(srv.Network))
+		connectionAlive := 0
+		totalNodes := len(srv.Network)
 		if srv.Network != nil {
 			for _, v := range srv.Network {
 				if v == "online" {
 					connectionAlive++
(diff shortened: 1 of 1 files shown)


klauspost (approved):
LGTM

cmd/admin-info.go
```diff
@@ -175,23 +175,28 @@ func (u clusterStruct) String() (msg string) {
...
+			displayNwInfo := strconv.Itoa(connectionAlive) + "/" + strconv.Itoa(totalNodes)
+			msg += fmt.Sprintf("   Network: %s %s\n", displayNwInfo, console.Colorize(clr, "OK "))
 		}
 
 		if backendType != "FS" {
 			// Info about drives on a server, only available for non-FS types
 			var OffDisks int
 			var OnDisks int
 			var dispNoOfDisks string
+			fmt.Println(len(srv.Disks))
```
harshavardhana, line 199: is this expected log? @vadmeste

harshavardhana (changes requested):

cmd/admin-info.go
vadmeste, line 199:
```suggestion
			fmt.Println(len(srv.Disks))

```

harshavardhana (approved):

PR #3819, opened 2021-10-01
ilm: Correctly show tags in tabular info when this non AND tag

vadmeste:
Tag can be set directly under Filter tag and not always under
Filter.And in the lifecycle document. In that case, the tag won't be
properly showed in the tabular info of the ILM listing. This PR fixes
it.


klauspost (approved):
LGTM

krisis (approved):

PR #3820, opened 2021-10-04
health: perform only lightweight tests by default

anjalshireesh:
Perform only lightweight tests by default, and include the long running
tests only if specified through a `--full` flag in the `admin subnet
health` command.


harshavardhana (approved):

klauspost (approved):
LGTM

PR #3822, opened 2021-10-05
fix: rb --force was not working properly

harshavardhana:


vadmeste:
@harshavardhana it is not very clear what wasn't working, can you give an example command ?

PR #3839, opened 2021-10-19
Update go.mod to use latest version of madmin-go

anjalshireesh:
So that the TLS config related data gets added in the health report.


klauspost:
Can't we do a proper madmin release?

PR #3857, opened 2021-11-03
admin-trace: Add missing asterisk to --path example in usage

### LINKED ISSUE #3856: Usage for mc admin trace command's --path switch is misleading
## Expected behavior

Running `mc admin trace` will output the following usage:

```bash
EXAMPLES:
  [...]

  4. Show console trace for a specific path
    $ mc admin trace --path my-bucket/my-prefix/ myminio

  [...]
```

When I run something like the given command I expect that any requests made to `myminio/my-bucket/my-prefix/` will be recorded and logged to Stdout.

## Actual behavior

When running the exact same command, requests made to `myminio/my-bucket/my-prefix/` are not shown. What does work is something like the following command: `mc admin trace --path my-bucket/my-prefix/* myminio`. From my experience the asterisk is not optional.

I have no idea what the actual behavior should be. Should it work as documented or should it work using an asterisk? Either way the documentation or implementation should be fixed. This could also be a regression depending on what framework (if any) is used to filter the requests by path.

1. Start tracing with `mc admin trace --path my-bucket/my-prefix/ myminio`
2. Open another shell session.
3. `mc mb myminio/my-bucket`
4. `mc cp --recursive <path-to-local-folder> myminio/my-bucket/my-prefix/`
5. Repeat the same process using `mc admin trace --path my-bucket/my-prefix/* myminio`

## mc --version
- mc version RELEASE.2021-10-07T04-19-58Z

## System information
Linux IT-FR-NB-RS-19 5.10.16.3-microsoft-standard-WSL2 Nr.1 SMP Fri Apr 2 22:23:49 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

rluetzner:
Limiting tracing to a specific path requires an asterisk at the end. I
don't know if this is a regression or the example was always wrong. I
only ever figured this out by trial and error.

This fixes #3856.

### DIFF
--- a/cmd/admin-trace.go
+++ b/cmd/admin-trace.go
@@ -104,7 +104,7 @@ EXAMPLES:
     {{.Prompt}} {{.HelpName}} -v --status-code 503 myminio
 
   4. Show console trace for a specific path
-    {{.Prompt}} {{.HelpName}} --path my-bucket/my-prefix/ myminio
+    {{.Prompt}} {{.HelpName}} --path my-bucket/my-prefix/* myminio
 
   5. Show console trace for requests with '404' and '503' status code
     {{.Prompt}} {{.HelpName}} --status-code 404 --status-code 503 myminio


harshavardhana (approved):

vadmeste (approved):
LGTM, thanks!

PR #3858, opened 2021-11-03
Fix `subnet health` crash on passing invalid flag

anjalshireesh:
The custom `String()` function written for the `HealthDataTypeFlag` was
resulting in a crash as it did not contain `\t` character on which the
usage gets split when printing the error message.

Fixed by reusing the `FlagStringer` from cli package, which is used by
`StringFlag.String()` as well.


vadmeste (approved):
LGTM

harshavardhana (approved):

PR #3859, opened 2021-11-03
Fix usage validation in `mc admin service restart`

anjalshireesh:
The command needs exactly one argument, but was allowing two and
ignoring the second if passed. Fixed the validation.


vadmeste (approved):
LGTM

harshavardhana (approved):

PR #3861, opened 2021-11-11
Better message on reset of dynamic config

anjalshireesh:
When a dynamic config is reset, the minio api response has `true` as
value of the header `x-minio-config-applied`. In such cases, mc should
not show the message asking the user to restart `minio`.


harshavardhana (approved):

klauspost (approved):
LGTM

PR #3869, opened 2021-11-15
feat: detect starting from windows explorer

harshavardhana:
Windows users often click on the binary without
knowing `mc` is a command-line tool and should be
run from a terminal. Throw a message to guide them
on what to do.


klauspost (approved):

vadmeste (approved):
LGTM

PR #3870, opened 2021-11-16
Fix health upload to subnet failing

anjalshireesh:
The writer needs to be closed before creating the request. Otherwise the
call to subnet api fails with error like:

HTTP/1.x transport connection broken: http: ContentLength=6235 with Body length 6303.


klauspost (approved):

vadmeste (approved):
LGTM

harshavardhana (approved):

PR #3871, opened 2021-11-17
honor cgroup memory limits in parallel-manager

harshavardhana:


cmd/parallel-manager.go
```diff
@@ -220,22 +222,52 @@ func (p *ParallelManager) stopAndWait() {
...
+		available = cgroupLimit
+	}
+
+	// No limit set, It's the highest positive signed 64-bit
+	// integer (2^63-1), rounded down to multiples of 4096 (2^12),
+	// the most common page size on x86 systems - for cgroup_limits.
+	if available != 9223372036854771712 {
+		// This means cgroup memory limit is configured.
+		return
+	} // no-limit set proceed to set the limits based on virtual memory.
```
klauspost, line 248:
```suggestion
	if runtime.GOOS == "linux" {
		buf, err := ioutil.ReadFile(cgroupLimitFile)
		if err != nil {
			return
		}
		cgroupLimit, err := strconv.ParseUint(string(buf), 10, 64)
		if err != nil {
			return 0
		}
		// No limit set, It's the highest positive signed 64-bit
		// integer (2^63-1), rounded down to multiples of 4096 (2^12),
		// the most common page size on x86 systems - for cgroup_limits.
		if cgroupLimit!= 9223372036854771712 {
			// This means cgroup memory limit is configured.
			return cgroupLimit
		} // no-limit set proceed to set the limits based on virtual memory.		
	}
```

klauspost, line 248:
```suggestion
	for runtime.GOOS == "linux" {
		buf, err := ioutil.ReadFile(cgroupLimitFile)
		if err != nil {
			break
		}
		cgroupLimit, err := strconv.ParseUint(string(buf), 10, 64)
		if err != nil {
			break
		}
		// No limit set, It's the highest positive signed 64-bit
		// integer (2^63-1), rounded down to multiples of 4096 (2^12),
		// the most common page size on x86 systems - for cgroup_limits.
		if cgroupLimit == 9223372036854771712 {
			// no-limit set proceed to set the limits based on virtual memory.
			break
		}
		return cgroupLimit
	}
```

klauspost (changes requested):

klauspost (approved):

PR #3874, opened 2021-11-26
[WIP] Add --check-etag in cat command

vadmeste:
Waiting for the command approval.

This doesn't work with server encrypted objects since ETag is not close to the md5sum here.

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -809,6 +809,7 @@ func (c *S3Client) Get(ctx context.Context, opts GetOptions) (io.ReadCloser, *pr
 		minio.GetObjectOptions{
 			ServerSideEncryption: opts.SSE,
 			VersionID:            opts.VersionID,
+			PartNumber:           opts.PartNumber,
 		})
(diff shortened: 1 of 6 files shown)


harshavardhana (changes requested):
We can also do this with `mc find` @vadmeste as a feature - like find me all corrupted objects.

vadmeste:
> We can also do this with `mc find` @vadmeste as a feature - like find me all corrupted objects.

We can do this way:
```
mc find myminio/testbucket/ -exec "mc cat --check-etag {}" >/dev/null
```

I don't want to push this feature too much since --check-etag won't work with server side encrypted objects.

abperiasamy:
Checking etag on the client side is an expensive operation.

vadmeste:
> Checking etag on the client side is an expensive operation.

Yes, this is a feature more for debugging purposes.

harshavardhana:
Let's close this, there is no scope for this at the moment - s3-check-md5 served its purpose.

PR #3875, opened 2021-11-29
speedtest: Ask the user to increase duration with empty result.

vadmeste:
Sometimes, with slow machines, speedtest will not able to accomplish
any PUT/GET operations though no errors is met. In that case, this
commit will show an adequate error message asking the user to increase
the --duration flag.


klauspost (approved):
LGTM

PR #3881, opened 2021-12-06
tags: Show how to assign two tags to an object

vadmeste:
mc cp --tags is able to assign more than one tag to an object, enhance
the example to show that.


klauspost (approved):
LGTM

harshavardhana (approved):

PR #3882, opened 2021-12-06
Rewind mode list directory bug fix

adfost:
Adds an extra parameter, IsDir to ListObjects, because folders would never end with '/' in rewind mode, causing them to be misinterpreted as files.

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -1699,11 +1699,9 @@ func (c *S3Client) List(ctx context.Context, opts ListOptions) <-chan *ClientCon
 // it falls back to the regular listing if not.
 func (c *S3Client) versionedList(ctx context.Context, contentCh chan *ClientContent, opts ListOptions) {
 	b, o := c.url2BucketAndObject()
-
 	if opts.IsDir {
 		o += "/"
 	}
-	
 	switch {
 	case b == "" && o == "":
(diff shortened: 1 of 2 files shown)


cmd/client-s3.go
```diff
@@ -1699,6 +1699,9 @@ func (c *S3Client) List(ctx context.Context, opts ListOptions) <-chan *ClientCon
 // it falls back to the regular listing if not.
 func (c *S3Client) versionedList(ctx context.Context, contentCh chan *ClientContent, opts ListOptions) {
 	b, o := c.url2BucketAndObject()
+	if opts.IsDir {
+		o += "/"
+	}
```
harshavardhana, line 1704: This kind of manual append is problematic @adfost is this not possible for the console to do this as necessary? - since you already know that `IsDir: true`

harshavardhana (changes requested):

cmd/client-s3.go
adfost: No it isn't possible, regardless of whether it is a directory it treats it like a file. It is necessary for mc to know that it is a directory, not a file.

dvaldivia: he's right, we don't need to do this, we can simply pass this on the prefix as we access the file

PR #3893, opened 2021-12-14
ux: Fix admin heal display with multiple zones

vadmeste:
mc admin heal was showing additional unnecessary entries of nodes with
empty disks listing when that node does not belong to the current pool
number. A node only belongs to one pool, so remove the extra entries.


harshavardhana:
@vadmeste linter errors

harshavardhana (approved):

harshavardhana (approved):

PR #3894, opened 2021-12-14
mirror: Sort bucket names with slash for site wide scenarios

vadmeste:
mc diff/mirror command fails to compare source & target when doing
an S3 site-wide listing.

The reason is that client-s3.List(), when doing a site wide listing,
lists buckets first and list objects inside each bucket with the order
returned by S3 ListBuckets API.

However this can produce a non properly sorted listing output,
e.g.:

```
$ mc ls -r myminio/
[2021-12-14 15:58:27 CET] 227KiB bucketname/object1
[2021-12-14 15:58:24 CET] 227KiB bucketname-0/object2
```

Though it should be
```
$ mc ls -r myminio/
[2021-12-14 15:58:24 CET] 227KiB bucketname-0/object2
[2021-12-14 15:58:27 CET] 227KiB bucketname/object1
```

because '/' has a higher ASCII value than '-'

This commit fixes the behavior by always sorting bucket names with
a fake trailing slash when doing a site wide recursive listing.

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -33,6 +33,7 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


harshavardhana:
Is this PR ready @vadmeste ?

harshavardhana (approved):

PR #3901, opened 2021-12-22
du: Fix double counting of objects

klauspost:
`mc du play` would count objects twice since it was doing a recursive listing AND traversing into folders.

Compare total to `mc du play --depth=2` and `mc du play --recursive`.


kannappanr (approved):
LGTM

vadmeste (approved):
LGTM

PR #3903, opened 2021-12-26
remove retry in difference() comparator

harshavardhana:
it is potentially harmful when retrying
a lot of buckets where in this might
trigger a retry - continous Listing()

however this retry is not needed as the
callers restart as needed when --watch
or --active-active are used.


vadmeste (approved):
LGTM

PR #3914, opened 2022-01-07
Add option to `mc admin policy info` to save policy JSON to file

donatello:


harshavardhana (approved):

cmd/admin-policy-info.go
```diff
@@ -93,6 +105,18 @@ func mainAdminPolicyInfo(ctx *cli.Context) error {
...
+	if policyFile != "" {
+		f, err := os.Create(policyFile)
+		if err != nil {
+			fatalIf(probe.NewError(err).Trace(args...), "Could not open given policy file")
+		}
+		_, err = f.Write(pinfo.Policy)
+		if err != nil {
+			fatalIf(probe.NewError(err).Trace(args...), "Could not write to given policy file")
+		}
+	}
```
harshavardhana, line 118: can you support `-` to write to os.Stdout? @donatello

harshavardhana (changes requested):

cmd/admin-policy-info.go
```diff
@@ -18,25 +18,34 @@
...
 	"github.com/fatih/color"
 	"github.com/minio/cli"
 	"github.com/minio/madmin-go"
 	"github.com/minio/mc/pkg/probe"
 	"github.com/minio/pkg/console"
 )
 
+var policyInfoFlags = []cli.Flag{
+	cli.StringFlag{
+		Name:  "policy-file, f",
```
vadmeste, line 32: `output-file, o` or `save, s` are better in my opinion.

cmd/admin-policy-info.go
```diff
@@ -18,25 +18,34 @@
...
+		Usage: "additionally (over-)write policy JSON to given file",
```
vadmeste, line 33: I think maybe let's avoid printing in the stdout when this flag is passed

vadmeste (commented):
Some comments

cmd/admin-policy-info.go
donatello: This is an option to save just policy definition, not the metadata that is output to stdout

cmd/admin-policy-info.go
harshavardhana: @donatello ^^

harshavardhana (approved):

poornas (approved):

PR #3915, opened 2022-01-08
Allow storageclass to be specified on mc ls

Willena:
The current version of the ls command omits GLACIER objects. 
Sometimes you need to get a list of those files that are located under another storage class. 

This commit allows filtering a given storage class and only list this class, or list all objects in all storage classes. The default value is STANDARD. 

Usage example: 
* List only GLACIER objects: `mc ls -r -sc GLACIER myAlias/ `
* List only STANDARD objects (default): `mc ls -r myAlias/ `
* List all objects in all storage classes: `mc ls -r -sc '*' myAlias/ `

output example:
```
[2022-01-05 21:17:32 CET] 237KiB STANDARD gu-952f-8404d1f1a068/bl78af97a0f4de9155e56d709e70e81db820eb90f397b39d547f9f9fa0ff84b.blk
[2022-01-05 21:17:49 CET]  19KiB STANDARD gu-952f-8404d1f1a068/bla1889f63d333c93f14f9930dd52526db68ea0073a8336316f24ac2486f98d.blk
[2022-01-05 21:18:06 CET] 160KiB STANDARD gu-952f-8404d1f1a068/bla05d10933062879e669c465f57e5445fbec4f7ed8a4c9690d139c89a05867.blk
[2022-01-08 10:26:41 CET]   635B STANDARD gu-952f-8404d1f1a068/data.cfg
[2022-01-08 10:45:40 CET]  15MiB GLACIER  gu-952f-8404d1f1a068/bkp.blk
```

### DIFF
--- a/cmd/ls-main.go
+++ b/cmd/ls-main.go
@@ -53,6 +53,10 @@ var (
 			Name:  "summarize",
 			Usage: "display summary information (number of objects, total size)",
 		},
+		cli.StringFlag{
+			Name:  "storage-class, sc",
+			Usage: "filter to specified storage class",
+		},
 	}
 )
 
@@ -102,6 +106,12 @@ EXAMPLES:
 
   9. List all objects on mybucket, summarize the number of objects and total size.
      {{.Prompt}} {{.HelpName}} --summarize s3/mybucket/
+
+  10. List all objects on mybucket, for all storageClass
+     {{.Prompt}} {{.HelpName}} --storage-classes '*' s3/mybucket
+  
+  11. List all objects on mybucket, for the GLACIER storage class
+     {{.Prompt}} {{.HelpName}} --storage-classes 'GLACIER' s3/mybucket 
 `,
 }
 
@@ -147,7 +157,7 @@ func parseRewindFlag(rewind string) (timeRef time.Time) {
 }
 
 // checkListSyntax - validate all the passed arguments
-func checkListSyntax(ctx context.Context, cliCtx *cli.Context) ([]string, bool, bool, bool, time.Time, bool) {
+func checkListSyntax(ctx context.Context, cliCtx *cli.Context) ([]string, bool, bool, bool, time.Time, bool, string) {
 	args := cliCtx.Args()
 	if !cliCtx.Args().Present() {
 		args = []string{"."}
@@ -168,7 +178,12 @@ func checkListSyntax(ctx context.Context, cliCtx *cli.Context) ([]string, bool, 
 		timeRef = time.Now().UTC()
 	}
 
-	return args, isRecursive, isIncomplete, isSummary, timeRef, withOlderVersions
+	storageClasss := cliCtx.String("storage-class")
+	if len(storageClasss) == 0 {
+		storageClasss = "STANDARD"
+	}
+
+	return args, isRecursive, isIncomplete, isSummary, timeRef, withOlderVersions, storageClasss
 }
 
 // mainList - is a handler for mc ls command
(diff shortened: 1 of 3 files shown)


harshavardhana (changes requested):
I think instead of providing storage-class we can bring in an option called `-a` to list all hidden files like Unix.

In case of object storage this prints storage class as  well.

Willena:
Not a bad idea. I'll add to -a option. 
But I do like the fact that we can also specify the storageClass we want to see.
It seems like AWS for example (https://aws.amazon.com/s3/storage-classes/?nc1=h_ls#Performance_across_the_S3_Storage_Classes) provides more than two (S3 Standard, S3 Intelligent-Tiering*, S3 Standard-IA, S3 One Zone-IA, S3 Glacier Instant Retrieval, S3 Glacier Flexible Retrieval, S3 Glacier, Deep Archive)  storageClass types. It might still be useful to filter a specific one.  

@harshavardhana  What do you think about it ?

harshavardhana:
> Not a bad idea. I'll add to -a option. 
> But I do like the fact that we can also specify the storageClass we want to see.
> It seems like AWS for example (https://aws.amazon.com/s3/storage-classes/?nc1=h_ls#Performance_across_the_S3_Storage_Classes) provides more than two (S3 Standard, S3 Intelligent-Tiering*, S3 Standard-IA, S3 One Zone-IA, S3 Glacier Instant Retrieval, S3 Glacier Flexible Retrieval, S3 Glacier, Deep Archive)  storageClass types. It might still be useful to filter a specific one.  
> 
> @harshavardhana  What do you think about it ? 

Yes buts not really useful to filter in this manner since Listing is slow and it may never provide any output if you incorrectly provide the storage-class. This in-turn might be assumed as listing is not working.

The best way is print everything -a and use grep instead to filter what you want.

There are many ways to do it - trying to see what makes sense for our command line. 

Note: mc ls is used for filesystem folders and files as well which may provide empty results here with storage-class.

I will circle back later on what makes sense let me think about it more.

Willena:
> Listing is slow and it may never provide any output if you incorrectly provide the storage-class. This in-turn might be assumed as listing is not working

Totally true, I agree with that. That exactly what could happen if the storage class provided is invalid or not objects are of this storage class.

> The best way is print everything -a and use grep instead to filter what you want.

Didn't thought about it...

> Note: mc ls is used for filesystem folders and files as well which may provide empty results here with storage-class.

As implemented in the PR the storage class filter only applies when a storage class is present ( != "" )

vadmeste:
Why not simply list GLACIER objects in `mc ls` without any flags but still avoid them in mc cp/mirror (as it is now) ?

harshavardhana:
> Why not simply list GLACIER objects in `mc ls` without any flags but still avoid them in mc cp/mirror (as it is now) ?

We can change the format and print always, and keep filtering for specific storage classes if needed.

Willena:
> We can change the format and print always, and keep filtering for specific storage classes if needed.

I applied a small change to have the described behavior. 
`mc ls` will always print the storageClass if available. 
If `-sc <STORAGECLASS>` is provided, then filter the result looking only for `<STORAGECLASS>`.

harshavardhana:
> I applied a small change to have the described behavior.
> `mc ls` will always print the storageClass if available.
> If `-sc <STORAGECLASS>` is provided, then filter the result looking only for `<STORAGECLASS>`.

Looks great @Willena

harshavardhana (approved):

cmd/ls-main.go
```diff
@@ -102,6 +106,12 @@ EXAMPLES:
 
   9. List all objects on mybucket, summarize the number of objects and total size.
      {{.Prompt}} {{.HelpName}} --summarize s3/mybucket/
+
+  10. List all objects on mybucket, for all storageClass
+     {{.Prompt}} {{.HelpName}} --storage-classes '*' s3/mybucket
```
harshavardhana, line 111:
```suggestion
     {{.Prompt}} {{.HelpName}} --storage-classes '*' s3/mybucket
```

This is default - so don't know if this needs an example @Willena

harshavardhana (changes requested):

cmd/ls-main.go
Willena: Fixed !

harshavardhana (approved):

PR #3917, opened 2022-01-09
sufficiently conlorize svacct command outputs

### LINKED ISSUE #3803: svcacct: Colorize output and make it consistent with other cmds
svcacct output is not consistent with other mc admin user output. 

We should do it.

harshavardhana:
also print help when no args passed, instead of asking
for `--help`

fixes #3803


kannappanr (approved):
LGTM

vadmeste (approved):
LGTM

PR #3919, opened 2022-01-14
Add `admin replicate remove` command to allow removal of sites

poornas:


cmd/admin-replicate-remove.go
```diff
@@ -0,0 +1,113 @@
...
+		argsNr := len(ctx.Args())
+		if ctx.IsSet("all") && argsNr > 1 {
+			fatalIf(errInvalidArgument().Trace(ctx.Args().Tail()...),
+				"")
+		}
+		if argsNr < 2 && !ctx.IsSet("all") {
+			fatalIf(errInvalidArgument().Trace(ctx.Args().Tail()...),
+				"Need at least two arguments to remove command.")
+		}
+		if !ctx.IsSet("force") {
```
klauspost, line 89: Always having to add `--force` makes it lose its meaning.

To be honest I don't really see a need for it. It is not destructive. Sure, you lose the config, but no data is ever in jeopardy. When you specify `--all` it can't really be a surprise it removes all.

poornas: With `remove`, the site that is removed cannot rejoin site replication with `mc admin replicate add` because there is a constraint requiring the clusters to be empty at the time of joining site replication. In most situations this would not be the case, hence the need for --force.  

There is a plan to add suspend|resume as a alternate where the intent is just to stop participation in site replication temporarily

harshavardhana: Force is more like a confirmation without throwing users y/n - questions. Since they need to mean it by typing `--force`.  But perhaps in future we can add "suspend/resume" that happens without --force, it means like a `soft remove` instead of `hard remove`.

cmd/admin-replicate-remove.go
```diff
@@ -0,0 +1,114 @@
...
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}
+
+EXAMPLES:
+  1. Remove site replication for all sites:
+     {{.Prompt}} {{.HelpName}} minio2 --all --force
+
+  2. Remove site replication for site with site names alpha, baker from active cluster minio2:
+     {{.Prompt}} {{.HelpName}} minio2 alpha baker
```
harshavardhana, line 62: This needs to have `--force`

cmd/admin-replicate-remove.go
```diff
@@ -0,0 +1,114 @@
...
+		argsNr := len(ctx.Args())
+		if ctx.IsSet("all") && argsNr > 1 {
+			fatalIf(errInvalidArgument().Trace(ctx.Args().Tail()...),
+				"")
+		}
+		if argsNr < 2 && !ctx.IsSet("all") {
+			fatalIf(errInvalidArgument().Trace(ctx.Args().Tail()...),
+				"Need at least two arguments to remove command.")
+		}
+		if !ctx.IsSet("force") {
```
harshavardhana, line 90: input validation should be in a function to follow most commands in `mc` style as `checkAdminReplicationRemoveStatusSyntax()`  function name.

cmd/admin-replicate-remove.go
```diff
@@ -0,0 +1,114 @@
...
+type srRemoveStatus madmin.ReplicateRemoveStatus
+
+func (i srRemoveStatus) JSON() string {
+	ds, e := json.MarshalIndent(madmin.ReplicateRemoveStatus(i), "", " ")
+	fatalIf(probe.NewError(e), "Unable to marshal into JSON.")
+	return string(ds)
+}
+
+func (i srRemoveStatus) String() string {
+	return console.Colorize("UserMessage", i.Status)
```
harshavardhana, line 75:
```suggestion
	return console.Colorize("UserMessage", i.Status)
```

We should word it like what all sites were removed?

harshavardhana (changes requested):

harshavardhana:
PTAL @poornas

klauspost (approved):
Nothing that isn't already mentioned 👍🏼

cmd/admin-replicate-remove.go
poornas: the Status here is sent from the server and will have more detailed info on which sites were not removed, if any

harshavardhana: Is it not possible to show what was asked to be removed since you get the input for the site name?

cmd/admin-replicate-remove.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (i srRemoveStatus) JSON() string {
+	ds, e := json.MarshalIndent(i.ReplicateRemoveStatus, "", " ")
+	fatalIf(probe.NewError(e), "Unable to marshal into JSON.")
+	return string(ds)
+}
+
+func (i srRemoveStatus) String() string {
+	return console.Colorize("UserMessage", fmt.Sprintf("%s were removed successfully from site replication", strings.Join(i.sites, ", ")))
```
harshavardhana, line 81: in case of when no sites are given do you want to say `all sites were removed` ?

harshavardhana (changes requested):

cmd/admin-replicate-remove.go
poornas: syntax validation would have caught that

harshavardhana: how what happens when I pass `--all` @poornas ?

harshavardhana (approved):

PR #3925, opened 2022-01-19
Update madmin-go libary

vadmeste:
mc needs to be released after the next release of MinIO server.


vadmeste:
This is not needed anymore. It looks like this is done in another PR.

PR #3927, opened 2022-01-20
heal: Add verbose flag to heal command

vadmeste:
For deployments which has many disks, it is annoying to list all disks
by default. Add --verbose flag to only show the status of all disks.


klauspost (approved):
LGTM

harshavardhana (approved):

Issue #3930, opened 2022-01-21
mc admin policy update toggles user PolicyName


ElliottMoos:
## Expected behavior
`mc admin policy update <alias> <policy> user=<user>` does nothing when the given user is already associated with the policy 
```bash
# First run
$ mc admin policy update <alias> <policy> user=<user> 

Policy <policy> is added to user <user>

$ mc admin user info <alias> <user>

AccessKey: <access_key>
Status: <status>
# PolicyName is correctly updated
PolicyName: <policy>
MemberOf: <groups>

# Alternating runs of this command (i.e. MinIO helm chart provisioning Job)
$ mc admin policy update <alias> <policy> user=<user> 

Policy <policy> is added to user <user>

$ mc admin user info <alias> <user>

AccessKey: <access_key>
Status: <status>
# PolicyName does not change
PolicyName: <policy>
MemberOf: <groups>
```
## Actual behavior
`mc admin policy update <alias> <policy> user=<user>` removes the association for the given user if it is already associated with the policy 
```bash
# First run
$ mc admin policy update <alias> <policy> user=<user> 

Policy <policy> is added to user <user>

$ mc admin user info <alias> <user>

AccessKey: <access_key>
Status: <status>
# PolicyName is correctly updated
PolicyName: <policy>
MemberOf: <groups>

# Alternating runs of this command (i.e. MinIO helm chart provisioning Job)
$ mc admin policy update <alias> <policy> user=<user> 

Policy <policy> is added to user <user>

$ mc admin user info <alias> <user>

AccessKey: <access_key>
Status: <status>
# PolicyName is now blank
PolicyName:
MemberOf: <groups>
```
## Steps to reproduce the behavior
1. Add an existing user to a policy via the `mc admin policy update` command
2. Run the same command again

## mc --version
`mc version DEVELOPMENT.2022-01-11T03-10-40Z`

## System information
`Linux minio 5.4.149-73.259.amzn2.x86_64 #1 SMP Mon Sep 27 12:48:12 UTC 2021 x86_64 GNU/Linux`

vadmeste:
@ElliottMoos thanks. I am able to reproduce. We will send a fix for this.

ElliottMoos:
Thanks for the quick response @vadmeste

PR #3931, opened 2022-01-21
typo: Error out if the policy is already set in policy update cmd

vadmeste:
Fix a typo when checking if a policy is already set to a user or a
group.

Fixes https://github.com/minio/mc/issues/3930


kannappanr (approved):
LGTM

harshavardhana (approved):

PR #3942, opened 2022-01-28
remove quota FIFO support

harshavardhana:


mshanmu:
@harshavardhana , This is very important feature. I would dare say a key differentiator. Any reason on why it is being removed?

harshavardhana:
> @harshavardhana , This is very important feature. I would dare say a key differentiator. Any reason on why it is being removed?

You need to use ILM instead of FIFO for expiring old content - this is not an important feature for any of our customers and it is not optimal. 

If you are an active customer please reach out to https://subnet.min.io if not we are removing this.

cmd/admin-bucket-quota.go
```diff
@@ -118,10 +111,7 @@ func checkAdminBucketQuotaSyntax(ctx *cli.Context) {
 		cli.ShowCommandHelpAndExit(ctx, ctx.Command.Name, 1) // last argument is exit code
 	}
 
-	if ctx.IsSet("hard") && ctx.IsSet("fifo") {
-		fatalIf(errInvalidArgument(), "Only one of --hard or --fifo flags can be set")
-	}
-	if (ctx.IsSet("hard") || ctx.IsSet("fifo")) && len(ctx.Args()) == 0 {
+	if (ctx.IsSet("hard")) && len(ctx.Args()) == 0 {
```
vadmeste, line 114: len(ctx.Args()) == 0 is not needed, we already exit in the beginning of the function if no args is passed

cmd/admin-bucket-quota.go
```diff
@@ -143,16 +133,11 @@ func mainAdminBucketQuota(ctx *cli.Context) error {
...
-		quotaStr = ctx.String("hard")
-	}
+
 	_, targetURL := url2Alias(args[0])
-	if ctx.IsSet("fifo") || ctx.IsSet("hard") && len(args) == 1 {
-		qType := madmin.FIFOQuota
-		if ctx.IsSet("hard") {
-			qType = madmin.HardQuota
-		}
+	if ctx.IsSet("hard") && len(args) == 1 {
```
vadmeste, line 138: I think len(args) == 1 is not needed as well as checkAdminBucketQuotaSyntax() ensures it already

vadmeste (commented):
Few comments

parkjeongryul:
I've read the ILM document, and it seems that FIFO quata cannot be replaced. 

If I want to delete old buckets by specifying a size (like 10Gi) as threshold instead of specific date, how can i replace it with ILM?

Is there a way to do this using ILM?

harshavardhana:
> I've read the ILM document, and it seems that FIFO quata cannot be replaced.
> 
> If I want to delete old buckets by specifying a size (like 10Gi) as threshold instead of specific date, how can i replace it with ILM?
> 
> Is there a way to do this using ILM?

Size based ILM will come in future, no ETA yet.

PR #3949, opened 2022-02-02
Move admin-subnet commands to support

anjalshireesh:
So that

`mc admin subnet register|health`

become

`mc support register|diagnostics`

The previous commands will print a message asking the user to use the
new ones.


cmd/admin-subnet-register.go
```diff
@@ -18,160 +18,20 @@
...
-	Username string `json:"username"`
-	Password string `json:"password"`
-}
-
-// SubnetMFAReq - JSON payload of the SUBNET mfa api
-type SubnetMFAReq struct {
-	Username string `json:"username"`
-	OTP      string `json:"otp"`
-	Token    string `json:"token"`
+	Name:               "register",
```
vadmeste, line 26: I think we have Hidden flag here, I guess you can use it, can you double check ?

cmd/admin-subnet-health.go
```diff
@@ -18,653 +18,20 @@
...
-
-func greenText(s string) string {
-	console.SetColor("GREEN", color.New(color.FgGreen))
-	return console.Colorize("GREEN", s)
-}
-
-func warnText(s string) string {
-	console.SetColor("WARN", color.New(color.FgRed, color.Bold))
-	return console.Colorize("WARN", s)
+	Name:               "health",
```
vadmeste, line 26: I think you can hide this omment with Hidden, can you check ?

vadmeste (commented):
Few comments

cmd/admin-subnet-register.go
anjalshireesh: Yes, thanks. Done.

cmd/admin-subnet-health.go
anjalshireesh: Done

harshavardhana (approved):

vadmeste (approved):
LGTM

PR #3950, opened 2022-02-02
du: Add object counts

klauspost:
Collect object count as well.

Example:

image

Without versions:

image

### DIFF
--- a/cmd/du-main.go
+++ b/cmd/du-main.go
@@ -95,7 +95,7 @@ type duMessage struct {
 	Size       int64  `json:"size"`
 	Objects    int64  `json:"objects"`
 	Status     string `json:"status"`
-	IsVersions bool   `json:"is_versions"`
+	IsVersions bool   `json:"isVersions"`
 }
 
 // Colorized message for console printing.
--- a/cmd/du-main.go
+++ b/cmd/du-main.go
@@ -91,16 +91,25 @@ EXAMPLES:
 
 // Structured message depending on the type of console.
 type duMessage struct {
-	Prefix string `json:"prefix"`
-	Size   int64  `json:"size"`
-	Status string `json:"status"`
+	Prefix     string `json:"prefix"`
+	Size       int64  `json:"size"`
+	Objects    int64  `json:"objects"`
+	Status     string `json:"status"`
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

cmd/du-main.go
```diff
@@ -91,16 +91,25 @@ EXAMPLES:
...
 // Structured message depending on the type of console.
 type duMessage struct {
-	Prefix string `json:"prefix"`
-	Size   int64  `json:"size"`
-	Status string `json:"status"`
+	Prefix     string `json:"prefix"`
+	Size       int64  `json:"size"`
+	Objects    int64  `json:"objects"`
+	Status     string `json:"status"`
+	IsVersions bool   `json:"is_versions"`
```
vadmeste, line 98: This looks like only useful for internal use, we don't have to export this field

vadmeste (commented):
One comment

cmd/du-main.go
harshavardhana: I think it's exported for the same reason how it's used internally. Exposing this is okay I think. @vadmeste 

However we follow JavaScript style for JSON  in mc so it is better tagged as `isVersion`  @klauspost

klauspost: @vadmeste It shows what you are looking at, if only presented with the JSON.

@harshavardhana The rest of the world has snake case for JSON, but I will ofc change it, so we at least consistent.

vadmeste: > It shows what you are looking at

AFAIK,  isVersions is set to true if the user passes --versions flag, it doesn't seem it is useful to output this again to the user.

vadmeste: Oh because we already have 'Objects' which can be confusing.

klauspost: @vadmeste The user may just send the output and not the commandline.

vadmeste (approved):
LGTM

cmd/du-main.go
vadmeste: > @vadmeste The user may just send the output and not the commandline.

Agree. In fact we introduced versioning in MinIO after mc du. That's why non breaking change is hard here.

PR #3954, opened 2022-02-04
add pool id as part of 'mc admin info' output

harshavardhana:
This also fixes `mc` output when it incorrectly displays offline disks.

### DIFF
--- a/cmd/admin-info.go
+++ b/cmd/admin-info.go
@@ -20,7 +20,6 @@ package cmd
 import (
 	"errors"
 	"fmt"
-	"reflect"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


klauspost (approved):
LGTM

cmd/admin-info.go
```diff
@@ -230,12 +226,19 @@ func (u clusterStruct) String() (msg string) {
...
-		msg += fmt.Sprintf("%s online, %s offline\n",
-			english.Plural(totalOnlineDisksCluster, "drive", ""),
-			english.Plural(totalOfflineDisksCluster, "drive", ""))
+		bkInfo, ok := u.Info.Backend.(madmin.ErasureBackend)
+		if ok {
+			msg += fmt.Sprintf("%s online, %s offline\n",
+				english.Plural(bkInfo.OnlineDisks, "drive", ""),
+				english.Plural(bkInfo.OfflineDisks, "drive", ""))
+		} else {
+			msg += fmt.Sprintf("%s online, %s offline\n",
```
vadmeste, line 238: Printing disks offline/online does not work for gateway. I think we don't have to print anything if this is not erasure.

vadmeste (commented):
One comment

cmd/admin-info.go
harshavardhana: How does gateway come into picture here @vadmeste 

This code never prints in Gateway. Did you test it with gateway?

harshavardhana: done fixed @vadmeste

vadmeste (approved):
LGTM

PR #3963, opened 2022-02-08
Add usage information to output if no arguments passed to mc alias set

jinapurapu:
Addresses https://github.com/miniohq/engineering/issues/269

### DIFF
--- a/cmd/alias-set.go
+++ b/cmd/alias-set.go
@@ -61,10 +61,8 @@ var aliasSetCmd = cli.Command{
 	HideHelpCommand: true,
 	CustomHelpTemplate: `NAME:
   {{.HelpName}} - {{.Usage}}
-
 USAGE:
   {{.HelpName}} ALIAS URL ACCESSKEY SECRETKEY
-
 FLAGS:
   {{range .VisibleFlags}}{{.}}
   {{end}}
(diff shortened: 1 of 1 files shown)


cmd/alias-set.go
```diff
@@ -100,19 +101,40 @@ EXAMPLES:
...
+			// Create a new MinIO Admin Client
+			client, err := newAdminClient(k)
+			fatalIf(err, "Unable to initialize admin connection.")
+			admInfo, e := client.ServerInfo(globalContext)
+			// this is not the right way to handle this error - find the correct way
+			if e != nil {
+				fmt.Println("Unable to retrieve usage information for `" + k + "`.")
+			} else {
+				fmt.Println("Usage for `" + k + "` is `" + humanize.IBytes(admInfo.Usage.Size) + "`.")
+			}
```
harshavardhana, line 124: If there are configured endpoints that are not reachable this command will never return @jinapurapu - I don't think this is a feasible behavior.

harshavardhana (changes requested):

harshavardhana:
is this still a WIP ? @jinapurapu

jinapurapu:
No, I have updated it, it is working now.

harshavardhana (approved):

klauspost (approved):

PR #3970, opened 2022-02-09
Error out from `support diag` if cluster not registered

anjalshireesh:
If user tries to run `mc support diag` for a cluster that is not yet
registered, error out asking the user to register the cluster first.


harshavardhana (approved):

harshavardhana:
PTAL @balamurugana

balamurugana (approved):

PR #3986, opened 2022-02-16
Add `mc support callhome get` command

anjalshireesh:
Will print all the callhome settings (currently just `logs=on|off`)

### DIFF
--- a/cmd/support-callhome-get.go
+++ b/cmd/support-callhome-get.go
@@ -89,7 +89,7 @@ func (m callhomeGetMessage) JSON() string {
 	return string(statusJSONBytes)
 }
 
-func retrieveCallhomeSetting(alias string, s string) (v string, e error) {
+func printCallhomeSetting(alias string, s string) {
 	switch s {
 	case "logs":
 		// Create a new MinIO Admin Client
@@ -118,7 +118,6 @@ func retrieveCallhomeSetting(alias string, s string) (v string, e error) {
 			Value: val,
 		})
 	}
-	return "", nil
 }
 
 func mainCallhomeGet(ctx *cli.Context) error {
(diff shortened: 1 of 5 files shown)


cmd/support-callhome-get.go
```diff
@@ -0,0 +1,146 @@
...
+
+// JSON jsonified service status Message message.
+func (m callhomeGetMessage) JSON() string {
+	statusJSONBytes, e := json.MarshalIndent(m, "", " ")
+	fatalIf(probe.NewError(e), "Unable to marshal into JSON.")
+
+	return string(statusJSONBytes)
+}
+
+func retrieveCallhomeSetting(alias string, s string) (v string, e error) {
```
vadmeste, line 92: Let's not return any value here if we are not going to use them, it is better to call this function printCallhomeSetting as well.

cmd/support-callhome-get.go
```diff
@@ -0,0 +1,146 @@
...
+		ep, found := kvs.Lookup("endpoint")
+
+		if found && len(ep) > 0 {
+			// subnet webhook is configured. check if it is enabled
+			enable, found := kvs.Lookup("enable")
+
+			if found {
+				val = enable
+			} else {
+				// if the 'enable' key is not found, it means that the webhook is enabled
```
vadmeste, line 111: :+1:

vadmeste (commented):
Few comments

cmd/support-callhome-get.go
anjalshireesh: Thanks, done.

vadmeste (approved):
LGTM

harshavardhana (approved):

PR #3987, opened 2022-02-16
fix: rm should reject --non-current with --recursive

harshavardhana:


cmd/rm-main.go
```diff
@@ -212,9 +212,9 @@ func checkRmSyntax(ctx context.Context, cliCtx *cli.Context, encKeyDB map[string
 			"You cannot specify --version-id with any of --versions, --rewind and --recursive flags.")
 	}
 
-	if isNoncurrentVersion && !isVersions {
+	if isNoncurrentVersion && !isVersions && !isRecursive {
```
poornas, line 215:
```suggestion
	if !(isNoncurrentVersion && isVersions && isRecursive) {
```

poornas (changes requested):
commit statement to be fixed

cmd/rm-main.go
klauspost: All false will trigger the error, which isn't intended, I suppose:
```
func main() {
	var isNoncurrentVersion, isVersions, isRecursive bool
	fmt.Println(isNoncurrentVersion && !isVersions && !isRecursive)
	fmt.Println(!(isNoncurrentVersion && isVersions && isRecursive))
}
```

poornas: yes, it should be if isNoncurrentVersion && !(isVersions && isRecursive) {

poornas (approved):

PR #3992, opened 2022-02-21
Refactor profiling and add goroutine profiling

vadmeste:
Running mc like this, `MC_PROFILER=mem,goroutine mc ...` will start profiling and save
profiling data when mc quits with Control-C.


harshavardhana (approved):
Awesome thanks for removing `pkg/profile` hated it for so long.

harshavardhana (approved):

PR #3995, opened 2022-02-22
fix: not allowing to contain double slash ('//') in target url in cp, mv, and mirror

ocworld:
I've experienced that when cp, targetPath containing double slash ('//') in minio cluster, copy is failure actually (0 bytes).

To solve it, checking containing '//' in targetPath is added.


harshavardhana:
please fix the PR title not sure what it is trying to fix, all I see is some special handling for `//` - why only `cp` why not other commands?

ocworld:
> please fix the PR title not sure what it is trying to fix, all I see is some special handling for `//` - why only `cp` why not other commands?

@harshavardhana Thanks for your comment.

- The title is fixed.
- in mirror command, it is also not supported now.  
- '//' is not supported in minio. However, it is allowed in aws s3. (https://github.com/minio/minio/issues/5874).
To support both flexibly, '//'  is handling as error in copy-like operations's target path, such as cp, mv, and mirror, but not in other operator such as rm, ls,  copy-like operations's source path. I think that just read and stat don't make the problem.

harshavardhana:
> * The title is fixed.
> * in mirror command, it is also not allowed now.
> * '//' is not supported in minio. However, it is allowed in aws s3. ([Object name contains unsupported characters when using double forward slash (//) minio#5874](https://github.com/minio/minio/issues/5874)).
>   To support both flexibly, '//'  is handling as error in copy-like operations's target path, such as cp, mv, and mirror, but not in other operator such as rm, ls,  copy-like operations's source path. I think that just read and stat don't make the problem.

The error is returned by the server why do you want to do special hacks on `mc`? @ocworld

Issue #3999, opened 2022-02-24
`mc share upload` should quote special characters in object's key


supriyo-biswas:
## Expected behavior

When generating upload commands through the `mc share upload` command, special characters in the object's key name, such as spaces or quotes are not quoted, which then has to be addressed separately by the user.

For example, running `mc share upload --json "s3/personal/Robert O'Neil.png"` should generate a properly quoted command, such as one of the following:

```
curl https://minio.example/personal/ ... -F bucket=personal -F key="Robert O'Neil.png" -F file=@<FILE>
curl https://minio.example/personal/ ... -F bucket=personal -F key=Robert\ O\'Neil.png -F file=@<FILE>
```

## Actual behavior

Minio does not quote special characters, leading to the following command:

```
curl https://minio.example/personal/ ... -F bucket=personal -F key=Robert O'Neil.png -F file=@<FILE>
```

## Steps to reproduce the behavior

Run `mc share upload --json "s3/personal/bar 'foo.png" | jq -r .share` to get the curl command for file uploads.

## mc --version
mc version RELEASE.2022-02-23T03-15-59Z

## System information
macOS 12.2.1

PR #4000, opened 2022-02-24
Quote special characters in share curl command

supriyo-biswas:
See #3999 for the justification behind this change.

### DIFF
--- a/cmd/share-upload-main.go
+++ b/cmd/share-upload-main.go
@@ -20,6 +20,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


harshavardhana (commented):
Please add unit tests for the function, that covers corner cases.

supriyo-biswas:
Are we looking for testcases around `makeCurlCmd` or around `doShareUploadURL`?

harshavardhana:
> Are we looking for testcases around `makeCurlCmd` or around `doShareUploadURL`?

makeCurlCmd, the regex part.

klauspost:
Doesn't it work with URL escaped paths? Seems safer than escaping.

supriyo-biswas:
As I understand URL-escaping is not a requirement of multipart/form-data requests. As a result, a file name of `foo $.pdf` is distinct from `foo%20%24.pdf` from Minio's perspective.

klauspost (approved):
👍🏼

harshavardhana (approved):

PR #4004, opened 2022-03-01
More validation for retention commands

vadmeste:
1. Print an error when --default with specified with any other flags, such as --version-id, --recursive, etc..
2. Print an error when --default is passed with object name in the aliased path


klauspost (approved):
LGTM

harshavardhana (approved):

PR #4006, opened 2022-03-01
Fix arm FreeBSD build

### LINKED ISSUE #4005: Build fails on armv7 FreeBSD 13
The build of minio-client fails on armv7 FreeBSD 13 with the following error message:

    # github.com/minio/mc/pkg/disk
    pkg/disk/stat_freebsd.go:38:101: cannot use st.Atimespec.Nsec (type int32) as type int64 in argument to strconv.FormatInt
    pkg/disk/stat_freebsd.go:51:101: cannot use st.Mtimespec.Nsec (type int32) as type int64 in argument to strconv.FormatInt

The cause of this error is that these fields have different types on different architectures.  To fix this problem, add an explicit conversion before passing to `strconv.FormatInt`:

    fileAttr.WriteString(strconv.FormatInt(st.Atimespec.Sec, 10) + "#" + strconv.FormatInt(int64(st.Atimespec.Nsec), 10))
    fileAttr.WriteString(strconv.FormatInt(st.Mtimespec.Sec, 10) + "#" + strconv.FormatInt(int64(st.Mtimespec.Nsec), 10))

klauspost:
Fixes #4005

Tested with `GOOS=freebsd GOARCH=arm go build .`.

### DIFF
--- a/pkg/disk/stat_freebsd.go
+++ b/pkg/disk/stat_freebsd.go
@@ -35,7 +35,8 @@ func GetFileSystemAttrs(file string) (string, error) {
 
 	var fileAttr strings.Builder
 	fileAttr.WriteString("atime:")
-	fileAttr.WriteString(strconv.FormatInt(st.Atimespec.Sec, 10) + "#" + strconv.FormatInt(st.Atimespec.Nsec, 10))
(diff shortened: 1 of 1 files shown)


vadmeste (approved):
LGTM

PR #4009, opened 2022-03-03
Add support for minio warm tier

krisis:


harshavardhana:
PTAL at the go mod updates @krisis

PR #4010, opened 2022-03-07
alias: fetch peer certificate when adding an alias

aead:
This commit adds an extension to the `alias set`
command. When adding an alias fails due to a TLS
trust error - i.e. remote peer certificate verification
fails - `mc` fetches the remote peer certificate,
computes the public key fingerprint and asks the user
to confirm the fingerprint.

When the user confirms the fingerprint, the certificate
is added to the `CAs` directory.

This allows users to use a TOFU authentication with
self-signed certificates - similar to SSH.

### DIFF
--- a/cmd/alias-set.go
+++ b/cmd/alias-set.go
@@ -19,10 +19,19 @@ package cmd
 
 import (
 	"bufio"
+	"bytes"
 	"context"
+	"crypto/sha256"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/hex"
+	"encoding/pem"
 	"fmt"
 	"math/rand"
+	"net"
+	"net/http"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -32,6 +41,7 @@ import (
 	"github.com/minio/minio-go/v7"
 	"github.com/minio/pkg/console"
 	"golang.org/x/crypto/ssh/terminal"
+	"golang.org/x/term"
 )
 
 const cred = "YellowItalics"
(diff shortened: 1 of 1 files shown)


cmd/alias-set.go
```diff
@@ -184,45 +194,81 @@ func probeS3Signature(ctx context.Context, accessKey, secretKey, url string) (st
...
-		if _, err := s3Client.Stat(ctx, StatOptions{}); err != nil {
-			e := err.ToGoError()
-			if _, ok := e.(BucketDoesNotExist); ok {
-				// Bucket doesn't exist, means signature probing worked successfully.
-				return stype, nil
+		if strings.Contains(pErr.ToGoError().Error(), "x509: certificate signed by unknown authority") {
+			// We don't need to fetch the peer certificate
+			// if STDOUT is not a terminal and the user cannot
+			// confirm the public key.
+			if !term.IsTerminal(int(os.Stdout.Fd())) {
```
harshavardhana, line 216: BTW we should also not ask for `globalIsJSON`  if set @aead

harshavardhana (changes requested):

cmd/alias-set.go
aead: 👍

cmd/alias-set.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2015-2021 MinIO, Inc.
+// Copyright (c) 2013-2021 MinIO, Inc.
```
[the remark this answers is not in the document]
aead: Not sure why this got changed... Reverted 👍

vadmeste:
@aead just for simplificaton purpose, maybe it is better to do this outside probeS3Configuration(), this latter is doing too many things already

it can be done in mainAliasSet() before BuildS3Config(). I believe, you can just issue a GET request

aead:
@vadmeste We can do that but that means `mc alias set` does one HTTP additional HTTP request in all cases.

harshavardhana (approved):

vadmeste (approved):
LGTM & tested

PR #4017, opened 2022-03-11
Error handling for remove alias request if alias not found 

jinapurapu:
Previously, `mc alias remove thisalias` would return a success message even if `thisalias` was not an existing alias. This checks if the requested alias is present in the configuration file, and if not, throws an appropriate error message.

### DIFF
--- a/cmd/alias-remove.go
+++ b/cmd/alias-remove.go
@@ -18,8 +18,6 @@
 package cmd
 
 import (
-	"errors"
-
 	"github.com/fatih/color"
 	"github.com/minio/cli"
 	"github.com/minio/mc/pkg/probe"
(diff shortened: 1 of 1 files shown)


cmd/alias-remove.go
```diff
@@ -80,11 +83,21 @@ func mainAliasRemove(ctx *cli.Context, deprecated bool) error {
...
+func isAliasIn(alias string, aliasMap map[string]aliasConfigV10) *probe.Error {
+	if _, ok := aliasMap[alias]; ok {
+		return nil
+	}
+	return probe.NewError(errors.New("Alias not found in alias list"))
+}
+
 // removeAlias - removes an alias.
 func removeAlias(alias string) aliasMessage {
 	conf, err := loadMcConfig()
```
harshavardhana, line 96: We could just add this here

```
        hostConfig := mustGetHostConfig(alias)
        if hostConfig == nil {
                fatalIf(errInvalidAliasedURL(alias), "No such alias `"+alias+"` found.")
                return nil
        }
```

harshavardhana (changes requested):

cmd/alias-remove.go
```diff
@@ -80,11 +81,24 @@ func mainAliasRemove(ctx *cli.Context, deprecated bool) error {
...
 	fatalIf(err.Trace(globalMCConfigVersion), "Unable to load config version `"+globalMCConfigVersion+"`.")
 
+	// check if alias is valid
+	err = isAliasIn(alias)
+
+	fatalIf(err.Trace(alias), alias+" not found in the configuration file. Please confirm the alias name you are trying to remove. You can use mc alias list to view a list of all aliases configured.")
```
harshavardhana, line 101: we already exit inside isAliasIn() there is no reason for this function to return a `nil` and this fatalIf() will never be called.

```
func aliasMustExist(alias string) {
hostConfig := mustGetHostConfig(alias)
	if hostConfig == nil {
		fatalIf(errInvalidAliasedURL(alias), "No such alias `"+alias+"` found.")
}
```

This is all we need

harshavardhana (changes requested):

klauspost (approved):

cmd/alias-remove.go
harshavardhana: @jinapurapu

harshavardhana (approved):

PR #4019, opened 2022-03-16
Fix error message in `mc admin subnet health`

anjalshireesh:
Change from

`Please use 'mc support diagnostics'`

to

`Please use 'mc support diag'`

### DIFF
--- a/cmd/admin-subnet-health.go
+++ b/cmd/admin-subnet-health.go
@@ -29,10 +29,10 @@ var adminSubnetHealthCmd = cli.Command{
 	Action:             mainSubnetHealth,
 	Before:             setGlobalsFromContext,
 	Hidden:             true,
-	CustomHelpTemplate: "Please use 'mc support diagnostics'",
+	CustomHelpTemplate: "Please use 'mc support diag'",
 }
 
 func mainSubnetHealth(ctx *cli.Context) error {
-	console.Infoln("Please use 'mc support diagnostics'")
+	console.Infoln("Please use 'mc support diag'")
 	return nil
 }
--- a/cmd/admin-subnet-health.go
+++ b/cmd/admin-subnet-health.go
@@ -18,7 +18,11 @@
 package cmd
 
 import (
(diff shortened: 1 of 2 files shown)


harshavardhana (changes requested):
while we are at this we should also support dummy flags until we fully deprecate these commands to ensure that we can throw appropriate message such as `mc support diag --flag`

anjalshireesh:
> while we are at this we should also support dummy flags until we fully deprecate these commands to ensure that we can throw appropriate message such as `mc support diag --flag`

Have updated the PR to support all flags supported by `mc support diag`. The behavior now will be:

```
> ./mc admin subnet health --offline
mc: Please use 'mc support diag'
```

anjalshireesh:
> > while we are at this we should also support dummy flags until we fully deprecate these commands to ensure that we can throw appropriate message such as `mc support diag --flag`
> 
> Have updated the PR to support all flags supported by `mc support diag`. The behavior now will be:
> 
> ```
> > ./mc admin subnet health --offline
> mc: Please use 'mc support diag'
> ```

Updated to include arguments and flags also in the new command

cmd/admin-subnet-health.go
```diff
@@ -29,10 +33,33 @@ var adminSubnetHealthCmd = cli.Command{
 	Action:             mainSubnetHealth,
 	Before:             setGlobalsFromContext,
 	Hidden:             true,
-	CustomHelpTemplate: "Please use 'mc support diagnostics'",
+	Flags:              append(supportDiagFlags, globalFlags...),
+	CustomHelpTemplate: "Please use 'mc support diag'\n",
```
ravindk89, line 37: Should we be more specific?

"This command is deprecated and will be removed in a future release. Use `mc support diag` instead."

anjalshireesh: Done

harshavardhana (approved):

ravindk89 (approved):

PR #4021, opened 2022-03-16
replicate add should enable all replication types by default

### LINKED ISSUE #4020: Confusing help docs for defaults in `replicate add`'s `--replicate` option
`mc replicate add` will not enable `existing-objects` feature(as the help docs say with regard to `default: ...`), but `mc replicate add --replicate existing-objects` will.

## Expected behavior
` --replicate value             comma separated list to enable replication of soft deletes, permanent deletes, existing objects and metadata sync (valid options are: "delete-marker,delete,existing-objects,metadata-sync"`

## Actual behavior
` --replicate value             comma separated list to enable replication of soft deletes, permanent deletes, existing objects and metadata sync (default: "delete-marker,delete,existing-objects,metadata-sync"`

`ExistingObjectReplication`'s Status is `Disabled`.
```
 mc replicate export AlphaReplication/src-bucket
{
 "Rules": [
  {
   "ID": "c8opc98b5u56ic6mij10",
   "Status": "Enabled",
   "Priority": 0,
   "DeleteMarkerReplication": {
    "Status": "Disabled"
   },
   "DeleteReplication": {
    "Status": "Disabled"
   },
   "Destination": {
    "Bucket": "arn:minio:replication::7b92ad2c-9d6f-4d84-a005-8049a960b009:dst-bucket"
   },
   "Filter": {
    "And": {},
    "Tag": {}
   },
   "SourceSelectionCriteria": {
    "ReplicaModifications": {
     "Status": "Enabled"
    }
   },
   "ExistingObjectReplication": {
    "Status": "Disabled"
   }
  }
 ],
 "Role": ""
}
```

`mc replicate add --help`
## mc --version
mc version RELEASE.2022-03-03T21-12-24Z

## System information
```
$ sw_vers
ProductName: Mac OS X
ProductVersion: 10.15.7
BuildVersion: 19H1615
```

harshavardhana:
fixes #4020

### DIFF
--- a/cmd/replicate-add.go
+++ b/cmd/replicate-add.go
@@ -167,21 +167,19 @@ func mainReplicateAdd(cliCtx *cli.Context) error {
 	deleteReplicationStatus := disableStatus
 	replicaSync := enableStatus
 	existingReplicationStatus := disableStatus
-	if cliCtx.IsSet("replicate") {
-		replSlice := strings.Split(cliCtx.String("replicate"), ",")
-		for _, opt := range replSlice {
-			switch strings.TrimSpace(strings.ToLower(opt)) {
-			case "delete-marker":
-				dmReplicateStatus = enableStatus
-			case "delete":
-				deleteReplicationStatus = enableStatus
-			case "metadata-sync", "replica-metadata-sync":
-				replicaSync = enableStatus
-			case "existing-objects":
-				existingReplicationStatus = enableStatus
-			default:
-				fatalIf(probe.NewError(fmt.Errorf("invalid value for --replicate flag %s", cliCtx.String("replicate"))), `--replicate flag takes one or more comma separated string with values "delete", "delete-marker", "metadata-sync", "existing-objects" or "" to disable these settings`)
-			}
+	replSlice := strings.Split(cliCtx.String("replicate"), ",")
+	for _, opt := range replSlice {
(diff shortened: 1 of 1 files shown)


cmd/replicate-add.go
```diff
@@ -167,21 +167,19 @@ func mainReplicateAdd(cliCtx *cli.Context) error {
...
-			case "delete":
-				deleteReplicationStatus = enableStatus
-			case "metadata-sync", "replica-metadata-sync":
-				replicaSync = enableStatus
-			case "existing-objects":
-				existingReplicationStatus = enableStatus
-			default:
-				fatalIf(probe.NewError(fmt.Errorf("invalid value for --replicate flag %s", cliCtx.String("replicate"))), `--replicate flag takes one or more comma separated string with values "delete", "delete-marker", "metadata-sync", "existing-objects" or "" to disable these settings`)
-			}
+	replSlice := strings.Split(cliCtx.String("replicate"), ",")
```
vadmeste, line 170: This means that --replicate will be always mandatory to pass in the cli, is that expected?

poornas: if --replicate is not passed, it will use the defaults.

vadmeste: > if --replicate is not passed, it will use the defaults.

Okay, but the new change breaks this behavior. replSlice slice will be always non empty here and it will fail with a fatal() in the next for loop

PR #4032, opened 2022-03-31
mirror: Do not exit upon errors when --watch is passed

vadmeste:
Currently, when mirror is not able to copy an object (e.g. corrupted)
and watch is specified, mirror will restart mirroring again to stop
again at the same problematic object.

This commit will not abort mirroring when --watch is specified, to
continue mirroring the rest of the object. An error message will still
be printed.


harshavardhana (approved):

kannappanr (approved):
LGTM

PR #4038, opened 2022-04-02
rm: Fix sending an object/version to removal with --no-current

vadmeste:
Do no miss sending object/version to contentCh when there is deletion
result available.

This commit fixes issues seen when some objects/versions are not deleted
when --no-current is passed.


kannappanr (approved):
LGTM

poornas (approved):

PR #4041, opened 2022-04-04
Do not rely on contentCh to print removal messages

vadmeste:
Removal is an asynchrounous operation. We should not use structures that
we send to the removal goroutine and use them in removal messages.

### DIFF
--- a/cmd/rm-main.go
+++ b/cmd/rm-main.go
@@ -167,12 +167,10 @@ EXAMPLES:
 
 // Structured message depending on the type of console.
 type rmMessage struct {
-	Status       string    `json:"status"`
-	Key          string    `json:"key"`
-	DeleteMarker bool      `json:"deleteMarker"`
-	VersionID    string    `json:"versionID"`
-	ModTime      time.Time `json:"modTime"`
-	Size         int64     `json:"size"`
+	Status       string `json:"status"`
+	Key          string `json:"key"`
+	DeleteMarker bool   `json:"deleteMarker"`
+	VersionID    string `json:"versionID"`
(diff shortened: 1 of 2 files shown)


cmd/rm-main.go
```diff
@@ -182,11 +180,7 @@ func (r rmMessage) String() string {
 		msg = console.Colorize("Remove", fmt.Sprintf("Creating delete marker `%s`", r.Key))
 	}
 	if r.VersionID != "" {
-		if !r.ModTime.IsZero() {
-			msg += fmt.Sprintf(" (versionId=%s, modTime=%s)", r.VersionID, r.ModTime)
```
harshavardhana, line 186: why are we not interested in printing modTime anymore?

harshavardhana (changes requested):

cmd/rm-main.go
vadmeste: > why are we not interested in printing modTime anymore?

Because it is buggy. It doesn't show the correct information sometimes.
```
                              select {
                                case contentCh <- content:
                                case result := <-resultCh:
                                        // using a mix of result & content variables
```

The old code was mixing the use of the element that we sent to deletion and the result deletion element (content & result above. However this is wrong, result doesn't always correspond to content here.

Basicaly deletion is asynchorounous in mc perspective.

prakashsvmx (approved):
Tested. Changes look good to me. :+1:

klauspost:
Was just looking at this.

kannappanr (approved):
LGTM

Issue #4051, opened 2022-04-15
HTTP Proxy support is broken


Alwaysin:
## Expected behavior

Tool is using proxy when defined, eg.:

` https_proxy=http://192.168.1.254:3128 mc alias set s3 https://...
`
## Actual behavior

Ptoxy is not used. Looks like the regression from #685 has come back

## Steps to reproduce the behavior

```
wget https://dl.min.io/client/mc/release/linux-amd64/mc
chmod +x mc
https_proxy=http://192.168.1.254:3128 ./mc alias set s3 https://...

```
## ./mc --version
`mc version RELEASE.2022-04-07T21-43-27Z
`
## System information
Deb10

harshavardhana:
Please provide more details such as --debug messages, we have nothing to go about in your issue report.

Alwaysin:
I went too quick on this one. The above command is adding the config correctly, and `./mc mv` seems to take into account the proxy.

So to be more precise the problem is just about the `alias `command that is not using it. Unfortunately, with `--debug`, I had nothing else printed on the CLI. I'm just seeing the requests going straight to the gateway without using the proxy, and thus rejected.

Fine by me if you want to close this issue, since it seems like a minor inconvenience and not blocking the whole thing.

PR #4052, opened 2022-04-15
Load proxy config when probing for server TLS certificate

vadmeste:
mc alias set recently started to probe the TLS certificate for
the remote server, however it should load the proxy environment
configuration if any.


harshavardhana (approved):

harshavardhana:
fixes #4051

PR #4053, opened 2022-04-15
mirror: Stop minio-go bucket listen when watch is cancelled

vadmeste:
When mirror in --watch mode is restarted, the previous minio-go
ListenNotification/ListenBucketNotification call is not properly
canceled. This commit creates a new context for that listen call and
cancels it when client-s3 Watch() is asked to be stopped.

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -772,19 +772,22 @@ func (c *S3Client) Watch(ctx context.Context, options WatchOptions) (*WatchObjec
 		DoneChan:      make(chan struct{}),
 	}
 
+	listenCtx, listenCancel := context.WithCancel(ctx)
+
 	var eventsCh <-chan notification.Info
 	if bucket != "" {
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

harshavardhana:
IMO we should also fix server-side @vadmeste such that - we should not allow more than X amount of active Listen Notifications.

vadmeste:
> IMO we should also fix server-side @vadmeste such that - we should not allow more than X amount of active Listen Notifications.

Yes, 1024 looks a fair number

vadmeste:
> Yes, 1024 looks a fair number

Maybe we can do 1024 per node instead for bucket notification and 8 per node for tracing.

harshavardhana:
> > Yes, 1024 looks a fair number
> 
> Maybe we can do 1024 per node instead for bucket notification and 8 per node for tracing.

Yeah that is an okay value but still high though.

PR #4061, opened 2022-04-25
modify support profile command

sinhaashish:
**Ask** : mc support profile myminio, download profile.

Results

-  **mc support profile myminio**

```
ashish@ashish:~/code/go/src/github/minio/mc$ ./mc support profile myminio
mc: Profile data successfully started. 
2022/04/25 18:52:57 Collecting data for 10 seconds
2022/04/25 18:53:07 Stopping profiling..
mc: Profile data successfully downloaded as profile.zip.2022-04-25T18-53-07.834808+05-00
```


cmd/support-profile.go
```diff
@@ -32,13 +64,175 @@ var supportProfileCmd = cli.Command{
...
+
+		getProfileData(client)
+		console.Infof("count %d : Profile data successfully downloaded as %s\n", i, getProfileData(client))
+
+		if count > 1 && i <= count-1 {
+			log.Printf("Waiting for %d seconds, before starting another profile\n", interval)
+			sleepBetweenIntervals := time.Duration(interval)
+			time.Sleep(time.Second * sleepBetweenIntervals)
+		}
+	}
```
harshavardhana, line 236: Do you know that we already have `madmin.Profile()` to do this?

harshavardhana (changes requested):

cmd/support-profile.go
```diff
@@ -32,13 +64,175 @@ var supportProfileCmd = cli.Command{
...
+				break
+			}
+		}
+	}
+
+	var interval int
+	if ctx.IsSet("interval") {
+		interval = ctx.Int("interval")
+		if interval <= 30 {
+			fatalIf(errInvalidArgument().Trace(ctx.Args()...), " the minimum interval between two profiling must be 30 seconds, for example: '--interval 30 --count 2' to get two profilers output at 30 seconds interval")
```
anjalshireesh, line 126:
```suggestion
		if interval < 30 {
			fatalIf(errInvalidArgument().Trace(ctx.Args()...), " the minimum interval between two profiling must be 30 seconds, for example: '--interval 30 --count 2' to get two profilers output at 30 seconds interval")
```

cmd/support-profile.go
```diff
@@ -32,13 +64,175 @@ var supportProfileCmd = cli.Command{
...
+		madmin.ProfilerBlock,
+		madmin.ProfilerMutex,
+		madmin.ProfilerTrace,
+		madmin.ProfilerThreads,
+		madmin.ProfilerGoroutines,
+		madmin.ProfilerCPUIO,
+	}
+	for _, profilerType := range supportedProfilerTypes {
+		s.Add(string(profilerType))
+	}
```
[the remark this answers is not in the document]
sinhaashish: IMO `set.CreateStringSet` wont be useful here as `set.CreateStringSet` needs a slice of string and we are using an abstract data type . 
set.CreateStringSet("cpu"  , "cpuio" ,"mem","block", "mutex" , "trace", "threads" , "goroutines") can lead to errors as we are hardcoding the values. The aliases can also be used to improve the readability of the code.

cmd/support-profile.go
```diff
@@ -32,13 +64,175 @@ var supportProfileCmd = cli.Command{
...
+	supportedProfiler := false
+	profilers := strings.Split(strings.ToLower(ctx.String("type")), ",")
+	for _, profiler := range profilers {
+		if profiler != "" {
+			if s.Contains(profiler) {
+				supportedProfiler = true
+				break
+			}
+		}
+	}
```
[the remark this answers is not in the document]
sinhaashish: done

cmd/support-profile.go
```diff
@@ -32,13 +64,175 @@ var supportProfileCmd = cli.Command{
...
+       {{.Prompt}} {{.HelpName}} --type cpu myminio/
+
+    2. Fetch CPU, Memory and Block profiling concurrently
+       {{.Prompt}} {{.HelpName}} --type cpu,mem,block myminio/
+
+	3. Fetch CPU, Memory and Block profiling concurrently 3 times
+       {{.Prompt}} {{.HelpName}} --type cpu,mem,block --count 3 myminio/ 
+
+	4. Fetch CPU, Memory and Block profiling concurrently 2 times with an interval of 10 mins
+       {{.Prompt}} {{.HelpName}} --type cpu,mem,block  --count 2 --interval 600 myminio/
```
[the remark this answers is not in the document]
sinhaashish: It woks
```
ashish@ashish:~/code/go/src/github/minio/mc$ ./mc support profile --help
NAME:
  mc support profile - generate profile data for debugging

USAGE:
  mc support profile COMMAND [COMMAND FLAGS | -h] [ARGUMENTS...]

COMMANDS:
  
FLAGS:
  --count value                 number of times the profiling is needed (default: 1)
  --interval value              interval between two profile data in seconds (default: 30)
  --duration value              start profiling for the specified duration in seconds (default: 10)
  --type value                  profiler type, possible values are 'cpu', 'cpuio', 'mem', 'block', 'mutex', 'trace', 'threads' and 'goroutines' (default: "cpu,mem,block,mutex,threads,goroutines")
  --config-dir value, -C value  path to configuration folder (default: "/home/ashish/.mc")
  --quiet, -q                   disable progress bar display
  --no-color                    disable color theme
  --json                        enable JSON lines formatted output
  --debug                       enable debug output
  --insecure                    disable SSL certificate verification
  --help, -h                    show help
```

cmd/support-profile.go
sinhaashish: :+1:

cmd/support-profile.go
sinhaashish: Thanks replaced  `StartProfiling` and `DownloadProfilingData` with `madmin.Profile()` . But the ask here is to have a count and interval . 
```
mc admin profile --count --interval
```
It  downloads profile file based on number of times specified in the count and sleep for the interval specified.

Refer https://docs.google.com/spreadsheets/d/1JbIi1JZCF04sH6MZ5X9zr6iCNYb2c4I8sISRNMguj6c/edit#gid=0

cmd/support-profile.go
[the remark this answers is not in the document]
harshavardhana: Correct but the examples are not showing up @sinhaashish - perhaps that's what @anjalshireesh meant.

cmd/support-profile.go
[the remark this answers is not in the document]
anjalshireesh: You can use `string(madmin.ProfilerCPU)` just like you are doing with `s.Add`

cmd/support-profile.go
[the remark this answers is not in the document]
anjalshireesh: Yes, what it's showing is the default help and not the `customHelpTemplate` defined here.

cmd/support-profile.go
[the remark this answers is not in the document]
sinhaashish: I still believe `s := set.NewStringSet()` is good.  If in future we need to add a new value then we just need to add that value in `supportedProfilerTypes` and the for loop will add that in set . But with the below approach
```
	s := set.CreateStringSet(string(madmin.ProfilerCPU), string(madmin.ProfilerMEM),
		string(madmin.ProfilerBlock), string(madmin.ProfilerMutex), string(madmin.ProfilerTrace),
		string(madmin.ProfilerThreads), string(madmin.ProfilerGoroutines), string(madmin.ProfilerCPUIO))
``` 
we need to add both in set and `supportedProfilerTypes`.

cmd/support-profile.go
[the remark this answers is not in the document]
sinhaashish: thanks done

cmd/support-profile.go
harshavardhana: the count is not that useful IMO // cc @klauspost - we only ever care about is `interval` -- we have no reason to take 10 different profiles at various intervals in succession.

klauspost: @harshavardhana 99% agree, but I kind of assumed you guys had discussed it already.

The 1% would be to grab goroutines at intervals. But mostly I don't see the point.

sinhaashish: cc @kannappanr

harshavardhana: Yeah only Goroutines make sense rest have little to no value in capturing in this manner

harshavardhana: The point of profiling is that it samples itself, we don't have to take many samples of profiles to compare. Interval is the sampler here - not the number of times we took the profile.

kannappanr: with count and interval, we will know when to stop and how many times to capture the profile

harshavardhana: yeah but its not going to be useful @kannappanr - it's a waste of capturing profiles.  all we care is `interval` not `count`

sinhaashish: @kannappanr If agreed then i can remove count and interval flag. The only flag which will be present will be `duration : start profiling for the specified duration in seconds` . This duration is the sleep time which we specify between start and stop profilers.

harshavardhana: AFAICS count is meaningless count is nothing but `count * duration` should be actual interval to be sent to the server, since sampling ratio is server-side for profiling, not client-side.

In a single sitting, no one will wait for 10 different profiles for the same interval.  There will be better data captured if we did `count * duration` instead

harshavardhana (approved):

anjalshireesh (changes requested):
Just running `mc support profile` without any arguments should show the usage. However, it currently fails with error `mc: <ERROR> Unable to initialize admin client. No valid configuration found for '' host alias.`

cmd/support-profile.go
```diff
@@ -18,27 +18,171 @@
...
+	profilers := strings.Split(strings.ToLower(ctx.String("type")), ",")
+	for _, profiler := range profilers {
+		if profiler != "" {
+			if !s.Contains(profiler) {
+				fatalIf(errDummy().Trace(ctx.String("type")),
+					"Profiler type %s unrecognized. Possible values are: %v.", profiler, s)
+			}
+		}
+	}
+	if len(ctx.Args()) == 0 || len(ctx.Args()) > 1 {
```
anjalshireesh, line 99:
```suggestion
	if len(ctx.Args()) != 1 {
```
This would be simpler

anjalshireesh (approved):
LGTM, one minor nit.

anjalshireesh (approved):
LGTM

PR #4064, opened 2022-04-27
Prefix version suspension

krisis:


harshavardhana (approved):

cmd/version-enable.go
```diff
@@ -28,13 +29,23 @@ import (
 	"github.com/minio/pkg/console"
 )
 
+var versionEnableFlags = []cli.Flag{
+	cli.StringFlag{
+		Name:  "excluded-prefixes",
```
vadmeste, line 34: s/excluded-prefixes/exclude-prefixes ?

cmd/version-enable.go
```diff
@@ -47,6 +58,13 @@ FLAGS:
 EXAMPLES:
   1. Enable versioning on bucket "mybucket" for alias "myminio".
      {{.Prompt}} {{.HelpName}} myminio/mybucket
+
+  2. Enable versioning on bucket "mybucket" while excluding versioning on a few select prefixes.
+     {{.Prompt}} {{.HelpName}} myminio/mybucket --excluded-prefixes "app1/*/_temporary/,app2/*/_staging/"
```
vadmeste, line 63: String array is easier for the users I guess:
`--exclude-prefix "app1/*/_temporary/" --exclude-prefix "app2/*/_staging/"`

vadmeste (commented):
Nit comments

cmd/version-enable.go
harshavardhana: Its written to mimic the XML @vadmeste

cmd/version-enable.go
harshavardhana: a comma-separated is fine for now I think

kannappanr (approved):
LGTM

PR #4065, opened 2022-04-28
Document setting multiple policies

ksonj:
This just adds an example for setting multiple policies for a user/group

### DIFF
--- a/cmd/admin-policy-set.go
+++ b/cmd/admin-policy-set.go
@@ -41,7 +41,7 @@ USAGE:
   {{.HelpName}} TARGET POLICYNAME [ user=username1 | group=groupname1 ]
 
 POLICYNAME:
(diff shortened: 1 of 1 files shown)


cmd/admin-policy-set.go
```diff
@@ -41,7 +41,7 @@ USAGE:
   {{.HelpName}} TARGET POLICYNAME [ user=username1 | group=groupname1 ]
 
 POLICYNAME:
-  Name of the policy on the MinIO server.
+  Name of the policy on the MinIO server. To set multiple policies separate names with a comma (,)
```
vadmeste, line 44:
```suggestion
  Name of the policy on the MinIO server. To set multiple policies, separate names with a comma (,).
```

vadmeste (approved):
LGTM

PR #4073, opened 2022-05-02
Introduce `--immediate` flag to support purging on removal

Praveenrajmani:
setting this flag would purge the objects immediately upon deletion

### DIFF
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -95,6 +95,14 @@ type CopyOptions struct {
 	storageClass     string
 }
 
+// RemoveOptions holds options for delete operation
+type RemoveOptions struct {
+	Incomplete   bool
+	RemoveBucket bool
+	Bypass       bool
+	Immediate    bool
+}
+
 // Client - client interface
 type Client interface {
 	// Common operations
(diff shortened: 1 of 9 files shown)


cmd/rm-main.go
```diff
@@ -255,7 +259,11 @@ func checkRmSyntax(ctx context.Context, cliCtx *cli.Context, encKeyDB map[string
 }
 
 // Remove a single object or a single version in a versioned bucket
-func removeSingle(url, versionID string, isIncomplete, isFake, isForce, isBypass bool, olderThan, newerThan string, encKeyDB map[string][]prefixSSEPair) error {
+func removeSingle(url, versionID string,
```
klauspost, line 262: This is completely impossible to read.

Can we pass options as:
```
type removeSingleOptions struct {
	versionID                       string
	incomplete, fake, force, bypass bool
	olderThan, newerThan            string
	encKeyDB                        map[string][]prefixSSEPair
}
```

cmd/client.go
```diff
@@ -138,7 +138,7 @@ type Client interface {
 	Watch(ctx context.Context, options WatchOptions) (*WatchObject, *probe.Error)
 
 	// Delete operations
-	Remove(ctx context.Context, isIncomplete, isRemoveBucket, isBypass bool, contentCh <-chan *ClientContent) (errorCh <-chan RemoveResult)
+	Remove(ctx context.Context, isIncomplete, isRemoveBucket, isBypass, IsImmediate bool, contentCh <-chan *ClientContent) (errorCh <-chan RemoveResult)
```
klauspost, line 141:
```
type RemoveOptions struct {
	Incomplete   bool
	RemoveBucket bool
	Bypass       bool
	Immediate    bool
}
```

Praveenrajmani:
closing this PR as we don't need to expose such special operations

PR #4076, opened 2022-05-05
Update: adding new flag --token when your AWS Credentials are temporary

shinitiandrei:
Adding new flag `--token` when your AWS Credentials are temporary. 
When you use SSO or OIDC for example, you gonna get a session token.
Example:
```./mc alias set s3 https://s3.amazonaws.com YourAccessKey YourSecretKey --token "YourVeryLongToken"```

### DIFF
--- a/cmd/admin-tier-add.go
+++ b/cmd/admin-tier-add.go
@@ -101,7 +101,7 @@ USAGE:
   {{.HelpName}} TYPE ALIAS NAME [FLAGS]
 
 TYPE:
-  Transition objects to supported cloud storage backend tier. Supported values are s3, azure and gcs.
+  Transition objects to supported cloud storage backend tier. Supported values are minio, s3, azure and gcs.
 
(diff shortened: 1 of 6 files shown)


harshavardhana:
This not needed @shinitiandrei you should just use https://github.com/minio/mc/blob/master/docs/minio-client-complete-guide.md#rotating-credentials

You should really ask us first on these type of changes, we can't really take these type of changes without prior discussion.

Token is not added or persisted on purpose since it is temporary.

shinitiandrei:
thanks @harshavardhana 
Sorry I didn't know about the existence of that docs, also when using the --help command it didn't provide further information on tokens.
Maybe you can add that information when typing `mc alias set --help`? also maybe adding that complete guide on the S3 section in https://github.com/minio/mc/blob/master/README.md would help.

Sorry to bother you, genuinely thought it didn't exist.

PR #4079, opened 2022-05-06
avoid cascading deletes for active-active setups

harshavardhana:
This PR in-turn ends up supporting cascading deletes
for `mc mirror` setups when you have
    
```
mc mirror --watch alias1 alias2
mc mirror --watch alias2 alias3
```
    
It still avoids cyclical deletes such as
    
```
mc mirror --watch alias1 alias2
mc mirror --watch alias2 alias1
```


kannappanr (approved):
LGTM

PR #4094, opened 2022-05-23
alias: Enable self signed TLS prompt in darwin platforms

vadmeste:
Not trusted TLS error message has another string representation in
Darwin platform. Since there is no another way to test for this specific
error with Golang API, we will need to support this alternate error message.


klauspost (approved):
LGTM

PR #4099, opened 2022-05-30
cleanup print message for 'support profile'

harshavardhana:
use single logging style, we do not use
`go` logger in `mc` avoid it.


sinhaashish (approved):

anjalshireesh (approved):
LGTM, tested

klauspost (approved):

PR #4104, opened 2022-06-03
Fix replicate status detail summary

poornas:
Remove PendingStatus from replication summary as this is no longer reported. Also refactoring to ensure target summary is associated with correct ARN

Fixes issue: https://github.com/minio/mc/issues/4105


harshavardhana (approved):

kaplanlior:
Thanks for the quick fix.

PR #4109, opened 2022-06-08
Add admin bucket info command

vadmeste:
Print bucket usage and configuration (encryption, policy, ILM, etc..)

e.g:

```
$ mc admin bucket info myminio/testbucket/
Usage:
      Total size: 17 MiB
   Objects count: 63
  Versions count: 63

Properties:
  Encryption: 
        Algorithm: AES256
        Key ID: 
  Versioning: Enabled
  LockConfiguration: 
    RetentionMode: GOVERNANCE
    Retention Until Date: 30DAYS
  Location: us-east-1
  Policy: readwrite
  ILM: Set
```

### DIFF
--- a/cmd/admin-bucket-info.go
+++ b/cmd/admin-bucket-info.go
@@ -1,0 +1,140 @@
+// Copyright (c) 2022 MinIO, Inc.
+//
+// This file is part of MinIO Object Storage stack
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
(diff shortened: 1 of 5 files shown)


harshavardhana:
Do what does mc stat do on the bucket ?

vadmeste:
> Do what does mc stat do on the bucket ?

Yes, plus the bucket usage information. We might need to discuss this internally. I am not sure if this is the exact requirement.

klauspost:
We have the time of the last partial update. Should we add that?

harshavardhana:
PTAL at the conflicts @vadmeste

vadmeste:
> We have the time of the last partial update. Should we add that?

I tried to do it, but it doesn't seem we have a last update of a particular bucket. Only when the scanner finished doing a scan. And it seems we already have that in the new mc admin scanner info command

poornas (approved):

PR #4110, opened 2022-06-09
Add `admin cluster data` command to backup bucket

poornas:


klauspost (commented):
Listing will be highly compressible. Let's do that.

Other minor stuff.

harshavardhana:
If the intent is to backup why don't we call it `mc admin cluster backup` ? Or simply `mc admin backup ` ?

klauspost:
@harshavardhana I completely agree that it needs a new name. Not sure if "backup" is the right name either. To me it implies a compact backup of a bucket to either local or remote storage.

It is a `mirror` operation, but with some restrictions.

Possibly `mc admin mirror` or a bit more hidden `mc admin bucket mirror`

`replicate` is also covering well, but is probably too confusing to use these since we already have it for serverside replication.

poornas:
Just discussed this with @abperiasamy, the backup requirement is on hold for now and not in line with this PR. As for the original intent of integrating the DR replication tool into mc -  `mc replicate heal` and `mc replicate diff` commands will call server side APIs to heal a prefix or list a diff of unreplicated content instead of performing this on the client side. Closing this PR for now - thanks for the feedback.

PR #4111, opened 2022-06-09
ignore remove errors for WORM protected objects

harshavardhana:
this error is similar to insufficient permission
on local folders, we can proceed to remove whatever
objects we can.


vadmeste (approved):
LGTM

PR #4125, opened 2022-06-16
Update version printing to include more information

vadmeste:
Show licence, copyright header and platform in addition to mc version.

```
$ ./mc --version
mc version RELEASE.2022-06-16T17-13-07Z (darwin/arm64)
Copyright (c) 2015-2022 MinIO, Inc.
Licence AGPLv3 <https://www.gnu.org/licenses/agpl-3.0.html>
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -453,11 +453,21 @@ var appCmds = []cli.Command{
 	updateCmd,
 }
 
+func printMCVersion(c *cli.Context) {
+	fmt.Fprintf(c.App.Writer, "%s version %s (commit-id=%s)\n", c.App.Name, c.App.Version, CommitID)
+	fmt.Fprintf(c.App.Writer, "Runtime: %s %s/%s\n", runtime.Version(), runtime.GOOS, runtime.GOARCH)
+	fmt.Fprintf(c.App.Writer, "Copyright (c) 2015-%s MinIO, Inc.\n", CopyrightYear)
+	fmt.Fprintf(c.App.Writer, "Licence AGPLv3 <https://www.gnu.org/licenses/agpl-3.0.html>\n")
+}
+
 func registerApp(name string) *cli.App {
 	cli.HelpFlag = cli.BoolFlag{
 		Name:  "help, h",
 		Usage: "show help",
 	}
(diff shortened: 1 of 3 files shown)


cmd/main.go
```diff
@@ -453,12 +453,21 @@ var appCmds = []cli.Command{
 	updateCmd,
 }
 
+func printMCVersion(c *cli.Context) {
+	fmt.Fprintf(c.App.Writer, "%s version %s (%s/%s)\n", c.App.Name, c.App.Version, runtime.GOOS, runtime.GOARCH)
+	fmt.Fprintf(c.App.Writer, "Copyright (c) 2015-2022 MinIO, Inc.\n")
```
harshavardhana, line 458: The end year should be dynamic @vadmeste ?

vadmeste: > The end year should be dynamic @vadmeste ?

yes, working on it

vadmeste: > yes, working on it

Done

cmd/main.go
```diff
@@ -453,12 +453,21 @@ var appCmds = []cli.Command{
 	updateCmd,
 }
 
+func printMCVersion(c *cli.Context) {
+	fmt.Fprintf(c.App.Writer, "%s version %s (%s/%s)\n", c.App.Name, c.App.Version, runtime.GOOS, runtime.GOARCH)
+	fmt.Fprintf(c.App.Writer, "Copyright (c) 2015-%s MinIO, Inc.\n", CopyrightYear)
+	fmt.Fprintf(c.App.Writer, "Licence AGPLv3 <https://www.gnu.org/licenses/agpl-3.0.html>\n")
+}
```
harshavardhana, line 460: We should also add `runtime.Version()` and as well as the `commit-id` I think - similar to `minio -v`

harshavardhana (changes requested):

cmd/main.go
vadmeste: > We should also add runtime.Version() and as well as the commit-id I think - similar to minio -v

Done

harshavardhana (approved):

PR #4132, opened 2022-06-23
change admin console to support logs show

sinhaashish:
the command `mc console log` is deprecated and is replaced with `mc support logs show` 

Now the `mc support logs` has 4 subcommands
```
$ ./mc support logs
NAME:
  mc support logs - configure/display MinIO console logs

USAGE:
  mc support logs COMMAND [COMMAND FLAGS | -h] [ARGUMENTS...]

COMMANDS:
  enable   enable uploading real-time MinIO logs to SUBNET
  disable  disable uploading MinIO logs to SUBNET
  status   show current status of MinIO logs
  show     show MinIO logs

```


cmd/support-logs-status.go
```diff
@@ -0,0 +1,53 @@
...
+
+package cmd
+
+import (
+	"github.com/minio/cli"
+)
+
+var supportLogsStatusCmd = cli.Command{
+	Name:         "status",
+	Usage:        "check logs status for cluster with alias 'play'",
```
harshavardhana, line 26:
```suggestion
	Usage:        "show current status of MinIO logs'",
```

cmd/support-logs-status.go
```diff
@@ -0,0 +1,53 @@
...
+	Flags:        logsConfigureFlags,
+	CustomHelpTemplate: `NAME:
+  {{.HelpName}} - {{.Usage}}
+USAGE:
+  {{.HelpName}} ALIAS
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}
+EXAMPLES:
+	1. Check logs status for cluster with alias 'play'
```
harshavardhana, line 39: some indentation issue please check `mc support logs status --help`

cmd/admin-console.go
```diff
@@ -47,173 +39,37 @@ var adminConsoleFlags = []cli.Flag{
...
-     {{.Prompt}} {{.HelpName}} play
-
-  2. Show last 5 log entries for node 'node1' on MinIO server with alias 'myminio'
-     {{.Prompt}} {{.HelpName}} --limit 5 myminio node1
-
-  3. Show application error logs on MinIO server with alias 'play'
-     {{.Prompt}} {{.HelpName}} --type application play
-`,
+	Name:               "console",
+	Usage:              "show console logs for MinIO server",
```
harshavardhana, line 43:
```suggestion
	Usage:              "show MinIO logs",
```

cmd/support-logs-print.go
```diff
@@ -0,0 +1,215 @@
...
+	},
+	cli.StringFlag{
+		Name:  "type, t",
+		Usage: "list error logs by type. Valid options are '[minio, application, all]'",
+		Value: "all",
+	},
+}
+
+var supportLogsPrintCmd = cli.Command{
+	Name:            "print",
```
harshavardhana, line 50:
```suggestion
	Name:            "show",
```

More like 

```
mc support logs show alias/
```

harshavardhana (changes requested):

docs/minio-client-complete-guide.md
```diff
@@ -1953,4 +1954,14 @@ mc support diag play
 Get CPU profiling for 2 minutes
 ```
 mc support profile  --type cpu --duration 120 myminio/
+```
+
+Print last 5 application error logs entries for node 'node1' on MinIO server with alias 'myminio'
+```
+mc support logs print --limit 5 --type application myminio node1
```
harshavardhana, line 1961:
```suggestion
mc support logs show --last 5 --type application myminio node1
```

harshavardhana (changes requested):
--limit --> `--last` - since it going to provide last `n` log entries.

cmd/admin-console.go
```diff
@@ -47,173 +39,35 @@ var adminConsoleFlags = []cli.Flag{
...
-	return fmt.Sprintf("%s\n", logMsg)
+	Name:               "console",
+	Usage:              "show MinIO logs",
+	Action:             mainAdminConsole,
+	OnUsageError:       onUsageError,
+	Before:             setGlobalsFromContext,
+	Flags:              append(adminConsoleFlags, globalFlags...),
+	Hidden:             true,
+	HideHelpCommand:    true,
+	CustomHelpTemplate: "This command is deprecated and will be removed in a future release. Use 'mc support logs show' instead.\n",
```
ebozduman, line 50: As the custom help says here, `mc admin console` command is only deprecated, we should still support it with a warning message like this. I think this is the right behavior until the time comes to kill the command completely.

But the current behavior is in between not supported "killed" state and still alive with a warning message state.

So, if the user uses the old command with some mistake in defining the flags, we show full supported usage with  adminConsole and global flags (because code in lines 28~39 is still live).
```
$ mc admin console --typ
mc: <ERROR> Invalid command usage, flag provided but not defined: -typ

SUPPORTED FLAGS:
   --limit value, -l value       show last n log entries (default: 10)
   --type value, -t value        list error logs by type. Valid options are '[minio, application, all]' (default: "all")
   --config-dir value, -C value  path to configuration folder (default: "/home/ersan/.mc")
   --quiet, -q                   disable progress bar display
   --no-color                    disable color theme
   --json                        enable JSON lines formatted output
   --debug                       enable debug output
   --insecure                    disable SSL certificate verification
   --help, -h                    show help
```

If the user chooses to get `--help`, then we display this warning message:
```
$ mc admin console -h
This command is deprecated and will be removed in a future release. Use 'mc support logs show' instead.
```

If the user puts together the old command in correct syntax, the command fails and we recommend the new command to be used, that is basically means the old command is already dead.
```
$ mc admin console --type all myminio
mc: Please use 'mc support logs show --type all myminio'
```

We should adjust our behavior and choose either deprecation or dead/not supported state in all these little scenarios.

cmd/support-logs-disable.go
```diff
@@ -0,0 +1,49 @@
...
+	Flags:        logsConfigureFlags,
+	CustomHelpTemplate: `NAME:
+  {{.HelpName}} - {{.Usage}}
+USAGE:
+  {{.HelpName}} ALIAS
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}
+EXAMPLES:
+  1. Disable logs for cluster with alias 'play'
```
ebozduman, line 39: Please match the more meaningful `Usage` message.

cmd/support-logs-enable.go
```diff
@@ -0,0 +1,50 @@
...
+	HideHelpCommand: true,
+	CustomHelpTemplate: `NAME:
+  {{.HelpName}} - {{.Usage}}
+USAGE:
+  {{.HelpName}} ALIAS
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}
+EXAMPLES:
+  1. Enable logs for cluster with alias 'play'
```
ebozduman, line 40: Please match the more meaningful `Usage` message.

ebozduman (changes requested):

cmd/admin-console.go
sinhaashish: Thank for the detailed analysis @ebozduman . The moto of this PR is to replace `mc admin console` with `mc support logs show`. We are stopping the use of old command and asking users to switch to new command. There is no deprecation. 

1. Corrected the help command for `mc admin console`
```
ashish@ashish:~/code/go/src/github/minio/mc$ ./mc admin console -h
This command is replace by 'support logs show' command. Please use 'mc support logs show'.
```
2. If some one type a wrong flag then they need to correct it first and then we show the new command. 

3. The limit flag is replace by last . so if some one writes `mc admin console --limit 5 myminio2` , the suggestion shows the correct command `lease use 'mc support logs show --last 5 myminio2`. 
 ```
ashish@ashish:~/code/go/src/github/minio/mc$ ./mc admin console --limit 5 myminio2
mc: Please use 'mc support logs show --last 5 myminio2'
```

harshavardhana (approved):

PR #4143, opened 2022-07-02
reset timer after the function body is finished 

harshavardhana:
this is needed to make sure to wait "expected" amount
of time, otherwise the timer has already fired (negating
the wait of your function body)


vadmeste (approved):
LGTM

PR #4147, opened 2022-07-06
fix unable to validate source error

### LINKED ISSUE #4088: copy object rewind gives error unable to validate source
I needed to roll-back an object foo.
I tried to call cp with `--rewind` which resulted in an error.
From documentation of the flag I would suspect that something like this should be possible
`  --rewind value                     roll back object(s) to current version at specified time`

## Expected behavior
`mc cp  --rewind 1s s3/mlad-test/foo s3/mlad-test-2/`
shall rollback to version of foo older than 1s

Interestingly enough, it works in combination with --recursive when setting the SOURCE as bucket
```
$ mc cp  --rewind 1s --recursive s3/mlad-test/ s3/mlad-test-2/
`s3/mlad-test/foo` -> `s3/mlad-test-2/foo`
```
## Actual behavior
```
$ mc cp  --rewind 1s s3/mlad-test/foo s3/mlad-test-2/
mc: <ERROR> Unable to validate source `s3/mlad-test/foo`.
```
```
mc mb s3/mlad-test
mc mb s3/mlad-test-2
touch foo
mc cp foo s3/mlad-test
mc version enable s3/mlad-test
mc cp foo s3/mlad-test/foo
mc ls --versions s3/mlad-test/foo
```
## mc --version
`mc version RELEASE.2022-05-09T04-08-26Z`
## System information
I am using docker version of the the mc command.
The endpoint is Amazon AWS.

dormanze:
fix #4088 
I suggest deleting this content, because if I pass a file instead of a directory and take the --rewind parameter, it will add a delimiter to cause  The s3.ListObjectVersions interface returned an empty object.
image
is ok:
image
not ok:
image

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -1589,9 +1589,9 @@ func (c *S3Client) Stat(ctx context.Context, opts StatOptions) (*ClientContent, 
 
 	// No object found, start looking for a prefix with the same name
 	// or a directory marker. Add a trailing slash if it is not in the path
-	if !strings.HasSuffix(path, string(c.targetURL.Separator)) {
-		path += string(c.targetURL.Separator)
-	}
+	//if !strings.HasSuffix(path, string(c.targetURL.Separator)) {
+	//	path += string(c.targetURL.Separator)
+	//}
 
 	nonRecursive := false
 	maxKeys := 1


dormanze:
@vadmeste @harshavardhana Please pay attention to this problem.This will cause mc stat with --rewind to return null.
image
When calling the stat interface when mc ls --rewind 1s , the rewind parameter is not passed.
image

harshavardhana (commented):
Not sure what you are trying to fix but this is not a correct fix.

PR #4149, opened 2022-07-08
Add mc cat --tail and --offset

klauspost:
Allow range requests for cat operations, either specifying tail or an absolute offset.

If tail is longer than file, the entire file is listed.

Example:
```
λ mc cat --tail=100 play/testbucket/go1/src/README.vendor
e current module. If a new package is needed,
it should be imported before running 'go mod vendor'.
```

### DIFF
--- a/cmd/cat-main.go
+++ b/cmd/cat-main.go
@@ -47,6 +47,14 @@ var catFlags = []cli.Flag{
 		Name:  "zip",
 		Usage: "Extract from remote zip file (MinIO server source only)",
 	},
+	cli.Int64Flag{
+		Name:  "offset",
+		Usage: "Start offset",
+	},
+	cli.Int64Flag{
+		Name:  "tail",
+		Usage: "Tail number of bytes at ending of file",
+	},
 }
 
 // Display contents of a file.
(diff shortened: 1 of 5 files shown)


cmd/cat-main.go
```diff
@@ -165,11 +173,23 @@ func parseCatSyntax(ctx *cli.Context) (args []string, versionID string, timeRef
...
+	}
+	if isZip && (tailO != 0 || startO != 0) {
+		fatalIf(errInvalidArgument().Trace(), "You cannot combine --zip with --tail or --offset")
+	}
 	return
 }
 
 // catURL displays contents of a URL to stdout.
-func catURL(ctx context.Context, sourceURL, sourceVersion string, timeRef time.Time, encKeyDB map[string][]prefixSSEPair, isZip bool) *probe.Error {
+func catURL(ctx context.Context, sourceURL, sourceVersion string, timeRef time.Time, encKeyDB map[string][]prefixSSEPair, isZip bool, startO, tailO int64) *probe.Error {
```
harshavardhana, line 192:
```suggestion
func catURL(ctx context.Context, sourceURL, sourceVersion string, timeRef time.Time, encKeyDB map[string][]prefixSSEPair, isZip bool, startO, tailO int64) *probe.Error {
```

Perhaps time to convert this into `opts` struct.

harshavardhana (changes requested):
should we do an `mc tail` command instead - compliments with `mc head` ?

cmd/cat-main.go
klauspost: Added 👍🏼  Was getting annoyed myself.

harshavardhana (approved):

PR #4150, opened 2022-07-11
Move `support register` to `license register`

anjalshireesh:
Also introduce a new command `license info` to print information from
the license (when license is available)


sinhaashish:
Dont you think that removing `support register` to  `license register` will cause confusion without any deprecation info for `support register` ?

anjalshireesh:
> > Dont you think that removing `support register` to `license register` will cause confusion without any deprecation info for `support register` ?
> 
> Yes, good idea to print a message on `support register` asking user to use `license register`. Will implement.

@sinhaashish this is done. PTAL.

sinhaashish (approved):
LGTM

harshavardhana (approved):

PR #4151, opened 2022-07-12
Move mc admin top scanner -> mc admin scanner info

klauspost:
Also add `mc admin scanner trace` which gives simplified trace with only relevant options.


harshavardhana (approved):

vadmeste (approved):
LGTM

PR #4155, opened 2022-07-14
Add `license update` command

anjalshireesh:
Accepts the license token, validates it, and stores in the minio config.

### DIFF
--- a/cmd/license-update.go
+++ b/cmd/license-update.go
@@ -19,6 +19,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/fatih/color"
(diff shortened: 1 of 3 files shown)


cmd/license-update.go
```diff
@@ -0,0 +1,137 @@
...
+	li, e := parseLicense(lic, airgap)
+	if e != nil {
+		lum.Status = "error"
+		lum.Error = e.Error()
+		return lum
+	}
+
+	if li.ExpiresAt.Before(time.Now()) {
+		lum.Status = "error"
+		lum.Error = "License has expired"
```
harshavardhana, line 131:
```suggestion
		lum.Error = fmt.Sprintf("License has expired on %s", li.ExpiresAt)
```

cmd/license-update.go
```diff
@@ -0,0 +1,137 @@
...
+USAGE:
+  {{.HelpName}} ALIAS LICENSE-TOKEN
+
+FLAGS:
+  {{range .VisibleFlags}}{{.}}
+  {{end}}
+
+EXAMPLES:
+  1. Update license for cluster with alias 'play'
+     {{.Prompt}} {{.HelpName}} play eyJ0eXAiOiJKV1QiLCJhbGciOiJFUzM4NCJ9.eyJsaWQiOiJjMGRlNDk4OC0wYzA3LTRiYTMtYjY5NS00MDQ2NjllZDQwMWYiLCJzdWIiOiJzaGlyZWVzaCtjMSthZG1pbkBtaW5pby5pbyIsImNhcCI6MjAwMDAsIm9yZyI6IlNoaXJlZXNoLUMxIiwiaXNzIjoic3VibmV0QG1pbi5pbyIsImFpZCI6MSwiaWF0IjoxLjY1NzA4NDEwNDAzNjQxMzQ3MmU5LCJwbGFuIjoiRU5URVJQUklTRSIsImV4cCI6MS42ODg2MjAxMDQwMzY0MTM0NzJlOSwiZGlkIjoiY2Q2NTIwZTctYjg2Ny00YWU2LTkyY2EtNDc5MWI2OWEwY2M3In0.Ya9_HSpog8EhPY1Ckcay5J70_Rms1dNnu4xNlKvwy-8fF6lyF2bqsQMvuDOKaCIYj5w4May8l-1SJ5tC2mQ9Z_ycgCVWpwHGx6h2b7EOAtjGiN6yFEWBLedEScUx34u8
```
harshavardhana, line 50: I think pasting an entire JWT won't really be good for the bash prompt. I think we should take the input slightly differently in this.

This also looks like something is wrong here that we are taking in JWT in this manner.

harshavardhana: It's very easy to get JWTs wrong when taken as input, because of their sheer length and they wrap around terminals. We have to be careful about this visually as this can simply lead to an uptick in support activity.

harshavardhana (changes requested):
was this already decided to have a separate `update` sub-command? - this does look like a flag for `register` to me? 

I am not really opposed to this but I am just wondering.

cmd/license-update.go
anjalshireesh: Yes, will change to accept it as input after running the command.

anjalshireesh:
> was this already decided to have a separate `update` sub-command? - this does look like a flag for `register` to me?
> 
> I am not really opposed to this but I am just wondering.

Yes it has been discussed to add a separate command to handle the new flow of offline registration. ref: https://github.com/minio/mc/pull/4156

cmd/license-update.go
anjalshireesh: Done

harshavardhana (approved):

PR #4160, opened 2022-07-20
Update examples for admin-config-reset

taran-p:
Update examples to reflect minio server-side features


donatello (changes requested):
The USAGE string should say to indicate an optional set of sub-keys may be given:

```
{{.HelpName}} TARGET [CONFIG-KEY...]
```

donatello (changes requested):
Also, we need to add some validation to avoid this type of error:

```
$ mc admin config reset myminio scanner max_wait=30s
mc: <ERROR> Unable to reset 'scanner max_wait=30s' on the server: key max_wait=30s doesn't exist.
```

In this case, `mc` should validate the command and say that new settings may not be provided for sub-system keys.

harshavardhana (approved):

donatello (approved):

PR #4161, opened 2022-07-21
Validate license against deployment id

anjalshireesh:
The command `mc license update` should error out in case the deployment
id from the license doesn't match the cluster's.

### DIFF
--- a/cmd/license-update.go
+++ b/cmd/license-update.go
@@ -109,6 +109,10 @@ func performLicenseUpdate(licFile string, alias string, airgap bool) licUpdateMe
 		fatalIf(errDummy().Trace(), fmt.Sprintf("License has expired on %s", li.ExpiresAt))
 	}
 
+	if li.DeploymentID != getAdminInfo(alias).DeploymentID {
+		fatalIf(errDummy().Trace(), fmt.Sprintf("Invalid license"))
+	}
+
 	setSubnetCreds(alias, "", lic)
 	return lum
 }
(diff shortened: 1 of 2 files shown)


kanagarajkm (approved):

cmd/license-update.go
```diff
@@ -109,6 +109,10 @@ func performLicenseUpdate(licFile string, alias string, airgap bool) licUpdateMe
 		fatalIf(errDummy().Trace(), fmt.Sprintf("License has expired on %s", li.ExpiresAt))
 	}
 
+	if li.DeploymentID != getAdminInfo(alias).DeploymentID {
+		fatalIf(errDummy().Trace(), fmt.Sprintf("Invalid license"))
```
harshavardhana, line 113:
```suggestion
		fatalIf(errDummy().Trace(), fmt.Sprintf("Invalid license"))
```

should be we more prescriptive about the error on why the license is invalid here?

harshavardhana (changes requested):

cmd/license-update.go
anjalshireesh: Isn't it better to keep credentials related error messages vague for security reasons? Providing information when license has expired is fine, but if it's not a valid license for the deployment, I feel it's better to just say it's invalid.

harshavardhana: Yeah but I was thinking we should say something like

```
license is invalid for the deployment '%s' 
```

Is what I would expect as an error, %s is just the alias name.

anjalshireesh: sure, will do

harshavardhana (approved):

PR #4172, opened 2022-07-29
fix: 'mc stat' should exit properly if server returns 503

harshavardhana:


vadmeste:
@harshavardhana can you elaborate what this PR fixes exactly ? I know there is another related PR in minio/minio but still it is not clear how this PR is related

harshavardhana:
> @harshavardhana can you elaborate what this PR fixes exactly ? I know there is another related PR in minio/minio but still it is not clear how this PR is related

IT fixes a situation where `mc stat alias/bucket/obj` if it returns 503 that gets displayed - the reason for this is because it is skipped by `isURLPrefixExists()`  because this returns success instead.

vadmeste:
I think you can remove isURLPrefixExists() overall. It is always called after url2Stat(), the latter returns success if the prefix exists as well, 'prefix' or 'prefix/'

I can send a PR if you want

klauspost (approved):

harshavardhana:
> I think you can remove isURLPrefixExists() overall. It is always called after url2Stat(), the latter returns success if the prefix exists as well, 'prefix' or 'prefix/'
> 
> I can send a PR if you want

@vadmeste sure that can be done it's a larger change to look at

PTAL the referenced PR on for MinIO as we need to make a new release.

PR #4180, opened 2022-08-01
Add `od` command

taran-p:
`od` mimics the `dd` command in linux terminals. It copies a given local file to a given remote location in a specified number of parts and part sizes, and returns the time it took to upload the file. 

This screenshot shows different ways to use `mc od` on a 92MiB file.
Screenshot from 2022-08-01 17-59-31


cmd/od-main.go
```diff
@@ -0,0 +1,235 @@
...
+
+func mainOD(cliCtx *cli.Context) error {
+	ctx, cancelCopy := context.WithCancel(globalContext)
+	defer cancelCopy()
+
+	if !cliCtx.Args().Present() {
+		cli.ShowCommandHelpAndExit(cliCtx, "mb", 1) // last argument is exit code
+	}
+
+	mapArgs, err := parseKVArgs(strings.Join(cliCtx.Args(), ","))
```
harshavardhana, line 189: Use a more typed representation of parsed values using, map is buggy and can lead to silent bugs hard to debug.

taran-p: Would using madmin KVS make sense for this?

harshavardhana:
Some issues

Issue1: Not allowing non-multipart uploads

```
~ mc od if=/etc/hosts of=play/testbucket/hosts parts=1 
mc: <ERROR> Unable to put target stream. Input part size is smaller than allowed minimum of 5MiB.
```

This is not correct a single part upload will work < 5MiB - this error is premature - you shouldn't be setting the PartSize here because we didn't provide the size and it must not be set.

Size must be honored only if it's set.

Issue2: 
```
~ mc od if=play/testbucket/hosts of=bigfile1 parts=1
mc: <ERROR> Unable to get source and target URLs: invalid target path bigfile1.
```

We need to support copy back from the server and parts when specified would download N parts from a file if not specified it downloads the entire object from Object Storage to the local disk.

If the part is specified GetObject() actually as PartNumber configuration to get individual parts.

donatello:
This is still now working:
```
$ ./mc od if=/etc/issue of=play/bucket/test
mc: <ERROR> Unable to put object. Input part size is smaller than allowed minimum of 5MiB.
```

I checked the server->server (same server), server->local, local->local and local->server. All seem to work fine. 

Output looks like this:
```
Transferred: 23 MiB, Parts: 1, Time: 9.15s
```

We can add an MiB/s to show the speed here too.

donatello (approved):
Working as expected

harshavardhana (approved):

cmd/client.go
```diff
@@ -173,6 +173,10 @@ type Client interface {
 
 	// Restore an object
 	Restore(ctx context.Context, versionID string, days int) *probe.Error
+
+	// OD operations
+	ODGet(ctx context.Context, part int) (io.ReadCloser, *probe.Error)
+	PutN(ctx context.Context, reader io.Reader, size int64, progress io.Reader, opts PutOptions) (n int64, err *probe.Error)
```
harshavardhana, line 179: if this is specific to OD rename each functions appropriately, you can write it as 

```
GetPart()
PutPart()
```

harshavardhana (changes requested):

harshavardhana (approved):

PR #4185, opened 2022-08-04
Add support for inspect data v2

vadmeste:
requires https://github.com/minio/madmin-go/pull/104

The new format is zip file which contains 'data.zip' which is encrypted;
'data.zip' will content the data from the server backend disks.

The code is also backward compatible with the older inspect format as
well


go.mod
```diff
@@ -108,10 +108,12 @@ require (
...
-	golang.org/x/sys v0.0.0-20220811171246-fbc7d0a398ab // indirect
+	golang.org/x/sync v0.0.0-20220722155255-886fb9371eb4 // indirect
+	golang.org/x/sys v0.0.0-20220928140112-f11e5e49a4ec // indirect
 	google.golang.org/genproto v0.0.0-20220302033224-9aa15565e42a // indirect
 	google.golang.org/grpc v1.44.0 // indirect
 	google.golang.org/protobuf v1.27.1 // indirect
 	gopkg.in/ini.v1 v1.66.6 // indirect
 )
+
+replace github.com/minio/madmin-go => github.com/klauspost/madmin-go v1.0.15-0.20221012163854-fd15cf0ad215
```
klauspost, line 119: This is released now, so upgrading to latest madmin should get rid of this.

klauspost:
@vadmeste Thinking about it maybe we should add a `--legacy` switch, that will request old format?

vadmeste:
> @vadmeste Thinking about it maybe we should add a --legacy switch, that will request old format?

Done. I also tested with older MinIO version. This will necessarly return a key for the v1 format. I believe this is also what we want as well.

klauspost (approved):

harshavardhana (approved):

PR #4190, opened 2022-08-08
restart: Use health check instead of service info API

vadmeste:
mc admin service restart currently uses ServerInfo() API which is
costly.

Use healthcheck with read quorum instead.

### DIFF
--- a/cmd/admin-service-restart.go
+++ b/cmd/admin-service-restart.go
@@ -106,6 +106,9 @@ func mainAdminServiceRestart(ctx *cli.Context) error {
 	// Validate serivce restart syntax.
 	checkAdminServiceRestartSyntax(ctx)
 
+	ctxt, cancel := context.WithCancel(globalContext)
+	defer cancel()
+
 	// Set color.
 	console.SetColor("ServiceOffline", color.New(color.FgRed, color.Bold))
 	console.SetColor("ServiceInitializing", color.New(color.FgYellow, color.Bold))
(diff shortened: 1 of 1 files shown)


cmd/admin-service-restart.go
```diff
@@ -144,22 +152,22 @@ func mainAdminServiceRestart(ctx *cli.Context) error {
...
+		case <-ctxt.Done():
+			return ctxt.Err()
 		case <-timer.C:
-			ctx, cancel := context.WithTimeout(globalContext, 3*time.Second)
-			// Fetch the service status of the specified MinIO server
-			info, e := client.ServerInfo(ctx)
-			cancel()
+			healthCtx, healthCancel := context.WithTimeout(ctxt, 3*time.Second)
+			// Fetch the health status of the specified MinIO server
+			healthResult, healthErr := anonClient.Healthy(healthCtx, madmin.HealthOpts{ClusterRead: true})
```
harshavardhana, line 160: We should only use Write since that is what everyone is interested

harshavardhana (changes requested):

cmd/admin-service-restart.go
vadmeste: > We should only use Write since that is what everyone is interested

Yes, we don't have an API for that yet. I can add '/minio/health/cluster/write' for that purpose, what do you think?

harshavardhana: If you don't have `ClusterRead` set then its Write @vadmeste `/minio/health/cluster` returns 200 OK if there is write quorum.

vadmeste: > If you don't have ClusterRead set then its Write @vadmeste /minio/health/cluster returns 200 OK if there is write quorum.

Yes right, updated the PR

harshavardhana (approved):

PR #4191, opened 2022-08-08
rename  mc admin top  to mc  support top 

sinhaashish:
The `mc admin top` commands has been rename to `mc support top` command.
There is no message for deprecation , instead the old command suggest to use the new command as shown below 
```
$ ./mc admin top api 
mc: Please use 'mc support top api'
$ ./mc admin top api  -h
Please use 'mc support top api'
$ ./mc admin top locks
mc: Please use 'mc support top locks'
$ ./mc admin top locks -h
Please use 'mc support top locks'

```


harshavardhana (approved):

vadmeste (approved):
LGTM

PR #4193, opened 2022-08-08
Fix example description for `mc support perf`

anjalshireesh:
The current description seems incomplete as it doesn't mention object
storage performance.


vadmeste (approved):
LGTM

PR #4199, opened 2022-08-10
sort by Rx or Tx in admin top api command

sinhaashish:
```
$ ./mc admin top api --sort-by-rx myminio

API                             RX      TX      CALLS   ERRORS 
s3.PutObject                    76 MiB  27 KiB  84      0     
s3.HeadBucket                   231 B   942 B   3       0     
s3.GetBucketLocation            77 B    444 B   1       0     
s3.GetBucketObjectLockConfig    77 B    649 B   1       0

### DIFF
--- a/cmd/top-api-spinner.go
+++ b/cmd/top-api-spinner.go
@@ -79,6 +79,8 @@ type traceUI struct {
 	result      topAPIResult
 	lastResult  topAPIResult
 	apiStatsMap map[string]*topAPIStats
+	sortByRx    bool
+	sortByTx    bool
 }
 
 type topAPIResult struct {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cmd/top-api-spinner.go [hunk 1]
```diff
@@ -174,15 +178,23 @@ func (m *traceUI) View() string {
 	for k, stats := range m.apiStatsMap {
 		data = append(data, []string{
 			k,
-			whiteStyle.Render(humanize.IBytes(stats.loadAPIBytesRX())),
```
harshavardhana, line 177: Is removal of coloring expected here ?

harshavardhana: You should sort this before if necessary, otherwise it's simply expensive to parse the strings again and again.

harshavardhana (changes requested):

klauspost (commented):
`--sort=RX` would be cleaner IMO. Also eliminates the possibility of both being selected, and you can add `CALLS` and `ERRORS` as well.

PR #4201, opened 2022-08-10
Introduce disk top IO

vadmeste:
This shows IO stats of MinIO backend disks

Depends on https://github.com/minio/madmin-go/pull/111

### DIFF
--- a/cmd/auto-complete.go
+++ b/cmd/auto-complete.go
@@ -436,6 +436,7 @@ var completeCmds = map[string]complete.Predictor{
 	"/support/status":       aliasCompleter,
 	"/support/top/locks":    aliasCompleter,
 	"/support/top/api":      aliasCompleter,
+	"/support/top/disk":     aliasCompleter,
 
 	"/license/register": aliasCompleter,
 	"/license/info":     aliasCompleter,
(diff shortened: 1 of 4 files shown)


harshavardhana:
How does it look do you have an asciinema?

cmd/admin-top-disk.go
```diff
@@ -0,0 +1,129 @@
...
+	cli.IntFlag{
+		Name:  "count, c",
+		Usage: "show the maximum top disks",
+		Value: 10,
+	},
+}
+
+var adminTopDiskCmd = cli.Command{
+	Name:            "disk",
+	Usage:           "Show disk stats in real-time",
```
harshavardhana, line 42:
```suggestion
	Usage:           "show current disk statistics",
```

harshavardhana (changes requested):

vadmeste:
> How does it look do you have an asciinema?

https://asciinema.org/a/hDH7gZXzYK5Q9BIuvK0XlVUeW

I actually failed to find 4 different disks. Overall there is sorting by column, disk usage, tps, read/write speed, await and utilisation.

There is also left right arrow to change the current pool

harshavardhana:
> I actually failed to find 4 different disks. Overall there is sorting by column, disk usage, tps, read/write speed, await and utilisation.

Well you can do `losetup`  @vadmeste

harshavardhana (approved):

Issue #4179, opened 2022-08-01
Error message for admin user policy can be more clear.


bh4t:
## Expected behavior
Error message could be "Policy not found for the user  username"

## Actual behavior
The current error message is displayed as below:

./mc admin user policy myminio preeti2
mc: <ERROR> Unable to fetch user policy document. Only a single policy may be specified here.

## Steps to reproduce the behavior
- Create a user
- Run the command admin user policy myminio username

## mc --version
RELEASE.2022-07-29T19-17-16Z

kannappanr:
fixed by #4206

PR #4206, opened 2022-08-12
Clarify error when user has no policy

taran-p:
Changes error when `mc admin user policy` is used on a user with no policy

Changed from `Unable to fetch user policy document. Only a single policy may be specified here.` to 'Unable to fetch user policy document. Policy not found for user username.`

Fixes issue #4179


vadmeste (approved):
LGTM

harshavardhana (approved):

PR #4208, opened 2022-08-14
fix: ping must call AdminInfo only with '--distributed'

harshavardhana:


harshavardhana:
PTAL @sinhaashish @kannappanr

sinhaashish (approved):
Tested & LGTM

PR #4209, opened 2022-08-15
Add --disable, --enable and --prefix to ilm commands

vadmeste:
Passing the prefix in the aliased url is problematic for ilm edit
command when the user wants to modify something in the ILM rule: no
prefix in the url means the user wants to set an empty prefix or maybe
nothing. To fix this, add --prefix flag and make it a default choice in
the document.

Also, ilm edit will try to only apply passed flags. LifecycleOptions
fields are all pointers to apply only non nil fields to the ILM rule.


harshavardhana (approved):

harshavardhana:
We may need update of docs as well @vadmeste

PR #4211, opened 2022-08-17
ping: Show DNS resolution time

vadmeste:
```
$ ./mc ping play           
1: https://play.min.io:443   min=644.476ms   max=644.476ms   avg=644.476ms   dns=148.371ms   errors=0   roundtrip=644.476ms
2: https://play.min.io:443   min=207.834ms   max=644.476ms   avg=426.155ms   dns=0s   errors=0   roundtrip=207.834ms
```

### DIFF
--- a/cmd/ping.go
+++ b/cmd/ping.go
@@ -113,11 +113,11 @@ var colorMap = template.FuncMap{
 }
 
 // PingDist is the template for ping result in distributed mode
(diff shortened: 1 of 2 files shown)


klauspost:
Since we now have DNS separately, we should exclude it from the roundtrip time. This will make it more constant (and match regular ping operations as well, which resolves DNS prior to performing the ping).

vadmeste:
> Since we now have DNS separately, we should exclude it from the roundtrip time. This will make it more constant (and match regular ping operations as well, which resolves DNS prior to performing the ping).

Yes, roundtrip already doesn't count the DNS resolution here. Basically here, min, max, avg doesn't count DNS resolution but yeah you just flagged a UX issue. 

I wanted to send DNS info here because sometimes people mistakenly set DNS entries TTL to a very low level, such as 5 seconds or something like that and this would help us flag the issue.

sinhaashish:
Have incorporated these in PR https://github.com/minio/mc/pull/4218. the DNS resolution time is only shown in json response .

vadmeste:
@sinhaashish you can simply add the missing dns in your PR

PR #4214, opened 2022-08-18
trim response to two-decimal places in ping command

sinhaashish:
This PR parses the string o/p and then trims the result to two decimal places. A better approach than this will be really good to have .

### DIFF
--- a/cmd/ping.go
+++ b/cmd/ping.go
@@ -29,7 +29,6 @@ import (
 	"text/tabwriter"
 	"text/template"
 	"time"
-	"unicode"
 
 	"github.com/fatih/color"
 	"github.com/minio/cli"
@@ -218,12 +217,12 @@ func ping(ctx context.Context, cliCtx *cli.Context, anonClient *madmin.Anonymous
 		stat := getPingInfo(cliCtx, result, endPointMap)
 		endPointStat := EndPointStats{
 			Endpoint:  endPoint,
-			Min:       trimToTwoDecimal(time.Duration(stat.min).Round(time.Microsecond).String()),
-			Max:       trimToTwoDecimal(time.Duration(stat.max).Round(time.Microsecond).String()),
-			Average:   trimToTwoDecimal(time.Duration(stat.avg).Round(time.Microsecond).String()),
+			Min:       trimToTwoDecimal(time.Duration(stat.min).Round(time.Microsecond)),
+			Max:       trimToTwoDecimal(time.Duration(stat.max).Round(time.Microsecond)),
+			Average:   trimToTwoDecimal(time.Duration(stat.avg).Round(time.Microsecond)),
 			CountErr:  strconv.Itoa(stat.errorCount),
 			Error:     stat.err,
-			Roundtrip: trimToTwoDecimal(result.ResponseTime.Round(time.Microsecond).String()),
+			Roundtrip: trimToTwoDecimal(result.ResponseTime.Round(time.Microsecond)),
 		}
 		endPointStats = append(endPointStats, endPointStat)
 		endPointMap[result.Endpoint.Host] = stat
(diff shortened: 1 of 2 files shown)


cmd/ping.go
```diff
@@ -236,6 +238,27 @@ func ping(ctx context.Context, cliCtx *cli.Context, anonClient *madmin.Anonymous
 	time.Sleep(time.Duration(cliCtx.Int("interval")) * time.Second)
 }
 
+func trimToTwodecimal(inputString string) string {
```
harshavardhana, line 241: If you really want to do this we should basically simply try to always print `x seconds`  - https://go.dev/play/p/-vHAK0SrF8v

```
package main

import (
	"fmt"
	"time"
)

func main() {
	start := time.Now()
	time.Sleep(time.Duration(146198861))
	fmt.Printf("Execution time: %.2f s", time.Since(start).Seconds())
}
```

harshavardhana (changes requested):

cmd/ping.go
sinhaashish: I had already tried this but i didn't like the o/p as shown below . If we agree to always print in seconds then i can make the change.   
```
ashish@ashish:~/code/go/src/github/minio/mc$ ./mc ping play
1: https://play.min.io:443   min=1.77s   max=1.77s   average=1.77s   errors=0   roundtrip=1.77s
2: https://play.min.io:443   min=1.00s   max=1.77s   average=1.38s   errors=0   roundtrip=1.00s
3: https://play.min.io:443   min=0.31s   max=1.77s   average=1.02s   errors=0   roundtrip=0.31s
4: https://play.min.io:443   min=0.31s   max=1.86s   average=1.23s   errors=0   roundtrip=1.86s
5: https://play.min.io:443   min=0.30s   max=1.86s   average=1.05s   errors=0   roundtrip=0.30s
```

cmd/ping.go
```diff
@@ -236,6 +238,27 @@ func ping(ctx context.Context, cliCtx *cli.Context, anonClient *madmin.Anonymous
 	time.Sleep(time.Duration(cliCtx.Int("interval")) * time.Second)
 }
 
+func trimToTwoDecimal(inputString string) string {
```
vadmeste, line 241: Maybe let's do this, this is easier:
```
func trimToTwoDecimal(d time.Duration) string {
	var f float64
	var unit string

	switch {
	case d >= time.Second:
		f = float64(d) / float64(time.Second)
		unit = "s"
	default:
		f = float64(d) / float64(time.Millisecond)
		unit = "ms"
	}

	return fmt.Sprintf("%.02f%s", f, unit)
}
```

vadmeste (commented):
One comment

go.sum
```diff
@@ -191,6 +191,7 @@ github.com/google/go-cmp v0.5.3/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/
 github.com/google/go-cmp v0.5.4/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
```
harshavardhana, line 191:
```
go mod tidy -compat=1.17
```

You should run this..

harshavardhana (changes requested):

go.sum
sinhaashish: done

vadmeste (commented):
LGTM, only one nit comment

harshavardhana (approved):

cmd/ping.go
```diff
@@ -216,12 +217,12 @@ func ping(ctx context.Context, cliCtx *cli.Context, anonClient *madmin.Anonymous
 		stat := getPingInfo(cliCtx, result, endPointMap)
 		endPointStat := EndPointStats{
 			Endpoint:  endPoint,
-			Min:       time.Duration(stat.min).Round(time.Microsecond).String(),
-			Max:       time.Duration(stat.max).Round(time.Microsecond).String(),
-			Average:   time.Duration(stat.avg).Round(time.Microsecond).String(),
+			Min:       trimToTwoDecimal(time.Duration(stat.min).Round(time.Microsecond)),
```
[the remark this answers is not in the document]
sinhaashish: removed

vadmeste (approved):
LGTM

kannappanr (approved):

PR #4221, opened 2022-08-25
Fix `support logs status` command

anjalshireesh:
Earlier, the parsed config contained the `target` of the form
`subsystem:target` e.g. `logger_webhook:subnet`. Now it contains only
the target e.g. `subnet`.

Updating the `support logs status` command code accordingly to ensure
that it prints correct output.


harshavardhana (approved):

vadmeste (approved):
LGTM

PR #4223, opened 2022-08-26
Fix mb with-versioning

klauspost:
`mc mb --with-versioning` no longer works after https://github.com/minio/mc/commit/f70b7e537c908e2c426f87218d5a44f807588074 (no PR)

Use long param name.

### DIFF
--- a/cmd/mb-main.go
+++ b/cmd/mb-main.go
@@ -156,7 +156,7 @@ func mainMakeBucket(cliCtx *cli.Context) error {
 			continue
 		}
 
-		if cliCtx.Bool("s") {
+		if cliCtx.Bool("with-versioning") {
 			fatalIf(clnt.SetVersion(ctx, "enable", []string{}, false), "Unable to enable versioning")
 		}
 


cmd/mb-main.go
```diff
@@ -156,7 +156,7 @@ func mainMakeBucket(cliCtx *cli.Context) error {
 			continue
 		}
 
-		if cliCtx.Bool("s") {
+		if cliCtx.Bool("with-versioning") {
```
vadmeste, line 159: I was wondering the same when I reviewed the PR that introduced this, but then I saw L132 and L131 so..

klauspost: @vadmeste L132 and L131 are defined shorthands, see L37 & L41. 

The PR deleted `-s` as a valid option, so this no longer works.

vadmeste: oh I understand now, thanks!

vadmeste (approved):
LGTM

harshavardhana (approved):

PR #4232, opened 2022-09-04
cleanup commands to use Go 'err' as 'e'

harshavardhana:
many commands seem to have sort of had
stopped following the expected convention
in `mc` to differentiate *probe.Error.

To keep things readable this PR is the
first attempt at such a cleanup.

### DIFF
--- a/cmd/admin-bucket-quota.go
+++ b/cmd/admin-bucket-quota.go
@@ -133,9 +133,12 @@ func mainAdminBucketQuota(ctx *cli.Context) error {
 		quotaStr := ctx.String("hard")
 		quota, e := humanize.ParseBytes(quotaStr)
 		fatalIf(probe.NewError(e).Trace(quotaStr), "Unable to parse quota")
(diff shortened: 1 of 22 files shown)


klauspost:
So 'e' is now always a `*probe.Error`?

harshavardhana:
> So 'e' is now always a `*probe.Error`?

No 'e' is always Go error in `mc` and `err` is `*probe.Error` - a distinction we should always follow otherwise it's confusing.

klauspost:
> No 'e' is always Go error in `mc` and `err` is `*probe.Error` - a distinction we should always follow otherwise it's confusing.

Not having Go errors as 'err' is the confusing part for me (and probably everyone else looking at the code).

harshavardhana:
> > No 'e' is always Go error in `mc` and `err` is `*probe.Error` - a distinction we should always follow otherwise it's confusing.
> 
> Not having Go errors as 'err' is the confusing part for me (and probably everyone else looking at the code).

Yeah this convention has been long followed in mc, this PR is simply making it apparent on parts that didn't use it this way.

klauspost (approved):

vadmeste (approved):
LGTM

PR #4233, opened 2022-09-06
ls: Fix ls exit code related to some errors in FS

vadmeste:
The code that is being modified is from 2016. Currently it does not set
an exit code status in the following cases, all of them are returned by
the FS layer:

    BrokenSymlink
    TooManyLevelsSymlink
    PathNotFound
    PathInsufficientPermission

There is no reason to avoid setting an exit code in this case. The only
functional change of this PR is by setting an exit code


klauspost (approved):
Sounds reasonable to remove.

harshavardhana (approved):

PR #4237, opened 2022-09-07
allow terminal control back after bubbletea UI

harshavardhana:
## Description
allow terminal control back after bubbletea UI

## Motivation and Context
terminal control was not relinquished since ui.Start() 
must not be run in a go-routine as it is expected
to be a blocking call, and must be serialized with when 
an application exits.

refactor the 'scanner info/status' to ensure we have 
correct behavior.

This PR also renames 'info-> status', keeps the older 
one as alias for now.

## How to test this PR?
Test `mc admin scanner info alias/ -n 2`  the terminal
control is never relinquished, causes the terminal to
sort of hang and we have to manually perform `reset`

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
- [ ] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fmc%2fpull%2fNNNNN)

PR #4239, opened 2022-09-09
allow terminal control back for 'top api/disk'

harshavardhana:
## Description
allow terminal control back for 'top api/disk'

## Motivation and Context
similar to #4237 fix other command TUI's as well,
also no locking is needed to protect maps.

Update()/View() are serial so need to add mutexes.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/top-disk-spinner.go
+++ b/cmd/top-disk-spinner.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"sort"
 	"strings"
-	"sync"
 
 	"github.com/charmbracelet/bubbles/spinner"
 	tea "github.com/charmbracelet/bubbletea"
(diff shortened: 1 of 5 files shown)


cmd/top-disk-spinner.go
```diff
@@ -40,7 +39,6 @@ type topDiskUI struct {
 
 	disksInfo map[string]madmin.Disk
 
-	mu         sync.Mutex
```
vadmeste, line 43: @harshavardhana it is not clear in the bubbletea doc if View() and Update() can be called concurrently, that's why I added the mutex, LGTM otherwise

harshavardhana: View() is never called concurrent to Update() 

```
	// Update is called when a message is received. Use it to inspect messages
	// and, in response, update the model and/or send a command.
	Update(Msg) (Model, Cmd)

	// View renders the program's UI, which is just a string. The view is
	// rendered after every Update.
	View() string
```

vadmeste (approved):
LGTM

PR #4245, opened 2022-09-13
Use a pager for help output

vadmeste:
## Description
Automatically use a pager for the help

Depends on https://github.com/minio/cli/pull/8

## Motivation and Context
Add pagination when printing long help message

## How to test this PR?
mc cp -h

## Types of changes

## Checklist:

### DIFF
--- a/cmd/term-pager.go
+++ b/cmd/term-pager.go
@@ -1,0 +1,176 @@
+// Copyright (c) 2015-2022 MinIO, Inc.
+//
+// This file is part of MinIO Object Storage stack
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/charmbracelet/bubbles/viewport"
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/charmbracelet/lipgloss"
+	"github.com/muesli/reflow/wordwrap"
+)
+
+var percentStyle = lipgloss.NewStyle().Width(5).Align(lipgloss.Left)
+
+type model struct {
+	viewport viewport.Model
+	content  string
+
+	ready        bool
+	renderedOnce chan struct{}
+}
+
+func (m model) Init() tea.Cmd {
+	return nil
+}
+
+func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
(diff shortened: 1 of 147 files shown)


klauspost:
@vadmeste Would you mind if I try to implement without invoking less?

vadmeste:
> @vadmeste Would you mind if I try to implement without invoking less?

No problem, do you know any good library for that ? also I found it this PR doesn't work in some cases, unfortunately the cli library calls os.Exit() in certain cases, this is maybe not easy to implement

klauspost:
Yeah, `os.Exit` is always going to be a problem. So this only for help (for now?)

While a full 'less' would be great, maybe just paging would be fine.

External programs are always going to be flaky, so while it has the upside of all the "less" features, if they need that then `mc --help|less` is the way to go.

vadmeste:
> Rather than using less you can use this https://github.com/charmbracelet/bubbletea/tree/master/examples/pager @vadmeste

Yes, I didn't see this when I went through their docs, will try it out

vadmeste:
I made some progress, the code currently is little bit complex because I had to connect cli with bubbletea but overall it is fine, can you test this @harshavardhana before I make the remaining changes ? hopefully no surprises after that

harshavardhana:
It works fine @vadmeste however we have to show how to `quit` at the bottom page otherwise users won't know how to quit the pager.

vadmeste:
> It works fine @vadmeste however we have to show how to quit at the bottom page otherwise users won't know how to quit the pager.

okay, there are still more edge cases that I need to address, they should be fixed today

vadmeste:
This PR https://github.com/minio/cli/pull/9/files is also needed. Otherwise all of this won't work :)

cmd/term-pager.go
```diff
@@ -0,0 +1,159 @@
+package cmd
```
harshavardhana, line 1: license header missing @vadmeste

main.go
```diff
@@ -24,5 +24,7 @@ import (
 )
 
 func main() {
-	mc.Main(os.Args)
+	if e := mc.Main(os.Args); e != nil {
+		os.Exit(1)
```
harshavardhana, line 28:
```suggestion
		console.Fatalln(e)
```

harshavardhana (changes requested):

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/klauspost/compress v1.15.9
 	github.com/mattn/go-ieproxy v0.0.1
 	github.com/mattn/go-isatty v0.0.14
-	github.com/minio/cli v1.23.0
+	github.com/minio/cli v1.23.1-0.20220921164421-38105e9ed822
```
harshavardhana, line 19: tagged v1.24.0 @vadmeste

harshavardhana (changes requested):

harshavardhana:
PTAL at the conflict @vadmeste

harshavardhana (approved):

zeeZ:
Is there an option included to disable this?

It's pretty cumbersome if the help output disappears while you're trying to cobble together a command with multiple flags and one can't look at the output history because it disappears with the pager.

`mc cp -h | cat` is a dumb workaround.

harshavardhana:
We are going to fix the pager for it to stay the output..

vadmeste:
> Is there an option included to disable this?

@zeeZ I think you want this PR, https://github.com/minio/mc/pull/4289, it keeps the help screen after quitting it

PR #4270, opened 2022-09-28
avoid expensive threads profile by default

harshavardhana:


klauspost (approved):
👍🏼  I rarely, if ever, use the data from it.

PR #4276, opened 2022-09-29
remove 'mc policy' command

harshavardhana:


vadmeste (commented):
LGTM but wondering why removing it

harshavardhana:
> LGTM but wondering why removing it

what is the point of having it we have had `mc anonymous` for 3 yrs now already and `policy` has been hidden since.

vadmeste (approved):
LGTM

webjunkie:
> and `policy` has been hidden since

I'm sorry, but is this the `policy` command that is mentioned in the main readme file that everybody reads when visiting the repo?

harshavardhana:
> > and `policy` has been hidden since
> 
> I'm sorry, but is this the `policy` command that is mentioned in the main readme file that everybody reads when visiting the repo?

This has been deprecated for 3+yrs please move to 'mc anonymous'

harshavardhana:
> Yeah, this broke my CI build which was using `mc policy`, and I went to update it but there is no documentation of `mc anonymous` in the docs: https://min.io/docs/minio/linux/reference/minio-mc.html - only `mc policy`.

This has been deprecated for 3+yrs please move to 'mc anonymous'  - documentation will be updated.

webjunkie:
Seems like the command was renamed in 7dafdaf88e2d48ba0167a8fbcecaf66cb2da39b9, which was a little over a year ago. I cannot find a deprecation notice for it either.

harshavardhana:
Locking this conversation, please use the `mc anonymous` command for similar functionality.

PR #4302, opened 2022-10-13
Fixing Destination bucket

adfost:
Showing destination bucket instead of arn for destination bucket.


poornas (approved):

kannappanr (approved):
LGTM

harshavardhana (changes requested):
This kind of parsing logic is not future-proof, you need to rely on something minio-go to do this for you.  `r.Destination.Bucket` must already be just `bucketName` as sent by the server.

The ARN must `r.Destination.ARN`

PR #4311, opened 2022-10-19
Improve help text for `support profile|perf`

anjalshireesh:
## Description

- Use `play` as the example cluster to be consistent with `support diag`
- Mention `upload` in usage
- Replace `start profiling` with `profile` in usage of `--duration` flag
- Downcase first letter of `--airgap` help text
- Add `--airgap` examples

## Motivation and Context

Better help text

## How to test this PR?

```
mc support profile --help
mc support perf --help
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/support-perf.go
+++ b/cmd/support-perf.go
@@ -82,7 +82,7 @@ var supportPerfFlags = append([]cli.Flag{
 
 var supportPerfCmd = cli.Command{
 	Name:            "perf",
-	Usage:           "analyze object, network and drive performance",
+	Usage:           "upload object, network and drive performance analysis",
 	Action:          mainSupportPerf,
 	OnUsageError:    onUsageError,
 	Before:          setGlobalsFromContext,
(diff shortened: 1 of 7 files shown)


cmd/support-perf.go
```diff
@@ -99,8 +99,10 @@ FLAGS:
   {{end}}
 
 EXAMPLES:
-  1. Run object storage, network, and drive performance tests on 'myminio' cluster:
-     {{.Prompt}} {{.HelpName}} myminio/
+  1. Upload object storage, network, and drive performance analysis for 'play' (https://play.min.io by default) to SUBNET
+     {{.Prompt}} {{.HelpName}} play
+  2. Run object storage, network, and drive performance tests on 'play' (https://play.min.io by default), save and upload to SUBNET manually
+     {{.Prompt}} {{.HelpName}} --airgap play
```
harshavardhana, line 105: I think we can avoid talking about `play/`  @anjalshireesh or maybe have it as a single reference.  More we mention `play/`  it kind of leads to abuse of the public server.

We can limit it here in this situation. - is the idea you want play because it is a working setup?

harshavardhana (changes requested):

cmd/support-perf.go
anjalshireesh: @harshavardhana it's not my preference to use `play` as such. I was just trying to be consistent with what `mc support diag` shows. I can switch to `myminio` in all support sub-commands if that's preferred.

harshavardhana: Yeah we should use myminio instead, we can have one example with play

anjalshireesh: Done. PTAL.

PR #4317, opened 2022-10-20
Fixing typo in mc

adfost:


prakashsvmx (approved):

cmd/sql-main.go
```diff
@@ -106,7 +106,7 @@ EXAMPLES:
 
   3. Run a query on an encrypted object with customer provided keys.
      {{.Prompt}} {{.HelpName}} --encrypt-key "myminio/iot-devices=32byteslongsecretkeymustbegiven1" \
-         --query "select count(s.power) from S3Object s" myminio/iot-devices/power-ratio-encrypted.csv
+         --query "select count(s.power) from S3Object" myminio/iot-devices/power-ratio-encrypted.csv
```
donatello, line 109: This is not right @adfost - the existing it correct. The `s` after `S3Object` defines an alias for the table name, which allows a column reference by alias like `s.power` - it is actually examples 2 and 4 that need to be fixed (with the alias to be added).

donatello (changes requested):

PR #4327, opened 2022-10-27
Adding mc replicate update

adfost:
Adding the mc replicate update command, comment on this PR if you want the command this is replacing (mc replicate edit) to be deleted. I'm going to leave it in for now for backwards compatibility.


cmd/replicate-update.go
```diff
@@ -105,7 +105,7 @@ EXAMPLES:
 // checkReplicateUpdateSyntax - validate all the passed arguments
 func checkReplicateUpdateSyntax(ctx *cli.Context) {
 	if len(ctx.Args()) != 1 {
-		cli.ShowCommandHelpAndExit(ctx, "edit", 1) // last argument is exit code
+		cli.ShowCommandHelpAndExit(ctx, "update", 1) // last argument is exit code
```
vadmeste, line 108: You can use 'ctx.Command.Name' as well, we discovered this recently

vadmeste (commented):
One comment

vadmeste:
```
auto-complete_test.go:56: Missing completion function: Completion for `/replicate/update` not found
```

@adfost you just need to fix auto-complet_test.go to recognize the new command name. This is a unit test to ensure that all mc commands have a bash completion rule

PR #4330, opened 2022-10-27
Improve format of perf tests output

anjalshireesh:
## Description

The current JSON format of the perf tests output is same as what mc receives in response from the minio api. This is tailored more for effectively rendering the real-time output on screen.

In order to retain this existing behavior, and to ensure backward compatibility with older versions of minio, we will transform the minio response into a more readable format to be used for uploading to subnet and printing when `--json` is passed.

Following are the changes made:

- Remove unnecessary fields:
  - final
  - type
  - version
- Remove `ttfb` from `PUTStats`
- Rename `throughputPerSec` to `throughput`
- Rename `disks` to `drives`
- Rename `err` to `error`

## Motivation and Context

More readable JSON output from `mc support perf`

## How to test this PR?

- Run `mc support perf {alias} --airgap` (use `--dev` also if cluster is not registered)
- Unzip the generated file and verify the format of the `{alias}-perf_{timestamp}.json`
- Run the same command with `--json` also and verify that it's output is also in the new format

## Types of changes

## Checklist:

### DIFF
--- a/cmd/support-perf-drive.go
+++ b/cmd/support-perf-drive.go
@@ -68,11 +68,12 @@ func mainAdminSpeedTestDrive(ctx *cli.Context, aliasedURL string, outCh chan<- P
 
 	if globalJSON {
 		if e != nil {
-			printMsg(PerfTestResult{
+			printMsg(convertPerfResult(PerfTestResult{
 				Type:  DrivePerfTest,
 				Err:   e.Error(),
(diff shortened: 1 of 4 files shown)


harshavardhana (approved):

prakashsvmx (approved):
Tested , LGTM :+1:

PR #4332, opened 2022-10-28
Adding replicate update

adfost:
Replacing replicate edit with replicate update, with edit as an alias for backwards compatibility.

### DIFF
--- a/cmd/admin-bucket-info.go
+++ b/cmd/admin-bucket-info.go
@@ -142,7 +142,7 @@ EXAMPLES:
 // checkAdminBucketInfoSyntax - validate all the passed arguments
 func checkAdminBucketInfoSyntax(ctx *cli.Context) {
 	if len(ctx.Args()) != 1 {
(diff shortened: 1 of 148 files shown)


poornas (approved):

vadmeste:
@adfost are we going to have both 'mc replicate edit' and 'mc replicate update' ? they look similar to me

klauspost (approved):
Please add a description to the PR of your change.

cmd/replicate-update.go
```diff
@@ -171,7 +172,7 @@ func mainReplicateEdit(cliCtx *cli.Context) error {
 				dmReplicate = enableStatus
 			case "delete":
 				vDeleteReplicate = enableStatus
-			case "metadata-sync", "replica-metadata-sync":
+			case "replica-metadata-sync":
```
poornas, line 175: @adfost , `replica-metadata-sync` was shortened to `metadata-sync` - we can just use metadata-sync here and in the help

PR #4360, opened 2022-11-10
Changing 2021 to 2022 in license

adfost:
Adding a script to change 2021 to 2022 in license and running it on the files. The script is included in this PR for easy future modification, comment if you would like it removed.


klauspost (approved):

vadmeste (approved):
LGTM

harshavardhana:
> Adding a script to change 2021 to 2022 in license and running it on the files. The script is included in this PR for easy future modification, comment if you would like it removed.

do you have the script added in the PR? @adfost

klauspost:
@harshavardhana It is somewhere in the middle.

cmd/change_years.py
```diff
@@ -0,0 +1,12 @@
+import os
```
harshavardhana, line 1: @adfost this kind of script is not going to work we should use something like 

- https://github.com/palantir/go-license
- https://github.com/elastic/go-licenser

harshavardhana (changes requested):

cmd/change_years.py
harshavardhana: no, we can run these tools to update the license on all files automatically, this way we need to only manage a single template for the license. 

Look at https://github.com/palantir/go-license for this project - it would be better to do that instead.  Lets take this PR I can send a change that handles this via CI/CD automatically.

harshavardhana (approved):

PR #4361, opened 2022-11-10
Fixing wrong number of items in mc du

adfost:
Fixing error where mc du would often return one extra item than was actually in a folder, as . was being counted.


cmd/du-main.go
```diff
@@ -136,6 +137,11 @@ func du(ctx context.Context, urlStr string, timeRef time.Time, withVersions bool
 	// just do a recursive listing
 	recursive := depth == 1
 
+	targetAbsolutePath := clnt.GetURL().String()
+	if strings.HasSuffix(targetAbsolutePath, "/") {
+		targetAbsolutePath = targetAbsolutePath[:len(targetAbsolutePath)-1]
+	}
```
harshavardhana, line 143: do we want to do something like path.Clean()?

harshavardhana (changes requested):

harshavardhana (approved):

PR #4377, opened 2022-11-23
Implement 'mc admin user sts info' command

vadmeste:
## Description
This command shows information about the STS account, 
such as parent user and the associated policy.

## Motivation and Context
Retrieve parent user from a given sts access key

## How to test this PR?
./mc admin user sts info <alias> <sts-access-key>

## Types of changes

## Checklist:


cmd/admin-user-sts-info.go
```diff
@@ -0,0 +1,131 @@
...
+
+	if ctx.Bool("policy") {
+		if stsInfo.Policy == "" {
+			fatalIf(errDummy().Trace(args...), "No policy found associated to the specified service account. Check the policy of its parent user.")
+		}
+		p, e := iampolicy.ParseConfig(strings.NewReader(stsInfo.Policy))
+		fatalIf(probe.NewError(e).Trace(args...), "Unable to parse policy.")
+		enc := json.NewEncoder(os.Stdout)
+		enc.SetIndent("", " ")
+		fatalIf(probe.NewError(enc.Encode(p)).Trace(args...), "Unable to write policy to stdout.")
```
harshavardhana, line 115: can you not use `colorjson` ?

cmd/admin-user-sts-info.go
```diff
@@ -0,0 +1,131 @@
...
+		return nil
+	}
+
+	fmt.Println(stsInfo.ImpliedPolicy)
```
harshavardhana, line 119:
```suggestion
	fmt.Println(stsInfo.ImpliedPolicy)
```

is this json formatted already? this is an odd place to print something like this before you print the actual info

harshavardhana (changes requested):

cmd/admin-user-sts-info.go
vadmeste: yes, this was a typo

cmd/admin-user-sts-info.go
vadmeste: > can you not use colorjson ?

why is that ? by the way, if mc is piped, the color is automatically disabled

harshavardhana:
@vadmeste can you fix the conflicts?

vadmeste:
> @vadmeste can you fix the conflicts?

Done

cmd/admin-user-sts-info.go
harshavardhana: colorjson pretty-fies your JSON output - it is used everywhere.

vadmeste: > colorjson pretty-fies your JSON output - it is used everywhere.

Are you asking to use or not to use colorjson ?

harshavardhana: My bad, It looks like it is being used.

vadmeste:
@donatello please take a look.

Issue #4315, opened 2022-10-19
mc stat --rewind flag isnt working intended


bh4t:
## Expected behavior
mc stat --rewind flag should display objects created before the specified date.
## Actual behavior
mc stat --rewind flag sint displaying any results
## Steps to reproduce the behavior
Have multiple objects created prior to the below mentioned date.
mc stat --rewind 2022.10.15 --versions myminio/hellotest/

Another thing that I noticed, when you use the --rewind flag with ls it seems to be working.

![image](https://user-images.githubusercontent.com/109562845/196772794-d64b4293-0e06-49e3-bd4b-6fa457ea8fd6.png)

## mc --version
mc version RELEASE.2022-10-12T18-12-50Z

dormanze:
I findthis problem in this pr #4147

PR #4396, opened 2022-12-08
Stat: Do not append a slash when --rewind is specified

vadmeste:
## Description
--rewind should trigger a listing operation to look for a list
 of versions of a specified object name and picks the right 
version based on its modtime.

Do not attempt to add a trailing slash to the object name 
when HEAD fails in that case.

## Motivation and Context
Fix https://github.com/minio/mc/issues/4315

## How to test this PR?
- Upload an object in a bucket
- ./mc stat --rewind 2023.01.01T00:00 <alias>/<bucket>/object

## Types of changes

## Checklist:

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -1600,13 +1600,9 @@ func (c *S3Client) Stat(ctx context.Context, opts StatOptions) (*ClientContent, 
 	//     - /path/to/directory_marker
 	//     - /path/to/directory_marker/
 
-	// First an HEAD call is issued, this is faster than doing listing even if the object exists
-	// because the list could be very large. At the same time, the HEAD call is avoided if the
-	// object already contains a trailing prefix or we passed rewind flag to know the object version
-	// created just before the rewind parameter.
+	// Start with a HEAD request first to return object metadata information.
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

cmd/client-s3.go
```diff
@@ -1615,16 +1611,11 @@ func (c *S3Client) Stat(ctx context.Context, opts StatOptions) (*ClientContent,
...
 		// Ignore object missing error but return for other errors
 		if !errors.As(err.ToGoError(), &ObjectMissing{}) && !errors.As(err.ToGoError(), &ObjectIsDeleteMarker{}) {
 			return nil, err
 		}
-	}
-
-	// No object found, start looking for a prefix with the same name
-	// or a directory marker. Add a trailing slash if it is not in the path
-	if !strings.HasSuffix(path, string(c.targetURL.Separator)) {
+		// The object is not found, look for a directory marker or a prefix
```
klauspost, line 1618: This is the only code change. (removing the `if`). Can you explain how that does what your description says?

I don't really understand.

klauspost (approved):
Took a while for me to figure out, but I see that moving it into the if statement does what you say.

PR #4406, opened 2022-12-13
Print 'Cancel upon user request' when the global context is cancelled

vadmeste:
## Description
The code creates new context and cancels them and it is not 
correct to show 'Cancel upon user request' when an error 
is context.Cancelled.

Fix this by requiring globalContext error to be context.Cancelled. 
The global context is only cancelled after receiving a signal.

## Motivation and Context
Fix this error message: 'Cancel upon user request' error message
when the user did not really kill mc

## How to test this PR?
Not trivial

## Types of changes

## Checklist:

### DIFF
--- a/cmd/error.go
+++ b/cmd/error.go
@@ -84,11 +84,12 @@ func fatal(err *probe.Error, msg string, data ...interface{}) {
 	msg = fmt.Sprintf(msg, data...)
 	errmsg := err.String()
 	if !globalDebug {
-		e := err.ToGoError()
-		if errors.Is(e, context.Canceled) {
-			// This will replace context canceled error message
-			// that the user is seeing to a better one.
+		var e error
+		if errors.Is(globalContext.Err(), context.Canceled) {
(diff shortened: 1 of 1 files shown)


cmd/error.go
```diff
@@ -84,11 +84,12 @@ func fatal(err *probe.Error, msg string, data ...interface{}) {
 	msg = fmt.Sprintf(msg, data...)
 	errmsg := err.String()
 	if !globalDebug {
-		e := err.ToGoError()
-		if errors.Is(e, context.Canceled) {
-			// This will replace context canceled error message
-			// that the user is seeing to a better one.
+		var e error
+		if globalContext.Err() == context.Canceled {
```
harshavardhana, line 88: You can still use errors.Is()

harshavardhana (changes requested):

harshavardhana (approved):

krisis (approved):

PR #4407, opened 2022-12-14
fix: support unlocks

dreamCodeMan:
## Description
support unlocks


harshavardhana:
This is not implemented on purpose @dreamCodeMan we do want users to be in manually mucking around with locks, let alone it be scriptable at scale.

PR #4424, opened 2022-12-22
avoid converting API types to local forms in 'mc'

harshavardhana:
## Description
avoid converting API types to local forms in 'mc'

## Motivation and Context
keeping the API type to be exactly what they
are supposed to be, instead of calling

azure -> blob

blob is mostly meant for objects and not for the 
API itself, we could have called it as AzureBlob 
but having a cleaner differentiation while 
printing is perfectly normal.

## How to test this PR?
Nothing would change it's a cosmetic UI change.

## Types of changes

## Checklist:


krisis (approved):

vadmeste (approved):
LGTM

poornas (approved):

PR #4437, opened 2023-01-09
disable sha256 for 'piped' uploads() to reduce CPU usage

harshavardhana:
## Description
disable sha256 for 'piped' uploads() to reduce CPU usage

## Motivation and Context
performance optimization

## How to test this PR?
Nothing special, must-see greatly reduced CPU usage with this PR.

## Types of changes

## Checklist:


vadmeste (approved):

klauspost (changes requested):
This is only used for plain HTTP and not HTTPS.

As such, I don't think we should do this, and instead just promote TLS.

We can add it as an option, so it can be disabled - but I am not sure it is worth it.

PR #4443, opened 2023-01-13
Improve message when profile upload fails

anjalshireesh:
## Description

When profile upload to SUBNET fails because of any error, currently just the error message is shown, creating an impression that the profile file was not saved anywhere. (we delete the file only if upload goes through, otherwise it is stored locally)

Improve the message(s) printed on terminal in case of such failure, clearly indicating that the file has been stored locally.

## Motivation and Context

More meaningful output in case of error.

## How to test this PR?

- Register a cluster with subnet
- Run the `mc support profile {alias}` and verify that it works fine and uploads the profile file to SUBNET (it becomes visible in the `Profile` tab on the deployment page)
- Modify following line in `support-profile.go` to replace `"profile"` with `"profile1"` (this will try to upload to a wrong URL and hence will simulate upload error)
`uploadURL := subnetUploadURL("profile", profileFile)`
- Build `mc` with this modified code and run the profile command again.
- Verify that apart from the upload error message, it also tells the user that the profile file has been stored locally.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/support-profile.go
+++ b/cmd/support-profile.go
@@ -203,15 +203,17 @@ func execSupportProfile(ctx *cli.Context, client *madmin.AdminClient, alias stri
 
 	saveProfileFile(data)
 
-	clr := color.New(color.FgGreen, color.Bold)
+	successClr := color.New(color.FgGreen, color.Bold)
+	failureClr := color.New(color.FgRed, color.Bold)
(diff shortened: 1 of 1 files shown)


klauspost (approved):
LGTM

harshavardhana (approved):

PR #4450, opened 2023-01-22
cleanup mc to use 'go' errs to be 'e'

harshavardhana:
## Description
cleanup mc to use 'go' errs to be 'e'

## Motivation and Context
by design we have agreed to follow

go errors to be 'e'
probe.Error to be 'err'

in 'mc' fix it everywhere this is not
followed appropriately.

## How to test this PR?
Nothing should change

## Types of changes

## Checklist:

### DIFF
--- a/cmd/parallel-manager.go
+++ b/cmd/parallel-manager.go
@@ -225,19 +225,19 @@ func (p *ParallelManager) stopAndWait() {
 const cgroupLimitFile = "/sys/fs/cgroup/memory/memory.limit_in_bytes"
 
 func cgroupLimit(limitFile string) (limit uint64) {
-	buf, err := os.ReadFile(limitFile)
-	if err != nil {
+	buf, e := os.ReadFile(limitFile)
+	if e != nil {
 		return 9223372036854771712
 	}
-	limit, err = strconv.ParseUint(string(buf), 10, 64)
-	if err != nil {
+	limit, e = strconv.ParseUint(string(buf), 10, 64)
+	if e != nil {
 		return 9223372036854771712
 	}
 	return limit
 }
 
 func availableMemory() (available uint64) {
-	available = 8 << 30 // Default to 8 GiB when we can't find the limits.
+	available = 4 << 30 // Default to 4 GiB when we can't find the limits.
 
 	if runtime.GOOS == "linux" {
 		available = cgroupLimit(cgroupLimitFile)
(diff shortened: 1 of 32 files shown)


cmd/parallel-manager.go
```diff
@@ -252,11 +252,7 @@ func availableMemory() (available uint64) {
 
 	} // for all other platforms limits are based on virtual memory.
 
-	memStats, err := mem.VirtualMemory()
-	if err != nil {
-		return
```
vadmeste, line 257: maybe it is better to keep this to return 8Gb instead of 0Gb as available memory, 8GB is better than zero.

cmd/replicate-update.go
```diff
@@ -66,11 +66,13 @@ var replicateUpdateFlags = []cli.Flag{
 	},
 	cli.StringFlag{
 		Name:  "sync",
-		Usage: "enable synchronous replication for this target. Valid values are enable,disable.Defaults to disable if unset",
+		Usage: "enable synchronous replication for this target, valid values are ['enable', 'disable'].",
+		Value: "unset",
```
vadmeste, line 70: I think you mean 'disable' here

cmd/support-profile.go
```diff
@@ -114,22 +114,20 @@ func checkAdminProfileSyntax(ctx *cli.Context) {
...
 // be handled appropriately use this function instead of os.Rename()
 func moveFile(sourcePath, destPath string) error {
-	inputFile, err := os.Open(sourcePath)
-	if err != nil {
-		return err
+	inputFile, e := os.Open(sourcePath)
+	if e != nil {
+		return e
 	}
+	defer inputFile.Close()
```
vadmeste, line 121: I am not sure in Windows if we can move a file which is open, /cc @klauspost

vadmeste (commented):
Few comments

cmd/support-profile.go
klauspost: Good spot. It must be closed before delete can happen.

cmd/parallel-manager.go
harshavardhana: 8gb is too high right? @vadmeste I will perhaps fit this into 4Gb

harshavardhana:
PTAL @vadmeste @klauspost

harshavardhana:
PTAL @vadmeste

vadmeste (approved):
LGTM

prakashsvmx (approved):

PR #4458, opened 2023-01-27
Use --api-key flag on register when provided

kanagarajkm:
## Description

When `--api-key` is provided in the `mc license register` command, the existing key and the license stored on the config can be ignored since they can be expired.

## Motivation and Context

Some of the older deployments may have an expired license and no api-key. They needs to ignored when `--api-key` flag is provided.

## How to test this PR?
Register a new deployment using `mc license register ALIAS --api-key=API_KEY`
Check if it is successfule.

Update a deployment using `mc license register ALIAS` but with no api-key
Check if the deployment is updated succesfully.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/license-register.go
+++ b/cmd/license-register.go
@@ -175,14 +175,11 @@ func mainLicenseRegister(ctx *cli.Context) error {
 		fatalIf(probe.NewError(e), "Error in fetching subnet API Key")
 		if len(apiKey) > 0 {
 			alreadyRegistered = true
-			if len(accAPIKey) == 0 {
-				accAPIKey = apiKey
-			}
+			accAPIKey = apiKey
 		}
 	} else {
 		apiKey := getSubnetAPIKeyFromConfig(alias)
-		lic := getSubnetLicenseFromConfig(alias)
-		if len(apiKey) > 0 || len(lic) > 0 {
+		if len(apiKey) > 0 {
 			alreadyRegistered = true
 		}
 	}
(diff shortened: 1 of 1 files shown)


cmd/license-register.go
```diff
@@ -170,12 +170,20 @@ func mainLicenseRegister(ctx *cli.Context) error {
...
-	if len(apiKey) > 0 {
-		alreadyRegistered = true
-		if len(accAPIKey) == 0 {
-			accAPIKey = apiKey
+	if len(accAPIKey) == 0 {
+		apiKey, _, e := getSubnetCreds(alias)
+		fatalIf(probe.NewError(e), "Error in fetching subnet API Key")
+		if len(apiKey) > 0 {
+			alreadyRegistered = true
+			if len(accAPIKey) == 0 {
```
anjalshireesh, line 178: This condition is redundant as it is already checked in the first `if` condition.

cmd/license-register.go
```diff
@@ -170,12 +170,20 @@ func mainLicenseRegister(ctx *cli.Context) error {
...
+				accAPIKey = apiKey
+			}
+		}
+	} else {
+		apiKey := getSubnetAPIKeyFromConfig(alias)
+		lic := getSubnetLicenseFromConfig(alias)
+		if len(apiKey) > 0 || len(lic) > 0 {
+			alreadyRegistered = true
```
anjalshireesh, line 186:
```suggestion
		if len(getSubnetAPIKeyFromConfig(alias)) > 0 {
			alreadyRegistered = true
```
Now api key is the main indicator of registration. So we need not look for license when checking if the cluster is already registered.

cmd/license-register.go
kanagarajkm: done. thanks

cmd/license-register.go
kanagarajkm: done, thanks

harshavardhana (approved):

anjalshireesh (approved):
LGTM

PR #4478, opened 2023-02-21
subnet: Avoid general hard timeout for requests

vadmeste:
## Description
There is a ten seconds timeout deadline for subnet requests. T
his is now wrong after introducing Subnet inspect upload feature. 
One user can see deadline timeout error when trying to upload 
a large inspect file.

Disabling the timeout for all subnet operations for now.

We can introduce context.Context in subnet calls later in 
order to configure a deadline per use case.

## Motivation and Context
Fixing 
```
mc: <ERROR> Unable to upload inspect data to SUBNET portal: Post "https://subnet.min.io/api/inspect/upload?filename=inspect-data.enc": context deadline exceeded (Client.Timeout exceeded while awaiting hea
ders)  
```

## How to test this PR?
Setup a MinIO cluster and register it in Subnet
Upload many files under a bucket
mc support inspect "<alias>/bucket/**"

## Types of changes

## Checklist:

### DIFF
--- a/cmd/subnet-utils.go
+++ b/cmd/subnet-utils.go
@@ -175,7 +175,7 @@ func subnetAPIKeyAuthHeaders(apiKey string) subnetHeaders {
 }
 
 func getSubnetClient() *http.Client {
-	client := httpClient(10 * time.Second)
+	client := httpClient(0)
 	if globalSubnetProxyURL != nil {
 		client.Transport.(*http.Transport).Proxy = http.ProxyURL(globalSubnetProxyURL)
 	}
--- a/cmd/utils.go
+++ b/cmd/utils.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"math"
 	"math/rand"
+	"net"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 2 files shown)


harshavardhana:
Why can't we get those users to use --airgap and then download as workaround?

vadmeste:
> Why can't we get those users to use --airgap and then download as workaround?

Yes, we can do that for now. But the current code will never be able to upload large files. It always quit after 10 seconds. There is a hard timeout for the whole request operation.

harshavardhana:
> > Why can't we get those users to use --airgap and then download as workaround?
> 
> Yes, we can do that for now. But the current code will never be able to upload large files. It always quit after 10 seconds. There is a hard timeout for the whole request operation.

Actually we do have connection level timeouts, so seems like this change is okay.

vadmeste:
> Actually we do have connection level timeouts, so seems like this change is okay.

This PR mainly fixes the issue of uploading large files. The old code gives 10 seconds maximum limit to upload a file.

klauspost (approved):
LGTM

harshavardhana (approved):

PR #4494, opened 2023-03-08
replicate status - drop stale ARNs from stats

poornas:
This info is still accessible in --json mode

Fixes: #4493

## Motivation and Context
Users get confused by metrics from ARN's that are no longer part of replication config. 
Continue to show the metrics for all ARNs in --json mode as this historical info represents data replicated under the old ARN. ARN change will not cause re-replication of already synced content.

## How to test this PR?
set up bucket replication to target 1, upload some content. Delete the replication config and re-add target and upload some more 
`mc replicate status` would show only metrics for valid target in replication config whereas the --json output should show both ARN's

## Types of changes

## Checklist:

### DIFF
--- a/cmd/replicate-status.go
+++ b/cmd/replicate-status.go
@@ -20,6 +20,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"math"
 	"sort"
 
 	humanize "github.com/dustin/go-humanize"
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

harshavardhana (approved):

vadmeste (approved):
LGTM

PR #4550, opened 2023-04-13
heal: Calculate ETA based on used inodes

vadmeste:
## Description

The used space ETA can be wrong if a cluster has a mix of 
large objects and small objects.

Add healing finish estimation based on the number of consumed 
inodes in the healing disk compared to other disks in the 
same erasure set.

## Motivation and Context
Better ETA when a cluster has a mix of large and tiny files

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-heal.go
+++ b/cmd/admin-heal.go
@@ -327,7 +327,7 @@ func estimateHealDiskETA(now time.Time, healDisk madmin.Disk, set setInfo) (rema
 	// calculation if ETA is furtherst or pct is lower
 	healSpeed = float64(healDisk.UsedInodes) / float64(now.Sub(healDisk.HealInfo.Started))
 	if t := time.Duration(float64(set.maxUsedInodes-healDisk.UsedInodes) / healSpeed); t > remainingTime {
-		remainingTime = t
+		remainingTime = (t + remainingTime) / 2
 	}
 	if p := float64(healDisk.UsedInodes) / float64(set.maxUsedInodes); p < pct {
 		pct = p
--- a/cmd/admin-heal.go
+++ b/cmd/admin-heal.go
@@ -145,6 +145,7 @@ type poolInfo struct {
 
 type setInfo struct {
 	maxUsedSpace   uint64
+	maxUsedInodes  uint64
 	totalDisks     int
 	incapableDisks int
 }
@@ -202,6 +203,9 @@ func generateSetsStatus(disks []madmin.Disk) map[setIndex]setInfo {
 		setSt.totalDisks++
 		if d.UsedSpace > setSt.maxUsedSpace {
 			setSt.maxUsedSpace = d.UsedSpace
+		}
+		if d.UsedInodes > setSt.maxUsedInodes {
+			setSt.maxUsedInodes = d.UsedInodes
 		}
 		if d.State != "ok" || d.Healing {
 			setSt.incapableDisks++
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

klauspost:
@vadmeste This is the 5th or something heal estimate change. We are just switching the problem from one case to another.

This will fix *some* cases, but likely regress even more cases.

You can combine all the metrics (object size, object count, space used, inodes) and use each for 25% of the estimate. That way, at least you even out the estimate for different cases.

vadmeste:
> @vadmeste This is the 5th or something heal estimate change. We are just switching the problem from one case to another.

This is the third time and I think it is better each time.

What we got right is the percentage of completion, only the ETA doesn't still work correctly with some use cases.

> You can combine all the metrics (object size, object count, space used, inodes) and use each for 25% of the estimate.

- Object usage should be always proportional to disk usage, I don't think we should include it in the formula
- Object count is not proportional with inodes in some use cases but still the number of inodes will skew the result to a better approximation (large objects will have higher inodes because of higher number of parts)

The current PR gives the most pessimistic ETA, but the real ETA will never exceed it which I believe is a good thing. If we don't agree, we can use 50% of both disk usage and inodes, but I still don't see clearly why object count/size should be included in the ETA formula;

Otherwise, maybe another idea I just thought about is to use the histogram of objects reported by the scanner and calculate the average time to heal each range of the histogram, however I wanted to avoid the scanner stats in the first place since it is not really real ltime but maybe I am exagerating.

klauspost:
> the most pessimistic ETA

We don't really need the most pessimistic. Seeing a 3 month ETA is just going to raise more questions than answer. We need the most realistic - or remove it completely. Maybe instead we provide should provide a "last minute" to show the progress.

Healing is either limited on objects/s or mb/sec - and a bucket may not even be uniform, and you have the complexity of "sidechannel healing". We may or may not know the object sizes and free space/inodes may or may not be a reliable.

Therefore use whatever metric tools we have available for a combined estimate.

harshavardhana:
> We don't really need the most pessimistic. Seeing a 3 month ETA is just going to raise more questions than answer. We need the most realistic - or remove it completely. Maybe instead we provide should provide a "last minute" to show the progress.

I think we can remove ETA altogether because we can never honestly provide a valid value here. Instead, provide the speed.

Just like Batch replication that can never provide an ETA since thats impossible to do.

vadmeste:
>  Maybe instead we provide should provide a "last minute" to show the progress.

@klauspost , Umm, I am not sure, maybe ETA will just be fluctuating more. Estimating since when healing started seems to be better to me, especially for large healing task.

> We don't really need the most pessimistic. Seeing a 3 month ETA is just going to raise more questions than answer. 

@klauspost it is only pessismistic in the worst case scenario, like healing bucket 1 with 50 millions of large objects, then healing bucket 2 with 50 millions of small objects

> I think we can remove ETA altogether because we can never honestly provide a valid value here. Instead, provide the speed.

We already have the speed. We also have the progress percentage which is fairly accurate.

I would say let's try this PR and see.. this is ETA and everybody knows it is not highly accurate. I can also make ETA less pessimistic by calculating the ETA avg based on both disk usage and disk inodes consumption.

klauspost:
@vadmeste Averaging the metrics we have will minimize the error to the extent we can with the information given. I don't really see any reason not to.

vadmeste:
> @vadmeste Averaging the metrics we have will minimize the error to the extent we can with the information given. I don't really see any reason not to.

I sent a separate commit just to average between inodes and used space. It should be enough without using the scanner metrics or let me know why think those are needed @klauspost

harshavardhana:
This won't make sense as per all the discussions we have here. Lets close this since there is no "realistic" view possible here.

PR #4559, opened 2023-04-24
Cp star

taran-p:
## Types of changes

## Checklist:


cmd/cp-url-syntax.go
```diff
@@ -30,9 +31,6 @@ import (
 
 func checkCopySyntax(ctx context.Context, cliCtx *cli.Context, encKeyDB map[string][]prefixSSEPair, isMvCmd bool) {
 	if len(cliCtx.Args()) < 2 {
-		if isMvCmd {
-			showCommandHelpAndExit(cliCtx, 1) // last argument is exit code.
-		}
```
harshavardhana, line 35: if this is removed remove `isMvCmd bool` argument as well.

harshavardhana (changes requested):

cmd/cp-url-syntax.go
taran-p: isMvCmd is used for some error messages, like "Could not guess type of "+operation+" operation", should these be changed to "Could not guess type of move/copy operation" or something else?

PR #4553, opened 2023-04-19
Added `--recursive` option to `mc tag set` and `mc tag list` command

shtripat:
## Description
If while setting tags we select `--recursive` it would set tags for all objects of the bucket and it's subdirs recursively. Additionally if this is combined with `--versions` recursively for all versions of objects of bucket and its subdirs would get the same tags set.

## Motivation and Context
Earlier there was an issue where with `--versions` option, the tags were set for the bucket level again and again, no of subdirs times under the bucket. This PR fixes the issue and also introduces `--recursive` option for setting the tags for objects/versions of subdirs recursively (if need be).
Similarly tags listing also used to list again and again for bucket of all it's subdirs. Fixed the same with `mc tag list` command.

## How to test this PR?
Execute the below commands for initial setup
```
# Create Buckets
./mc mb myminio/test1
./mc mb myminio/test2
./mc mb myminio/test3
./mc mb myminio/test4

# Enable versioning
./mc version enable myminio/test1
./mc version enable myminio/test2
./mc version enable myminio/test3

# Objects for bucket `test1`
./mc cp /etc/issue myminio/test1
./mc cp /etc/issue myminio/test1

# Objects for bucket `test2`
./mc cp /etc/hosts myminio/test2
./mc cp /etc/hosts myminio/test2
./mc cp /etc/issue myminio/test2/1/issue
./mc cp /etc/issue myminio/test2/1/issue
./mc cp /etc/issue myminio/test2/2/issue
./mc cp /etc/issue myminio/test2/2/issue
./mc cp /etc/issue myminio/test2/3/issue
./mc cp /etc/issue myminio/test2/3/issue

# Objects for bucket `test3`
./mc cp /etc/issue myminio/test3/1/issue
./mc cp /etc/issue myminio/test3/1/issue
./mc cp /etc/issue myminio/test3/2/issue
./mc cp /etc/issue myminio/test3/2/issue
./mc cp /etc/issue myminio/test3/3/issue
./mc cp /etc/issue myminio/test3/3/issue

# Objects for bucket `test4`
./mc cp /etc/issue myminio/test4/1/issue
./mc cp /etc/issue myminio/test4/2/issue
./mc cp /etc/issue myminio/test4/3/issue
```

Verify the below scenarios

### Scenario-1: Bucket has objects at bucket level only and no subdirs

```
./mc tag set myminio/test1 "key=val-bkt-test1"
Tags set for http://localhost:9000/test1.

./mc tag list myminio/test1
[...]

PR #4563, opened 2023-04-28
Remove tags for object versions

shtripat:
## Description
Earlier there was an issue where with `--versions` option where the tags were removed for the bucket level again and again, no of subdirs times under the bucket. This PR fixes the issue and also introduces `--recursive` option for removing the tags for objects/versions of subdirs recursively (if need be).

## How to test this PR?
Follow the below steps for verifying the changes
Step-1: Create a bucket with subdirs and enable versioning
Step-2: Load multiple versions of object to the subdirs of the bucket
Step-3: Set tags for different versions of the bucket objects
Step-3: Use `mc tag remove ALIAS/BUCKET --recursive --versions` and it should remove tags for all objects and all their versions in all subdirs
Step-4: Use command `mc tag remove ALIAS/BUCKET --recursive` and it should remove tags for all objects and only its top version in all subdirs

## Types of changes

## Checklist:

### DIFF
--- a/cmd/tag-list.go
+++ b/cmd/tag-list.go
@@ -48,7 +48,7 @@ var tagListFlags = []cli.Flag{
 	},
 	cli.BoolFlag{
 		Name:  "recursive, r",
-		Usage: "recursivley show tags for all objects of subdirs",
+		Usage: "recursivley show tags for all objects",
 	},
 }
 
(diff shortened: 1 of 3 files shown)


shtripat:
Depends on https://github.com/minio/mc/pull/4553

harshavardhana (approved):

PR #4581, opened 2023-05-26
Leave GC to runtime

### LINKED ISSUE #4580: mc consumes ~45% of 24 cores to produce 1.4MB/s of output.
I have a digitalocean droplet with two volumes, one has the files from my backup, roughly 750GB for under 3 million files. On the other volume I have my minio directory.

I just need to restore all my files from my backup, basically just copying from one volume to the other through minio.
The max I can get is 1.4MB/s with `mc mirror`, it starts around 100MB/s but quickly drops.

I ended up testing with a large VM with 24 cores and 190GB ram, all 24 cores are under around 45% load to produce that 1.4MB/s.
All that processing power is taken by mc, not the server and the server shows no error in the trace.

Running on:

`Linux TemporaryMinio-TO-DELETE 5.10.0-11-amd64 minio/minio#1 SMP Debian 5.10.92-1 (2022-01-18) x86_64 GNU/Linux`

Minio version:
```
minio version RELEASE.2023-05-18T00-05-36Z (commit-id=9d96b18df03f5978873705ac55959161f018fb48)

Runtime: go1.19.9 linux/amd64

License: GNU AGPLv3 <https://www.gnu.org/licenses/agpl-3.0.html>

Copyright: 2015-2023 MinIO, Inc.
```

klauspost:
## Description

Running manual GC for every file uploaded is extremely resource intensive on many small files.

Leave GC to the runtime unless dealing with a big file.

Fixes #4580

## How to test this PR?

See linked issue.

## Types of changes

### DIFF
--- a/cmd/parallel-manager.go
+++ b/cmd/parallel-manager.go
@@ -195,7 +195,10 @@ func (p *ParallelManager) enoughMemForUpload(uploadSize int64) bool {
 	}
 
 	smem := runtime.MemStats{}
-	runtime.GC()
+	if uploadSize > 50<<20 {
+		// GC if upload is bigger than 50MB.
+		runtime.GC()
+	}
 	runtime.ReadMemStats(&smem)
 
 	return estimateNeededMemoryForUpload(uploadSize)+smem.Alloc < p.maxMem


cmd/parallel-manager.go
```diff
@@ -195,7 +195,10 @@ func (p *ParallelManager) enoughMemForUpload(uploadSize int64) bool {
 	}
 
 	smem := runtime.MemStats{}
-	runtime.GC()
```
vadmeste, line 198: Maybe we don't need to run GC at all.. This is only to prevent mc crash due to the automatic parallel uploads

klauspost: @vadmeste I don't know all the logic here - that was also my initial solution, but since it wouldn't hurt on big files I just added the check.

vadmeste (approved):
LGTM

cmd/parallel-manager.go
klauspost: @vadmeste I will do what you prefer :)

vadmeste: > @vadmeste I will do what you prefer :)

Let's keep what you did right now. I just remembered the context more, it is mainly for users who are using mc in VMs with low memory (2 or 4 GB). We had multiple issues before but now it is quiet. 50MB is a good trade off.

harshavardhana (approved):

PR #4588, opened 2023-05-31
print in tabular form for callhome status

harshavardhana:


shtripat (approved):
LGTM. Tested and Verified.
```
$ ./mc support callhome status myminio                                                                                                                                    
┌─────────────────────────────┐
│ Features                    │
│─────────────────────────────│
│ Diagnostics      disabled   │
│ Logs             enabled    │
└─────────────────────────────┘
```

klauspost (approved):

cmd/support-callhome.go
```diff
@@ -91,20 +92,43 @@ type supportCallhomeMessage struct {
...
 		}
 		if len(s.Logs) > 0 {
-			msgs = append(msgs, "Logs is "+s.Logs)
+			rows = append(rows, table.Row{licInfoField("Logs"), licInfoVal(s.Logs)})
 		}
-		msg = strings.Join(msgs, "\n")
-	} else {
-		msg = s.Feature + " is now " + s.Action
+
+		t := table.New(
```
anjalshireesh, line 110: We need to call `initLicInfoColors()` in the beginning; otherwise the two rows (diagnostics and logs) get printed in different colors. The side effect of this is that the table column widths will have to be a little bigger; otherwise the values in cells get truncated. `20, 15` instead of `15, 10` would work.

anjalshireesh (changes requested):

cmd/support-callhome.go
anjalshireesh: callhome-status
screenshot of the rows in different colors

harshavardhana: updated @anjalshireesh

anjalshireesh (approved):
LGTM, tested

PR #4607, opened 2023-06-20
Improve Consistency in mc Command Behavior

r-scheele:
## Description
Modified the argument handling of 'mc cat' and 'mc pipe' commands to display help text when no arguments are passed, aligning them with the behavior of other mc commands.

## Motivation and Context
`mc stat` and `mc pipe` commands are executed without any arguments, instead of returning the expected help message or error notification, the commands cause the program to hang.

## How to test this PR?
- checkout to the branch `mc-command-consistency-fix`
- Build the program
- run `/mc cat` and `/mc pipe`

## Types of changes

### DIFF
--- a/cmd/cat-main.go
+++ b/cmd/cat-main.go
@@ -104,8 +104,8 @@ EXAMPLES:
 
 // checkCatSyntax - validate all the passed arguments
 func checkCatSyntax(ctx *cli.Context) {
-	if len(ctx.Args()) == 0 {
-		cli.ShowCommandHelpAndExit(ctx, "cat", 1) // last argument is exit code
+	if len(ctx.Args()) < 1 || len(ctx.Args()) > 2 {
+		showCommandHelpAndExit(ctx, 1) // last argument is exit code
 	}
 }
 
--- a/cmd/cat-main.go
+++ b/cmd/cat-main.go
@@ -104,7 +104,7 @@ EXAMPLES:
 
 // checkCatSyntax - validate all the passed arguments
 func checkCatSyntax(ctx *cli.Context) {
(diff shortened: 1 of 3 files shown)


klauspost (approved):
Thanks! LGTM

klauspost:
```
Running test_cat_stdin()
mc: test_cat_stdin: unable to redirect stdin to stdout using 'mc cat'
Error: make: *** [Makefile:57: verify] Error 1
Error: Process completed with exit code 2.
```
It seems like it is expected behavior that it just acts as pass-through on no args...

Not sure I see the point of that - could you fix up the test, and we can either take the PR as is, or not if someone decides it is the behavior we want to keep.

r-scheele:
I had a conversation with @kannappanr yesterday, so I believe fixing up the test is the right move. I've fixed it in my next commit. Please review to make sure it's the correct behavior.

klauspost (approved):
LGTM

cmd/cat-main.go
```diff
@@ -102,6 +102,13 @@ EXAMPLES:
 `,
 }
 
+// checkCatSyntax - validate all the passed arguments
+func checkCatSyntax(ctx *cli.Context) {
+	if len(ctx.Args()) < 1 || len(ctx.Args()) > 2 {
```
vadmeste, line 107: @r-scheele I think 'mc cat' supports multiple arguments, like `mc cat arg1 arg2 arg3 arg4, ..`

can you double check ?

r-scheele: You're right. My bad.

harshavardhana:
Don't create branches on `origin` create branches in your fork only - closing this PR - please re-submit @r-scheele

PR #4623, opened 2023-07-05
admin replicate update: rename --sync flag to --mode

poornas:
## Motivation and Context
ui 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-replicate-update.go
+++ b/cmd/admin-replicate-update.go
@@ -39,9 +39,16 @@ var adminReplicateUpdateFlags = []cli.Flag{
 		Usage: "endpoint for the site",
 	},
 	cli.StringFlag{
-		Name:  "sync",
-		Usage: "enable synchronous replication for this target, valid values are ['enable', 'disable'].",
-		Value: "disable",
+		Name:  "mode",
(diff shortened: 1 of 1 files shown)


harshavardhana (changes requested):
We still have to preserve previous flag for few releases and deprecate it later

cmd/admin-replicate-update.go
```diff
@@ -39,9 +39,9 @@ var adminReplicateUpdateFlags = []cli.Flag{
 		Usage: "endpoint for the site",
 	},
 	cli.StringFlag{
-		Name:  "sync",
-		Usage: "enable synchronous replication for this target, valid values are ['enable', 'disable'].",
-		Value: "disable",
+		Name:  "mode",
```
vadmeste, line 42: I think you can add 'sync' to Aliases field, to keep backward compatibility

poornas: well, the field changed to take on different values - so will have to validate this separately

harshavardhana (approved):

shtripat (approved):
lgtm, verified

PR #4626, opened 2023-07-08
Support mc top net

jiuker:
## Description

Support mc top net

## Types of changes

## Checklist:

### DIFF
--- a/cmd/support-top-net.go
+++ b/cmd/support-top-net.go
@@ -96,16 +96,16 @@ func mainSupportTopNet(ctx *cli.Context) error {
 	opts := madmin.MetricsOptions{
 		Type:     madmin.MetricNet,
 		Interval: time.Second,
-		ByDisk:   true,
+		ByHost:   true,
 		N:        ctx.Int("count"),
 	}
 
 	p := tea.NewProgram(initTopNetUI(endpoint, ctx.Int("count")))
 	go func() {
 		out := func(m madmin.RealtimeMetrics) {
-			for _, metric := range m.ByNet {
+			for endPoint, metric := range m.ByNet {
 				p.Send(topNetResult{
-					endPoint: metric.EndPoint,
+					endPoint: endPoint,
 					stats:    metric,
 				})
 			}
--- a/cmd/support-top-net.go
+++ b/cmd/support-top-net.go
@@ -134,7 +134,6 @@ func mainSupportTopNet(ctx *cli.Context) error {
 			}
 			p.Quit()
 		}
-
 	}()
 
 	if _, e := p.Run(); e != nil {
--- a/cmd/support-top-net.go
+++ b/cmd/support-top-net.go
@@ -132,8 +132,8 @@ func mainSupportTopNet(ctx *cli.Context) error {
 			if e != nil {
 				fatalIf(probe.NewError(e), "Unable to fetch top net events")
 			}
-			p.Quit()
 		}
+		p.Quit()
 	}()
 
 	if _, e := p.Run(); e != nil {
--- a/cmd/support-top-net.go
+++ b/cmd/support-top-net.go
@@ -1,0 +1,127 @@
+// Copyright (c) 2015-2022 MinIO, Inc.
+//
+// This file is part of MinIO Object Storage stack
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package cmd
+
+import (
+	"context"
+	"time"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/minio/cli"
+	"github.com/minio/madmin-go/v3"
+	"github.com/minio/mc/pkg/probe"
+)
+
+var supportTopNetFlags = []cli.Flag{
+	cli.IntFlag{
+		Name:  "count, c",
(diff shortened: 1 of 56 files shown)


cmd/support-top-net.go
```diff
@@ -0,0 +1,127 @@
+// Copyright (c) 2015-2022 MinIO, Inc.
```
klauspost, line 1:
```suggestion
// Copyright (c) 2015-2023 MinIO, Inc.
```

inb4 Harsha!

cmd/support-top-net.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/minio/madmin-go/v3"
+	"github.com/minio/mc/pkg/probe"
+)
+
+var supportTopNetFlags = []cli.Flag{
+	cli.IntFlag{
+		Name:  "count, c",
+		Usage: "show up to N Nets",
+		Value: 10,
+	},
```
klauspost, line 35: Is this copied from somewhere?

```suggestion
	cli.IntFlag{
		Name:  "count, c",
		Usage: "show top count interfaces",
		Value: 10,
	},
```
How is this determined?

cmd/support-top-net.go
```diff
@@ -0,0 +1,127 @@
...
+	for _, srv := range info.Servers {
+		endpoint = append(endpoint, srv.Endpoint)
+	}
+
+	// MetricsOptions are options provided to Metrics call.
+	opts := madmin.MetricsOptions{
+		Type:     madmin.MetricNet,
+		Interval: time.Second,
+		ByHost:   true,
+		N:        ctx.Int("count"),
```
klauspost, line 100: This means something else. This return "count" full results and stop returning more. There is no ordering on the server side, so you will need to do sorting on the client.

Maybe lift 'n' from https://github.com/minio/mc/blob/5f39522e69025bea6be0d350b65fdec2de2785c8/cmd/admin-scanner-status.go#L41

Adding host filtering (`--nodes`) should be free. Configurable interval is nice as well for scripts.

cmd/top-net-spinner.go
```diff
@@ -0,0 +1,146 @@
...
+
+	var data []metricNetMetricsWrap
+
+	for endPoint, metric := range m.currTopMap {
+		data = append(data, metricNetMetricsWrap{NetMetrics: metric, EndPoint: endPoint})
+	}
+
+	sort.Slice(data, func(i, j int) bool {
+		if m.sortAsc {
+			return data[i].NetStats.RxBytes < data[j].NetStats.RxBytes
```
klauspost, line 112: Why RxBytes?

RxBytes+TxBytes probably provides the best sorting value.

cmd/top-net-spinner.go
```diff
@@ -0,0 +1,146 @@
...
+
+type topNetUI struct {
+	spinner  spinner.Model
+	quitting bool
+
+	sortAsc   bool
+	count     int
+	endPoints []string
+
+	prevTopMap map[string]*madmin.NetMetrics
```
klauspost, line 41: What is this used for? I can't see it being read from anywhere.

cmd/top-net-spinner.go
```diff
@@ -0,0 +1,146 @@
...
+	currTopMap map[string]*madmin.NetMetrics
```
klauspost, line 42:
```suggestion
	currTopMap map[string]madmin.NetMetrics
```
Having a pointer provides no value for you, since a nil value doesn't make sense.

cmd/top-net-spinner.go
```diff
@@ -0,0 +1,146 @@
...
+	table.SetCenterSeparator("")
+	table.SetColumnSeparator("")
+	table.SetRowSeparator("")
+	table.SetHeaderLine(false)
+	table.SetBorder(false)
+	table.SetTablePadding("\t") // pad with tabs
+	table.SetNoWhiteSpace(true)
+	table.SetHeader([]string{"SERVER", "Face", "RECEIVE", "TRANSMIT"})
+
+	var data []metricNetMetricsWrap
```
klauspost, line 104:
```suggestion
	data := make([]metricNetMetricsWrap, 0, len(m.currTopMap))
```
Good habit: When you know the size you will need, allocate the exact size.

klauspost (commented):
Minor things.

klauspost:
Please add sample output with `--json` and without `--json`.

jiuker:
```./mc support top net alais
SERVER  FACE    RECEIVE         TRANSMIT
:9000      eth0     15 EiB/s        7.2 EiB/s
```
@klauspost

cmd/top-net-spinner.go
```diff
@@ -0,0 +1,147 @@
...
+	table.SetHeaderAlignment(tablewriter.ALIGN_CENTER)
+	table.SetAlignment(tablewriter.ALIGN_CENTER)
+	table.SetCenterSeparator("")
+	table.SetColumnSeparator("")
+	table.SetRowSeparator("")
+	table.SetHeaderLine(false)
+	table.SetBorder(false)
+	table.SetTablePadding("\t") // pad with tabs
+	table.SetNoWhiteSpace(true)
+	table.SetHeader([]string{"SERVER", "Face", "RECEIVE", "TRANSMIT", ""})
```
shtripat, line 98: Why `Face` is not all caps in header?

klauspost:
@jiuker Yes - that is expected in environments with limited permissions. That is why we check errors :)

(I don't think there is a solution - there may be, but I don't know - feel free to investigate)

jiuker:
> @jiuker Yes - that is expected in environments with limited permissions. That is why we check errors :)
> 
> (I don't think there is a solution - there may be, but I don't know - feel free to investigate)
> 
> (or did you have another Q?)

You mean what if the file doesn't have read access? @klauspost

klauspost:
@jiuker Yes - it may require more privileges than the user has. Not a linux expert here.

jiuker:
> @jiuker Yes - it may require more privileges than the user has. Not a linux expert here.

These errors are running on windows, on my machine currently linux is fine. @klauspost But I'm not sure what happens inside the container.

klauspost:
Just leave it for Windows. No need for special code for that. People will have to live with that.

klauspost:
> Please add sample output with `--json` and without `--json`.

Just so I can see what it outputs.

harshavardhana (approved):

PR #4663, opened 2023-08-17
Added `mc admin kms key list ALIAS` command

shtripat:
## Contribution License
All contributions in this pull request are licensed to the project maintainers 

## Description
This change adds an option to list all KMS keys for a MinIO instance.

## How to test this PR?
1. Deploy a MinIO instance with KES. Make sure KES is up and running

```
MINIO_KMS_KES_ENDPOINT=https://127.0.0.1:7373 MINIO_KMS_KES_CERT_FILE=minio.crt MINIO_KMS_KES_KEY_FILE=minio.key MINIO_KMS_KES_KEY_NAME=minio-default-key MINIO_KMS_KES_CAPATH=public.crt ./minio server ./test-kes
```

2. Create alias for the MinIO instance
3. Create few more KMS keys using command `mc admin kms  key create`
4. List all keys as below

```
$ ./mc admin kms key list myminio
┌─────────────────────────┐
│ KMS Keys                │
├─────┬───────────────────┤
│ S N │ NAME              │
├─────┼───────────────────┤
│   1 │ minio-default-key │
│   2 │ my-key-1          │
│   3 │ my-key-2          │
│   4 │ my-key-new        │
│   5 │ my-key-new1       │
└─────┴───────────────────┘

$ ./mc admin kms key list myminio --json
{
    "status": "success",
    "target": "myminio",
    "keys": [
        "minio-default-key",
        "my-key-1",
        "my-key-2",
        "my-key-new",
        "my-key-new1"
    ]
}

```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/admin-kms-key-list.go
+++ b/cmd/admin-kms-key-list.go
@@ -19,8 +19,11 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/fatih/color"
+	"github.com/jedib0t/go-pretty/v6/table"
+	"github.com/jedib0t/go-pretty/v6/text"
 	"github.com/minio/cli"
 	json "github.com/minio/colorjson"
 	"github.com/minio/mc/pkg/probe"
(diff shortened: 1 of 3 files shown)


aead (approved):

cmd/admin-kms-key-list.go
```diff
@@ -0,0 +1,92 @@
...
+
+	keys, e := client.ListKeys(globalContext, "*")
+	fatalIf(probe.NewError(e).Trace(args...), "Unable to list KMS keys")
+
+	for _, k := range keys {
+		printMsg(kmsKeyMsg{
+			Key: k.Name,
+		})
+	}
+	return nil
```
harshavardhana, line 76: We should actually print all the keys together @shtripat in a table form not one key per line.

We should also print when the key was created there is lot more information in kes.KeyInfo that can be pretty printed here.

It also looks like our `madmin-go` is behind what kes-go package can provide so we must update madin-go also to comply with the kes-go changes.

harshavardhana (changes requested):

cmd/admin-kms-key-list.go
shtripat: Ok, will check and send PR for `madmin-go` and use the updated info here for displaying.

aead: I agree about the table format.

Unfortunately, listing keys does not return any CreatedAt data, key algorithm etc.
These fields are not set by current stateless KES servers since they would have to fetch each key from the backend KMS to send that information to the client - which is expensive.

Instead the API (stateful and stateless) KES servers expose will be: https://github.com/minio/kes/blob/854064f68ab89c895d2eea9d03f41829bfba08fd/internal/api/types.go#L99

So, listing - as of now - will only return the name of the item (key, policy, secret, ...)
In the future this might be expended but for now, only the key names are returned.

harshavardhana: That is fine then - let's put it into a table format for now.

harshavardhana (approved):

PR #4691, opened 2023-09-21
Updates/merge multiple policy

r-scheele:
## Description
This PR merges multiple policies in `mc admin user policy TARGET USERNAME`    command, if more than one exists.

```bash
./mc admin user add myminio minio1 minio123
Added user `minio1` successfully.
```

```bash
$ ./mc admin policy list myminio
consoleAdmin
diagnostics
readonly
readwrite
writeonly
```

```bash
$ ./mc admin user policy myminio minio1
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "admin:*"
      ]
    },
    {
      "Effect": "Allow",
      "Action": [
        "kms:*"
      ]
    },
    {
      "Effect": "Allow",
      "Action": [
        "s3:*"
      ],
      "Resource": [
        "arn:aws:s3:::*"
      ]
    }
  ]
}
```


cmd/admin-user-policy.go
```diff
@@ -76,12 +78,33 @@ func mainAdminUserPolicy(ctx *cli.Context) error {
...
+
+	var policies []policy.Policy
+	for _, policyName := range policyNames {
+		policyInfo, err := getPolicyInfo(client, policyName)
+		if err != nil {
+			fatalIf(probe.NewError(err).Trace(), "Unable to fetch user policy document for policy "+policyName)
+			return nil
+		}
+
+		if len(policyNames) == 1 {
```
shtripat, line 91: This check can be done before for loop and print only policy and return.

cmd/admin-user-policy.go
```diff
@@ -76,12 +78,34 @@ func mainAdminUserPolicy(ctx *cli.Context) error {
...
-		fmt.Println(string(pinfo.Policy))
+	policyNames := strings.Split(user.PolicyName, ",")
+
+	if len(policyNames) == 1 {
+		policyInfo, err := getPolicyInfo(client, policyNames[0])
+		if err != nil {
+			fatalIf(probe.NewError(err).Trace(), "Unable to fetch user policy document")
+		}
+		fmt.Println(string(policyInfo.Policy))
+		return nil
```
harshavardhana, line 89:
```suggestion
	if len(policyNames) == 1 {
		policyInfo, err := getPolicyInfo(client, policyNames[0])
		if err != nil {
			fatalIf(probe.NewError(err).Trace(), "Unable to fetch user policy document")
		}
		fmt.Println(string(policyInfo.Policy))
		return nil
```

This is not needed not sure why we have this.

cmd/admin-user-policy.go
```diff
@@ -76,12 +78,34 @@ func mainAdminUserPolicy(ctx *cli.Context) error {
...
 	}
 
+	var policies []policy.Policy
+	for _, policyName := range policyNames {
```
harshavardhana, line 93:
```suggestion
	for _, policyName := range policyNames {
	         if policyName == "" {
	              continue
	         }
	}
```

harshavardhana (changes requested):

cmd/admin-user-policy.go
r-scheele: Thanks for the suggestion @harshavardhana `MergePolicies()` works for a policy too

cmd/admin-user-policy.go
r-scheele: Hi @harshavardhana i already submitted a fix for that in the next commit

harshavardhana (approved):

donatello (changes requested):
I think output should indicate that this is a merged policy output of the policies applied to the user to reduce chances of confusion.

cmd/admin-user-policy.go
```diff
@@ -76,12 +78,28 @@ func mainAdminUserPolicy(ctx *cli.Context) error {
...
-		fatalIf(probe.NewError(e).Trace(args...), "Unable to fetch user policy document for policy "+policy)
-		fmt.Println(string(pinfo.Policy))
+	policyNames := strings.Split(user.PolicyName, ",")
+
+	var policies []policy.Policy
+	for _, policyName := range policyNames {
+		if policyName == "" {
+			continue
+		}
+		policyInfo, err := getPolicyInfo(client, policyName)
```
harshavardhana, line 88:
```suggestion
		policyInfo, e := getPolicyInfo(client, policyName)
		fatalIf(probe.NewError(e).Trace(), "Unable to fetch user policy document for policy "+policyName)
```

cmd/admin-user-policy.go
```diff
@@ -76,12 +78,28 @@ func mainAdminUserPolicy(ctx *cli.Context) error {
...
+		if err != nil {
+			fatalIf(probe.NewError(err).Trace(), "Unable to fetch user policy document for policy "+policyName)
+		}
+
+		var policyObj policy.Policy
+		if err := json.Unmarshal(policyInfo.Policy, &policyObj); err != nil {
```
harshavardhana, line 94:
```suggestion
		if e := json.Unmarshal(policyInfo.Policy, &policyObj); e != nil {
```

cmd/admin-user-policy.go
```diff
@@ -76,12 +78,28 @@ func mainAdminUserPolicy(ctx *cli.Context) error {
...
+			fatalIf(probe.NewError(err).Trace(), "Unable to unmarshal policy")
```
harshavardhana, line 95:
```suggestion
			fatalIf(probe.NewError(e).Trace(), "Unable to unmarshal policy")
```

cmd/admin-user-policy.go
```diff
@@ -76,12 +78,28 @@ func mainAdminUserPolicy(ctx *cli.Context) error {
...
+		}
+		policies = append(policies, policyObj)
 	}
 
+	mergedPolicy := policy.MergePolicies(policies...)
+	policyStr, _ := json.Marshal(mergedPolicy)
```
harshavardhana, line 101:
```suggestion
	policyStr, _ := json.Marshal(mergedPolicy)
```

don't ignore errors or use `json.NewEncoder(os.Stdout).Encode(mergedPolicy)`

harshavardhana (changes requested):
In mc we treat `go` err as `e` and `probe.Error` as `err`

cmd/admin-user-policy.go
r-scheele: This makes sense, thanks @harshavardhana

harshavardhana (approved):

harshavardhana:
PTAL @donatello

harshavardhana:
PTAL @donatello @shtripat

shtripat (approved):
lgtm

PR #4712, opened 2023-10-11
Capture network device info in health report

anjalshireesh:

## Description

Upgrade to latest version of madmin-go - this will start capturing the network device information in the health report if used with latest version of minio.

Also add code to handle the new version (4) of the health info struct and introduce spinner for the net info.

## Motivation and Context

Useful for debugging, can be used to add check(s) related to network device driver and revision

## How to test this PR?

Requires https://github.com/minio/minio/pull/18213

- Generate health report (`mc support diag <alias> --airgap`)
- Verify that
  - the header of the JSON data contains version `4`
  - there is a new node for each server under `sys -> netinfo`

Note: on single-server setups where loopback device gets used, there will be an error `operation not supported`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/support-diag.go
+++ b/cmd/support-diag.go
@@ -296,6 +296,7 @@ func fetchServerDiagInfo(ctx *cli.Context, client *madmin.AdminClient) (interfac
 	diskHw := spinner("Disk Info", madmin.HealthDataTypeSysDriveHw)
 	osInfo := spinner("OS Info", madmin.HealthDataTypeSysOsInfo)
 	mem := spinner("Mem Info", madmin.HealthDataTypeSysMem)
+	netInfo := spinner("Net Info", madmin.HealthDataTypeSysNet)
 	process := spinner("Process Info", madmin.HealthDataTypeSysLoad)
 	config := spinner("Server Config", madmin.HealthDataTypeMinioConfig)
 	syserr := spinner("System Errors", madmin.HealthDataTypeSysErrors)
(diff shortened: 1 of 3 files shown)


cmd/support-diag.go
```diff
@@ -388,6 +403,20 @@ func fetchServerDiagInfo(ctx *cli.Context, client *madmin.AdminClient) (interfac
 			progressV2(info)
 		}
 		healthInfo = info
+	case madmin.HealthInfoVersion3:
```
klauspost, line 406: Why is a new version needed here? Aren't you just adding a new field?

cmd/support-diag.go
```diff
@@ -325,9 +326,23 @@ func fetchServerDiagInfo(ctx *cli.Context, client *madmin.AdminClient) (interfac
 			admin(len(info.Minio.Info.Servers) > 0)
 	}
 
+	progressV3 := func(info madmin.HealthInfoV3) {
```
shtripat, line 329: No difference in entries for `progressV2` and `progressV3`? Shouldn't we add `netInfo` here?

anjalshireesh: V3 is the existing version. Previous `progress` (that was handling v3) becomes `progressV3` and new `progress` (that handles v4) now has `netInfo` also.

shtripat (approved):
LGTM. Verified with https://github.com/minio/minio/pull/18213

cmd/support-diag.go
klauspost: But if it is a nil-able struct can't it just check if it is there or not?

harshavardhana (approved):

cmd/support-diag.go
anjalshireesh: If we don't introduce the new version, then with same health info version (v3), some (newer) minio versions will send the net info, whereas others won't. And mc won't know when to _expect_ this data in the response and when not to.

If this was just a part of some other top level data category (disk, os, mem, process, etc) then we did not have to show it as a separate line item in the progress, and would not have required a new version.

However, since this is a new category of data for which we want to show the progress (waiting -> received), but do it only when we know that minio is going to send this data, it is required to introduce the new version. (mc first reads only the header of the health info to get the version so that it knows what data to expect)

anjalshireesh:
As per discussion with @klauspost , the progress/spinner mechanism needs to be fixed in a way that we don't need to introduce a new version of healthinfo struct when adding new data categories. This enhancement will be taken up afresh, without adding a new version, after the spinner is fixed.

cmd/support-diag.go
anjalshireesh: Closing this PR now. Will revisit after the spinner mechanism is fixed.

PR #4705, opened 2023-10-02
scanner: Limit the active disks if no enough term space

vadmeste:
## Community Contribution License
All community contributions in this pull request are licensed to the project maintainers
under the terms of the [Apache 2 license] (https://www.apache.org/licenses/LICENSE-2.0). 
By creating this pull request I represent that I have the right to license the 
contributions to the project maintainers under the Apache 2 license.

## Description
If the terminal height is not enough to show the number of active disks, 
which is likley the case in a large deployment, print only few entries
until the number of lines reaches the terminal height.

## Motivation and Context
Do not overload mc admin scanner status screen with large deployments

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (provides speedup with no functional changes)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Fixes a regression (If yes, please add `commit-id` or `PR #` here)
- [ ] Unit tests added/updated
- [ ] Internal documentation updated
- [ ] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fmc%2fpull%2fNNNNN)

PR #4721, opened 2023-10-17
Fix missing colours on windows

klauspost:
## Description

#4705 removed all color from Windows output.

## Motivation and Context

The world should have color diversity.

## How to test this PR?

Run any command on Windows. Observe greyness.

## Types of changes

## Checklist:
- [x] Fixes a regression #4705

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -116,9 +116,9 @@ func Main(args []string) error {
 	probe.SetAppInfo("Commit", ShortCommitID)
 
 	// Fetch terminal size, if not available, automatically
-	// set globalQuiet to true.
+	// set globalQuiet to on non-window.
 	if w, h, e := term.GetSize(int(os.Stdin.Fd())); e != nil {
-		globalQuiet = true
+		globalQuiet = runtime.GOOS != "windows"
 	} else {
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

vadmeste (approved):
LGTM

cmd/main.go
```diff
@@ -116,9 +116,9 @@ func Main(args []string) error {
 	probe.SetAppInfo("Commit", ShortCommitID)
 
 	// Fetch terminal size, if not available, automatically
-	// set globalQuiet to true.
+	// set globalQuiet to true on non-window.
 	if w, h, e := term.GetSize(int(os.Stdin.Fd())); e != nil {
-		globalQuiet = true
+		globalQuiet = runtime.GOOS != "windows"
```
vadmeste, line 121: @klauspost https://cs.opensource.google/go/x/term/+/refs/tags/v0.13.0:term_windows.go, it looks like the term package we are using supports Windows; I am not sure what is going on

harshavardhana: May be @klauspost has one of those special windows versions :-)

klauspost: @vadmeste I don't know. It fails on mine.

PR #4745, opened 2023-11-04
set-json to fail on custom policies without .json

r-scheele:
Fix set-json command to fail on custom policies without .json

This commit addresses the bug where the `set-json` command incorrectly
succeeds for custom policies even when the policy file doesn't have a .json
extension. The fix ensures that the command fails as expected, prompting the user to use the correct file.

`./mc anonymous set-json myminio/mydata ~/test/anonymous.json`

Now, `set-json` will return an error message if the provided policy file is not a .json, aligning the behavior with the documented functionality.

Related to the discussion with @kannappanr

### DIFF
--- a/cmd/access-perms.go
+++ b/cmd/access-perms.go
@@ -17,7 +17,12 @@
 
 package cmd
 
-import "path/filepath"
+import (
+	"os"
+	"path/filepath"
+
+	json "github.com/minio/colorjson"
+)
 
 // isValidAccessPERM - is provided access perm string supported.
 func (b accessPerms) isValidAccessPERM() bool {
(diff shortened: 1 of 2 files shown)


klauspost (commented):
File extension shouldn't really matter IMO.

r-scheele:
> File extension shouldn't really matter IMO.

The previous code only validates the file ext, so i think only `.json` policies are valid, i've added more validations asides the file ext and better error handling that prompted the discussion with @kannappanr. PTAL @klauspost @harshavardhana 


> File extension shouldn't really matter IMO.

cmd/access-perms.go
```diff
@@ -28,12 +33,50 @@ func (b accessPerms) isValidAccessPERM() bool {
 	return false
 }
 
+type PolicyDocument struct {
```
harshavardhana, line 37: why a new struct - can you not use `pkg/policy` ?

harshavardhana (changes requested):

harshavardhana:
> > File extension shouldn't really matter IMO.
> 
> The previous code only validates the file ext, so i think only `.json` policies are valid, i've added more validations asides the file ext and better error handling that prompted the discussion with @kannappanr. PTAL @klauspost @harshavardhana
> 
> > File extension shouldn't really matter IMO.

the file extension check can be relaxed though.

r-scheele:
> > > File extension shouldn't really matter IMO.
> > 
> > The previous code only validates the file ext, so i think only `.json` policies are valid, i've added more validations asides the file ext and better error handling that prompted the discussion with @kannappanr. PTAL @klauspost @harshavardhana
> > 
> > > File extension shouldn't really matter IMO.
> 
> the file extension check can be relaxed though. 

Do you mean I should remove it?

harshavardhana:
> > > > File extension shouldn't really matter IMO.
> > > 
> > > 
> > > The previous code only validates the file ext, so i think only `.json` policies are valid, i've added more validations asides the file ext and better error handling that prompted the discussion with @kannappanr. PTAL @klauspost @harshavardhana
> > > > File extension shouldn't really matter IMO.
> > 
> > 
> > the file extension check can be relaxed though.
> 
> Do you mean I should remove it?

Yes @r-scheele

cmd/anonymous-main.go
```diff
@@ -52,7 +52,7 @@ var anonymousCmd = cli.Command{
 
 USAGE:
   {{.HelpName}} [FLAGS] set PERMISSION TARGET
-  {{.HelpName}} [FLAGS] set-json TARGET FILE
+  {{.HelpName}} [FLAGS] set-json FILE TARGET
```
vadmeste, line 55: Can you update the example below as well ?

cmd/access-perms.go
```diff
@@ -29,7 +34,32 @@ func (b accessPerms) isValidAccessPERM() bool {
...
+		fatalIf(errDummy().Trace(), "Invalid policy version. Only 2012-10-17 is supported.")
+		return false
+	}
+
+	for _, statement := range policy.Statements {
+		if statement.Effect != "Allow" && statement.Effect != "Deny" {
+			fatalIf(errDummy().Trace(), "Invalid policy effect. Only Allow and Deny are supported.")
+			return false
+		}
+	}
```
vadmeste, line 60: I am not too much into client side validation; because the server already does that

cmd/access-perms.go
```diff
@@ -17,7 +17,12 @@
 
 package cmd
 
-import "path/filepath"
+import (
+	"os"
+
+	json "github.com/minio/colorjson"
+	"github.com/minio/minio-go/v7/pkg/policy"
```
harshavardhana, line 24: No don't use this use this https://pkg.go.dev/github.com/minio/pkg/v2@v2.0.3/policy

harshavardhana (changes requested):

harshavardhana (approved):

cmd/access-perms.go
harshavardhana: I think he is adding that to fail early when we do not know the file extension, its just a more stricter check.

klauspost (approved):

PR #4747, opened 2023-11-08
Check that destination server is MinIO for `mc replicate add`

taran-p:

## Description
Verifies that the target server for a bucket replication command is MinIO

## Motivation and Context
It is currently possible to replicate buckets to non-MinIO hosts, which can lead to unexpected behavior

## How to test this PR?
Attempt to replicate to MinIO and non-MinIO servers.

## Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/cmd/replicate-add.go
+++ b/cmd/replicate-add.go
@@ -284,6 +284,20 @@ func mainReplicateAdd(cliCtx *cli.Context) error {
 	fatalIf(cerr, "unable to initialize admin connection.")
 
 	bktTarget := fetchRemoteTarget(cliCtx)
+
+	// Check if target url is MinIO
+	healthClt, e := madmin.NewAnonymousClient(bktTarget.Endpoint, bktTarget.Secure)
+	fatalIf(probe.NewError(e).Trace(bktTarget.Endpoint), "unable to parse remote target")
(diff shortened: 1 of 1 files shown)


harshavardhana:
This needs to be also on the server side @taran-p `client` side you don't have much context. Some times certain endpoints might return 200 OK. You should be able to perform "MinIO" only checks to make sure this is a MinIO-only server for that it needs to be done via an authenticated endpoint.

vadmeste:
What happens if you this command is executed against AWS S3 ? no error at all ?

taran-p:
@harshavardhana do you mean I need to add an authenticated endpoint for this or does it already exist?

When I tested on S3 and GCS it errored properly.

harshavardhana:
> @harshavardhana do you mean I need to add an authenticated endpoint for this or does it already exist?

It would be best if you let the MinIO server decide this where we can do specific API checks. Anyone can fake this endpoint if its purely on `mc` to decide.

We do not have to do a heuristical approach, we can be precise. For that, you need to let this information be sent to MinIO and then MinIO validates whether it that it's the right endpoint.

PR #4752, opened 2023-11-09
add STS support for `mc admin` command

pjuarezd:

## Description
add STS support for `mc admin` commands

## Motivation and Context

To allow invoke subcommands in the `admin` section such as `user`, `group`, `policy`, etc.

Example:

```
MC_HOST_demomc=http://localhost:9000  \
MC_STS_ENDPOINT=https://sts.minio-operator.svc.cluster.local:4223/sts/ns-1 \
MC_WEB_IDENTITY_TOKEN_FILE=/var/run/secrets/kubernetes.io/serviceaccount/token \
./mc admin policy create demomc bucket-rw policy.json 
```

## Types of changes

## Checklist:

### DIFF
--- a/cmd/client-admin.go
+++ b/cmd/client-admin.go
@@ -20,7 +20,6 @@ package cmd
 import (
 	"crypto/tls"
 	"fmt"
-	"github.com/klauspost/compress/gzhttp"
 	"hash/fnv"
 	"net"
 	"net/http"
@@ -28,6 +27,8 @@ import (
 	"os"
 	"sync"
 	"time"
+
+	"github.com/klauspost/compress/gzhttp"
 
 	"github.com/minio/pkg/v2/env"
 
--- a/cmd/client-admin.go
+++ b/cmd/client-admin.go
@@ -24,13 +24,17 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"sync"
 	"time"
+
+	"github.com/minio/pkg/v2/env"
 
 	"github.com/klauspost/compress/gzhttp"
 	"github.com/mattn/go-ieproxy"
 	"github.com/minio/madmin-go/v3"
 	"github.com/minio/mc/pkg/httptracer"
+	"github.com/minio/mc/pkg/limiter"
 	"github.com/minio/mc/pkg/probe"
 	"github.com/minio/minio-go/v7/pkg/credentials"
 )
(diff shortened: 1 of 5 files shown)


cmd/client-admin.go
```diff
@@ -67,8 +71,74 @@ func NewAdminFactory() func(config *Config) (*madmin.AdminClient, *probe.Error)
...
+
+			// V4 Credentials
+			credsV4 := &credentials.Static{
+				Value: credentials.Value{
+					AccessKeyID:     config.AccessKey,
+					SecretAccessKey: config.SecretKey,
+					SessionToken:    config.SessionToken,
+					SignerType:      credentials.SignatureV4,
+				},
+			}
```
harshavardhana, line 138: We could make this a common function? only thing that needs to not be added is `signature.v2` which is not supported with Admin APIs.

harshavardhana (changes requested):

cmd/client-admin.go
pjuarezd: right, thats why I left v2 outside.

Should also reuse `transport` instead of overwrite it? there is a difference in couple of different settings:

```
MaxIdleConnsPerHost:   1024,
WriteBufferSize:       32 << 10, // 32KiB moving up from 4KiB default
ReadBufferSize:        32 << 10, // 32KiB moving up from 4KiB default
```

harshavardhana: > right, thats why I left v2 outside.
> 
> Should also reuse `transport` instead of overwrite it? there is a difference in couple of different settings:
> 
> ```
> MaxIdleConnsPerHost:   1024,
> WriteBufferSize:       32 << 10, // 32KiB moving up from 4KiB default
> ReadBufferSize:        32 << 10, // 32KiB moving up from 4KiB default
> ```

`mc` just runs and dies so there is nothing to re-use really. The transport is re-used always until `mc` is running.

and it is not created multiple times as such there is a `clientCache` of the `mc` config which is loaded per target in memory.

harshavardhana: Yes these settings are okay we should combine both transports just that add v2 optionally when its used by an S3 client.

cmd/client-admin.go
```diff
@@ -92,7 +162,7 @@ func NewAdminFactory() func(config *Config) (*madmin.AdminClient, *probe.Error)
 				tlsConfig.InsecureSkipVerify = true
 			}
 
-			var transport http.RoundTripper = &http.Transport{
+			transport = &http.Transport{
```
dvaldivia, line 165: this is duplicate code, it's better to move this above where you start the credentials chain

dvaldivia, line 165: this is duplicate code, it's better to move this above where you start the credentials chain

dvaldivia (changes requested):

pjuarezd:
PR is not getting updated with new commits on the branch, nor able to reopen it. Will open a new one

PR #4816, opened 2024-01-11
Removing termPager - its causing mc --help to not exit back to terminal on finish

zveinn:
The problem here is that when you run the --help it won't exit back to the terminal. We've had some complaints about this. The code removed is trying to adjust to random screen sizes but we should probably just be printing to stdout.


harshavardhana:
Term pager was implemented to provide a way to scroll through the help such as `mc cp --help` which can get lengthier. 

What we can do is disable this pager on dumb terminals such as TERM= values that are not `xterm` or `xterm-256color`

vadmeste:
> Term pager was implemented to provide a way to scroll through the help such as mc cp --help which can get lengthier.
> What we can do is disable this pager on dumb terminals such as TERM= values that are not xterm or xterm-256color

Agree, it is better to fix users issues rather than completely remove the term pager.

zveinn:
That's not really a solution either.. 
Changing the term value to disable this paging functionality breaks a whole lot of other things. Sure, the TERM env is reset when you open a new terminal but it's reaaalllyyy annoying. 
image

harshavardhana:
no, it is kind of inspired by this `aws s3 help` , and also like `git commit --help` - it is not very unusual. 

Why? Because `mc` commands are self-documented, we kind of have to show them in some scrollable manner by default; otherwise, the entire help goes away from terminal screens, it's painful to read all that text when we can't read it like a page.

Just like other tools do as shown above ^^.

zveinn:
hmm.. then maybe make a flag to disable it instead ?

zveinn:
Created a new PR for this: https://github.com/minio/mc/pull/4819

PR #4819, opened 2024-01-12
Making cli pager optional + better visual indicators for scrolling

zveinn:

## Description
This PR is a replacement for [another PR](https://github.com/minio/mc/pull/4816) which completely removed the pager. The decision was made to keep the pager but instead trigger on the TERM environment variable. 

I added an additional way to manually disable the pager using a `--disable-pager` flag.
**The name and/or usage of the flag is up for discussion in this PR**

## Motivation and Context
We had some clients talking about the mc client not exiting straight to terminal. Many cli tools have this scroll behavior but not everyone is a fan of it. This way we can prevent the behavior in non-supported terminals and let the user decide whether they want it or not. 

# Additional formatting changes
Currently the scroll indicator is located on the bottom right of the screen, which is normally the last part of the screen users look at when it comes to GUIs/Terminal usage. I've moved the percentage sign to the bottom left, where user is more likely to see it and added a scroll status indicator for the footer. 

NO SCROLL: 
image

SCROLL:
image

## How to test this PR?
```bash
$ ./mc --help --disable-pager
```

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -39,11 +39,6 @@ var globalFlags = []cli.Flag{
 		EnvVar: envPrefix + "QUIET",
 	},
 	cli.BoolFlag{
-		Name:   "disable-pager, dp",
-		Usage:  "disable mc internal pager and print to raw stdout",
-		EnvVar: envPrefix + "DISABLE_PAGER",
-	},
-	cli.BoolFlag{
 		Name:   "no-color",
 		Usage:  "disable color theme",
 		EnvVar: envPrefix + "NO_COLOR",
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -39,6 +39,12 @@ var globalFlags = []cli.Flag{
 		EnvVar: envPrefix + "QUIET",
 	},
 	cli.BoolFlag{
+		Name:   "disable-pager, dp",
+		Usage:  "disable mc internal pager and print to raw stdout",
+		EnvVar: envPrefix + "DISABLE_PAGER",
+		Hidden: true,
+	},
+	cli.BoolFlag{
 		Name:   "no-color",
 		Usage:  "disable color theme",
 		EnvVar: envPrefix + "NO_COLOR",
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -39,6 +39,11 @@ var globalFlags = []cli.Flag{
 		EnvVar: envPrefix + "QUIET",
 	},
 	cli.BoolFlag{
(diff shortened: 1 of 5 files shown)


zveinn:
Found a small bug, pager disappears when terminal is too small, fixing now.

harshavardhana (approved):

zveinn:
> @zveinn I like the progress bar, it is it expected that the / char does not expand to the whole screen width ? looks good to me otherwise.

It does once you reach 100% viewPortHeight .. aka once you reach the bottom.

cmd/term-pager.go
```diff
@@ -100,10 +101,31 @@ func (m model) headerView() string {
...
+	totalLength := m.viewport.Width - lipgloss.Width(info)
+	var finishedCount int
+	var onePart int
+	if totalLength <= 99 {
+		onePart = m.viewport.Width / 5
+		if !math.IsNaN(m.viewport.ScrollPercent()) {
+			finishedCount = onePart * min(5, int(m.viewport.ScrollPercent()*100)/20)
+		}
+	} else {
+		onePart = max(1, totalLength/100)
```
vadmeste, line 120: I think onePart needs to be float so the percentage calculation will be more correct. Can you check 100% here with different terminal widh in your machine ? it does not work very well for me

zveinn: The precision won't be better since it's all ints. Even if I transform the ints to floats they still need to go back. and it's a 1-100 so decimals won't really matter. 

However, there is a big in it .. the part where I break it down into 5 parts to make it fit smaller terminals better seems to have been done wrong so it's using 1/5 paging for all sizes. 

I will fix that on monday

zveinn: You were right, a float conversion gave me the accuracy needed for both small and big terminals. Thank you.

cmd/flags.go
```diff
@@ -38,6 +38,11 @@ var globalFlags = []cli.Flag{
 		Usage:  "disable progress bar display",
 		EnvVar: envPrefix + "QUIET",
 	},
+	cli.BoolFlag{
+		Name:   "disable-pager, dp",
+		Usage:  "disable mc internal pager and print to raw stdout",
+		EnvVar: envPrefix + "DISABLE_PAGER",
```
klauspost, line 44: Make this hidden. It is not needed for 99% of the time.

klauspost (commented):
Rest LGTM

vadmeste (approved):
LGTM

Issue #4082, opened 2022-05-08
mc mirror --exclude seems not to be working on windows


CrashLaker:
## Expected behavior
excluding folder when uploading
actually idk if it's actually only listing then excluding or both

## Actual behavior
not excluding

## Steps to reproduce the behavior
```
C:\Users\User>mc.exe mirror --exclude "AppData\*"  . windows-pc/backup/
 0 B / ? [                                                                                                           =]
mc.exe: <ERROR> Failed to perform mirroring Insufficient permissions to access this file `C:\Users\User\AppData\Local\ElevatedDiagnostics`
```

## mc --version
- (paste output of `mc --version`)
mc version RELEASE.2022-05-04T06-07-55Z

## System information
windows 10

thebetauser:
Is there any update on this? Seems the --exclude only works on files however when backing up some python projects I want to exclude all the directories that are not needed such as .git and __pycache__ and the exclude commands doesn't work on folders

This command doesn't filter any of the folders

`mc mirror --exclude ".*" --exclude "venv" --exclude "*pycache*" --exclude "*-info" "E:\Python Projects" s3/work_backup`

MC Version:
`mc version RELEASE.2022-08-23T05-45-20Z (commit-id=07fffc3bc88f70a73c3714d9af23122f611b76cf)
Runtime: go1.18.5 windows/amd64
Copyright (c) 2015-2022 MinIO, Inc.
License GNU AGPLv3 <https://www.gnu.org/licenses/agpl-3.0.html>`

cwinters8:
I can confirm this is still an issue. I'm unable to specify a directory to exclude using either the `dir` or `dir/*` syntax.

![exclude tmp from mirror](https://github.com/minio/mc/assets/40615752/652004b3-856e-4f57-80fc-73cb1c9ec1b3)

<img width="947" alt="exclude tmp/* from mirror" src="https://github.com/minio/mc/assets/40615752/23026d41-2372-4c34-97cb-c7284b509651">

cwinters8:
Further experimentation seems to reveal `--exclude` is not functional with file names or specific paths to files either

<img width="938" alt="exclude hello.txt from mirror" src="https://github.com/minio/mc/assets/40615752/f45609e2-24ab-4faf-b5c9-b839a27c765b">

Also tried removing `--remove` and enclosing the path in quotes just in case those might be factors, but saw the same behavior

[...]

PR #4825, opened 2024-01-17
fix: mirror with local path

jiuker:

## Description

There have path with `/` start when local file path.
before:
`mc mirror --exclude "*.yaml" localPath` will not work.
now:
it work.

## Types of changes
- [x] Bug fix https://github.com/minio/mc/issues/4082

## Checklist:

### DIFF
--- a/cmd/difference_test.go
+++ b/cmd/difference_test.go
@@ -24,26 +24,40 @@ import (
 var testCases = []struct {
 	pattern []string
 
-	object string
+	srcSuffix string
 
 	match bool
+
+	typ ClientURLType
 }{
-	{nil, "testfile", false},
-	{[]string{"test*"}, "testfile", true},
-	{[]string{"file*"}, "file/abc/bcd/def", true},
-	{[]string{"*"}, "file/abc/bcd/def", true},
-	{[]string{""}, "file/abc/bcd/def", false},
-	{[]string{"abc*"}, "file/abc/bcd/def", false},
-	{[]string{"abc*", "*abc/*"}, "file/abc/bcd/def", true},
-	{[]string{"*.txt"}, "file/abc/bcd/def.txt", true},
-	{[]string{".*"}, ".sys", true},
-	{[]string{"*."}, ".sys.", true},
+	{nil, "testfile", false, objectStorage},
+	{[]string{"test*"}, "testfile", true, objectStorage},
+	{[]string{"file*"}, "file/abc/bcd/def", true, objectStorage},
+	{[]string{"*"}, "file/abc/bcd/def", true, objectStorage},
+	{[]string{""}, "file/abc/bcd/def", false, objectStorage},
+	{[]string{"abc*"}, "file/abc/bcd/def", false, objectStorage},
+	{[]string{"abc*", "*abc/*"}, "file/abc/bcd/def", true, objectStorage},
+	{[]string{"*.txt"}, "file/abc/bcd/def.txt", true, objectStorage},
+	{[]string{".*"}, ".sys", true, objectStorage},
+	{[]string{"*."}, ".sys.", true, objectStorage},
+	{nil, "testfile", false, fileSystem},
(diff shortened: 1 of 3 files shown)


cmd/difference_test.go
```diff
@@ -24,26 +24,40 @@ import (
 var testCases = []struct {
 	pattern []string
 
-	object string
+	srcSuffix string
 
 	match bool
+
+	typ ClientURLType
```
shtripat, line 31:
```suggestion
	type ClientURLType
```

klauspost: @shtripat  `type` is a keywork.. Pretty sure you can't use it here.

shtripat: ah yes. we are fine then

vadmeste (commented):
@jiuker not sure, but would this work here, 'clnt.GetURL().Separator'  instead of passing the client type ? worth a try

jiuker:
> @jiuker not sure, but would this work here, 'clnt.GetURL().Separator' instead of passing the client type ? worth a try

```go
func newClientURL(urlStr string) *ClientURL {
	scheme, rest := getScheme(urlStr)
	if strings.HasPrefix(rest, "//") {
		// if rest has '//' prefix, skip them
		var authority string
		authority, rest = splitSpecial(rest[2:], "/", false)
		if rest == "" {
			rest = "/"
		}
		host := getHost(authority)
		if host != "" && (scheme == "http" || scheme == "https") {
			return &ClientURL{
				Scheme:          scheme,
				Type:            objectStorage,
				Host:            host,
				Path:            rest,
				SchemeSeparator: "://",
				Separator:       '/',
			}
		}
	}
	return &ClientURL{
		Type:      fileSystem,
		Path:      rest,
		Separator: filepath.Separator,
	}
}
```
They can be the same.    @vadmeste

cmd/mirror-url.go
```diff
@@ -89,7 +89,15 @@ func checkMirrorSyntax(ctx context.Context, cliCtx *cli.Context, encKeyDB map[st
...
 }
 
-func matchExcludeOptions(excludeOptions []string, srcSuffix string) bool {
+func matchExcludeOptions(excludeOptions []string, srcSuffix string, typ ClientURLType) bool {
+	// if type is file system, remove leading slash
+	if typ == fileSystem {
+		if strings.HasPrefix(srcSuffix, "/") {
+			srcSuffix = srcSuffix[1:]
+		} else if runtime.GOOS == "windows" && strings.HasPrefix(srcSuffix, "\\") {
+			srcSuffix = srcSuffix[2:]
```
klauspost, line 98: Why are you trimming 2 chars here? `"\\"` is just one char: `\`.

cmd/mirror-url.go
```diff
@@ -89,7 +89,15 @@ func checkMirrorSyntax(ctx context.Context, cliCtx *cli.Context, encKeyDB map[st
 	return
 }
 
-func matchExcludeOptions(excludeOptions []string, srcSuffix string) bool {
+func matchExcludeOptions(excludeOptions []string, srcSuffix string, typ ClientURLType) bool {
+	// if type is file system, remove leading slash
+	if typ == fileSystem {
+		if strings.HasPrefix(srcSuffix, "/") {
+			srcSuffix = srcSuffix[1:]
+		} else if runtime.GOOS == "windows" && strings.HasPrefix(srcSuffix, "\\") {
```
harshavardhana, line 97:
```suggestion
		} else if runtime.GOOS == "windows" && strings.HasPrefix(srcSuffix, `\`) {
```

cmd/mirror-url.go
harshavardhana, line 98:
```suggestion
			srcSuffix = srcSuffix[1:]
```

harshavardhana (changes requested):

klauspost (approved):

harshavardhana (approved):

PR #4853, opened 2024-02-21
removing deprecated encryption keys ( new keys will be added soon )

zveinn:
AB wants us to remove the old encryption keys from the mc. This will happen in a PR I will make next week. Because of the deprecation happening we should not include the old keys in these new commands. I will add the new keys to put and get in the coming PR.

### DIFF
--- a/cmd/put-main.go
+++ b/cmd/put-main.go
@@ -51,7 +51,7 @@ var putCmd = cli.Command{
 	Action:       mainPut,
 	OnUsageError: onUsageError,
 	Before:       setGlobalsFromContext,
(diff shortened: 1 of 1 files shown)


jiuker (approved):

bh4t:
@harshavardhana @kannappanr can you please review this?

jiuker (changes requested):
`get-main.go` have that too.

zveinn:
It´s fine.. I´ll just close this PR and adjust it in the upcoming PR. A few days probably won´t matter.

PR #4854, opened 2024-02-22
Add command for uploading file to SUBNET issue

anjalshireesh:

## Description

`mc support upload --issue <issueNum> --comment <msg> <alias> </path/to/file>`

Will upload the given file to the given SUBNET issue
The file will be compressed using `zstd` before upload.

## Motivation and Context

For customer's ease of sharing required logs to SUBNET issues

## How to test this PR?

`mc support upload --issue <issueNum> --comment <msg> <alias> </path/to/file>`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/subnet-utils.go
+++ b/cmd/subnet-utils.go
@@ -61,12 +61,20 @@ func subnetBaseURL() string {
 	return subnet.BaseURL(globalDevMode)
 }
 
+func subnetIssueURL(issueNum int) string {
+	return fmt.Sprintf("%s/issues/%d", subnetBaseURL(), issueNum)
+}
+
 func subnetLogWebhookURL() string {
 	return subnetBaseURL() + "/api/logs"
 }
 
-func subnetUploadURL(uploadType, filename string) string {
-	return fmt.Sprintf("%s/api/%s/upload?filename=%s", subnetBaseURL(), uploadType, filename)
+func subnetUploadURL(uploadType, filename string, params url.Values) string {
+	if params == nil {
+		params = url.Values{}
+	}
+	params.Add("filename", filename)
+	return fmt.Sprintf("%s/api/%s/upload?%s", subnetBaseURL(), uploadType, params.Encode())
 }
 
 func subnetRegisterURL() string {
@@ -719,9 +727,6 @@ func uploadFileToSubnet(alias, filename, reqURL string, headers map[string]strin
 		return "", e
 	}
 
-	// Delete the file after successful upload
-	os.Remove(filename)
-
 	// ensure that both api-key and license from
 	// SUBNET response are saved in the config
 	extractAndSaveSubnetCreds(alias, resp)
@@ -729,7 +734,7 @@ func uploadFileToSubnet(alias, filename, reqURL string, headers map[string]strin
 	return resp, e
 }
 
-func subnetUploadReq(url, filename string) (*http.Request, error) {
+func subnetUploadReq(url, filePath string) (*http.Request, error) {
 	r, w := io.Pipe()
 	mwriter := multipart.NewWriter(w)
 	contentType := mwriter.FormDataContentType()
@@ -744,12 +749,12 @@ func subnetUploadReq(url, filename string) (*http.Request, error) {
 			w.CloseWithError(e)
 		}()
 
-		part, e = mwriter.CreateFormFile("file", filepath.Base(filename))
+		part, e = mwriter.CreateFormFile("file", filepath.Base(filePath))
 		if e != nil {
 			return
 		}
 
-		file, e := os.Open(filename)
+		file, e := os.Open(filePath)
 		if e != nil {
 			return
 		}
--- a/cmd/subnet-utils.go
+++ b/cmd/subnet-utils.go
@@ -35,6 +35,7 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/klauspost/compress/zstd"
 	"github.com/minio/cli"
 	"github.com/minio/madmin-go/v3"
 	"github.com/minio/mc/pkg/probe"
@@ -69,12 +70,8 @@ func subnetLogWebhookURL() string {
 	return subnetBaseURL() + "/api/logs"
 }
 
-func subnetUploadURL(uploadType, filename string, params url.Values) string {
-	if params == nil {
-		params = url.Values{}
-	}
-	params.Add("filename", filename)
-	return fmt.Sprintf("%s/api/%s/upload?%s", subnetBaseURL(), uploadType, params.Encode())
+func subnetUploadURL(uploadType string) string {
+	return fmt.Sprintf("%s/api/%s/upload", subnetBaseURL(), uploadType)
 }
 
 func subnetRegisterURL() string {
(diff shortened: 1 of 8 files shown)


kanagarajkm (approved):

kanagarajkm (approved):

cmd/support-upload.go
```diff
@@ -0,0 +1,146 @@
+// Copyright (c) 2015-2022 MinIO, Inc.
```
harshavardhana, line 1:
```suggestion
// Copyright (c) 2015-2024 MinIO, Inc.
```

cmd/support-upload.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// JSON jsonified upload message
+func (s supportUploadMessage) JSON() string {
+	return toJSON(s)
+}
+
+var supportUploadCmd = cli.Command{
+	Name:            "upload",
+	Usage:           "upload file to SUBNET issue",
```
harshavardhana, line 68:
```suggestion
	Usage:           "upload file to a SUBNET issue",
```

harshavardhana (changes requested):

cmd/subnet-utils.go
```diff
@@ -807,7 +812,15 @@ func initSubnetConnectivity(ctx *cli.Context, aliasedURL string, forUpload bool,
...
 
 func validateSubnetFlags(ctx *cli.Context, forUpload bool) error {
-	if !globalAirgapped {
+	supportsAirgap := false
+	for _, flag := range ctx.FlagNames() {
+		if flag == "airgap" {
+			supportsAirgap = true
+			break
+		}
+	}
```
vadmeste, line 821: this way work, ctx.IsSet("airgap") ?

cmd/subnet-utils.go
```diff
@@ -807,7 +812,15 @@ func initSubnetConnectivity(ctx *cli.Context, aliasedURL string, forUpload bool,
 }
 
 func validateSubnetFlags(ctx *cli.Context, forUpload bool) error {
-	if !globalAirgapped {
+	supportsAirgap := false
```
vadmeste, line 815: This may work, ctx.Bool("airgap") ?

cmd/support-upload.go
```diff
@@ -0,0 +1,146 @@
...
+	filePath := ctx.Args().Get(1)
+	issueNum := ctx.Int("issue")
+	msg := ctx.String("comment")
+
+	f, e := os.Open(filePath)
+	fatalIf(probe.NewError(e), "Unable to open file '%s'", filePath)
+	defer f.Close()
+	fi, e := f.Stat()
+	fatalIf(probe.NewError(e), "Unable to get file info for '%s'", filePath)
+	filename := fi.Name()
```
vadmeste, line 131: L126 to L131 is not needed in my opinion. You are not using f later,  filename is just the name of the file and not its path. I think this will not work if specify a full path of a file

vadmeste (changes requested):

cmd/support-upload.go
shtripat: `filename` is used for creation of upload URL only and while upload we are using `filePath` at L141, so works. tested with `./mc support upload --issue 62 --comment "file txt" test-shireesh /home/shubhendu/Downloads/after-stop.txt --dev` and works as expected.

vadmeste: okay, it just looks an overkill. We can use os.Stat() and filepath.Base()

cmd/subnet-utils.go
harshavardhana: ctx.IsSet("airgap") would also work

anjalshireesh: > This may work, ctx.Bool("airgap") ?

I think this will give the value of the `--airgap` flag

> ctx.IsSet("airgap") would also work

and this will tell if the `--airgap` flag was set

But what I'm trying to do is to check if this flag is even supported by the command being executed. The subsequent validation (that says `--json is applicable only when --airgap is also passed`) should be done only for commands that support the `--airgap` flag. 

The command I'm adding (`mc support upload`) doesn't support `--airgap` flag but supports `--json`

cmd/support-upload.go
anjalshireesh: thanks, fixed

cmd/support-upload.go
anjalshireesh: done

klauspost:
Should we apply compression if not already compressed?

Other than the reduced transfer size, it has the added benefit that it will wrap potentially harmful files and not be directly executable.

The server should not do anything do decompress, but we add a field just to be nice.

Adding it is pretty trivial:

```
	go func() {
		var (
			part io.Writer
			e    error
		)
		defer func() {
			mwriter.Close()
			w.CloseWithError(e)
		}()

		file, e := os.Open(filePath)
		if e != nil {
			return
		}
		defer file.Close()
		doComp := true
			if strings.HasSuffix(strings.ToLower(filePath), ext) {
				doComp = false
				break
			}
		}
		if doComp {
			mwriter.WriteField("auto-compression", "zstd")
			filePath += ".zst"
		}
		part, e = mwriter.CreateFormFile("file", filepath.Base(filePath))
		if e != nil {
			return
		}

		if doComp {
			z, _ := zstd.NewWriter(part, zstd.WithEncoderConcurrency(2))
			defer z.Close()
			_, e = z.ReadFrom(file)
		} else {
			_, e = io.Copy(part, file)
		}
	}()
```

cmd/subnet-utils.go
harshavardhana: --json must be present for even successful operations towards SUBNET - not just in --airgap mode.

anjalshireesh: @harshavardhana sure, I can change it that way. In that case, say for example in case of `mc support diag`, `--json` will print the diagnostics report to screen in case of `--airgap`, and something like `{"status": "success"}` in non-airgapped execution. Does that sound ok?

anjalshireesh: @harshavardhana I have sent https://github.com/minio/mc/pull/4859 for removing this validation (and other related changes)

anjalshireesh: this function has been removed now.

anjalshireesh:
> Should we apply compression if not already compressed?
> 
> Other than the reduced transfer size, it has the added benefit that it will wrap potentially harmful files and not be directly executable.
> 
> The server should not do anything do decompress, but we add a field just to be nice.
> 
> Adding it is pretty trivial:
> 
> ```
> 	go func() {
> 		var (
> 			part io.Writer
> 			e    error
> 		)
> 		defer func() {
> 			mwriter.Close()
> 			w.CloseWithError(e)
> 		}()
> 
> 		file, e := os.Open(filePath)
> 		if e != nil {
> 			return
> 		}
> 		defer file.Close()
> 		doComp := true
> 			if strings.HasSuffix(strings.ToLower(filePath), ext) {
> 				doComp = false
> 				break
> 			}
> 		}
> 		if doComp {
> 			mwriter.WriteField("auto-compression", "zstd")
> 			filePath += ".zst"
> 		}
> 		part, e = mwriter.CreateFormFile("file", filepath.Base(filePath))
> 		if e != nil {
> 			return
> 		}
> 
> 		if doComp {
> 			z, _ := zstd.NewWriter(part, zstd.WithEncoderConcurrency(2))
> 			defer z.Close()
> 			_, e = z.ReadFrom(file)
> 		} else {
> 			_, e = io.Copy(part, file)
> 		}
> 	}()
> ```

Have implemented this. PTAL.

shtripat (approved):
lgtm

harshavardhana (approved):

PR #4891, opened 2024-04-02
pipe: support append

dormanze:

## Description
As the title.

## How to test this PR?

1. creat file
```
echo "test1" | mc pipe --append myminio/bucket/test.file
```
2. upload new content with append
```
echo "test2" | mc pipe --append myminio/bucket/test.file
```
3. cat test.file
```
mc cat myminio/bucket/test.file
```

## Types of changes

## Checklist:


cmd/pipe-main.go
```diff
@@ -95,20 +100,36 @@ EXAMPLES:
   3. Copy an ISO image to an object on Amazon S3 cloud storage.
      {{.Prompt}} cat debian-8.2.iso | {{.HelpName}} s3/opensource-isos/gnuos.iso
 
-  4. Stream MySQL database dump to Amazon S3 directly.
+  4. Write contents of stdin append to an object on Amazon S3 cloud storage.
```
shtripat, line 103:
```suggestion
  4. Append contents of stdin to an existing object on Amazon S3 cloud storage.
```

klauspost:
I am not against the feature per se - but please make it *very* clear in the documentation that this works by reading back the existing object. 

Users will expect to be able to append to a 1GB file instantly - and it should be noted everywhere that this will not be the case. (and even so users will still complain).

harshavardhana:
I think this may not be what we want to encourage as a practice @dormanze it is better called as `mc compose` or something where you create a new object with existing 2 or more objects.

dormanze:
> I think this may not be what we want to encourage as a practice @dormanze it is better called as `mc compose` or something where you create a new object with existing 2 or more objects.

That's a good idea. I'll put it on my to-do list. However, I added the append functionality initially because Linux streaming supports appending. Do you think this is necessary?

dormanze:
> I am not against the feature per se - but please make it _very_ clear in the documentation that this works by reading back the existing object.
> 
> Users will expect to be able to append to a 1GB file instantly - and it should be noted everywhere that this will not be the case. (and even so users will still complain).

I'll add some notes to the documentation.Online documentation also needs attention.

PR #4909, opened 2024-04-22
enabling MC_DISABLE_PAGER and --dp

zveinn:

## Types of changes

### DIFF
--- a/cmd/globals.go
+++ b/cmd/globals.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"crypto/x509"
 	"net/url"
+	"os"
 	"time"
 
 	"github.com/charmbracelet/lipgloss"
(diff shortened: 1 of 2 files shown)


prakashsvmx (approved):

cmd/globals.go
```diff
@@ -89,17 +90,23 @@ var (
...
 
 	// CA root certificates, a nil value means system certs pool will be used
 	globalRootCAs *x509.CertPool
 )
 
 func parsePagerDisableFlag(args []string) {
+	disableEnv := os.Getenv(envPrefix + globalDisablePagerEnv)
+	if disableEnv == "true" {
+		globalPagerDisabled = true
+	}
```
klauspost, line 107: Simpler/Better?

```suggestion
	globalPagerDisabled, _ = strconv.ParseBool(os.Getenv(envPrefix + globalDisablePagerEnv))
```

I don't see it being written to elsewhere, so this seems good to me. 

ParseBool has a bit more options... and this matches other bool parsing AFAICT:

```
func ParseBool(str string) (bool, error) {
	switch str {
	case "1", "t", "T", "true", "TRUE", "True":
		return true, nil
	case "0", "f", "F", "false", "FALSE", "False":
		return false, nil
	}
	return false, syntaxError("ParseBool", str)
}
```

klauspost (changes requested):

shtripat (approved):

klauspost (approved):

jiuker (approved):

PR #4938, opened 2024-05-20
Fix regression that causes `mc idp ldap accesskey create` to error

taran-p:

## Description
Fix error that causes any `mc idp ldap accesskey create` without a specified expiration time to fail

## Types of changes

## Checklist:
- [x] Fixes a regression minio/madmin-go#277


cmd/idp-ldap-accesskey-create.go
```diff
@@ -143,6 +143,7 @@ func accessKeyCreateOpts(ctx *cli.Context, targetUser string) madmin.AddServiceA
 	}
 
 	var exp time.Time
+	var expPointer *time.Time
```
vadmeste, line 146: @taran-p can we avoid adding a new local variable here and change 'exp' type to `* time.Time` ? it will be easier that way.

taran-p: I'm not sure how to do this without adding a different way to avoid null reference issues

vadmeste: can you elaborate more about the null reference issue ?

taran-p: If exp is declared as a pointer, setting its value directly will be a null dereference and fault, so using an additional variable avoids this issue.

harshavardhana:
> Fix error that causes any `mc idp ldap accesskey create` without a specified expiration time to fail

what is the error?

taran-p:
Error was `mc: <ERROR> Unable to add service account. the expiration time should be in the future.` The command previously sent the unix epoch time if the expiration time was not defined, but now it just sends a nil pointer.

cmd/idp-ldap-accesskey-create.go
```diff
@@ -165,8 +167,7 @@ func accessKeyCreateOpts(ctx *cli.Context, targetUser string) madmin.AddServiceA
 		}
 	} else if expDurVal != 0 {
 		exp = time.Now().Add(expDurVal)
```
harshavardhana, line 169: how is this value passed? I see that this being ignored.

harshavardhana:
https://github.com/minio/mc/pull/4939 should address, the code itself was not readable using too many variables to do one thing. 

I have restructured things and made it more readable PTAL lets close this PR.

PR #4945, opened 2024-05-28
Wait for object to upload properly

shtripat:

## Description
Sometimes while mint tests, removal of object immediately after upload command fails with object not found. As the object may not be uploaded and reflected yet to the cluster, this can fail. Added a small sleep after upload to make sure object is available before any remove call.

## How to test this PR?
mint tests would take care

## Types of changes

## Checklist:

### DIFF
--- a/functional-tests.sh
+++ b/functional-tests.sh
@@ -1028,6 +1028,7 @@ function test_bucket_replication() {
 			break
 		fi
 		assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd cp "$FILE_0_B" "${SERVER_ALIAS}/${bkt1}/${object_name}"
+		sleep 5s # to make sure objects reflects
 		assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd rm "${SERVER_ALIAS}/${bkt1}/${object_name}"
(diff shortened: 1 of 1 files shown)


functional-tests.sh
```diff
@@ -1028,6 +1028,7 @@ function test_bucket_replication() {
 			break
 		fi
 		assert_success "$start_time" "${FUNCNAME[0]}" mc_cmd cp "$FILE_0_B" "${SERVER_ALIAS}/${bkt1}/${object_name}"
+		sleep 5s # to make sure objects reflects
```
klauspost, line 1031: Looks more like you are trying to hide an underlying issue.

shtripat: in local testing and even with play it always worked. only during CI while mint tests sometimes it fails.

klauspost: @shtripat A wait is never the solution. It may make it less likely - but AFAICT this shouldn't be needed in the first place, unless we are not consistent.

shtripat: I agree, but not sure as some cases it just fails immediate rm for the object which was uploaded using `mc cp`.
This test anyway was added for a scenario which in itself is very rare. May be we can just disable this test for the time being.

klauspost: We should figure out why it is failing in the first place.

harshavardhana:
As discussed removing this from `mc` instead write these tests directly on the server.

PR #4947, opened 2024-05-30
fix: avoid stat call for mc get

jiuker:
fix: avoid stat call for mc get

## Description

fix: avoid stat call for mc get
revert https://github.com/minio/mc/pull/4866 for `Stat` call

## Types of changes

## Checklist:


cmd/common-methods.go
```diff
@@ -576,4 +577,5 @@ type uploadSourceToTargetURLOpts struct {
 	multipartSize       string
 	multipartThreads    string
 	updateProgressTotal bool
+	preventStat         bool
```
shtripat, line 580:
```suggestion
	skipStat         bool
```
I think `skipStat` is better name

ramondeklein (changes requested):
Why is it necessary to remove the `Stat()` call? Also it now passes an empty object to the `cmd.objectInfo2ClientContent()` and all parent methods receive uninitialized content in the `cmd.ClientContent` structure. The code that follows the `getSourceStream()` method also uses the content to access metadata:

https://github.com/minio/mc/blob/a8fdcbe7cb2f85ce98d60e904717aa00016a7d37/cmd/common-methods.go#L428-L451

So if you specify `preventStat`, then you'll lose metadata during the copy operation, so I guess we need the `Stat()` call.

ramondeklein:
I checked the code, because the information that `Stat` returns should also be available during a normal `GET` request. The reason that is fires a separate `HEAD` request is, because nobody started reading yet. If we can "start" the read operation, then the `Stat()` function will be at zero-cost. I think we should persue that option instead.

ramondeklein:
> Stat will call HeadObject first time. And `mc get` should call GetObject only @ramondeklein

`HeadObject` should be identical to `GetObject` with the exception that `HeadObject` won't return the actual data. So, if we send `GetObject` first, then you also get all the metadata. If you look at the `Stat()` code, you can see this. If the read was already started, then it doesn't invoke the `HeadObject`. We don't initiate a read operation before the `Stat()`, so this results in invoking the `HeadObject`. I think if you try my "hack" as shown https://github.com/minio/mc/pull/4947#issuecomment-2139555549, then it won't send multiple requests.

ramondeklein:
> `mc get remoteObject localObject` that means s3 to local file. That's indeed @ramondeklein

@justinclift When copying data to the local filesystem the metadata isn't important (it's lost anyway). But when looking at the code, it also seems possible to copy data from one S3 source to another and you don't want to lose the metadata in that case.

I guess you also lose the progress bar when setting this option and that doesn't seem to be necessary.

jiuker:
> > Stat will call HeadObject first time. And `mc get` should call GetObject only @ramondeklein
> 
> @jiuker `HeadObject` should be identical to `GetObject` with the exception that `HeadObject` won't return the actual data. So, if we send `GetObject` first, then you also get all the metadata. If you look at the `Stat()` code, you can see this. If the read was already started, then it doesn't invoke the `HeadObject`. We don't initiate a read operation before the `Stat()`, so this results in invoking the `HeadObject`. I think if you try my "hack" as shown [#4947 (comment)](https://github.com/minio/mc/pull/4947#issuecomment-2139555549), then it won't send multiple requests.

Hmm. I

> > `mc get remoteObject localObject` that means s3 to local file. That's indeed @ramondeklein
> 
> @justinclift When copying data to the local filesystem the metadata isn't important (it's lost anyway). But when looking at the code, it also seems possible to copy data from one S3 source to another and you don't want to lose the metadata in that case.
> 
> I guess you also lose the progress bar when setting this option and that doesn't seem to be necessary.

Yes. So this only true for `mc get`

ramondeklein:
> Hmm. The code is unreadable maybe. But it work. You can debug it. @ramondeklein And `headObject` return metadata, getObject return the file body.

As you can see, the underlying [GetObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html) API also returns all the metadata. You only call HeadObject if you want metadata only. If you want both, then you should use GetObject instead. This is also in our Go SDK:

https://github.com/minio/minio-go/blob/14b3aa6d56708fdf2071346aadcdb085069d8937/api-get-object.go#L407-L433

The problem that we have is that if you call `Stat()` before doing an actual read operation, then it needs to do either a `HeadObject` or a `GetObject` and because it's not trying to read data, the logical choice is `HeadObject`. But we start reading directly after the `Stat()`, so that triggers another `GetObject`.

>> I guess you also lose the progress bar when setting this option and that doesn't seem to be necessary.
> 
> Yes. So this only true for `mc get`

So you need to specify an additional option with `mc` and you get less. I don't see the added value of that.

klauspost (approved):

harshavardhana (approved):

PR #4962, opened 2024-06-12
Adding support for Hex keys

zveinn:

## Description
Due to the nature of S3 APIs and the way we declare encryption keys, we can not support encryption keys with the symbol = in them. This makes support for base64 encoding, which includes padding, impossible. 

To provide an additional option to the RawBase64Encoding option, this PR adds support for Hex Encoding.

## Motivation and Context
Most tools (openssl for example) generate base64 encoded keys with padding, so when they are used in scripts along with mc, they tend to fail. An alternative is to offer support for Hex encoding. 

## How to test this PR?
The functional test suite includes two tests for hex support.

## Types of changes

## Checklist:
- [x] Unit tests added/updated
- [x] Internal documentation updated
- [x] Create a documentation update request [here](https://github.com/minio/docs/issues/new?label=doc-change,title=Doc+Updated+Needed+For+PR+github.com%2fminio%2fmc%2fpull%2fNNNNN)

### DIFF
--- a/cmd/encryption-methods.go
+++ b/cmd/encryption-methods.go
@@ -198,19 +198,19 @@ func parseSSEKey(sseKey string, keyType sseKeyType) (
 ) {
 	sseKeyBytes := []byte(sseKey)
 
-	seperatorIndex := bytes.LastIndex(sseKeyBytes, []byte("="))
-	if seperatorIndex < 0 {
+	separatorIndex := bytes.LastIndex(sseKeyBytes, []byte("="))
+	if separatorIndex < 0 {
 		err = errSSEKeyMissing().Trace(sseKey)
 		return
 	}
 
-	encodedKey := string(sseKeyBytes[seperatorIndex+1:])
-	if seperatorIndex == len(sseKeyBytes)-1 {
+	encodedKey := string(sseKeyBytes[separatorIndex+1:])
+	if separatorIndex == len(sseKeyBytes)-1 {
 		err = errSSEKeyMissing().Trace(sseKey)
 		return
 	}
 
-	alias, prefix = splitKey(string(sseKeyBytes[:seperatorIndex]))
+	alias, prefix = splitKey(string(sseKeyBytes[:separatorIndex]))
 
 	if keyType == sseS3 {
 		return
--- a/cmd/encryption-methods.go
+++ b/cmd/encryption-methods.go
@@ -18,9 +18,11 @@
 package cmd
 
 import (
+	"bytes"
 	"encoding/base64"
+	"encoding/hex"
+	"fmt"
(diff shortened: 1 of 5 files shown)


jiuker (approved):

shtripat (approved):
lgtm

cmd/encryption-methods.go
```diff
@@ -194,43 +196,48 @@ func parseSSEKey(sseKey string, keyType sseKeyType) (
 	key string,
 	err *probe.Error,
 ) {
-	if keyType == sseS3 {
-		alias, prefix = splitKey(sseKey)
+	sseKeyBytes := []byte(sseKey)
+
+	seperatorIndex := bytes.LastIndex(sseKeyBytes, []byte("="))
```
vadmeste, line 201: s/seperatorIndex/separatorIndex

cmd/encryption-methods.go
```diff
@@ -194,43 +196,48 @@ func parseSSEKey(sseKey string, keyType sseKeyType) (
...
-			err = errSSEKMSKeyFormat("One of the inserted keys was " + strconv.Itoa(len(key)) + " bytes and did not have a valid KMS key name.").Trace(sseKey)
+		if !validKMSKeyName(encodedKey) {
+			err = errSSEKMSKeyFormat(fmt.Sprintf("Key (%s) is badly formatted.", encodedKey)).Trace(sseKey)
+		}
+		return
+	}
+
+	keyB, de := hex.DecodeString(encodedKey)
+	if de != nil {
+		keyB, de = base64.RawStdEncoding.DecodeString(encodedKey)
```
vadmeste, line 228: I remember this one, 

base64.RawStdEncoding.DecodeString() will reject the = padding, but base64.StdEncoding.DecodeString() will not reject it.

Also I think there is a possibility that when a user intents to pass base64, it will still be interpreted as hex, for example, both:

	base64.RawStdEncoding.DecodeString("abcd")
	hex.DecodeString("abcd")

will not return an error

I think it is okay to trim all the trailing = in the passed argument rather than introducing a new format.

vadmeste (changes requested):

cmd/encryption-methods.go
zveinn: You are right about some base64 encodings being valid hex, I'll fix that. 

But trimming the padding presents the same problem as supporting the padding to begin with. In certain circumstances there would be a possibility that we would use the wrong key. There is an old discussion on this on slack, if you want we can also discuss it there since it tends to be a long discussion.

klauspost (approved):
LGTM

vadmeste (approved):

PR #4983, opened 2024-07-10
Fix non-recursive compares against alias+key on Windows

klauspost:
## Description

Some listings would stop after first version had been processed on windows, since for example `mytls/testbucket/trace.json` != `mytls\testbucket\trace.json` would cause `mc undo` to only see one version, since this would fail.

Use `getStandardizedURL` for both sides of these compares.

## How to test this PR?

Used `mc undo --last=1 myminio/testbucket/trace.json`.

## Types of changes


ebozduman (approved):
LGTM

ebozduman:
@klauspost 
I guess you just need to resync your branch...

shtripat (approved):

PR #4989, opened 2024-07-17
Show a warning when the server certificate duration is expiring soon

vadmeste:
When 90% of the validity time of a server is elapsed, print a warning remainding the user to change the certificate.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -512,7 +512,7 @@ func registerApp(name string) *cli.App {
 			host := k.(string)
 			expires := v.(time.Time)
 			fmt.Fprintf(os.Stderr, "\n")
-			fmt.Fprintf(os.Stderr, "== WARN: `%s` certificate will expire in %s. Consider renewing it soon.\n", host, expires)
+			fmt.Fprintf(os.Stderr, "== WARN: `%s` certificate will expire in %s. Renew soon to avoid outage.\n", host, expires)
 			fmt.Fprintf(os.Stderr, "\n")
 			return true
 		})
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -507,6 +507,17 @@ func registerApp(name string) *cli.App {
 	app.CustomAppHelpTemplate = mcHelpTemplate
 	app.EnableBashCompletion = true
 	app.OnUsageError = onUsageError
+	app.After = func(*cli.Context) error {
+		globalExpiringCerts.Range(func(k, v interface{}) bool {
+			host := k.(string)
+			expires := v.(time.Time)
+			fmt.Fprintf(os.Stderr, "\n")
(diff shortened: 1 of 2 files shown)


cmd/main.go
```diff
@@ -507,6 +507,17 @@ func registerApp(name string) *cli.App {
 	app.CustomAppHelpTemplate = mcHelpTemplate
 	app.EnableBashCompletion = true
 	app.OnUsageError = onUsageError
+	app.After = func(*cli.Context) error {
+		globalExpiringCerts.Range(func(k, v interface{}) bool {
+			host := k.(string)
+			expires := v.(time.Time)
+			fmt.Fprintf(os.Stderr, "\n")
+			fmt.Fprintf(os.Stderr, "== WARN: `%s` certificate will expire in %s. Consider renewing it soon.\n", host, expires)
```
harshavardhana, line 515:
```suggestion
			fmt.Fprintf(os.Stderr, "== WARN: `%s` certificate will expire in %s. Renew soon to avoid outage.\n", host, expires)
```

harshavardhana (changes requested):

cmd/main.go
```diff
@@ -507,6 +507,17 @@ func registerApp(name string) *cli.App {
 	app.CustomAppHelpTemplate = mcHelpTemplate
 	app.EnableBashCompletion = true
 	app.OnUsageError = onUsageError
+	app.After = func(*cli.Context) error {
+		globalExpiringCerts.Range(func(k, v interface{}) bool {
+			host := k.(string)
+			expires := v.(time.Time)
+			fmt.Fprintf(os.Stderr, "\n")
+			fmt.Fprintf(os.Stderr, "== WARN: `%s` certificate will expire in %s. Renew soon to avoid outage.\n", host, expires)
+			fmt.Fprintf(os.Stderr, "\n")
```
harshavardhana, line 516: should we using printMsg() like loggers? so that we can have a --json version of this info as well?

harshavardhana (changes requested):

cmd/main.go
vadmeste: I do not see high value doing that, first this warning is printed at the end of all commands and it should be redirected to os.Stderr so it will not confuse other tools parsing mc output

harshavardhana: I think it will, if I pass --json flag and I am redirecting stderr to stdout this is a problem. It will break scripts.

vadmeste: so currently, --debug and --json will disable this check. Currently the warning is meant for users not for users. But I can do printMsg() for stderr as well. The most important thing for me is that the warning needs to be sent to os.Stderr. Let me know what do you think

harshavardhana: Okay so if --json and --debug disable this then we are good.

harshavardhana (approved):

PR #5004, opened 2024-07-31
combine STS, SVC accounts into access keys with or without expiration

harshavardhana:

## Description
combine STS, SVC accounts into access keys with or without expiration

## Motivation and Context
from an end-user point of view, they are no different

## How to test this PR?
Configure LDAP and perform all the necessary activities to create accounts
with or without expiration. Then, observe the pretty printed output.

## Types of changes

## Checklist:


taran-p:
Would the inability to delete STS access keys be an issue with listing them together?

donatello:
Also sts keys need tokens so I think there should be an indication like a type so that users don't go looking for them in apps or scripts (for an audit for example).

harshavardhana:
Fundamentally they are all the same from an API point of view, the idea here is just to show them a hint. Adding an sts tag might be okay let me see if I can add it cleanly. 

The current UI is quite wrong we need to address it.

harshavardhana:
> Would the inability to delete STS access keys be an issue with listing them together?

There is no API contract requiring that all access keys be allowed to be deleted. Service Accounts is our extended implementation in MinIO. We just chose to allow it, in AWS STS credentials do not have that feature so in MinIO we didn't implement a DELETE behavior 
for them either.

harshavardhana:
Added STS differentiation

taran-p (approved):

donatello (approved):

PR #5013, opened 2024-08-08
add global `--resolve HOST[:PORT]=IP` flag

aead:
## Description
This commit adds support for custom DNS overwrites via one or multiple `--resolve` flags. It allows `mc` users to define custom DNS mappings from a HOST (and optional port) to an IP address.

For example:
```
mc --resolve foo.com:9000:10.1.2.3 ls myminio/mybucket
```
The `--resolve` syntax is taken from cURL. Ref. `curl --help dns`

Use cases:

  1. DNS overwrites. If there are no DNS resolvers available, e.g. due to a temp. outage, this allows `mc` to connect to a defined alias as long as the IP of the one or multiple cluster nodes are known. It also allows `mc` to overwrite the current DNS responses. This is useful for various tests in dev and staging environments.
  2. TLS certificate verification. Often TLS certificates are issued for some DNS names (via SAN) but not for IP addresses. IP addresses are usually not static and might change anytime. Accessing a MinIO cluster that serves a certificate containing only DNS SANs via an IP address results in a TLS certificate verification error - even if the certificate is issued by a trusted CA. The current workaround is the `--insecure` flag which disables TLS certification completely.

Example for TLS verification:

```
$ mc alias set 'myminio' 'https://192.168.188.118:9000' 'minioadmin' 'minioadmin'
mc: <ERROR> Unable to initialize new alias from the provided credentials.
Get "https://192.168.188.118:9000": tls: failed to verify certificate:
x509: cannot validate certificate for 192.168.188.118 because it doesn't contain any IP SANs.
```

```
mc alias set --resolve foo.com:9000:192.168.188.118 'myminio' 'https://foo.com:9000' 'minioadmin' 'minioadmin'
Added `myminio` successfully.
```

## Motivation and Context
TLS

## Types of changes

## Checklist:

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -59,6 +59,11 @@ var globalFlags = []cli.Flag{
 		Usage:  "enable debug output",
 		EnvVar: envPrefix + "DEBUG",
 	},
+	cli.StringSliceFlag{
+		Name:   "resolve",
+		Usage:  "resolves HOST[:PORT] to an IP address. Example: minio.local:9000=10.10.75.1",
+		EnvVar: envPrefix + "RESOLVE",
+	},
 	cli.BoolFlag{
 		Name:   "insecure",
 		Usage:  "disable SSL certificate verification",
--- a/cmd/globals.go
+++ b/cmd/globals.go
@@ -21,9 +21,13 @@ package cmd
 import (
 	"context"
 	"crypto/x509"
+	"fmt"
+	"net"
+	"net/netip"
 	"net/url"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/charmbracelet/lipgloss"
@@ -69,6 +73,7 @@ var (
 	globalDebug        = false               // Debug flag set via command line
 	globalNoColor      = false               // No Color flag set via command line
 	globalInsecure     = false               // Insecure flag set via command line
+	globalResolvers    map[string]netip.Addr // Custom mappings from HOST[:PORT] to IP
 	globalAirgapped    = false               // Airgapped flag set via command line
 	globalSubnetConfig []madmin.SubsysConfig // Subnet config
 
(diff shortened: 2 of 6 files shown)


cmd/globals.go
```diff
@@ -170,5 +175,30 @@ func setGlobalsFromContext(ctx *cli.Context) error {
 		}
 	}
 
+	dnsEntries := ctx.StringSlice("resolve")
+	if len(dnsEntries) > 0 {
+		globalResolvers = make(map[string]netip.Addr, len(dnsEntries))
+
+		// Each entry is a HOST[:PORT]:IP pair. This matches cURL's syntax.
+		for _, e := range dnsEntries {
+			i := strings.LastIndexByte(e, ':')
```
klauspost, line 184: Why the last?  Wouldn't that break on an IPv6? Wouldn't you enter that like this: `minio.local:2001:0db8:85a3:0000:0000:8a2e:0370:7334`.  Hmmm. That makes the optional port clunky.

Maybe splitting on `=` would make it clearer. `minio.local:9000=2001:0db8:85a3:0000:0000:8a2e:0370:7334`.

With `=` it would be a lot easier easier to parse - and I don't see either side having an `=` char - unlike other places ;)

cmd/flags.go
```diff
@@ -59,6 +59,11 @@ var globalFlags = []cli.Flag{
 		Usage:  "enable debug output",
 		EnvVar: envPrefix + "DEBUG",
 	},
+	cli.StringSliceFlag{
+		Name:   "resolve",
+		Usage:  "resolves HOST[:PORT] to an IP address",
```
klauspost, line 64:
```suggestion
		Usage:  "resolves HOST[:PORT] to an IP address. Example: minio.local:9000=10.10.75.1",
```

(if you do go for `=`)

cmd/globals.go
aead: That is true. 👍 I wanted to stick to `curl` for being able to copy paste things. (useful for debugging instead of having to adjust things)
However, no strong opinion. We can also use `=`.

harshavardhana (approved):

allanrogerr (approved):

allanrogerr:
Tested
```
minio server /tmp/data &
mc alias set --resolve minio.train:9000=127.0.0.1 resolved http://minio.train:9000 minioadmin minioadmin
mc admin info resolved --resolve minio.train:9000=127.0.0.1
```

PR #5022, opened 2024-08-15
fix:  Use TLS for STS endpoint when STS endpoint uses https scheme.

jiuker:
fix: use https's transport to call sts endpoint when alias is http

## Description

export MC_HOST_myminio=http://$(ACCESS_KEY):$(SECRET_KEY)@minio.minio-operator.svc.cluster.local
export MC_STS_ENDPOINT_myminio=https://sts.minio-operator.svc.cluster.local:4223/sts/minio-operator
export MC_WEB_IDENTITY_TOKEN_FILE_myminio=....

operator show:
```
2024/08/13 13:48:54 http: TLS handshake error from 10.42.0.170:35032: remote error: tls: bad certificate
2024/08/13 13:48:55 http: TLS handshake error from 10.42.0.171:46770: remote error: tls: bad certificate
2024/08/13 13:49:35 http: TLS handshake error from 10.42.0.170:39536: remote error: tls: bad certificate
2024/08/13 13:49:46 http: TLS handshake error from 10.42.0.171:34764: remote error: tls: bad certificate
2024/08/13 13:51:06 http: TLS handshake error from 10.42.0.170:54302: remote error: tls: bad certificate
2024/08/13 13:51:17 http: TLS handshake error from 10.42.0.171:58744: remote error: tls: bad certificate
2024/08/13 13:53:56 http: TLS handshake error from 10.42.0.170:41560: remote error: tls: bad certificate
2024/08/13 13:54:10 http: TLS handshake error from 10.42.0.171:37956: remote error: tls: bad certificate
```

## How to test this PR?

1. Deploy a http tenant.
2. Deploy a minioJob
3. minioJob will failed. (But success with a https tenant)

## Types of changes

## Checklist:


cmd/client-s3.go
```diff
@@ -285,10 +286,12 @@ func getCredentialsChainForConfig(config *Config, transport http.RoundTripper) (
 				os.Setenv("AWS_ROLE_SESSION_NAME", val)
 			}
 		}
-
 		stsEndpointURL, err := url.Parse(stsEndpoint)
 		if err != nil {
-			return nil, probe.NewError(fmt.Errorf("Error parsing sts endpoint: %v", err))
+			return nil, nil, probe.NewError(fmt.Errorf("Error parsing sts endpoint: %v", err))
```
ramondeklein, line 291: Nested errors are typically wrapped using `%w`.

cmd/client-s3.go
```diff
@@ -207,10 +207,10 @@ func (n notifyExpiringTLS) RoundTrip(req *http.Request) (res *http.Response, err
 
 // getTransportForConfig returns a corresponding *http.Transport for the *Config
 // set withS3v2 bool to true to add traceV2 tracer.
-func getTransportForConfig(config *Config, withS3v2 bool) http.RoundTripper {
+func getTransportForConfig(config *Config, withS3v2 bool, chainsTLS bool) http.RoundTripper {
```
ramondeklein, line 210: Why is this variable named `chainsTLS`? I guess `forceTLS` would be a better name, because it forces a TLS transport.

ramondeklein (changes requested):
I think this PR description should be changed into: **Use TLS for STS endpoint when STS endpoint uses `https` scheme**.

cmd/client-s3.go
ramondeklein: No, if you set `forceTLS` to `false`, then it may/may not be TLS. If it's set to `true` then you're actually forcing it to `true`. I'm not sure what `chainsTLS` means, but I don't think it's suitable in this case. You typically use _chains_ with a list of TLS certificates to the CA root. That's called a _chain_, but that's not what is being used here.

shtripat (commented):
lgtm, no specific comments

harshavardhana (approved):

cmd/client-admin.go
```diff
@@ -57,9 +57,7 @@ func NewAdminFactory() func(config *Config) (*madmin.AdminClient, *probe.Error)
 		var found bool
 		if api, found = clientCache[confSum]; !found {
 
-			transport := getTransportForConfig(config, true)
-
-			credsChain, err := getCredentialsChainForConfig(config, transport)
+			transport, credsChain, err := getTransportAndCredentialsChainForConfig(config)
```
harshavardhana, line 60: For a cleaner implementation move this as a method for the `config` such as 

```
config.getTransport()
```

```
config.getCredsChain()
```

Just more cleaner separation of functions.

cmd/client-s3.go
```diff
@@ -285,10 +286,12 @@ func getCredentialsChainForConfig(config *Config, transport http.RoundTripper) (
...
 		}
-
 		stsEndpointURL, err := url.Parse(stsEndpoint)
 		if err != nil {
-			return nil, probe.NewError(fmt.Errorf("Error parsing sts endpoint: %v", err))
+			return nil, nil, probe.NewError(fmt.Errorf("Error parsing sts endpoint: %w", err))
+		}
+		if stsEndpointURL.Scheme == "https" {
+			transport = getTransportForConfig(config, false, true)
 		}
```
harshavardhana, line 295:
```suggestion
		transport = getTransportForConfig(config, false, stsEndpointURL.Scheme == "https")
```

harshavardhana (changes requested):

cmd/client.go
```diff
@@ -242,6 +252,129 @@ type Config struct {
...
+			}
+			if val := env.Get("MC_ROLE_SESSION_NAME_"+config.Alias, randString(32, rand.NewSource(time.Now().UnixNano()), "mc-session-name-")); val != "" {
+				os.Setenv("AWS_ROLE_SESSION_NAME", val)
+			}
+		}
+		stsEndpointURL, err := url.Parse(stsEndpoint)
+		if err != nil {
+			return nil, probe.NewError(fmt.Errorf("Error parsing sts endpoint: %w", err))
+		}
+		config.forceTLS = true
```
ramondeklein, line 279: I don't think it's a good idea to use a struct field to hide parameter transfer. Just pass it to `config.getTransport(forceTLS bool)` would be much cleaner. This may also introduce a bug, because you're also calling `config.getTransport()` in `cmd/client-admin.go`. Because you set `config.forceTLS` to `true` here, it will store it in the `config` and subsequent calls to `config.getTransport()` may always return TLS enabled transports.

cmd/client.go
```diff
@@ -242,6 +252,129 @@ type Config struct {
 	UploadLimit       int64
 	DownloadLimit     int64
 	Transport         http.RoundTripper
+	// Force TLS
+	forceTLS bool
+}
+
+// getCredsChain returns an []credentials.Provider array for the config
+// and the STS configuration (if present)
+func (config *Config) getCredsChain() ([]credentials.Provider, *probe.Error) {
```
ramondeklein, line 261: Not sure if it really adds anything to move the existing function inside the `struct`. It's pretty clean to keep `Config` just a struct that holds data and doesn't provide functionality. It also makes the PR more complex to review (more changes).

It's not a deal-breaker, but I think it was fine the way it was.

ramondeklein (changes requested):

cmd/client.go
ramondeklein: It wasn't in your previous code of the PR, so functionality has changed during this refactor. Suppose you have the following code (pseudo code):
```
cfg := Config{}
factory := NewAdminFactory(cfg)
transport1 := factory(cfg)     // may return a transport without TLS
_ = cfg.getCredsChain()
transport2 := factory(cfg)     // may return a transport with TLS
```
Because `Config.getCredsCain()` changes the internals of the `Config` structure, it will have side-effects that are unexpected. If you would pass `forceTLS` via parameters, then the `Config` structure will be untouched and `transport2` would have had the same result as the first call.

Even if this is the desired behaviour, I still consider it a bug because it's very unexpected that the order of these calls would be relevant.

ramondeklein: I think this code is hard to read and it doesn't need to be that way...

cmd/client.go
ramondeklein: I agreed that it's better to split the functions. Moving them into `Config` is probably personal preference.

cmd/client.go
ramondeklein: PS: This will now force TLS always. I thought it should only enable TLS when `stsEndpointURL.Scheme == "https"`.

cmd/client.go
```diff
@@ -242,6 +252,127 @@ type Config struct {
...
+			}
+			if val := env.Get("MC_ROLE_SESSION_NAME_"+config.Alias, randString(32, rand.NewSource(time.Now().UnixNano()), "mc-session-name-")); val != "" {
+				os.Setenv("AWS_ROLE_SESSION_NAME", val)
+			}
+		}
+		stsEndpointURL, err := url.Parse(stsEndpoint)
+		if err != nil {
+			return nil, probe.NewError(fmt.Errorf("Error parsing sts endpoint: %w", err))
+		}
+		config.forceTLS = stsEndpointURL.Scheme == "https"
```
ramondeklein, line 279: Sorry to bring this up again, but `config.forceTLS` shouldn't be in here. It should be parameter that is fed to `config.getTransport()`. You shouldn't store parameters as persistent state.

ramondeklein: Setting this makes the following code result in different transports:
```
credsChain, _ := cfg.getCredsChain()           // set config.forceTLS to true
transport, _ := cfg.getTransport()               // TLS based transport
```
opposed to:
```
transport, _ := cfg.getTransport()               // non-TLS based transport
credsChain, _ := cfg.getCredsChain()           // set config.forceTLS to true, but too late...
```
It's good to have separation of concerns, but in this case the order is important and that's non-trivial when looking at the code. The proper way to fix this is to let `getTransport()` check whether or not TLS is required.

cmd/client.go
```diff
@@ -242,6 +252,127 @@ type Config struct {
...
+				os.Setenv("AWS_ROLE_ARN", val)
+			}
+			if val := env.Get("MC_ROLE_SESSION_NAME_"+config.Alias, randString(32, rand.NewSource(time.Now().UnixNano()), "mc-session-name-")); val != "" {
+				os.Setenv("AWS_ROLE_SESSION_NAME", val)
+			}
+		}
+		stsEndpointURL, err := url.Parse(stsEndpoint)
+		if err != nil {
+			return nil, probe.NewError(fmt.Errorf("Error parsing sts endpoint: %w", err))
+		}
```
ramondeklein, line 278: This code could be refactored into a private function (i.e. `func (config *Config) stsEndPoint() (*url.URL, err)` that can also be called from `getTransport()` to detect whether or not to use TLS.

harshavardhana (approved):

cmd/client.go
```diff
@@ -244,6 +254,135 @@ type Config struct {
...
+		return isHostTLS(config) || stsEndpointURL.Scheme == "https"
+	}
+	return isHostTLS(config)
+}
+
+func (config *Config) initTransport(withS3v2 bool) {
+	var transport http.RoundTripper
+
+	useTLS := config.isTLS()
```
ramondeklein, line 327: This could be moved to the inner scope (line 349), but it's not an expensive operation, so fine with me to leave it here. As it is used only once, you could also replace the check in line 349 with `config.isTLS()`.

ramondeklein (approved):

PR #5036, opened 2024-09-13
idp-ldap: Fix create-with-login command

vadmeste:

## Description
Currently, LDAP create-with-login command is not working properly when the LDAP user does not have admin:CreateServiceAccount permission.

The permission is normally not needed since a user is allowed to create a service account for itself. In that case, a temporary account should be created and a new service account should be issued to the temporary account access key, and not to the LDAP username as what the code currently does.

This commit will send the correct target user in that case to make create-with-login always successful.

## Motivation and Context
Fix creating a new service account of an LDAP user

## How to test this PR?
Run a MinIO cluster with LDAP enabled and run this: `mc idp ldap accesskey create-with-login http://localhost:9000`

## Types of changes

## Checklist:

### DIFF
--- a/cmd/idp-ldap-accesskey-create-with-login.go
+++ b/cmd/idp-ldap-accesskey-create-with-login.go
@@ -90,6 +90,9 @@ func mainIDPLdapAccesskeyCreateWithLogin(ctx *cli.Context) error {
 func loginLDAPAccesskey(ctx *cli.Context) (*madmin.AdminClient, madmin.AddServiceAccountReq) {
 	urlStr := ctx.Args().First()
 
+	u, e := url.Parse(urlStr)
+	fatalIf(probe.NewError(e), "unable to parse server URL")
+
 	console.SetColor(cred, color.New(color.FgYellow, color.Italic))
 	reader := bufio.NewReader(os.Stdin)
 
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

donatello:
Does this seem ok, the user is not permitted create-service-account, but we are creating one anyway (via an sts cred)?

donatello (changes requested):
> The permission is normally not needed since a user is allowed to create a service account for itself. In that case, a temporary account should be created and a new service account should be issued to the temporary account access key, and not to the LDAP username as what the code currently does.

Generally, create-service-account permission is implicit, but when it's explicitly denied is it right to create on here? I think i may be missing something.

harshavardhana:
> Generally, create-service-account permission is implicit, but when it's explicitly denied is it right to create on here? I think i may be missing something.

you will be rejected by the server @donatello what @vadmeste is doing is how Console UI does `'create-with-login` instead of passing the LDAP username directly we must pass the STS creds and let the server tell us yay or nay.

donatello (approved):
looks good. CC @taran-p

taran-p (approved):
LGTM

PR #5077, opened 2024-11-05
Only warn if less than 28 days to cert expires

### LINKED ISSUE #5076: WARN message is suddenly printed on each mc calls
This is related to the PR #4989 which was merged 4 months ago.
Related code is: [cmd/main.go#L511-L521](https://github.com/minio/mc/blame/master/cmd/main.go#L511-L521).

IMHO, this is very bad for the user experience.
At least, please make this option an opt-in, even though I still don't see any reason to keep it.

## Expected behavior

Do not print WARN messages not related to mc command.
There is no reason to print a WARN when I `mc ls` if not related to the ls itself.

## Actual behavior

When I use `mc ls`, I now have a WARN message reporting the certificate will expire soon (in 2 months).
This is not relevant, as there is nothing I can do here.

```
[0] > mc ls www/mybucket/myfile
[2024-10-16 12:19:45 CEST] ...

== WARN: `xxx.xxx.xxx` certificate will expire in 2025-01-25 10:16:45 +0000 UTC. Renew soon to avoid outage.
```

Hopefully, this is not yet happening for Amazon S3 or with the `-json` flag.

## mc --version

```
[0] > mc --version
mc version RELEASE.2024-10-08T09-37-26Z (commit-id=cf128de2cf42e763e7bd30c6df8b749fa94e0c10)
Runtime: go1.22.8 linux/amd64
Copyright (c) 2015-2024 MinIO, Inc.
License GNU AGPLv3 <https://www.gnu.org/licenses/agpl-3.0.html>
```

klauspost:
## Description

Fixes #5076 as much as reasonable.

## How to test this PR?

Long-lived certs..

## Types of changes

### DIFF
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -196,7 +196,8 @@ func (n notifyExpiringTLS) RoundTrip(req *http.Request) (res *http.Response, err
 
 	cert := res.TLS.PeerCertificates[0] // leaf certificate
 	validityDur := cert.NotAfter.Sub(cert.NotBefore)
-	if time.Since(cert.NotBefore) > time.Duration(0.9*float64(validityDur)) {
+	// Warn if less than 10% of time left and it is less than 28 days.
+	if time.Since(cert.NotBefore) > time.Duration(min(0.9*float64(validityDur), 28*24*float64(time.Hour))) {
 		globalExpiringCerts.Store(req.Host, cert.NotAfter)
 	}
 
--- a/cmd/client-s3.go
+++ b/cmd/client-s3.go
@@ -197,7 +197,7 @@ func (n notifyExpiringTLS) RoundTrip(req *http.Request) (res *http.Response, err
 	cert := res.TLS.PeerCertificates[0] // leaf certificate
 	validityDur := cert.NotAfter.Sub(cert.NotBefore)
 	// Warn if less than 10% of time left and it is less than 28 days.
-	if time.Since(cert.NotBefore) > time.Duration(min(0.9*float64(validityDur), 28*24*float64(time.Hour))) {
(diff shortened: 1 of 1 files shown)


cmd/client-s3.go
```diff
@@ -196,7 +196,8 @@ func (n notifyExpiringTLS) RoundTrip(req *http.Request) (res *http.Response, err
 
 	cert := res.TLS.PeerCertificates[0] // leaf certificate
 	validityDur := cert.NotAfter.Sub(cert.NotBefore)
-	if time.Since(cert.NotBefore) > time.Duration(0.9*float64(validityDur)) {
+	// Warn if less than 10% of time left and it is less than 28 days.
+	if time.Since(cert.NotBefore) > time.Duration(min(0.9*float64(validityDur), 28*24*float64(time.Hour))) {
```
vadmeste, line 200: I think you mean (validityDur-28*24*float64(time.Hour))

klauspost: Yeah, though changing it for clarity.

harshavardhana (approved):

xtian-socgen:
I'm adding a comment here as my original issue was closed too fast...

Although I understand the point behind "mc is just an admin tool that is build only to be used with minio and not by end users", I wonder if this is just reality. I know plenty of people using it instead of the aws cli, because it's just 1M times easier to use.
This is why, I created the issue at first, as it looked like to me that this was a real problem for non-admin usage.

Now that its usage is clarified, I understand this was not the good way to use it, which is sad, but its like this.
As this feature is a "customer" need, I can't go against it.

xtian-socgen:
Thank you any way for such a fast answer!

PR #5083, opened 2024-11-19
Return delete markers in mc find

allanrogerr:

## Description
Default to list delete flags if they are part of the result set, so that they are not hidden from the find function

## Motivation and Context
The default behaviour of `mc find` is to not return delete flags. However, for example in the case of ILM, this makes it difficult to detect if there are delete marker candidates for processing, even when the `versions` flag is specified.

## How to test this PR?
Run `mc find --versions` on a specific bucket where delete markers are known to be present. These delete markers should be part of the result set.
e.g.
```
➜  mc git:(find-versions) ./mc find --smaller 1B local/find --json --versions                     
{
 "status": "success",
 "type": "",
 "lastModified": "2024-11-19T06:35:14.294-04:00",
 "size": 0,
 "key": "local/find/pics/47110640341_df7a79b41b_o.jpg",
 "etag": "",
 "versionId": "3ad022fa-230a-4bbd-86cf-b4dd12a3f4b0"
}
➜  mc git:(find-versions) ./mc find --smaller 1B local/find --json           
➜  mc git:(find-versions) 
```

## Types of changes
- [X] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/cmd/find-main.go
+++ b/cmd/find-main.go
@@ -45,10 +45,6 @@ var (
 			Name:  "versions",
 			Usage: "include all objects versions",
 		},
-		cli.BoolFlag{
-			Name:  "with-delete-marker",
-			Usage: "include delete markers when used with the `versions` flag",
-		},
 		cli.StringFlag{
 			Name:  "name",
 			Usage: "find object names matching wildcard pattern",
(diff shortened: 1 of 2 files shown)


klauspost:
Wouldn't this change existing behavior? Maybe do a `--with-dm` flag?

allanrogerr:
> Wouldn't this change existing behavior? Maybe do a `--with-dm` flag?

It would however @vadmeste opinion was that although it may break in some rare corner cases in the community, it is better than adding a new flag. Your thoughts @klauspost ?

klauspost:
I see potential scripts breaking... And weird stuff people use it for breaking in strange ways... So personally uncomfortable changing it.

allanrogerr:
Simplified the implementation by enabling delete marker listing when using `--versions`

harshavardhana (approved):

PR #5104, opened 2025-01-06
Change enable to "enable="

taran-p:

## Description
Change `mc idp ldap/openid enable` to change config file to "enable=" rather than "enable=on"

If configuration is returned by `mc idp ldap/openid info` but `enable` is missing, print that `enable` is on since that is the default.

## Motivation and Context
Currently if LDAP or OpenID is configured through environment variables, it is possible to disable the config but not possible to reenable.

miniohq/eos#276

## How to test this PR?
Configure LDAP or OpenID using environment variables, disable the config, then enable it. This fails previously, now succeeds.

## Types of changes

## Checklist:


taran-p:
@harshavardhana @vadmeste Not sure if this is the "right" way to do this, but it's simple and works. The issue stems from the fact that enable/disable simply edits the configuration file, and that editing process checks the validity of the config in the file but not env variables. This fix works because enable is on by default as long as there is a valid configuration. If this isn't the appropriate way to fix this, I can explore other options.

harshavardhana (approved):

harshavardhana:
> @harshavardhana @vadmeste Not sure if this is the "right" way to do this, but it's simple and works. The issue stems from the fact that enable/disable simply edits the configuration file, and that editing process checks the validity of the config in the file but not env variables. This fix works because enable is on by default as long as there is a valid configuration. If this isn't the appropriate way to fix this, I can explore other options.

You don't need enable=on, in SetKV it is expected default when you are setting it, when not specified. 

enable=off is needed when you want to turn it off explicitly.

zveinn (approved):

harshavardhana (approved):

PR #5107, opened 2025-01-13
Don't show Bucket created confirmation message if bucket creation failed

jinapurapu:

## Description
Skips displaying "Bucket created successfully" confirmation if bucket creation fails.

## Motivation and Context

Before, UI displayed "Bucket created successfully" confirmation despite failure of bucket creation. Confirmation no longer displayed if bucket creation fails. 

Before
Screenshot 2025-01-13 at 3 04 25 PM

Screenshot 2025-01-13 at 3 02 09 PM
## How to test this PR?
Attempt to create a bucket in an incorrect way (ex - "mc mb ALIAS BUCKETNAME" vs  "mc mb ALIAS/BUCKETNAME"). No bucket should be created (can verify with mc ls), and only error message should show.

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/cmd/mb-main.go
+++ b/cmd/mb-main.go
@@ -153,7 +153,7 @@ func mainMakeBucket(cliCtx *cli.Context) error {
 				errorIf(err.Trace(targetURL), "Unable to make bucket `%s`.", targetURL)
 			}
 			cErr = exitStatus(globalErrorExitStatus)
-			continue
+			break
 		}
 
 		if cliCtx.Bool("with-versioning") {


donatello (changes requested):
mc mb works the same as mkdir for local filesystem. So when multiple args are given it will try to create all the buckets given. So in this case it's more of a user error.

jinapurapu:
> mc mb works the same as mkdir for local filesystem. So when multiple args are given it will try to create all the buckets given. So in this case it's more of a user error.

Yes, but in the error case it still was confirming that the bucket was created when it was not, printing both the correct error and the incorrect confirmation message.

donatello:
badbucket would have been created as a local directory. So the success part of the message is also correct

PR #5113, opened 2025-01-20
Add "mc support upload" encryption

klauspost:
## Description

* Add `--enc` parameter to `mc support upload` to encrypt the file content.
* Add file size as zstd frame content size.
* Test opening the file before doing more.
* Uses same key as `mc support inspect`.
* 
## How to test this PR?

@anjalshireesh - would it be possible for you to test the upload? We will probably need to make changes to `inspect`  but I would need a test file to verify

## Types of changes


zveinn (approved):

anjalshireesh:
Sure @klauspost I will test and update

cmd/subnet-file-uploader.go
```diff
@@ -80,6 +87,10 @@ func (i *SubnetFileUploader) updateParams() {
 		i.filename += ".zst"
 		i.Params.Add("auto-compression", "zstd")
 	}
+	if i.AutoEncrypt || len(i.PubKey) >= 0 {
```
anjalshireesh, line 90:
```suggestion
	if i.AutoEncrypt || len(i.PubKey) > 0 {
```

This is causing the filename to become `<file>.enc.enc` in case of inspect.

klauspost: ok, is this used for inspect.... Yeah. That should be fixed, since it probably is also double encrypted.

anjalshireesh (approved):
LGTM, tested, working fine. (tested `mc support inspect` also)

@klauspost the inspect command allows using custom keys. https://github.com/minio/mc/blob/423ef1ba2ab62f30e220bc9fd2bc8a232cf5cde3/cmd/support-inspect.go#L165 I feel it will be useful to support that here also.

klauspost:
@anjalshireesh We never use it, so didn't bother. We can always add it, if we ever should get a use for it.

harshavardhana:
we need this in `miniohq/ec` as well.

PR #5116, opened 2025-01-21
Giving mirror the option to exit early when --remove is not set

zveinn:
This PR allowed `mc mirror` to skip full listing of the target bucket if `--remove` is not set. 

NOTE: I marked this as a speed up with no functional changes, but technically mirror will change it's behavior if `--remove` is not used 

## Types of changes


cmd/difference.go
```diff
@@ -232,8 +243,14 @@ func differenceInternal(sourceURL string, srcCh <-chan *ClientContent, targetURL
...
 		// No objects from source AND target: Finish
-		if srcEOF && tgtEOF {
-			break
+		if opts.isRemove {
+			if srcEOF && tgtEOF {
+				break
+			}
+		} else {
+			if srcEOF {
+				break
```
vadmeste, line 252: in fact, we cannot exit here, 

if the source listing is finished and there are still more entries in the target listing, we will generate a wrong difference here because we are exiting early.

isn't an early quitting here when we detect a canceled context is  enough to fix the issue ?

vadmeste (changes requested):

cmd/difference.go
zveinn: ok.. I think I see what you mean. I'm probably exiting to early because we need to do all the diff check stuff. 
I'll check if there is a better way.

I don't think cancel context is enough, but I'll re-visit this tomorrow

zveinn: NOTE: we ended up with a similar result, but taking into account the active-active flag.

cmd/mirror-url.go
```diff
@@ -277,6 +277,7 @@ type mirrorOptions struct {
 	storageClass                                          string
 	userMetadata                                          map[string]string
 	checksum                                              minio.ChecksumType
+	localListingOnly                                      bool
```
vadmeste, line 280: s/sourceListingOnly/localListingOnly/g

vadmeste (commented):
one variable renaming, LGTM otherwise

vadmeste (approved):

Issue #5162, opened 2025-03-10
mc mirror --watch should exit with non-zero code upon failure


jadolg:
## Expected behavior
When any error occurs during the mirror operation, **mc** should exit with a non-zero status code instead of just logging the error and continuing.

## Actual behavior
mc logs the errors but continues to run.

## Steps to reproduce the behavior
Try anything that would cause a problem like non existing bucket in the destination.

## mc --version
mc version RELEASE.2025-02-15T10-36-16Z (commit-id=383560b1c3d6912042e8c8c275bb78e83e67ef2b)
Runtime: go1.23.6 linux/amd64

## System information
Ubuntu 22.04 
5.15.0-122-generic

jadolg:
I think changing [monitorMirrorStatus](https://github.com/minio/mc/blob/ce874fdc6a059d5376558cf108c96132a8f3ac4d/cmd/mirror-main.go#L559) to use `fatalIf` instead of `errorIf` should solve this.

PR #5163, opened 2025-03-11
Exit mc mirror when errors occur if retry is disabled and it's a watch operation

jadolg:

## Description

When running `mc mirror --watch` the cli exits with a non-zero exit code instead of just printing the error.

## Motivation and Context

I'm running `mc mirror --watch` in a **systemd** unit to constantly mirror my minio buckets to AWS S3. In the current state, it's not possible to monitor **mc** effectively since the process will just continue running.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/mirror-main.go
+++ b/cmd/mirror-main.go
@@ -561,6 +561,10 @@ func (mj *mirrorJob) monitorMirrorStatus(cancel context.CancelFunc) (errDuringMi
 	mj.status.Start()
 	defer mj.status.Finish()
 
+	// if the operation is not retriable and is a watch operation, then
+	// we should exit on the first error.
+	useFatal := !mj.opts.isRetriable && mj.opts.isWatch
+
 	var cancelInProgress bool
 
 	defer func() {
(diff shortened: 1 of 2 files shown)


jadolg:
Solves https://github.com/minio/mc/issues/5162

klauspost (commented):
I may be missing something, but this seems very likely to break all kind of stuff  already running?

jadolg:
@klauspost it is a breaking change but I think it makes sense.
`mc mirror --watch` is a long running operation. Without directly looking at the logs, there's currently no way to know that the application is failing to do it's job.
In my case, I was trying to mirror a bucket into an unexisting bucket in another system. I'm running `mc mirror --watch` in an "unatended" way (systemd unit) so I'm not looking at the logs unless something goes wrong. `mc mirror --watch` will fail and I have no idea my bucket is not being mirrored.
I could technically monitor the logs and trigger some alert if errors occur, but then I'd need to account for all possible errors and error format making this not really bulletproof.
My point is: I think it should break if an error occurs so it is obvious the operation can't be completed.
If you are concerned about breaking other people's already existing workflows, we could also hide this behavior behind a flag (Eg: `--fail-on-error`)
wdyt?

klauspost:
> If you are concerned about breaking other people's already existing workflows, we could also hide this behavior behind a flag (Eg: `--fail-on-error`)

Yeah. I think that is the only reasonable approach. `mc` is running in a lot of places, so we need to be reasonable about back-compat.

jadolg:
New flag added and removed the need for the `--watch` parameter since the new flag should make it explicit.

cmd/mirror-main.go
```diff
@@ -1011,6 +1015,7 @@ func runMirror(ctx context.Context, srcURL, dstURL string, cli *cli.Context, enc
 		isMetadata:            isMetadata,
 		isSummary:             cli.Bool("summary"),
 		isRetriable:           cli.Bool("retry"),
+		failOnError:           cli.Bool("fail-on-error"),
```
klauspost, line 1018: You must add this to the [flags](https://github.com/minio/mc/blob/391616f9cfb9af5799a5cea63ace1a48b4c11481/cmd/mirror-main.go#L49) as well AFAICT.

jadolg: Added in [cf8fbed](https://github.com/minio/mc/pull/5163/commits/cf8fbed54fefdb34279076029284c855ade7da16)

klauspost:
ok, so my only remaining reservation is that it may not exit in all cases where people would expect it, but I can live with that.

jadolg:
> ok, so my only remaining reservation is that it may not exit in all cases where people would expect it, but I can live with that.

It's tricky to get it to fail everywhere where it might fail because everything happens separately but I think `monitorMirrorStatus` should cover most cases since the errors were only logged there anyway.

jadolg:
@klauspost What would be the next steps for this PR to get merged or at least looked at?

PR #5262, opened 2025-11-12
feat: allow to list objects sorted by size

iTrooz:

## Description & context
Add a `--sort=size` option to `mcli ls` to be able to sort files by size

Usually, you would do that by piping to `sort`, or using the `--sort` flag of ls

## How to test this PR?
Compare `go run . ls <path>` and `go run . ls <path> --sort=size`

## Types of changes

## Checklist:
- [ ] Internal documentation updated (??)

### DIFF
--- a/cmd/ls-main.go
+++ b/cmd/ls-main.go
@@ -60,6 +60,10 @@ var (
 			Name:  "zip",
 			Usage: "list files inside zip archive (MinIO servers only)",
 		},
+		cli.StringFlag{
+			Name:  "sort",
+			Usage: "sort by field. Only possible value for now: size",
+		},
 	}
 )
 
@@ -177,6 +181,11 @@ func checkListSyntax(cliCtx *cli.Context) ([]string, doListOptions) {
 
 	timeRef := parseRewindFlag(cliCtx.String("rewind"))
 
+	sortBy := cliCtx.String("sort")
+	if sortBy != "" && sortBy != "size" {
+		fatalIf(errInvalidArgument().Trace(args...), "Unsupported sort option '"+sortBy+"'. Only 'size' is supported.")
+	}
+
 	if listZip && (withVersions || !timeRef.IsZero()) {
 		fatalIf(errInvalidArgument().Trace(args...), "Zip file listing can only be performed on the latest version")
 	}
@@ -189,6 +198,7 @@ func checkListSyntax(cliCtx *cli.Context) ([]string, doListOptions) {
 		withVersions: withVersions,
 		listZip:      listZip,
 		filter:       storageClasss,
+		sortBy:       sortBy,
 	}
 	return args, opts
 }
--- a/cmd/ls-main.go
+++ b/cmd/ls-main.go
@@ -182,7 +182,10 @@ func checkListSyntax(cliCtx *cli.Context) ([]string, doListOptions) {
 	timeRef := parseRewindFlag(cliCtx.String("rewind"))
 
 	sortBy := cliCtx.String("sort")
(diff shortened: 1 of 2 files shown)


iTrooz:
I'm waiting for initial feature approval before adding tests/making a PR to the docs repo

cmd/ls.go
```diff
@@ -257,7 +259,19 @@ func doList(ctx context.Context, clnt Client, o doListOptions) error {
 		totalObjects++
 	}
 
-	printObjectVersions(clnt.GetURL(), perObjectVersions, o.withVersions)
+	objects = append(objects, getObjectVersions(clnt.GetURL(), perObjectVersions, o.withVersions)...)
```
klauspost: This definitely shouldn't be the default behaviour. There can be hundreds of millions of results. 

A) You likely just OOMs the client.
B) There will be no output for a long-long time. Some listings can take hours.

cmd/ls.go
```diff
@@ -257,7 +259,19 @@ func doList(ctx context.Context, clnt Client, o doListOptions) error {
...
+
+	// Sort by size if requested
+	if o.sortBy == "size" {
+		slices.SortFunc(objects, func(a, b contentMessage) int {
+			return int(a.Size - b.Size)
```
klauspost: Clever, but breaks on 32 bit platforms above 2GiB.

Just do simple if.

cmd/ls-main.go
```diff
@@ -177,6 +181,11 @@ func checkListSyntax(cliCtx *cli.Context) ([]string, doListOptions) {
 
 	timeRef := parseRewindFlag(cliCtx.String("rewind"))
 
+	sortBy := cliCtx.String("sort")
+	if sortBy != "" && sortBy != "size" {
+		fatalIf(errInvalidArgument().Trace(args...), "Unsupported sort option '"+sortBy+"'. Only 'size' is supported.")
+	}
```
klauspost: Do a `switch sortBy {`

klauspost (commented):
Still not sure it should be there. But leaving for others to decide.

cmd/ls.go
iTrooz: Oh, indeed thank you.

I just pushed the fix for that

iTrooz:
@shtripat @harshavardhana Hello ! Would it be possible to review this PR please ? Thank you

shtripat:
I am confused about the value add of this feature. If we really need such a thing, we can use `--json` output and then sort as and when needed. Something like `mc ls ALIAS/BUCKET -r --json | jq -s 'sort_by(.size) | .[] | {key, size}'`

iTrooz:
We could do that, but but the `--json | jq -s 'sort_by(.size) | .[] | {key, size}'` part is a bit cumbersome (I personally wouldn't be able to pull it off in a reasonable time), which is why I added the `--sort`. It is purely a QoL feature

shtripat:
> We could do that, but but the `--json | jq -s 'sort_by(.size) | .[] | {key, size}'` part is a bit cumbersome (I personally wouldn't be able to pull it off in a reasonable time), which is why I added the `--sort`. It is purely a QoL feature

IIUC this kind of sorting is needed when we are programmatically processing the output and `--json` is meant to be used for this purpose. But lets hear from Harsha as well.

iTrooz:
My use case was "I want to see what the biggest files are in this bucket" (not in the context of a script, purely me out of curiosity)
Maybe there is already a way to do that that I missed ?

> But lets hear from Harsha as well.

Ok !

